PR #14, opened 2017-10-27
support giving a repo in the manifest

KristofferC:


Keno:
So this just adds a `repo` field and drops versions to allow non-registered packages. Upon further consideration, that might not be the proper thing to do. We should probably have a commit hash for version, so we still place it within a history (e.g. for upgrading to newer commits). One thing this also brought up is that the lack of compatibility information in the repo now, makes this kind of thing significantly more awkward.

StefanKarpinski:
I'm equal parts thrilled and terrified to have anyone else contributing to this repo 😁 😨

KristofferC:
The implementation process was to add the repo field to the project file and fixing errors until it worked :D.

StefanKarpinski:
I need to review this – will look at it soon.

KristofferC:
I was planning on looking at cloning after the pin free stuff is done (it seems to work now, so just have to do some finishing touches) and this project is relevant for that.
.
Instead of focusing too much on the implementation in this PR I think it makes sense to figure out a good design, what information hold exist in the manifest and how should we interact with it. The implementation will probably fall out by itself when the design imfeels solid.

KristofferC:
This works now.

PR #15, opened 2017-11-08
use HTTP downloads if possible

KristofferC:
A proof of concept for how we could use tarballs (or some other archive) instead of LibGit2 to download. Improves performance greatly from what I can see.

I'm currently using some utilities from BinaryProvider, they have very nice stuff to query for download and uncompress commands, it could perhaps be internalized.

https://github.com/StefanKarpinski/Pkg3.jl/pull/15/files?w=1 for no whitespace diff

### DIFF
--- a/ext/BinaryProvider/.codecov.yml
+++ b/ext/BinaryProvider/.codecov.yml
@@ -1,0 +1,1 @@
+comment: false
--- a/ext/BinaryProvider/.gitignore
+++ b/ext/BinaryProvider/.gitignore
@@ -1,0 +1,5 @@
+*.jl.cov
+*.jl.*.cov
+*.jl.mem
+
+global_prefix
--- a/ext/BinaryProvider/.travis.yml
+++ b/ext/BinaryProvider/.travis.yml
@@ -1,0 +1,30 @@
+# Documentation: http://docs.travis-ci.com/user/languages/julia/
+language: julia
+os:
+  - linux
+  - osx
+julia:
+  - 0.6
+  - nightly
+notifications:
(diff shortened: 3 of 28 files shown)


StefanKarpinski:
Very cool. We should decide on an interface for configuring this that's more official. @KristofferC, what do you think, should we merge this before figuring out an interface or the other way?

KristofferC:
If we had a `.config/julia` folder we could have put a `PkgConfig.toml` file there.

Also, as I said, this requires some Binary provider stuff as is. I could just internalise either the whole package or just the parts needed.

KristofferC:
Ok, I have internalized BinaryProvider. The best way to review is https://github.com/StefanKarpinski/Pkg3.jl/pull/15/commits/ff0704c8b9fbd6daba2dec3d23969781a882f7da?w=1.

The negative things with this PR are:

* A bit longer startup time, ~0.6 seconds is due to https://github.com/JuliaPackaging/BinaryProvider.jl/issues/18. We could fix this with a build script that caches the results.

* If one presses Ctrl + C while downloading a tarball, the same package will be tried to download with LibGit2. The reason is that there is no way to distinguish a download failure with an interruption exception (tracked at https://github.com/JuliaPackaging/BinaryProvider.jl/issues/17).

StefanKarpinski:
> Ok, I have internalized BinaryProvider.

Nice, this is the direction we need to go in any case, so the down sides are kind of inevitable. Seems like we should just fix the two BinaryProvider issues.

KristofferC:
Subsumed by #25

PR #16, opened 2017-11-08
internalize the needed parts from Pkg2

KristofferC:
This makes Pkg3 (including the registry generation) work without using Base.Pkg.
Since Base.Pkg will be removed, this needs to happen at some point, and it is good to see how much code that we rely on from Pkg2.

I have tried this on a locally built Julia 0.6 with Base.Pkg removed (except Pkg.dir because it is needed for code loading) and things seem to work fine.

### DIFF
--- a/bin/loadmeta.jl
+++ b/bin/loadmeta.jl
@@ -1,14 +1,15 @@
 #!/usr/bin/env julia
 
 using Base: thispatch, thisminor, nextpatch, nextminor
-using Base.Pkg.Reqs: Reqs, Requirement
-using Base.Pkg.Types
 using Base.LinAlg: checksquare
 using Base.Random: UUID
 using SHA
 
-## Computing UUID5 values from (namespace, key) pairs ##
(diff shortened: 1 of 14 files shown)


bin/loadmeta.jl
```diff
@@ -1,14 +1,15 @@
...
 
 using Base: thispatch, thisminor, nextpatch, nextminor
-using Base.Pkg.Reqs: Reqs, Requirement
-using Base.Pkg.Types
 using Base.LinAlg: checksquare
 using Base.Random: UUID
 using SHA
 
-## Computing UUID5 values from (namespace, key) pairs ##
+import Pkg3.Pkg2
```
ararslan: 😄

KristofferC: In `Pkg4` we will have `Pkg4.Pkg3.Pkg2`.

StefanKarpinski: The Russian doll of package managers.

PR #21, opened 2017-11-10
WIP: free and pin

KristofferC:
TODO: Check that the package is in the project. Or should we also support pinning freeing things that are only in the manifest?

https://asciinema.org/a/ua8H2J80S1bJN77xomEklra0l

### DIFF
--- a/src/Display.jl
+++ b/src/Display.jl
@@ -72,16 +72,19 @@ end
 struct VerInfo
     hash::SHA1
     ver::Union{VersionNumber,Void}
+    pinned::Union{Bool, Void}
 end
 
 vstring(a::VerInfo) =
-    a.ver == nothing ? "[$(string(a.hash)[1:16])]" : "v$(a.ver)"
+    string(a.ver == nothing ? "[$(string(a.hash)[1:16])]" : "v$(a.ver)",
+           a.pinned == true ? "⚲" : "")
 
 Base.:(==)(a::VerInfo, b::VerInfo) =
-    a.hash == b.hash && a.ver == b.ver
+    a.hash == b.hash && a.ver == b.ver && a.pinned == b.pinned
 
 ≈(a::VerInfo, b::VerInfo) = a.hash == b.hash &&
-    (a.ver == nothing || b.ver == nothing || a.ver == b.ver)
+    (a.ver == nothing || b.ver == nothing || a.ver == b.ver) &&
+    (a.pinned == b.pinned)
 
 struct DiffEntry
     uuid::UUID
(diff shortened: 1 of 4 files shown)


StefanKarpinski:
This is looking good. I have to review more thoroughly, but it's a good start and I'm glad you're able to make sense of my code and modify it. Re project versus manifest... On the one hand, if you don't have a direct dependency on something, why do you care what version it's at? On the other hand, why shouldn't we let you insist that it stay at the same version? Of course, that always introduces the somewhat weird corner case: what if changing versions removes the need for something in the manifest entirely? In that case, does having it pinned (a) prevent that resolution from being considered, or (b) go ahead, deleting the pinned package entirely, or (c) leave the pinned package in place even though it's no longer needed by any project dependency? These are all kind of weird solutions. For (a) maybe you wanted the package at that version because it's the only one that works for some reason; but not needing it at all seems better in that case. In case (b) if you do something that re-adds the previously-pinned, now-deleted package, it will no longer be pinned, which seems undesirable. For (c), it's obviously kind of strange to have orphaned manifest entries that are no longer needed by anything.

My plan for expressing compatibility is to have a `Compat.toml` file that records which commits of a package are broken/fixed by which commits of other packages – not necessarily dependencies, since you can record incompatibilities with other packages that you don't depend on. From there, pointwise compatibility ranges that are recorded in registries can be computed for easier resolution. This would be a more permanent place to record the incompatibilities that caused you to need to pin something, and it would allow effectively pinning manifest dependencies in a sensible way whether they are present or not. Perhaps this is a better way to approach this problem? Pinning was always a bit of a hack to work around the fact that the old package manager tended up want to upgrade everything all the time. The new one only upgrades the things you ask it to, so maybe this is less of a problem?

KristofferC:
> My plan for expressing compatibility is to have a Compat.toml file that records which commits of a package are broken/fixed by which commits of other packages – not necessarily dependencies, since you can record incompatibilities with other packages that you don't depend on. From there, pointwise compatibility ranges that are recorded in registries can be computed for easier resolution. This would be a more permanent place to record the incompatibilities that caused you to need to pin something, and it would allow effectively pinning manifest dependencies in a sensible way whether they are present or not. Perhaps this is a better way to approach this problem?

I must say that this sounds quite complicated and thinking about the timeline when Pkg3 should be available for general use, perhaps it could be a 1.1 feature or something.

My opinion of how we can get the excellent improvements of Pkg3 out in a timely manner is the following:

* The Registry is the new METADATA, it will (like old METADATA) hold version history for packages, and the dependencies of each version. This is pretty much exactly what the script generates (except the script seems to also add upper bounds when the original REQUIRE did not have one, and it also seems to ignore OS-specific bounds).

* The REQUIRE file for packages has a quite similar format as currently. We could potentially allow for giving unregistered packages in the REQURIE but perhaps not allow these to be registered?

* Tagging a package works quite similar to how it is currently done. Attobot creates a PR to the Registry (perhaps a command in Pkg3 as well). The fact that dependencies are compressed make manual tagging a bit more difficult.

I feel that questions like "How should tagging of a package be done?", and "What information should the registry contain?", needs to be finalized quite soon so that, tools and implementation can start properly.

KristofferC:
Hmm, is `pin` really needed anymore though? Since we now have more control over what packages gets updated and the "update level" you can just chose not to update the package that would be pinned.

I guess it is convenient if you just one to have one package version fixed, you could pin it and then to a full upgrade... Perhaps this can be left to cool for a while since it is not needed urgently anyway.

PR #25, opened 2017-11-14
Asynchronously install packages

KristofferC:
I moved (and changed) the printing a bit because printing inside a task can get pretty confusing.

Build on top of https://github.com/StefanKarpinski/Pkg3.jl/pull/15.

The commit to look at is https://github.com/StefanKarpinski/Pkg3.jl/pull/25/commits/26983596f51b965c069b9ab020ecd479fbc73425.

### DIFF
--- a/ext/BinaryProvider/.codecov.yml
+++ b/ext/BinaryProvider/.codecov.yml
@@ -1,0 +1,1 @@
+comment: false
--- a/ext/BinaryProvider/.gitignore
+++ b/ext/BinaryProvider/.gitignore
@@ -1,0 +1,5 @@
+*.jl.cov
+*.jl.*.cov
+*.jl.mem
+
+global_prefix
--- a/ext/BinaryProvider/.travis.yml
+++ b/ext/BinaryProvider/.travis.yml
@@ -1,0 +1,30 @@
+# Documentation: http://docs.travis-ci.com/user/languages/julia/
+language: julia
+os:
+  - linux
+  - osx
+julia:
+  - 0.6
+  - nightly
+notifications:
+  email: false
+
+matrix:
+  include:
+    # Make sure to override to "wget" at least once
+    - julia: 0.6
+      os: linux
+      env: BINARYPROVIDER_DOWNLOAD_ENGINE="wget"
+
+# Ironic.  He could provide binaries for others but not himself...
+addons:
+  apt:
+    packages:
+      - curl
+      - wget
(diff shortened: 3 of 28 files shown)


StefanKarpinski:
When you internalize, did you make a commit where the three (and thus its hash) is identical to the upstream version? If changes need to be made after that, they can be, but this is helpful for tracking what version we got the code from upstream.

KristofferC:
IIRC, I cloned it, deleted `.git`-folder and then `git add`ed it, then commit. I can do it some other way if that is better.

StefanKarpinski:
No, that's exactly how I did it too.

KristofferC:
Simon tested and there are some problems on Windows. I'm fixing a VM now so I can check it out.

StefanKarpinski:
FWIW, I haven't tested at all on Windows, so it may not be the fault of this PR. You may want to test on master first to make sure that works.

KristofferC:
It was this PR. I think I fixed it though. The solution was to use as little of BinaryProvider as possible :P. It works reliably on my VM now but let's wait for Simon to check it as well.

SimonDanisch:
But shouldn't we fix that in BinaryProvider as well then?

KristofferC:
Feel free to fix it :)

SimonDanisch:
got me there :P I guess we should at least ping @keno and @staticfloat, or are they already aware of that issue?

Btw, the latest commit fixed it on my machine as well :)

staticfloat:
What issue are you seeing?

KristofferC:
Deadlocks on Windows when starting a bunch of tasks that download and unpack files using the `download` and `unpack` commands in BinaryProvider. Quite difficult to debug because print statements in different positions seemed to make it work again.

The consumer at https://github.com/StefanKarpinski/Pkg3.jl/pull/25/commits/760ceca113431f4514fd131c2140059eb6e9769f#diff-920ed338dcbd9d62489c6d12dc2fbb18R386 waits for the `put!´ at https://github.com/StefanKarpinski/Pkg3.jl/pull/25/commits/760ceca113431f4514fd131c2140059eb6e9769f#diff-920ed338dcbd9d62489c6d12dc2fbb18R375 which never arrives.

It did seem that usually it was the first task that started that never finishes.

KristofferC:
Since this works for Simon now, perhaps we should try it out on master, barring any requests from review.

PR #27, opened 2017-11-16
enable precompilation for Pkg3

KristofferC:
Otherwise loading precompiled packages does not work (funnily enough).


StefanKarpinski:
I assume that loading this works though?

Issue #6, opened 2017-10-10
REPL mode error on backspace


Keno:
```
pkg> up JSOERROR (in the keymap): type Void has no field options
Stacktrace:
 [1] edit_backspace(::Base.LineEdit.PromptState) at .\repl\LineEdit.jl:609
 [2] edit_backspace(::Base.LineEdit.MIState) at .\repl\LineEdit.jl:178
 [3] (::getfield(Base.REPL, Symbol("##18#21")){Base.LineEdit.Prompt})(::Base.LineEdit.MIState, ::Void, ::Vararg{Any,N} where N) at .\repl\REPL.jl:701
 [4] (::getfield(Base.LineEdit, Symbol("##19#20")){getfield(Base.REPL, Symbol("##18#21")){Base.LineEdit.Prompt},String})(::Base.LineEdit.MIState, ::Void) at .\repl\LineEdit.jl:1059
 [5] prompt!(::Base.Terminals.TTYTerminal, ::Base.LineEdit.ModalInterface, ::Base.LineEdit.MIState) at .\repl\LineEdit.jl:2061
 [6] run_interface(::Base.Terminals.TTYTerminal, ::Base.LineEdit.ModalInterface) at .\repl\LineEdit.jl:1962
 [7] run_frontend(::Base.REPL.LineEditREPL, ::Base.REPL.REPLBackendRef) at .\repl\REPL.jl:986
 [8] run_repl(::Base.REPL.LineEditREPL, ::getfield(Base, Symbol("##542#543"))) at .\repl\REPL.jl:182
 [9] _start() at .\client.jl:439
```
This is on 0.7

StefanKarpinski:
Oh, I haven't tested on 0.7 at all – I shocked it even loads 😂

KristofferC:
Fixed by #30

PR #30, opened 2017-11-17
Upgrade Pkg3 and ext dependencies to 0.7

KristofferC:
PRs have been filed for removing deprecations to the dependencies.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -134,7 +134,7 @@ function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec})
                     d = get_or_make(Dict{String,UUID}, dependencies, v)
                     r = get_or_make(Dict{String,VersionSpec}, compatibility, v)
                     q = Dict(u => get_or_make(VersionSpec, r, p) for (p, u) in d)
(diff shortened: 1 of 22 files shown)


src/Operations.jl
```diff
@@ -135,7 +135,7 @@ function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec})
                     d = get_or_make(Dict{String,UUID}, dependencies, v)
                     r = get_or_make(Dict{String,VersionSpec}, compatibility, v)
                     q = Dict(u => get_or_make(VersionSpec, r, p) for (p, u) in d)
-                    VERSION in get_or_make(VersionSpec, r, "julia") || continue
+                    # VERSION in get_or_make(VersionSpec, r, "julia") || continue
```
KristofferC: @StefanKarpinski

What should we do about this? On julia 0.7 nothing will match because the VersionSpec is upper bounded to julia 0.6 on everything.

KristofferC:
Fixes #6

KristofferC:
Tests passes locally on 0.6 and 0.7. I'll open an issue about disabling the julia version check but for now I think it is important that people can test this package on 0.7

PR #33, opened 2017-11-17
start implementing REPL test command

KristofferC:
Put most of the scaffolding in place for doing `pkg> test Pakcage`.

What is left (to match Pkg2) is to do resolution and support test only dependencies.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -606,5 +606,65 @@ function up(env::EnvCache, pkgs::Vector{PackageSpec})
     build_versions(env, new)
 end
 
+function test(env::EnvCache, pkgs::Vector{PackageSpec}; coverage=false)
+    # See if we can find the test files for all packages
+    missing_runtests = String[]
+    testfiles        = String[]
+    version_paths    = String[]
+    for pkg in pkgs
+        info = manifest_info(env, pkg.uuid)
+        haskey(info, "hash-sha1") || cmderror("Could not find hash-sha for package $(pkg.name)")
+        version_path = find_installed(pkg.uuid, SHA1(info["hash-sha1"]))
+        testfile = joinpath(version_path, "test", "runtests.jl")
+        if !isfile(testfile)
+            push!(missing_runtests, pkg.name)
+        end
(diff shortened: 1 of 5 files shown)


src/Operations.jl
```diff
@@ -577,5 +577,57 @@ function up(env::EnvCache, pkgs::Vector{PackageSpec})
...
+        cmderror(length(missing_runtests) == 1 ? "Package " : "Packages ",
+                join(missing_runtests, ", "),
+                " did not provide a `test/runtests.jl` file")
+    end
+
+    pkgs_errored = []
+    for (pkg, testfile, version_path) in zip(pkgs, testfiles, version_paths)
+        # TODO: Test only dependencies
+        info("Testing $(pkg.name) located at $version_path")
+        # TODO, cd to test folder (need to be careful with getting the same EnvCache
```
KristofferC: Pkg2 does a `cd` here but we need to be careful because if we are in a path with its own Project file then this will not be respected when the `EnvCache` is created in the process we start below.

We could send a path to the process below pointing to the place where the current Project is located.

src/REPLMode.jl
```diff
@@ -388,6 +402,37 @@ function do_status!(env::EnvCache, tokens::Vector{Tuple{Symbol,Vararg{Any}}})
...
+# TODO , test recursive dependencies as on option.
+function do_test!(env::EnvCache, tokens::Vector{Tuple{Symbol,Vararg{Any}}})
+    pkgs = PackageSpec[]
+    coverage = false
+    while !isempty(tokens)
+        token = shift!(tokens)
+        if token[1] == :pkg
+            if length(token) == 2
+                pkg = PackageSpec(token[2])
+                pkg.mode = :manifest
```
KristofferC: Is this correct? So that it is possible to test packages only in the Manifest.

src/Operations.jl
StefanKarpinski: > We could send a path to the process below pointing to the place where the current Project is located.

Yes, that sounds like the way to do it – should be an absolute path.

Issue #34, opened 2017-11-17
Dry runs (non modificating) execution of commands


KristofferC:
Perhaps it would be a good idea to be able to run the commands in a "dry mode" where the command is run but no modifications are made to anything.

This could be a good way to allow people to explore different commands without being scared of messing up something.

StefanKarpinski:
I was kind of hoping that checking project and manifest files into git would address that since you can always get back to a previous state just by doing a `git checkout Project.toml Manifest.toml`. Thought about having a `pkg> reset` or something. Perhaps we could have some kind of `pkg> undo` command? This would require keeping a per-environment log of operations and then updoing them, but we could do that. Would only operate on the project and manifest files since everything else should not need undoing, being immutable and all.

KristofferC:
Yeah, perhaps `git checkout` is enough. I was thinking that it would skip downloading, building etc as well but perhaps this is just overkill.

StefanKarpinski:
I guess that could be useful, kind of a "preview" feature? So maybe `pkg> preview add Gadfly` would show you all the things that *would* be installed and added to the project & manifest. I kind of like that – sometimes you do really want to know what effect something will have *before* doing it.

KristofferC:
Fixed by #35

PR #35, opened 2017-11-18
add preview mode

KristofferC:
As a response to #34, this shows a possible implementation.

Any command can be called as `preview cmd` and the effects of the command will go through but with (hopefully) no side effects. This removes downloading, building, and writing to any Project or Manifest files.

Example:

image

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -231,6 +231,8 @@ struct EnvCache
     uuids::Dict{String,Vector{UUID}}
     paths::Dict{UUID,Vector{String}}
 
+    preview::Ref{Bool}
+
     function EnvCache(env::Union{Void,String})
         project_file, git_repo = find_project(env)
         project = read_project(project_file)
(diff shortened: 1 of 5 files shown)


KristofferC:
Thanks for review! Don't have merge rights on this repo though.

StefanKarpinski:
Arg. GitHub's read-only default when you add people/teams. Everyone with Julia commit bit should now be able to merge PRs on Pkg3 as well.

src/Types.jl
```diff
@@ -231,6 +231,8 @@ struct EnvCache
     uuids::Dict{String,Vector{UUID}}
     paths::Dict{UUID,Vector{String}}
 
+    preview::Ref{Bool}
```
StefanKarpinski: I should have mentioned this the first time around but it might be better to just make this strict mutable. There’s no real reason for it to be immutable.

StefanKarpinski:
This is really nice to use  💯👍

PR #36, opened 2017-11-18
make a basic functional API and add some very basic tests

KristofferC:
Gief Travis plz?

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1,0 +1,56 @@
+module API
+
+import Pkg3
+using Pkg3.Types
+using Base.Random.UUID
+
+add(pkg::String)               = add([pkg])
+add(pkgs::Vector{String})      = add([PackageSpec(pkg) for pkg in pkgs])
+add(pkgs::Vector{PackageSpec}) = add(EnvCache(), pkgs)
+
+function add(env::EnvCache, pkgs::Vector{PackageSpec})
+    project_resolve!(env, pkgs)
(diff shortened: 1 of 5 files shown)


test/runtests.jl
```diff
@@ -1,5 +1,34 @@
...
+    end
+end
+
+temp_pkg_dir() do
+    Pkg3.API.add("Example")
+    @eval import Example
+    Pkg3.API.up()
+    Pkg3.API.rm("Example")
+
+    nonexisting_pkg = randstring(14)
```
KristofferC: YOLO

src/API.jl
```diff
@@ -0,0 +1,56 @@
+module API
+
+import Pkg3
+using Pkg3.Types
+using Base.Random.UUID
+
+add(pkg::String)               = add([pkg])
+add(pkgs::Vector{String})      = add([PackageSpec(pkg) for pkg in pkgs])
+add(pkgs::Vector{PackageSpec}) = add(EnvCache(), pkgs)
```
StefanKarpinski: Perhaps `add(pkgs::Vector{<:Union{String,PackageSpec}})`? Ditto with the others. Of course, this is a stricter API so we could always loosen to that if it turns out to be useful.

test/runtests.jl
StefanKarpinski: Safer than flying.

PR #41, opened 2017-11-23
add tokenization of strings

KristofferC:
We will probably need to accept paths as input (e.g. when checking out a package) and the easiest way to give them is probably in a quoted manner.


KristofferC:
Perhaps this is not so nice. It would be better if you could just write something like `--path ~/juliadev/´ instead of having to put quotes there. That would mean that tokenization couldn't really say if it is a pkg or a path so I guess that would be up to the "parser".

StefanKarpinski:
Agree. It's probably nicer to just allow bare paths and URLs. It does somewhat complicate things, but I think it should be doable. Will have to think about this a bit more.

Issue #28, opened 2017-11-16
add precompiles file


StefanKarpinski:
Startup time is really slow even with precompilation. Presumably having a precompiles file would help that out, so it's worth trying. I guess the way to do this these days is with SnoopCompile?

KristofferC:
I will try it out and see if it improves stuff.

KristofferC:
I wonder if the load time is long because we overwrite methods and thereby possible invalidate a 
lot of other methods?

Anyway, snoop compile by itself actually increased `import Pkg3` from 1.08 to 1.88 seconds. It also didn't really help with time to first call. However, adding a `Base.require(:Pkg3)` to `userimg.jl` and recompiling made stuff feel a lot faster (and this is probably how it will be released) so that's reassuring.

KristofferC:
Here's the branch https://github.com/KristofferC/Pkg3.jl/tree/kc/precompile

StefanKarpinski:
> I wonder if the load time is long because we overwrite methods and thereby possible invalidate a
> lot of other methods?

Oh yeah, that is entirely possible.

KristofferC:
I could try a version of Julia where it is possible to plugin into the loading system instead of overwriting methods. Would give a sense of what the cost of method overwrites are.

KristofferC:
Done

PR #42, opened 2017-11-23
add cmdline binary

KristofferC:
This allows one to execute Pkg3 commands without explicitly starting Julia with `juliapkg ...`. Perhaps this could be added in the bin directory of julia.

The startup time is still horrible but that is probably related to #28

image

### DIFF
--- a/bin/juliapkg
+++ b/bin/juliapkg
@@ -1,0 +1,7 @@
+#!/usr/bin/env julia
+
+import Pkg3
+
+@eval Base have_color = true
+Pkg3.REPLMode.do_cmd(join(ARGS, " "))
+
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Is this usable with the compilation time?

KristofferC:
Not as a package (method overwrites are really bad for compilation)  but as a stdlib package, where the method overrides have been swapped out for hooks into Base, it is at least decent. However, perhaps we should just encourage Repl use.

KristofferC:
Our REPL is awsome, we should make it the preferred way of using Pkg3.

StefanKarpinski:
Agree! Perl always had the best package manager and it was based on its REPL.

ExpandingMan:
As much as I agree that the REPL is awesome, it would be very useful to have a command line entry point to the package manager because this will be extremely useful for container set up scripts such as dockerfiles and AWS scripts.  I think a `juliapkg` command would be awesome, any chance of reviving this?

PR #44, opened 2017-11-26
add test for pkg name in resolution failure

KristofferC:


KristofferC:
Passes locally, will just merge since adding a test should be uncontroversial.

PR #46, opened 2017-11-27
update CI

KristofferC:


KristofferC:
We are now competing with base Julia and METADATA for CI :P

StefanKarpinski:
Sigh. Should we move to a different org? That’s always stuck me as an unfortunate reason for choosing orgs.

PR #47, opened 2017-11-27
remove sha field from Available

KristofferC:
This doesn't seem to have any purpose in the way Pkg3 structure things (SHA is readily available from the Manifest as opposed to the REQUIRE file in Pkg2).

Does this makes sense to you @StefanKarpinski.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -116,7 +116,7 @@ load_package_data(f::Base.Callable, path::String, version::VersionNumber) =
     get(load_package_data(f, path, [version]), version, nothing)
 
 function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec})
-    deps = Dict{UUID,Dict{VersionNumber,Tuple{SHA1,Dict{UUID,VersionSpec}}}}()
+    deps = Dict{UUID,Dict{VersionNumber,Dict{UUID,VersionSpec}}}()
     uuids = [pkg.uuid for pkg in pkgs]
     seen = UUID[]
     while true
@@ -135,7 +135,7 @@ function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec})
                     r = get_or_make(Dict{String,VersionSpec}, compatibility, v)
                     q = Dict(u => get_or_make(VersionSpec, r, p) for (p, u) in d)
                     # VERSION in get_or_make(VersionSpec, r, "julia") || continue
-                    deps[uuid][v] = (h, q)
+                    deps[uuid][v] = q
                     for (p, u) in d
                         u in uuids || push!(uuids, u)
                     end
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Agree, this doesn't make sense in Pkg3 but I thinking of replacing types rather than changing them in-place. There's a bunch of places where old Pkg types are used just because I didn't want to get into replacing them while getting things working initially, but we should probably rethink them a bit. The `Available` and `Requires` types are in that category, so are the old `VersionSet` types (which should be replaced with `VersionSpec`). Of course, that then needs to be threaded through all the version resolution code, which seems like a big headache. Hence the current state of affairs.

KristofferC:
Yeah, I just viewed this as a small step on the way of doing exactly that.

StefanKarpinski:
Is the idea that it's easier to gradually change the Pkg2 types in-place until they're ready to be moved over whole-sale?

carlobaldassi:
I think I could be of help in transitioning the version resolution parts to the new types. I'll just need a few days to get to it. In particular, moving from `VersionSet` to `VersionSpec` seems like a well-defined task, and relatively straightforward. For `Available` and `Require`, I would need to know what you had in mind though.

StefanKarpinski:
@carlobaldassi, would it be possible to just refactor the version resolution code so that it can stand alone and be used by either Pkg2 or Pkg3? To me that's the ideal outcome here since the tight coupling itself seems like an issue. I realize that using the built-in types here is convenient, but I found it a bit iffy that I had to use these Pkg2 types (e.g. [here](https://github.com/JuliaLang/Pkg3.jl/blob/6bf7d0d1df0a2ad8e5b1167c72a979ebc3dd6b50/src/Operations.jl#L118-L148) and [here](https://github.com/JuliaLang/Pkg3.jl/blob/6bf7d0d1df0a2ad8e5b1167c72a979ebc3dd6b50/src/Operations.jl#L165-L166)) to pose what is essentially an indepednent resolution problem completely separate from all the frontend stuff. But perhaps a greater decoupling would already be a natural result of what you had in mind!

KristofferC:
After this PR, `Available` is just a wrapper around a `Dict{String, VersionSet}` so if we swap out `VersionSet` -> `VersionRange`, we could just replace `Available` with `Dict{String, VersionRange}`?

StefanKarpinski:
Or maybe `Dict{UUID, VersionSpec}`? I've converted UUIDs to strings when calling this because strings is what it already accepted but going forward UUIDs are the canonical way to identify packages. `VersionSpec` is the most natural replacement for `VerionSet` – just like `VersionSet` is a basically vector of non-overlapping closed-open version intervals, `VersionSpec` is a vector of non-overlapping version ranges.

KristofferC:
Yes, sorry, I meant `VersionSpec`.

StefanKarpinski:
Re `VersionBound` – feel free to change. I just did what seemed easiest to implement at the time :)

StefanKarpinski:
Superseded by #73.

Issue #40, opened 2017-11-22
Stop auto creating Manifests + Projects in git repos


KristofferC:
After using Pkg3 for a while, this is quite annoying. I think one should explicitly have to instantiate these using some `init` command. If it is a git repo, we go to the base, otherwise just in the cwd.

lobingera:
@KristofferC Could you provide some context? In which git repos?

KristofferC:
In git repos that happens to be in your current working directory.

StefanKarpinski:
I commented in my review of https://github.com/JuliaLang/Pkg3.jl/pull/51#discussion_r153489270:

> Different scenario: what if you have `JULIA_ENV` set to a name like `v1.2.3` and `~/.julia/environments/v1.2.3` does not yet exist. Should this create those files? In that case it's more reasonable to just auto-create the directory and its project and manifest files. But what if someone explicitly does `pkg> init` in those circumstances? Or maybe all environments should be explicitly created...

Perhaps auto-discovery of the current environment is too clever. In Python virtualenv, you have to explicitly set your environment. Should we do something more like that? I do think this design makes it easy to hop back and forth between projects in a single terminal session, which is nice.

But this whole environment discovery step could use a little thought.

PR #51, opened 2017-11-28
WIP force explicit init using `init` for local env

KristofferC:
On its way to solve #40. The idea is to never automatically create a local env, instead you can run `init` which will create that env in the current folder or the base of a git folder.

WIP because it needs tests and more cleanup. cc @fredrikekre

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -400,7 +400,8 @@ function apply_versions(env::EnvCache, pkgs::Vector{PackageSpec})::Vector{UUID}
     end
 
     textwidth = VERSION < v"0.7.0-DEV.1930" ? Base.strwidth : Base.textwidth
-    max_name = maximum(textwidth(names[pkg.uuid]) for pkg in pkgs)
+    widths = [textwidth(names[pkg.uuid]) for pkg in pkgs]
+    max_name = length(widths) == 0 ? 0 : maximum(widths)
 
     for _ in 1:length(pkgs)
         r = take!(results)
@@ -676,12 +677,10 @@ function init(path::String)
         err isa LibGit2.GitError && err.code == LibGit2.Error.ENOTFOUND || rethrow(err)
     end
     path = gitpath == nothing ? path : dirname(dirname(gitpath))
-    try mkpath(path) end
-    for f in ("Project.toml", "Manifest.toml")
-        isfile(joinpath(path, f)) && cmderror("Project already initialized at $path")
-    end
+    mkpath(path)
+    isfile(joinpath(path, "Project.toml")) && cmderror("Environment already initialized at $path")
     touch(joinpath(path, "Project.toml"))
-    info("Initialized project in $path by creating the file Project.toml")
+    info("Initialized environment in $path by creating the file Project.toml")
 end
 
 end # module
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -5,6 +5,7 @@ using Base: LibGit2
 using Pkg3.TerminalMenus
 using Pkg3.Types
 import Pkg3: Pkg2, depots, BinaryProvider, USE_LIBGIT2_FOR_ALL_DOWNLOADS, NUM_CONCURRENT_DOWNLOADS
+import Pkg3: depots, BinaryProvider, USE_LIBGIT2_FOR_ALL_DOWNLOADS, NUM_CONCURRENT_DOWNLOADS
 
 const SlugInt = UInt32 # max p = 4
 const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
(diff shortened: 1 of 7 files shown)


src/REPLMode.jl
```diff
@@ -130,6 +131,7 @@ function do_cmd(repl::Base.REPL.AbstractREPL, input::String)
         cmd == :add    ?    do_add!(env, tokens) :
         cmd == :up     ?     do_up!(env, tokens) :
         cmd == :status ? do_status!(env, tokens) :
+        cmd == :init   ?  do_init!(env, tokens) :
```
StefanKarpinski: text alignment

src/REPLMode.jl
```diff
@@ -167,6 +169,8 @@ const help = Base.Markdown.parse("""
     `rm`: remove packages from project or manifest
 
     `up`: update packages in manifest
+
+    `init` initializes a project in the current, or git base, directory
```
StefanKarpinski: Different scenario: what if you have `JULIA_ENV` set to a name like `v1.2.3` and `~/.julia/environments/v1.2.3` does not yet exist. Should this create those files? In that case it's more reasonable to just auto-create the directory and its project and manifest files. But what if someone explicitly does `pkg> init` in those circumstances? Or maybe all environments should be explicitly created...

KristofferC: If you have an explicit `JULIA_ENV` (that is not `.`) then that will be automatically initialized on any pkg command. This PR should only change the behavior of creating new environments in the current working directory, opting to require an explicit initialization in those cases:

```
shell> ls .git
HEAD        config      description hooks       info        objects     refs

julia> import Pkg3

julia> ENV["JULIA_ENV"] = "v1.2.3"
"v1.2.3"

pkg> st
INFO: Status "~/.julia/environments/v1.2.3/Project.toml"

julia> delete!(ENV, "JULIA_ENV");

pkg> st
INFO: Status "~/.julia/environments/v0.6/Project.toml"

julia> ENV["JULIA_ENV"] = "."
"."

pkg> st
Could not find local environment in /Users/kristoffer/Documents/gitrepo, do you want to create it?
 > yes
   no
INFO: Initialized project in /Users/kristoffer/Documents/gitrepo by creating the file Project.toml
INFO: Status "Project.toml"
```

StefanKarpinski: I assume since you're making the change that this approach "feels right"?

KristofferC:
Another alternative is to force the use of `ENV["JULIA_ENV"] = "."` to use the cwd as an environment. And when that variable is set, automatically create the Project file for this environment when running a pkg command.

StefanKarpinski:
Hmm. Tough call but I think explicit initiation may be better. Will have to consider for a little.

fredrikekre:
I prefer an `init` call. It is a bit iffy to mess with environment variables IMO.

StefanKarpinski:
Agree, although I do think that if the environment variable is set to `ENV["JULIA_ENV"] ="."` or passed via an explicit `--env=.` argument *and* you're in a git repo where the project and manifest files don't exist, it would probably make sense to prompt the user if they want to create them since there the alternative is just to fail.

KristofferC:
Added some tests, this should be ok to review.

src/Operations.jl
```diff
@@ -666,5 +667,22 @@ function test(env::EnvCache, pkgs::Vector{PackageSpec}; coverage=false)
...
+
+function init(path::String)
+    gitpath = nothing
+    try
+        gitpath = git_discover(path, ceiling = homedir())
+    catch err
+        err isa LibGit2.GitError && err.code == LibGit2.Error.ENOTFOUND || rethrow(err)
+    end
+    path = gitpath == nothing ? path : dirname(dirname(gitpath))
+    try mkpath(path) end
```
StefanKarpinski: Seems better to us `ispath(path) || mkpath(path)`. I know this isn't performance sensitive code but avoiding try/catch wherever possible is still better.

src/REPLMode.jl
```diff
@@ -445,6 +454,13 @@ function do_gc!(env::EnvCache, tokens::Vector{Tuple{Symbol,Vararg{Any}}})
     Pkg3.API.gc(env)
 end
 
+function do_init!(tokens::Vector{Tuple{Symbol,Vararg{Any}}})
+    if !isempty(tokens)
+        cmderror("`init` does currently not take any arguments")
```
StefanKarpinski: I suppose it could take a path, but that's a feature and can be added later.

src/Operations.jl
```diff
@@ -666,5 +667,22 @@ function test(env::EnvCache, pkgs::Vector{PackageSpec}; coverage=false)
...
+    for f in ("Project.toml", "Manifest.toml")
+        isfile(joinpath(path, f)) && cmderror("Project already initialized at $path")
+    end
+    touch(joinpath(path, "Project.toml"))
```
StefanKarpinski: It's a bit strange to create an empty file here. I wonder if we shouldn't prompt to add some fields with good initial guesses. For example:
```
project = basename(path)
author = "$(readchomp(`git config user.name`)) <$(readchomp(`git config user.email`))>"
```
We could prompt and let the user edit or skip but at least they put something useful in the project file. There's also probably a LibGit2 way to do that instead of shelling out, but you get my point.

StefanKarpinski:
Looks like you deleted the Project.toml file here. Was that accidental?

KristofferC:
Wow, yes. I'll put it right back.

KristofferC:
It's back.

PR #53, opened 2017-11-29
look in LOAD_PATH for packages

fredrikekre:


KristofferC:
@StefanKarpinski could you explain a bit more about emptying the LOAD_PATH and what relation that has to precompilation (i.e. was there some bootstrapping problems otherwise)?

StefanKarpinski:
Open question: do we keep `LOAD_PATH` in Julia 1.0? Last I discussed it with @JeffBezanson, he was against having multiple code loading mechanisms.

KristofferC:
I guess doing `pkg> add /Users/kristoffer/juliapkgs/LocalPkg` isn't too bad? However, it is kind of convenient right now that you can have a folder where you can dump packages and then have them be loadable.

fredrikekre:
Perhaps use `add` for everything:
```
pkg> add Package                        # from the registry
pkg> add [Package] "path/to/Package.jl" # from some local folder (what we currently use LOAD_PATH for)
pkg> add [Package] "url/to/Package.jl"  # from url, i.e. current Pkg2.clone() behaviour
```

KristofferC:
There is a beauty to that I must say

stevengj:
Suppose that a sysadmin or a distro wants to have a `julia` package and wants to have its own system-wide package tree for distro-managed packages, e.g. in `/usr/lib/julia`.  Right now, it can do this by modifying `LOAD_PATH`, and users can still install their own versions of packages in `~/.julia` and it will override the system-wide packages.

How would that work in Pkg3 without `LOAD_PATH`?

StefanKarpinski:
There's a `Pkg3.DEPOTS` variable which lists locations where the package manager looks for entities (registries, environments and installed packages/libraries). The set of known entities is the union across all of these depots. Only the first one will be treated as writeable by the package manager – this will typically be `~/.julia` and accordingly the package manager will install package versions in `~/.julia/packages` and create named environments in `~/.julia/environments`. A standard configuration will be to have three `DEPOTS` entries:

- **user:** the writeable one, typically `~/.julia`
- **system:** read-only, managed by the system administrator, e.g. `/usr/lib/julia/depot`
- **standard:** read-only, standard library versions specific to your `julia` version.

The system administrator can update registries and install new versions of packages whenever they want *and* leave the old ones in place until no one needs them anymore. This arrangement minimizes disruption for users who don't want to upgrade packages, but makes new version conveniently available for users who want them. If the sysadmin does happen to delete a version of some package that you were using, all you need to do is ask Pkg3 to "hydrate" your project and it will install an identical copy in your user depot and you can carry on about your business.

Note that the "standard" depot ships with your `julia` version and is never touched after that so the standard library is always available if you want it. Neither the user nor the sysadmin is expected to touch it. Either one, however, can install newer (or older) versions of standard library packages, which projects can opt to use instead of the standard versions that the `julia` that's running shipped with. That balances the "standard library" concept with upgradeability.

Hopefully this makes sense to people. (I wrote something similar up [in the Pkg3 Julep](https://github.com/JuliaLang/Juleps/blob/master/Pkg3.md#depots).) Unless I'm missing something, it seems like this scheme addresses the goals that `LOAD_PATH` tries to address better than `LOAD_PATH` does, so it's tempting to just get rid of `LOAD_PATH` altogether.

vdayanand:
Could we have multiple user depots? Maybe a also mechanism to override the default location

StefanKarpinski:
You can have as many depots as you want just by changing the contents of the `DEPOTS` variable, which will be controllable by a `JULIA_DEPOTS` environment variable, but I'm not sure it would mean to have multiple user depots: when installing a new package version, where do you install it? The user depot is the one you install things in and I'm not clear on how you would decide where to install things if there was more than one user-writeable depot. But it's possible that some sensible way of doing this exists. Maybe install a package in the depot where the first writeable registry in `DEPOTS` that knows about it is? But what if that depot isn't writeable?

tpapp:
I currently do the following: my `~/.juliarc.jl` has 
```julia
const LOCAL_PACKAGES = expanduser("~/src/julia-local-packages/")
push!(LOAD_PATH, LOCAL_PACKAGES)

const LOCAL_PROJECTS = expanduser("~/research/julia-projects/")
push!(LOAD_PATH, LOCAL_PROJECTS)
```
Some of these are full-fledged packages, some of them (especially in `LOCAL_PROJECTS`) are just `.jl` files defining a module, no package structure. I would not want the package manager, either `Pkg2` or `Pkg3`, to add or install anything here, but the following two features are useful:

1. `using` and `import` find stuff here,
2. if there happens to be a package at these locations that was `REQUIRE`d, it satisfies the respective dependency.

KristofferC:
I think completely disabling LOAD_PATH will be annoying. Sometimes you just want to put packages wherever you want and have them be loadable without doing anything about the package manager.

However, we could easily have a flag that we can disable when running e.g. tests to prevent stuff in LOAD_PATH to intervene with testing.

KristofferC:
Don't think this is applicable anymore.

PR #54, opened 2017-11-30
API for status

vdayanand:
```
julia> Pkg3.API.status()
INFO: Status "~/.julia/environments/v0.6/Project.toml"
Dict{String,VersionNumber} with 1 entry:
  "IJulia" => v"1.6.2"

julia> Pkg3.API.status(:project)
INFO: Status "~/.julia/environments/v0.6/Project.toml"
Dict{String,VersionNumber} with 1 entry:
  "IJulia" => v"1.6.2"

julia> Pkg3.API.status(:manifest)
INFO: Status "~/.julia/environments/v0.6/Manifest.toml"
Dict{String,VersionNumber} with 16 entries:
  "Libz"            => v"0.2.4"
  "LibExpat"        => v"0.3.1"
  "WinRPM"          => v"0.3.0"
  "Conda"           => v"0.7.0"
  "SHA"             => v"0.5.2"
  "ZMQ"             => v"0.5.1"
  "LibCURL"         => v"0.2.2"
  "JSON"            => v"0.16.1"
  "Compat"          => v"0.37.0"
  "IJulia"          => v"1.6.2"
  "BufferedStreams" => v"0.3.3"
  "MbedTLS"         => v"0.5.1"
  "HTTPClient"      => v"0.2.1"
  "BinDeps"         => v"0.6.0"
  "Homebrew"        => v"0.6.1"
  "URIParser"       => v"0.2.0"

```


KristofferC:
When you use `status` in "API mode" don't you want to know everything about your status, and not just the diff, i.e. shouldn't `status` almost return exactly what reading the `Manifest.toml` file returns?

Also, the conversion seems to be quite lossy, what's the reason for only extracting the name of the package and converting the `.vers` field to a string?

vdayanand:
> When you use status in "API mode" don't you want to know everything about your status, and not just the diff, i.e. shouldn't status almost return exactly what reading the Manifest.toml file returns?

Ive tried using the same method used by status in command mode.

> Also, the conversion seems to be quite lossy, what's the reason for only extracting the name of the package and converting the .vers field to a string

I can fix this, I was trying to emulate Pkg.installed() which is used during JuliaPro build.

vdayanand:
I could change the API name to "installed", which will work for me as of now, with status of entire Manifest file.

KristofferC:
I think I was wrong with that `st` only gives the diff, it gives you all packages in the project and if you call it with `st --manifest` you see all "installed" packages (sorry, still new to this code).

StefanKarpinski:
Yes, the default mode gives everything in project with diff indicators for changes and diff-only changes to the manifest for indirect dependencies (so the project changes aren't repeated). The project mode shows the entire project contents with diff indicators and the manifest mode shows the entire manifest contents with diff indicators. I played around with different options and this seemed like the best compromise, although I admit it's a little ad hoc.

StefanKarpinski:
What do you guys need from the API? This is mostly an interactive function – it only prints info to the terminal. What kind of information about the status of the project and manifest do you need?

vdayanand:
I use this function to precompile all the installed packages. Maybe more work needs to be done, by making it more informative.

StefanKarpinski:
Seems like it would make more sense to have a way to request compilation of all packages in an environment, no?

vdayanand:
Yeah, that would work fine for me.

KristofferC:
This seems to more correspond to the Pkg2's `installed()`?. The Pkg2 version of `status` is AFAIU only a pretty printing function. So perhaps this should be called `installed` as well?

vdayanand:
Implementation is almost same as that of `status` command, do we need to change that as well?

KristofferC:
I think we should do like this:

- add a method `Pkg3.status()` that prints the same that `pkg> status` does. This is just a method to get a pretty printout (just like `Pkg.status()` does today)
- add a method `Pkg.installed(::Symbol)` that does similar to what this PR does. This is if you want a returned value that you can interact with. I think this should default to `:manifest`.

This should mirror Pkg2, and I think it makes sense.

StefanKarpinski:
Keeping the API as Pkg2-compatible as possible for now makes sense to me. We can always evolve it in the future since it's not part of Base.

src/API.jl
```diff
@@ -85,9 +85,9 @@ function convert(::Type{Dict{String, VersionNumber}}, diffs::Union{Array{DiffEnt
     return version_status
 end
 
-status() = status(:combined)
+installed() = installed(:manifest)
 
-function status(mode::Symbol)::Dict{String, VersionNumber}
+function installed(mode::Symbol)::Dict{String, VersionNumber}
```
StefanKarpinski: You could just add a default argument here: `mode::Symbol=:manifest`.

src/API.jl
```diff
@@ -126,6 +128,7 @@ function test(env::EnvCache, pkgs::Vector{PackageSpec}; coverage=false, preview=
     Pkg3.Operations.test(env, pkgs; coverage=coverage)
 end
 
+
```
StefanKarpinski: Extra blank line added.

src/Display.jl
```diff
@@ -16,36 +16,42 @@ const colors = Dict(
...
+    diff = nothing
+    
     if env.git != nothing
         git_path = LibGit2.path(env.git)
         project_path = relpath(env.project_file, git_path)
         manifest_path = relpath(env.manifest_file, git_path)
         project₀ = read_project(git_file_stream(env.git, "HEAD:$project_path", fakeit=true))
         manifest₀ = read_manifest(git_file_stream(env.git, "HEAD:$manifest_path", fakeit=true))
-    end
+    end
```
StefanKarpinski: Trailing whitespace.

PR #56, opened 2017-11-30
implement cloning (and the inverse, freeing)

KristofferC:
This is an attempt to implement `clone`. Whether this should just be called `add` with an url can easily be changed later.

Summarizing this does the following:

* The package is cloned at the master branch.
* When cloning a package, the path where the package is cloned gets written to the Manifest. We no longer write the hash because the package can be changed arbitrarily.
* If we find the name of the package in the registry, we set the version to the maximum registered.
* If the package is not registered, we give it a new random UUID, set the version to "0.0".
* For the resolver, we consider any package with a `path` to be fixed. We force the resolver to use exactly the current version of the package. Furthermore, the dependencies for a checked out package is looked up through the REQUIRE file. @carlobaldassi does that make sense? It is a bit different from how Base does it.
* `free` removes the path and we go back to the registered package.
* If the path exist where the package is supposed to be cloned to AND it looks like a package, we just point to that path, we don't upgrade the package at all. The idea is that we should never touch a package that is being developed.
* The `.travis` tests are now completely Pkg2 free. ~~To do that, I had to internalize SHA.~
* The printing has been changed to account for checked out packages: https://asciinema.org/a/zv1ahmVWdiTmrHFxikKP8o861
* I sort of accidentally changed to using the hooking system for loading packages into 0.6. I can remove that.

Some thoughts:

* The moment when the `.path` field is being assigned is pretty weird. Right now it is done when looking for dependencies for cloned packages.
* How should the UUID be generated for non registered packages. Right now, I just did generate a random one.
* The location where packages are cloned has a default and one can use an ENV variable, but a path option or something should probably be available.
* I'm using the string tokenization know where it requires quote delimited paths. This is for simplicity because I didn't wanna deal too much with the tokenizer in this PR.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -116,7 +116,7 @@ load_package_data(f::Base.Callable, path::String, version::VersionNumber) =
     get(load_package_data(f, path, [version]), version, nothing)
 
 function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec})
-    deps = Dict{UUID,Dict{VersionNumber,Tuple{SHA1,Dict{UUID,VersionSpec}}}}()
+    deps = Dict{UUID,Dict{VersionNumber,Dict{UUID,VersionSpec}}}()
     uuids = [pkg.uuid for pkg in pkgs]
     seen = UUID[]
     while true
@@ -135,7 +135,7 @@ function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec})
                     r = get_or_make(Dict{String,VersionSpec}, compatibility, v)
                     q = Dict(u => get_or_make(VersionSpec, r, p) for (p, u) in d)
                     # VERSION in get_or_make(VersionSpec, r, "julia") || continue
-                    deps[uuid][v] = (h, q)
+                    deps[uuid][v] = q
                     for (p, u) in d
                         u in uuids || push!(uuids, u)
                     end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -153,38 +153,26 @@ function collect_fixed!(env, pkgs, uuids)
     fix_deps_map = Dict{UUID, Vector{PackageSpec}}()
     uuid_to_pkg = Dict{UUID, PackageSpec}()
     for pkg in pkgs
-        path    = nothing
-        version = nothing
-        has_path(pkg) && (path = pkg.path)
-        has_version(pkg) && (version = pkg.version)
-        info = manifest_info(env, pkg.uuid)
-        version == nothing && info != nothing && haskey(info, "version") && (version = VersionNumber(info["version"]))
-        path == nothing && info != nothing && haskey(info, "path") && (path = info["path"])
-        if path != nothing # This is a fixed package
-            # Package is actually being freed, so it is not fixed
-            pkg.beingfreed && continue
-            # A package with a path should have a version
-            @assert version != nothing
-            pkg.path = path
-            pkg.version = version
+        !has_path(pkg) && continue
+        # A package with a path should have a version
+        @assert pkg.version != nothing
 
-            uuid_to_pkg[pkg.uuid] = pkg
-            # Load the dependencies if this package has a REQUIRE
-            reqfile = joinpath(path, "REQUIRE")
-            fix_deps_map[pkg.uuid] = valtype(fix_deps_map)()
-            !isfile(reqfile) && continue
-            for r in filter!(r->r isa Pkg2.Reqs.Requirement, Pkg2.Reqs.read(reqfile))
-                # @assert length(r.versions.intervals) == 1
-                pkg_name, version = r.package, r.versions.intervals[1]
-                # TODO: Check VERSION
-                pkg_name == "julia" && continue
-                # Convert to Pkg3 data types
-                vspec = VersionSpec([VersionRange(VersionBound(version.lower),
-                                                  VersionBound(version.upper))])
-                deppkg = PackageSpec(pkg_name, vspec)
-                push!(fix_deps_map[pkg.uuid], deppkg)
-                push!(fix_deps, deppkg)
-            end
+        uuid_to_pkg[pkg.uuid] = pkg
+        # Load the dependencies if this package has a REQUIRE
+        reqfile = joinpath(pkg.path, "REQUIRE")
+        fix_deps_map[pkg.uuid] = valtype(fix_deps_map)()
+        !isfile(reqfile) && continue
+        for r in filter!(r->r isa Pkg2.Reqs.Requirement, Pkg2.Reqs.read(reqfile))
+            # @assert length(r.versions.intervals) == 1
+            pkg_name, version = r.package, r.versions.intervals[1]
+            # TODO: Check VERSION
+            pkg_name == "julia" && continue
+            # Convert to Pkg3 data types
+            vspec = VersionSpec([VersionRange(VersionBound(version.lower),
+                                              VersionBound(version.upper))])
+            deppkg = PackageSpec(pkg_name, vspec)
+            push!(fix_deps_map[pkg.uuid], deppkg)
+            push!(fix_deps, deppkg)
         end
     end
 
@@ -221,6 +209,7 @@ function resolve_versions!(env::EnvCache, pkgs::Vector{PackageSpec})::Dict{UUID,
         ver = VersionNumber(info["version"])
         push!(pkgs, PackageSpec(name, uuid, ver))
     end
+    path_resolve!(env, pkgs)
 
     # Collect all fixed packages (have a path) and their dependencies
     fixed = collect_fixed!(env, pkgs, uuids)
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -2,9 +2,8 @@ module Operations
 
 using Base.Random: UUID
 using Base: LibGit2
-using Pkg3.TerminalMenus
-using Pkg3.Types
-import Pkg3: Pkg2, depots, BinaryProvider, USE_LIBGIT2_FOR_ALL_DOWNLOADS, NUM_CONCURRENT_DOWNLOADS
+using ..Pkg3: TerminalMenus, Types
+import ..Pkg3: Pkg2, depots, equalto, BinaryProvider, USE_LIBGIT2_FOR_ALL_DOWNLOADS, NUM_CONCURRENT_DOWNLOADS
 
 const SlugInt = UInt32 # max p = 4
 const chars = "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789"
@@ -115,7 +114,7 @@ end
 load_package_data(f::Base.Callable, path::String, version::VersionNumber) =
     get(load_package_data(f, path, [version]), version, nothing)
 
-function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec})
+function deps_graph(env::EnvCache, pkgs::Vector{PackageSpec}, reqs)
     deps = Dict{UUID,Dict{VersionNumber,Dict{UUID,VersionSpec}}}()
     uuids = [pkg.uuid for pkg in pkgs]
     seen = UUID[]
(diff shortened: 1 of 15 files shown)


carlobaldassi:
> For the resolver, we consider any package with a path to be fixed. We force the resolver to use exactly the current version of the package. Furthermore, the dependencies for a checked out package is looked up through the REQUIRE file. @carlobaldassi does that make sense? It is a bit different from how Base does it.

If I understand correctly, this is how dirty packages are treated in Base, whereas Base performs a git pull for non-dirty repos in master, right? Anyway it certainly makes sense to me.

KristofferC:
Thanks for your comment. The difference I was thinking about seems to be with the call to 

https://github.com/JuliaLang/julia/blob/19921aa00468edf11f3d61f8c9b8d639ec71cd23/base/pkg/entry.jl#L478

In `Query.requirements` it seems that the fixed packages are actually removed from the `reqs` variable, and instead, their dependencies are injected into `reqs`. In the code here I have no call to `Query.requirements` nor `Query.check_requirements`.

carlobaldassi:
Ah, I see what you mean. Well, `Query.requirements` performs 2 tasks right now.

One is the one you mentioned, it removes the fixed packages from resolution and instead adds their requirements as constraints for the resolver. This is helpful for simplifying the graph, and also detects impossible requirements right away (a fixed package requiring a non-exisiting package produces an error, conflicting requirements produce warnings). I'm not sure what we want to do about the latter point in relation to cloned packages, but it seems like a sensible behavior? As for the graph simplification, it's not crucial, but I think that the resolution process would become stricter in this way. Consider this: the way things work now in Base, if you have two fixed packages A and B in which A requires a version of B which isn't the one provided, you will get a warning; they will be removed from dependency resolution and the rest proceeds normally. If you pass them along for resolution though, it would detect an unsatisfiable requirement and bail.

The other operation is that it starts building a backtrace for the pre-solver, classifying some packages as `:fixed` and some others as `:required`. In contrast, if I understand correctly, `Pkg3.Operations.resolve_versions!` classifies everything as `:requred` (it skips the call to `Query.requirements` and does this in `Query.prune_dependencies` instead). The only problem with this would be in the reported backtrace in case of unsatisfiable conditions, which would be worded in a slightly misleading way. This issue can only be solved by keeping track of what was required and what was cloned in some way, at least until `prune_dependencies` is invoked.

KristofferC:
Thank you for your detailed response. I first tried to imitate the way Base did it but the code became quite a bit more convoluted. New requirements popping up didn't really fit well with the current code. However, if it gives better error messages it is probably worth doing sometime but perhaps this will work for now.

src/Operations.jl
```diff
@@ -115,34 +114,113 @@ end
...
-                compatibility = load_package_data(VersionSpec, joinpath(path, "compatibility.toml"), versions)
-                for (v, h) in version_info
-                    d = get_or_make(Dict{String,UUID}, dependencies, v)
-                    r = get_or_make(Dict{String,VersionSpec}, compatibility, v)
-                    q = Dict(u => get_or_make(VersionSpec, r, p) for (p, u) in d)
-                    # VERSION in get_or_make(VersionSpec, r, "julia") || continue
-                    deps[uuid][v] = (h, q)
-                    for (p, u) in d
-                        u in uuids || push!(uuids, u)
+    for i in 1:2
```
KristofferC: This is a bit ugly. I did it like this because I want `collect_fixed` to overwrite the stuff from the dependency resolver above, but `collect_fixed` can itself find new dependencies so we need to run this loop again to fully resolve. Perhaps this can be done in a better way though.

src/Operations.jl
```diff
@@ -217,7 +295,7 @@ function version_data(env::EnvCache, pkgs::Vector{PackageSpec})
                 end
             end
         end
-        @assert haskey(hashes, uuid)
+        # @assert haskey(hashes, uuid)
```
KristofferC: Unregistered packages triggers this assert.

src/Operations.jl
```diff
@@ -325,13 +403,41 @@ function update_manifest(env::EnvCache, uuid::UUID, name::String, hash::SHA1, ve
...
+    if !isempty(path)
+        reqfile = joinpath(path, "REQUIRE")
+
+        !isfile(reqfile) && return
+        deps = Dict{String, String}()
+        pkgs = PackageSpec[]
+        for r in filter!(r->r isa Pkg2.Reqs.Requirement, Pkg2.Reqs.read(reqfile))
+            push!(pkgs, PackageSpec(r.package))
+        end
+        # TODO: Get rid of this one?
```
KristofferC: This is quite inefficient. When everything has been moved to Pkg3, the UUID would be available from the `Dependencies.toml` or whatever it is called, and this look up wouldn't be needed.

src/REPLMode.jl
```diff
@@ -1,9 +1,7 @@
 module REPLMode
 
-import Pkg3
-using Pkg3.Types
```
KristofferC: TODO: Repl help for clone and free

StefanKarpinski:
> How should the UUID be generated for non registered packages.

They should be generated randomly. The current mechanism is based on the name, which means that there would be a UUID collision any time there's a name collision, which defeats the purpose of using UUIDs in the first place. So we should not be externally assigning UUIDs. Before we switch over to Pkg3, we have to make PRs to all registered packages giving them UUIDs – those can be done by name since they're *currently* unique, but after that, generating a package should generate a random UUID for it.

There are two things we could do here:

1. Require packages that we clone to have UUIDs in their `Project.toml` file;

2. Don't require cloned packages to have UUIDs. This would require some adjusting of the code but isn't fundamentally impossible. We could allow top-level dependencies on UUID-less cloned packages by just sticking a path in the `[deps]` stanza of `Project.toml` instead of a UUID.

StefanKarpinski:
I'll have to review this one carefully. Won't have any time for that until Monday though.

KristofferC:
I think I should try refactor this to use the way that Base does it. Good error messages and faster failures sounds like good things to keep. Most of the code should be fairly similar though so reviewing would this is still useful.

carlobaldassi:
Maybe wait until I port the changes from https://github.com/JuliaLang/julia/pull/24904 ? I have refactored a couple of functions to split the behavior of `Query.requirements` in light of this PR, it should make your job easier here.

KristofferC:
Yes, that is a good idea. Thank you helping out!

KristofferC:
I rebased this on #63 and changed the method used to the one Base does.

KristofferC:
Hmm, @carlobaldassi, I sometimes get this assertion error. But it is very "random" in the sense if I rerun the command I do not get it anymore. Any ideas?

```
pkg> clone "https://github.com/KristofferC/MMA.jl"
INFO: Path /Users/kristoffer/.julia/dev/MMA exists and looks like a package, using that.
ERROR: AssertionError: !(isempty(allowedp))
Stacktrace:
 [1] filter_versions(::Dict{String,Pkg3.Pkg2.Types.VersionSet}, ::Dict{String,Dict{VersionNumber,Pkg3.Pkg2.Types.Available}}, ::Dict{AbstractString,Pkg3.Pkg2.Types.ResolveBacktraceItem}, ::Dict{String,String}) at /Users/kristoffer/.julia/v0.6/Pkg3/src/Pkg2/query.jl:207
 [2] prune_versions(::Dict{String,Pkg3.Pkg2.Types.VersionSet}, ::Dict{String,Dict{VersionNumber,Pkg3.Pkg2.Types.Available}}, ::Dict{AbstractString,Pkg3.Pkg2.Types.ResolveBacktraceItem}, ::Dict{String,String}) at /Users/kristoffer/.julia/v0.6/Pkg3/src/Pkg2/query.jl:290
 [3] prune_dependencies(::Dict{String,Pkg3.Pkg2.Types.VersionSet}, ::Dict{String,Dict{VersionNumber,Pkg3.Pkg2.Types.Available}}, ::Dict{String,String}, ::Dict{AbstractString,Pkg3.Pkg2.Types.ResolveBacktraceItem}) at /Users/kristoffer/.julia/v0.6/Pkg3/src/Pkg2/query.jl:503
 [4] resolve_versions!(::Pkg3.Types.EnvCache, ::Array{Pkg3.Types.PackageSpec,1}) at /Users/kristoffer/.julia/v0.6/Pkg3/src/Operations.jl:244
...

pkg> clone "https://github.com/KristofferC/MMA.jl"
INFO: Path /Users/kristoffer/.julia/dev/MMA exists and looks like a package, using that.
INFO: Updating "Project.toml"
 [fe960c49] + MMA [/Users/kristoffer/.julia/dev/MMA]
INFO: Updating "Manifest.toml"
 [9e28174c] + BinDeps v0.6.0
 [49dc2e85] + Calculus v0.2.2
 [bbf7d656] + CommonSubexpressions v0.0.1
 [34da2185] ↑ Compat v0.31.0 ⇒ v0.37.0
...
```

KristofferC:
Perhaps this fixed it...

carlobaldassi:
I just checked with this branch, without you last commit. I could reproduce the assertion error only once. When I tried the exact same sequence of operations again (starting from scratch both times) it worked. Which is very weird, the code is deterministic, or supposedly so.

(By "starting from scratch" I mean that I had deleted "packages", "registries" and "environments" from "~/.julia". Is there anything else?)

Anyway, if you look at the code where the assertion is, the only way this could happen is if `deps[p]` is empty, which would mean that package `p` has no versions at all. In turn, `deps` is created trough `Operations.deps_graph`, which can produce an empty `deps[uuid]` if `version_info = load_versions(path)` comes up empty. Is it possible that there is some (occasional?) problem with `load_versions`, i.e. that it sometimes reads an empty `toml` dict from "versions.toml"?

KristofferC:
Hmm, if an unregistered package is cloned then it will be added to the list of packages `pkgs`. And then we call `load_versions` that would be empty and we would add that package with no dependencies to the dependency graph? My last commit removes fixed packages so that dependencies for these are never looked up anymore (because we already do that in `collect_fixed!`).

That could be some sort of explanation but I have no idea why it is not deterministic.

carlobaldassi:
I found that the non-deterministic behavior happens in `Operations.collect_fixed!`, which occasionally returns an empty dict. This in turn makes `Query.propagate_fixed!` return `reqs` unaltered, which allows a required package with no versions to go through and leads to the assertion error.

Note: I manage to trigger the behavior by just using the following commands:

```
clone "https://github.com/KristofferC/MMA.jl"
rm MMA
add MMA
```

If I do this enough times, at some point the cloning fails.

carlobaldassi:
It's the `pkg.beingfreed` line. Or at least it is one possible failure mode I could observe...

KristofferC:
Thank you! I will look into it more carefully tomorrow. With this observation I should be able to figure it out..

src/Types.jl
```diff
@@ -171,18 +172,30 @@ mutable struct PackageSpec
...
-    PackageSpec(name::String, uuid::UUID, version::VersionTypes) =
-        new(name, uuid, version, :project)
-end
-PackageSpec(name::String, uuid::UUID) =
+    path::String
+    url::String
+    beingfreed::Bool # TODO: Get rid of this field
+    PackageSpec(name::AbstractString, uuid::UUID, version::VersionTypes, project::Symbol=:project,
+            path::AbstractString="", url::AbstractString="", beingfreed=false) =
+    new(name, uuid, version, project, path, url)
```
KristofferC: This misses to initialize the `PackageSpec` meaning that `beingfreed` will contain garbage memory giving the bad result.

KristofferC:
What I don't really like about this PR is that the `.path` field for the `PackageSpec` is added in the `collect_fixed!` function. It seems like the current design tries to resolve all that before any function in `API` is called.

I could add something like `resolve_paths!` that looks through the manifest and populates the `.path` field instead.

KristofferC:
I remember now that the reason for doing `path_resolve` in the API calls is because you need to go through all dependencies. I added it in a separate call in `resolve_versions!` though and I think it made the clone quite a bit cleaner.

The `beingfreed` field in the `PackageSpec` is a bit of a wart but not sure how to fix it. Information from `free` need to be propagated somehow to `resolve_versions!` so we know that we shouldn't add a path to the package being freed.

StefanKarpinski:
I swear I'm working on reviewing this, but it's a tough one...

carlobaldassi:
Superseded by #83.

PR #59, opened 2017-12-02
track environment usage

KristofferC:
So we can know what stuff we can gc.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -266,16 +266,15 @@ end
 EnvCache() = EnvCache(get(ENV, "JULIA_ENV", nothing))
 
 function write_env_usage(manifest_file::AbstractString)
-    usage_file = joinpath(depots()[1], ".envusage")
+    logpath = joinpath(depots()[1], "logs")
+    !ispath(logpath) && mkpath(logpath)
+    usage_file = joinpath(logpath, "usage.toml")
     touch(usage_file)
     !isfile(manifest_file) && return
-    usage = TOML.parse(String(read(usage_file)))
-    usage[manifest_file] = now()
-    open(usage_file, "w") do io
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
I was thinking that we would have a `~/.julia/logs` directory and have `env.log` in there. I don't see much advantage to using a hidden file since this is not a directory people will typically be working in and using a non-hidden file means that it will be more obviously discoverable what's happening. We could also have `~/.julia/logs/versions.log` file that records the usage of package versions. That would allow only gc:ing versions when they are no longer referenced by any environment and only after a certain amount of time, which could be useful if you're switching between different versions of a project with different dependency versions.

StefanKarpinski:
Since these files are TOML format, they should be end ing `.toml` – e.g. `~/.julia/logs/usage.toml`. We may want to consider a more log-like format so that the file can just be appended – it would be a shame if after so much careful design to make Pkg3 fast we end up being slowed down by needing to parse, modify and write usage logs. A usage log entry in TOML format could look like this:
```toml
[[usage]]
project = "/path/to/Project.toml"
manifest = "/path/to/Manifest.toml"
timestamp = 1979-05-27T07:32:00Z
```
That way when writing these, one only needs to open the log file in append mode and write a stanza. Another approach might be using log file paths sharded on package UUID or something to avoid having to parse large TOML files.

KristofferC:
Sure, just incrementally adding stuff to the log and then letting the gc command clean it up seems like a good idea. I don't think sharding log files will be necessary.

KristofferC:
What's the purpose of storing the Project file as well?

StefanKarpinski:
> What's the purpose of storing the Project file as well?

Dunno, I guess none – just seemed more complete.

PR #60, opened 2017-12-02
implement gc

KristofferC:
Implements the command `gc` that deletes all packages that are not reachable from an environment (manifest) that has not been within a time period.

TODO:

- Set a period from the REPL
- ~Perhaps we should write how much space was gced.~ Done
- (More) Tests

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -264,15 +264,13 @@ struct EnvCache
 end
 EnvCache() = EnvCache(get(ENV, "JULIA_ENV", nothing))
 
-usage_hash_manifest(manifest_file, t) = string(manifest_file, "_", hash(t))
 function write_env_usage(manifest_file::AbstractString)
     !ispath(logdir()) && mkpath(logdir())
     usage_file = joinpath(logdir(), "usage.toml")
     touch(usage_file)
     !isfile(manifest_file) && return
     open(usage_file, "a") do io
-        t = now()
-        TOML.print(io, Dict(usage_hash_manifest(manifest_file, t) => t))
+        TOML.print(io, Dict(manifest_file => [Dict("time" => now())]))
     end
 end
 
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -2,11 +2,10 @@ module Types
 
 using Base.Random: UUID
 using Pkg3.Pkg2.Types: VersionSet, Available
-using Pkg3.TOML
-using Pkg3.TerminalMenus
+using Pkg3: TOML, TerminalMenus, Dates
 
 import Pkg3
-import Pkg3: depots, iswindows
+import Pkg3: depots, logdir, iswindows
 
 export SHA1, VersionRange, VersionSpec, PackageSpec, UpgradeLevel, EnvCache,
     CommandError, cmderror, has_name, has_uuid, write_env, parse_toml, find_registered!,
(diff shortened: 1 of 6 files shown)


KristofferC:
Updated this PR from the discussion in #59.

* At any command we append the current manifest file (with a hash appended to have a unique key) to the `usage.toml` file.
* The `gc` command goes through all these manifests, only taking into account manifests that have been changed in a time period (currently defaults to six weeks), find all the reachable packages from this manifest, and delete all others packages. 
* Writes a new `usage.toml` file where all old manifests are no longer there.

KristofferC:
Yeah, working around the unique keys was kind of annoying... Rolling our own format here might not be so bad.

StefanKarpinski:
> Rolling our own format here might not be so bad.

That's what I thought at every step of the way when designing Pkg1/2  :trollface:

StefanKarpinski:
I kind of like the business of storing the command that triggered the log entry in the log too. More verbose but also more useful. You can at least always figure out what the heck you've done. Makes sharding more appealing though so that you can have a per-environment command log.

KristofferC:
Array of tables seems like a good idea.

StefanKarpinski:
How about this format:
```toml
[["/Users/stefan/projects/Pkg3.jl/Manifest.toml"]]
timestamp = 2017-12-04T11:21:05.891Z
command = "add JSON"
```
That way the parsed `usage.toml` TOML file is a map from manifest file locations to commands executed with a timestamp and you can still emit entries one at a time.

Aside: I have spend a surprisingly large amount of time and effort in this project thinking about good TOML schemas to represent the data I need to record.

KristofferC:
Updated to that format. I didn't add the `command` field because I wasn't sure how to do it when the API was used and you don't just have a string as a command. Guess you could just put nothing there then. However, more fields can easily be added later and I don't think they are needed for anything right now.

src/Types.jl
```diff
@@ -264,6 +264,16 @@ struct EnvCache
 end
 EnvCache() = EnvCache(get(ENV, "JULIA_ENV", nothing))
 
+function write_env_usage(manifest_file::AbstractString)
+    !ispath(logdir()) && mkpath(logdir())
+    usage_file = joinpath(logdir(), "usage.toml")
+    touch(usage_file)
+    !isfile(manifest_file) && return
+    open(usage_file, "a") do io
+        TOML.print(io, Dict(manifest_file => [Dict("time" => now())]))
```
StefanKarpinski: For this kind of function I would recommend just printing out the content you want instead of bothering with using the TOML print function. I.e. just do something like this:
```jl
println(io, "[[$(tomlkey(manifest_file))]]")
println(io, "time = ", repr(time))
println(io, "command = ", repr(command))
```
It'll be faster, more reliable and the code is easier to understand. It's not like the TOML format is going to radically change how it prints this kind of information.

KristofferC: I updated, hopefully it is clearer. Had to use the `format` to get the `Z` in the end.

StefanKarpinski:
What about putting the logging line in here after the `env` is created:

https://github.com/JuliaLang/Pkg3.jl/blob/d59635b78c250d48d57550cfe424a48d646a0ae9/src/REPLMode.jl#L109-L112

Then you can record `command = $(repr(input))` as well as the timestamp and manifest path. I guess the issue is that this doesn't let you record usage via the API. API usages could either reconstruct a command or pretty print an appropriate API call.

KristofferC:
> API usages could either reconstruct a command or pretty print an appropriate API call.

Before we have a use case for the storing the command history, I am a little bit wary on putting too much effort into pretty printing of the commands because depending on what it is needed for it might need to look differently. Can't we punt on storing the commands until there is a use case for it, and then it can be implemented as part of that.

StefanKarpinski:
Ok, then this is good to go from my perspective.

PR #63, opened 2017-12-04
Solver fixes, tests, etc.

carlobaldassi:
This basically imports the fixes from https://github.com/JuliaLang/julia/pull/24904.

It also restores the `sanity_check` functionality and tests it (I just imported the tests from Base in bulk for now).

Plus, it imports a few functions (refactored from the old `Query.requires` into `init_resolve_backtrace`, `check_fixed` and `propagate_fixed!`) which should help with #56 (cc @KristofferC ).

From here, I will start harmonizing the code by using the new types (and removing/simplifying a lot of stuff, hopefully).


KristofferC:
IIUC, we never form the whole `available` dictionary up front in Pkg3. Instead we just look up things we actually need from the requirements. So perhaps we need to rething alittle bit the methods that take the `avail` dictionary?

carlobaldassi:
Good to know, thanks. But the objects called `avail` in Pkg2 are just the same as `deps`: they are just dependency graph, it doesn't matter where they come from. I've kept all the methods for now mostly to help debugging while things are transitioned.

carlobaldassi:
I had to rebase this before merging - @KristofferC you may want to rebase #56 too I guess.

PR #64, opened 2017-12-05
update registries on up

KristofferC:
Taken mostly from the Base implementation.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -44,7 +44,7 @@ function up(env::EnvCache, pkgs::Vector{PackageSpec};
     preview && previewmode_info()
 
     # Update the registry
-    errors = Tuple{Exception, String}[]
+    errors = Tuple{String, String}[]
     if env.preview[]
         info("Skipping updating registry in preview mode")
     else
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -42,6 +42,55 @@ function up(env::EnvCache, pkgs::Vector{PackageSpec};
...
+                info("Registry at $reg is not a git repo, skipping update")
+            end
+            info("Updating registry at $reg")
+            LibGit2.with(LibGit2.GitRepo, reg) do repo
+                try
+                    branch = "master"
+                    LibGit2.with(LibGit2.head(repo)) do h
+                        if LibGit2.branch(h) != branch
+                            if LibGit2.isdirty(repo)
+                                cmderror("Not on branch $branch")
```
StefanKarpinski: Shouldn't this say something about the repo being dirty instead of just "not on branch"?

KristofferC: yeah

KristofferC:
Yeah, I'll also try to not make this rely on try catch.

src/API.jl
StefanKarpinski: I'm also not sure we need to care about the branch so much – maybe just pull whatever branch we're on and print a message about it not being master if it's not on master?

KristofferC: I think it depends on how tagging packages should work. If we get rid of the current way where a branch of METADATA is checked out, then we could probably get rid of the branch check.

StefanKarpinski: The purpose of that branch was so that we could change default branches if we changed the format of METADATA in an incompatible way. This happened once, caused a bunch of annoyance, I added the explicit branch checkout feature to handle it in the future, and then we never did it again. Now that we're using TOML, which is fairly standard, I would imagine that we can evolve the format smoothly by just being careful to add information to the registry format in a way that will continue to work reasonably on older versions of Pkg3 but allow newer versions to do whatever they're supposed to do with it. So, in short, I think we could get rid of the branch stuff. I'm definitely open to ideas about how this should work.

KristofferC: Regarding tagging, I wasn't talking about the metadata-v2 branch but rather that tagging a package (using PkgDev) currently checks out a branch in the local repo of metadata which you are then supposed to make a PR from.

For tagging in Pkg3, I don't think we should touch the original registry repo, but instead, just duplicate it somewhere and to whatever branching need to be done at that place.

KristofferC:
Updated, made the code quite a lot simpler.

src/API.jl
StefanKarpinski: That's a good point – we can have as many copies of a registry as we need.

PR #66, opened 2017-12-06
Respect test only dependencies

KristofferC:
This PR (built on #56, #51) makes `Pkg3.test` respect test only dependencies in `test/REQUIRE`.
Furthermore, it creates a temporary environment where the requirements for the package being tested are put into. This makes it so that if the package being tested uses Compat but fails to have that in its REQUIRE file, AND you have Compat in your current environment when testing, it will still fail. This avoids the common error of using stuff in your local environment.

The relevant commit is https://github.com/JuliaLang/Pkg3.jl/commit/a446937bc2f6b5005cbd0ee304fb1450c9a6554f.

Since the functionality of `clone` and `init` is used, there are some unnecessary info prints from these functions but we can disable that with a flag.

TODO: tests

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -258,6 +258,28 @@ function clone(env::EnvCache, url::AbstractString; name=nothing, basepath=get(EN
     registry_resolve!(env, [pkg])
     project_resolve!(env, [pkg])
     manifest_resolve!(env, [pkg])
+    if has_uuid(pkg)
+        if haskey(env.uuids, pkg.name) && pkg.uuid in env.uuids[pkg.name]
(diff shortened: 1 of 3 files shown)


helgee:
Just out of curiosity, will the contents of `test/REQUIRE` at some point be merged into `Project.toml`?

KristofferC:
REQUIRE files will go away but exactly how things like OS-specific, test-specific, build-specific dependencies will be specified is not completely decided yet. Anyway, this works now (just need to make https://github.com/JuliaLang/julia/pull/26336 pass CI) so this PR can be closed.

PR #67, opened 2017-12-07
Use import instead of Base.require in require.jl 

vdayanand:
Can we replace `Base.require` with `import` require.jl?, this causes problem if we want to import Pkg3 in juliarc.jl

### DIFF
--- a/src/require.jl
+++ b/src/require.jl
@@ -1,2 +1,2 @@
 push!(empty!(LOAD_PATH), dirname(dirname(@__DIR__)))
-Base.require(:Pkg3)
+import Pkg3


StefanKarpinski:
This was *very* fiddly to get to work in all cases, so I would not make this change unless various people have tried it in various ways. Note that I originally made this change to better support Pkg3 being loaded via `include` which was, IIRC, how you were using it at the time...

vdayanand:
The issue back then was that PyCall build was failing when `include` was used to load Pkg3. I checked and this change doesn't break that. If I am not wrong, when Pkg is replaced with Pkg3, this change wouldn't matter.

KristofferC:
This shouldn't be needed anymore.

PR #68, opened 2017-12-07
exit pkg repl mode with ^C

fredrikekre:
I find `^C` to be a convenient way to exit the pkg repl mode. That is how other repl modes behave too (help, shell, cxx etc). What was the reason for explicitly removing this?

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -474,8 +474,6 @@ function create_mode(repl, main)
     end
 
     mk = REPL.mode_keymap(main)
-    # ^C should not exit prompt
-    delete!(mk, "^C")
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
I forget – I think it was causing some kind of problem. But if it's not anymore, then we can re-enable.

fredrikekre:
From what I have tried it seems to work just fine.

StefanKarpinski:
IIRC, it was exiting the REPL altogether, which was pretty bad. I guess it's not doing that anymore?

fredrikekre:
Nope, works just like the shell and help repl modes.

PR #69, opened 2017-12-07
use global settings struct for some other globals

KristofferC:


src/Pkg3.jl
```diff
@@ -8,12 +8,19 @@ else
...
 end
 
-const DEPOTS = [joinpath(homedir(), ".julia")]
-depots() = DEPOTS
-logdir() = joinpath(DEPOTS[1], "logs")
+@enum LoadErrorChoice LOAD_ERROR_QUERY LOAD_ERROR_INSTALL LOAD_ERROR_ERROR
 
-const USE_LIBGIT2_FOR_ALL_DOWNLOADS = false
-const NUM_CONCURRENT_DOWNLOADS      = 8
+Base.@kwdef mutable struct GlobalSettings
```
rfourquet: I didn't know `Base.@kwdef`, that's awesome! you should have told me I can use it for `Base.REPL.Options` :) Speaking of which, would `Base.REPL.Settings` be a better name for it?
And here... wouldn't `Settings` be a better name than `GlobalSettings`? for the `const` global object, `GLOBAL_SETTINGS` looks like a good name, but could we imagine people doing crazy things and using multiple distinct instances of `GlobalSettings` concurrently, making it more of a local thing? (I really don't know Pkg3's code, so I speak blindly here).

KristofferC: Yeah, perhaps `Settings` is a better name. I don't really think many instances of it will be used though.

StefanKarpinski:
Probably a good idea but was there some particular motivation for this change?

KristofferC:
If we have a settings object, seems to make sense to put the settings there. Also, this means that the options like how many concurrent downloads to use and if one should only use libgit2 for downloads are no longer constants.

StefanKarpinski:
It would be good to put that reasoning in the commit message :)

KristofferC:
The main intention was to put global options into the global struct object which is what the commit message says.

PR #73, opened 2017-12-08
Transition from Pkg2 to Pkg3

carlobaldassi:
This removes all "legacy" stuff from Pkg2, keeping only what's still needed, and merging that into Pkg3.

Highlights:
* Removed `VersionInterval` and `VersionSet`, use `VersionRange` and `VersionSpec` instead (also, `VersionSpec`s are normalized at construction time now).
* Use `UUID` everywhere in the package resolution code, instead of `String`s.
* Removed `Available`, use `Requires = Dict{UUID,VersionSpec}` instead

All of it is essentially "mechanical", the behavior should be unchanged.

It is a straight improvement in basically all respects. One sore point is that I had to assign a `UUID` to "julia" in order to create a corresponding `Requires` item. At the moment, it's just an arbitrary string that I wrote by hand (it's in "Types.jl"). I guess it would be better to filter out the versions earlier instead, such that the compatibility with the current version of julia can simply be assumed in the resolution-related code.

One decision that I have postponed at the moment is whether to switch from the simple alias `DepsGraph = Dict{UUID,Dict{VersionNumber,Requires}}` to a `struct`, so that it could contain the `uuid_to_name` dict too, which seems to make sense to me (I actually did this for `ResolveBacktrace`).

Another minor point is that in order to ease the transition I have allowed constructing `VersionRange`s with strings of the form `"2.0-*"`, i.e. with and open bound, in analogy with the form `"*"` for both open bounds.

Anyway I think these are points that could be addressed rather easily later, there is already a lot of stuff here. For reviewing, it's probably much easier to look at individual commits (each of which is a gradual step in the transition and results in a working code).

This PR supersedes #47. It will also need some coordination with #56 (cc @KristofferC), but they are largely independent; I believe that it would be easier and cleaner to have this first, and rebase #56 on top of it.


src/Types.jl
```diff
@@ -128,16 +263,154 @@ function Base.print(io::IO, s::VersionSpec)
...
+    for (pkg,vers) in B
+        A[pkg] = haskey(A,pkg) ? intersect(A[pkg],vers) : vers
+    end
+    return A
+end
+
+satisfies(pkg::UUID, ver::VersionNumber, reqs::Requires) =
+    !haskey(reqs, pkg) || in(ver, reqs[pkg])
+
+const DepsGraph = Dict{UUID,Dict{VersionNumber,Requires}}
```
KristofferC: Is this typealias worth it? I find it easier to read the code when this is spelled out.

carlobaldassi: It's used in quite a few places (around 20 including argument type declarations).
Mostly, however, I was planning to switch from a typealias to a true struct at some point, to include the `uuid_to_name` bit, since to me it makes sense to lump them together.

StefanKarpinski:
Excellent! Julia's official package UUID is `1222c4b2-2114-5bfd-aeef-88e4692bbb3e`, see:

https://github.com/JuliaLang/Pkg3.jl/blob/d335252ae8ea39b383daeb96217d6d8d3405e822/bin/loadmeta.jl#L92

carlobaldassi:
Thanks @StefanKarpinski, I have moved those definitions into Pkg3 so that we can have `uuid_julia` and use `uuid5` in tests too.
I have also restored a few pieces of `Pkg2` into the `bin` directory since they were used by "loadmeta.jl", so that now "bin/update.jl" works again.

StefanKarpinski:
Great. This is good to merge from my end whenever @KristofferC sees fit.

carlobaldassi:
@KristofferC thanks for merging, but could you please not squash things this big next time? The merge commit message is almost unintelligible, and we basically lost track of the individual changes and their reason.

KristofferC:
Sure, if you want I could just revert the merge and merge this again but without squashing.

carlobaldassi:
Yes, good idea, the history will be a bit messy but I think it's still better that way, thanks!

KristofferC:
Okay, I'll fix it in a few hours (gotta go do something)

StefanKarpinski:
I created a trivial merge commit with this full PR branch as one of its parents, which records the history. It leads to a kind of weird history DAG, but that's ok.

carlobaldassi:
Ah, well, git blame and the github history still only show the whole lump, but the information is at least retrievable now - thanks and apologies for the fuss.

StefanKarpinski:
No worries, it's a good idea to preserve the history.

PR #77, opened 2017-12-13
implement Pkg.dir() giving the path to a package in the manifest

KristofferC:
The same environment can't use two packages with the same name I presume?

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -228,5 +228,28 @@ function init(path = pwd())
     Pkg3.Operations.init(path)
 end
 
+dir() = error("`Pkg.dir()` is discontinued, use `Pkg.dir(\"Package\")` or `Pkg.dir([\"PackageA\", ...])`")
+
+dir(pkg::String) = dir([pkg])[1]
+function dir(pkgs_str::Vector{String})
+    env = EnvCache()
+    pkgs = [PackageSpec(pkg, :manifest) for pkg in pkgs_str]
+    manifest_resolve!(env, pkgs)
+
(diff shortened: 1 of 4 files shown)


KristofferC:
Hmm passes locally...

StefanKarpinski:
It can, but the project can only have one thing by a given name. Something that the project depends on can have a different dependency by that name, however. It might be best to implement this by looking it up in the project file instead. But I'm not really sure since dependencies might need it too.

KristofferC:
But the dependencies for each package in the Manifest is only given by name?

StefanKarpinski:
See the logic here:

https://github.com/JuliaLang/Pkg3.jl/blob/70ca74e21802bbe6bd1d7c4c9a12b9599ea7ffbd/src/Types.jl#L619-L628

The `deps` entry in a manifest stanza is only an array of names if all of those names uniquely identify the package in the manifest. If any of the names has multiple meanings in the environment, then any `deps` map that depends on a "two meaning" package needs to be spelled out as a dict from names to UUIDs. So usually you don't need the mapping, but sometimes you do.

KristofferC:
Alright, we definitely need tests covering these cases. It is also difficult to code against this when there is no example that has the data saved in this format.

StefanKarpinski:
Agree. That's one of the reasons I really need to sit down and write a bunch of tests.

KristofferC:
This is perhaps better done as `Base.locate_package(Base.identify_package(Main, "Pkg"))` but with a nicer API.

PR #79, opened 2017-12-14
Update package URL in README.md

i-apellaniz:
I edited the package URL for cloning. I think it points now to the official package


StefanKarpinski:
Thanks!

PR #83, opened 2017-12-17
rebase of #56 (implement clone/free) onto #81 (graph refactor)

carlobaldassi:
As per the title, this rebases #56 on top of #81, cc @KristofferC, @StefanKarpinski.
I had to change a few things, and this allowed me to catch a few issues on the graph refactoring that I fixed. It seems to work now, or at least tests pass.

One annoying thing is that I had to bring back some pieces of Pkg2 from `bin/` into `src/`, to allow for the parsing of `REQUIRE` files. I suppose that it's going to be needed as long as we support Pkg2-style dependency specifications ~~(that part of the code needs to be fixed BTW)~~.


carlobaldassi:
Update: fixed the parsing of Pkg2 REQUIRE files.

StefanKarpinski:
Thank you for doing this, @carlobaldassi!

carlobaldassi:
I was feeling guilty for the amount of disruptiveness of my PRs :)

The Travis job fails with `ERROR: LoadError: LoadError: ArgumentError: Module SHA not found in current path.`, while AppVeyor succeeds. Does somebody have a clue?

StefanKarpinski:
Unfortunately, I'm not sure what's up with CI on this and @KristofferC is on vacation, so :man_shrugging:

KristofferC:
When I made this PR, SHA was not needed for Pkg3 itself, only for the bin stuff. This PR changes the .travis.yml script to only use Pkg3. We thus never do Pkg.clone and thus not download SHA.

carlobaldassi:
Ok the issue with Travis testing is fixed now. It's rather crude, but then again this won't be a problem once Pkg3 is moved to stdlib I guess.

carlobaldassi:
This is now on top of master, i.e. on top of all the graph refactoring stuff.

KristofferC:
Lovely, thanks a lot!

andyferris:
FYI I made a rebase of this at `ajf/clonerepo2-rebased`.

I found a problem where the tokenizer can't understand git urls like `pkg> clone git@github.com:a/b` (while the `Pkg3.clone("git@github.com:a/b")` works fine).

andyferris:
Using this, this seems neat. 

I'll note a couple things that would help me to use Pkg3 at work (eventually - doesn't have to be this PR):

 1) Use something other than REQUIRE to get the dependencies (it can remain as a "backup option"). Apart from consistency with the rest of the .toml files, I would like to add private dependencies here without breaking Pkg2 compatibility with standard METADATA (this would help me getting a private registry running for Julia v0.6).
 2) Allow one to specify a path inside the git repo as the base path of the package, since Pkg3 can support multiple packages per git repository using trees (this is going to be awesome, but probably not immediately necessary).

KristofferC:
> I found a problem where the tokenizer can't understand git urls..

For now, you need to surround the url with quotes.

> Use something other than REQUIRE to get the dependencies (it can remain as a "backup option").

Definitely, but in order to make this PR right now (and to test with existing package) it needed to handle REQUIRE files.

> since Pkg3 can support multiple packages per git repository using trees

It can, I didn't know this?

src/Operations.jl
```diff
@@ -364,13 +411,41 @@ function update_manifest(env::EnvCache, uuid::UUID, name::String, hash::SHA1, ve
...
+        reqfile = joinpath(path, "REQUIRE")
+
+        !isfile(reqfile) && return
+        deps = Dict{String,String}()
+        pkgs = PackageSpec[]
+        for r in filter!(r->r isa Pkg2.Reqs.Requirement, Pkg2.Reqs.read(reqfile))
+            push!(pkgs, PackageSpec(r.package))
+        end
+        # TODO: Get rid of this one?
+        registry_resolve!(env, pkgs)
```
StefanKarpinski: How do we determine if this should or shouldn't be gotten rid if?

src/REPLMode.jl
```diff
@@ -486,6 +528,50 @@ function do_init!(tokens::Vector{Tuple{Symbol,Vararg{Any}}})
     Pkg3.API.init(pwd())
 end
 
+function do_clone!(env::EnvCache, tokens::Vector{Tuple{Symbol,Vararg{Any}}})
+    isempty(tokens) && cmderror("`clone` take an url to a package to clone")
+    local url
+    basepath = get(ENV, "JULIA_DEV_PATH", default_dev_path())
```
StefanKarpinski: So even though this is a "path" to a place, we use `_PATH` and `_PATH` as a suffix in `LOAD_PATH` and `DEPOT_PATH` to indicate a list of places to look for things, whereas this is a single location to put things, which seems more like `Sys.BINDIR` or `JULIA_PKGDIR` so I think this environment variable should be called `JULIA_DEVDIR` or `JULIA_DEV_DIR`. I favor the former since we're [going to have](https://github.com/JuliaLang/julia/issues/20899) a bunch of variables with names like `SRCDIR` and `LIBDIR` so using the name `DEVDIR` seems more consistent.

StefanKarpinski:
>> since Pkg3 can support multiple packages per git repository using trees
>
> It can, I didn't know this?

It's not yet implemented, but since a version is just a tree and a tree doesn't need to appear at the root of a repository, it could with the right tooling work. Since we don't really have any dev tooling for tagging Pkg3 packages yet, it doesn't support this but it also doesn't not support this – if you arranged for a Pkg3 version entry to map to a subtree in the repo, it would just work. Of course that subtree should look like a normal package, but that's certainly doable.

andyferris:
> since a version is just a tree and a tree doesn't need to appear at the root of a repository, it could with the right tooling work. Since we don't really have any dev tooling for tagging Pkg3 packages yet, it doesn't support this but it also doesn't not support this – if you arranged for a Pkg3 version entry to map to a subtree in the repo, it would just work. Of course that subtree should look like a normal package, but that's certainly doable.

Exactly. AFAICT, if I make my own head-written registry, I can have my own monorepo for all of my company's private packages - that *is* implemented. (After reading some blogposts on Google's monorepo, it seems rather appealing, but in any case it would be good to put some related libraries and/or tools/apps/projects into one repository).

I would be extremely happy if any new tooling regarding creating and editing registries, package templates, and developing packages could respect this possibility - since adding this later might be difficult after architectural decisions are made and lead to avoidable refactoring.

Regarding the functionality in this PR, I guess packages in development mode need a local git repository (cloned in this case) *and* a subdirectory of the repository, which can then be mapped to-and-from a tree hash?

StefanKarpinski:
My thinking is that if a MetaProject.toml file exists we look in that for where the actual projects are and dev tools all respect that.

andyferris:
OK, cool, so this isn't recorded in the registry at all... actually that's quite nice and lets developers move stuff around inside the repository quite easily.

src/Operations.jl
```diff
@@ -198,20 +198,38 @@ function collect_fixed!(env, pkgs, uuids)
...
+
+            uuid_to_pkg[pkg.uuid] = pkg
+            # Load the dependencies if this package has a REQUIRE
+            # TODO: this is still using Pkg2 facilities
+            reqfile = joinpath(path, "REQUIRE")
+            fix_deps_map[pkg.uuid] = valtype(fix_deps_map)()
+            !isfile(reqfile) && continue
+            for r in filter!(r->r isa Pkg2.Reqs.Requirement, Pkg2.Reqs.read(reqfile))
+                # @assert length(r.versions.intervals) == 1
+                pkg_name, version = r.package, r.versions.intervals[1]
```
carlobaldassi: I had fixed this (and the following lines) in https://github.com/JuliaLang/Pkg3.jl/pull/83/commits/bcdfe81c34bde723681f91f3ede2a62ad46329d1

KristofferC: Sorry, must have messed up in rebasing :/

carlobaldassi: It would be preferable to apply the fixes to https://github.com/JuliaLang/Pkg3.jl/pull/83/commits/c728e4bcaa2ae471792482281cb2fd9f14e3d8c1, where the rebase went wrong

KristofferC: alright, I'll try...

KristofferC: Hopefully good now

carlobaldassi: :+1:

src/Display.jl
```diff
@@ -12,7 +12,7 @@ const colors = Dict(
     '↑' => :light_yellow,
     '~' => :light_yellow,
     '↓' => :light_magenta,
-    '?' => :red,
+    '?' => :white,
```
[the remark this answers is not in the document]
KristofferC: There was some reason for this that made red look weird but I don't remember. Will change it back.

src/Operations.jl
KristofferC: The problem is using REQUIRE files that do not have UUIDs so the only way to get the UUIDS is to go to the registry. When we are not using REQUIRE files the UUIDs can just be read from the Manifest file.

StefanKarpinski: This should be harmless – if all the packages are already resolved, it won't hit the registry at all.

KristofferC:
Everything here should now be with similar functionality in master

PR #91, opened 2017-12-28
 Build dependent packages and support varargs in API.build

vdayanand:
Previously ```build``` command was not building dependent packages.
Added support for varargs in `Pkg3.API.build,` as present in `Pkg.build`
```
julia> Pkg3.build("PyCall", "DataFrames")
INFO: Building Conda [b26546424d5f3ac3]...
INFO:  → /home/osboxes/.julia/packages/N8e4/u0lj/deps/build.log
INFO: Building PyCall [2ba54612238a85ee]...
INFO:  → /home/osboxes/.julia/packages/mbAM/vGCc/deps/build.log
INFO: Building SpecialFunctions [a052147e601e7926]...
INFO:  → /home/osboxes/.julia/packages/FTLa/YKDc/deps/build.log

```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -242,6 +242,7 @@ function build(env::EnvCache, pkgs::Vector{PackageSpec})
     end
     manifest_resolve!(env, pkgs)
     ensure_resolved(env, pkgs)
+    Pkg3.Operations.resolve_versions!(env, pkgs, false)
     Pkg3.Operations.build_versions(env, [pkg.uuid for pkg in pkgs])
 end
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -224,6 +224,15 @@ function gc(env::EnvCache=EnvCache(); period = Week(6), preview=env.preview[])
     info("Deleted $(length(paths_to_delete)) package installations", byte_save_str)
 end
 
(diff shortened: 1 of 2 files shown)


KristofferC:
Couldn't

https://github.com/JuliaLang/Pkg3.jl/blob/317e3127c9ee2611fa3ec0d5cccd582ac3de3e5d/src/Operations.jl#L422-L440

just add the dependencies (if they have a build file) to `builds` defined at

https://github.com/JuliaLang/Pkg3.jl/blob/317e3127c9ee2611fa3ec0d5cccd582ac3de3e5d/src/Operations.jl#L445

?

It seems excessive to run the resolver and everything here because there shouldn't be anything to resolve?

vdayanand:
@KristofferC  Thanks for the suggestion.

vdayanand:
fixes code broken by #94

KristofferC:
This looks good to me now. We are not using the resolver nor hitting any of the registry files. Any thoughts @StefanKarpinski

StefanKarpinski:
Seems good to me. Are we expected CI to pass these days or not?

KristofferC:
On 0.6 yes (which this does)

StefanKarpinski:
Merge at will then!

KristofferC:
Although we do not test this command yet... But I think @vdayanand has tested this quite extensively locally.

PR #86, opened 2017-12-22
Resolution log - looking into what the solver is doing, and why

carlobaldassi:
(This is done on top of #83. I guess I could rebase it on top of #81 instead if the review of cloning/freeing needs more time.)

This greatly improves the ability to see what `resolve` does. Examples are at the bottom. Essentially, it logs everything. In case of errors, the log is represented in tree form, starting from the package with unsatisfiable conditions, like before. Otherwise, in case of success, the full log can be seen in different ways - the tree view, a plain view in which events are grouped by package, and a chronological view which shows a journal. This can very well be the basis for a "why" API command if the user wants to understand why some package is not getting updated etc. (or something along those lines). Cf. https://github.com/JuliaLang/julia/issues/13340 (cc @vtjnash).

Note that the cryptic messages about "mysterious", unspecified difficult dependencies is gone. This one:
```
"""
resolve is unable to satisfy package requirements.
  The problem was detected when trying to find a feasible version
  for package $p.
  However, this only means that package $p is involved in an
  unsatisfiable or difficult dependency relation, and the root of
  the problem may be elsewhere.
"""
```
Instead, the solver tries to get to the bottom of the matter. The starting point is still the heuristic, so this can be improved (and it will be), but I have tried *really* hard to find some problematic situations with the packages in the current Uncurated registry and couldn't find any (I decided I'll have to resort to artificially generated nasty problem instances). Incidentally, I have also improved a tiny bit the heuristic solver.

Also, the log can be detached and passed around/displayed/saved to a file etc., independently of the `Graph` it came from.

One thing I'd like to add probably is some filtering functionality to the log display.

Another thing which is kind of annoying is that many log lines are long, and something like https://github.com/carlobaldassi/TextWrap.jl would be required to make them nicer, especially in the tree view. But then it would need to get into the stdlib.

--------------

Example of a resolve log in plain/group format, with test package names `A`, `B`, ..., `F`:

```
INFO: Resolve log:
 Global events:
 ├─propagating constraints
 ├─disabling unreachable nodes
 ├─pruned graph — stats (n. of packages, mean connectivity): before = (7,2.857142857142857) after = (0,NaN)
 ├─computing version equivalence classes
 ├─computed version equivalence classes, stats (total n. of states): before = 0 after = 0
 └─the solver found a feasible configuration
 A [77b1b708] log:
 ├─possible versions are: [1.0.0, 2.0.0, 3.0.0] or uninstalled
[...]

Issue #52, opened 2017-11-28
levels of incompatibility


StefanKarpinski:
Problem: We've had a recent situation where JuliaPro ships with a fixed set of package versions which were compatible at the time it was released, but a subsequent update to METADATA (https://github.com/JuliaLang/METADATA.jl/pull/12106, see discussion in https://github.com/JuliaLang/METADATA.jl/pull/12264) caused that set of versions to become incompatible. Unfortunately, this means that doing almost anything with that set of packages causes the resolver to say that the operation is not allowed because it leads to an incompatible ("infeasible") set of package versions.

Potential solution: In general, it would be good to allow operations that don't make the current situation any worse, and in general to afford users the option to pick their poison in situations where no fully compatible resolution exists. This seems to lead to a notion of "levels of incompatibility" for sets of versions rather than the current black-and-white notion of compatible or not. If you're already at a level of incompatibility, then an operation which doesn't make that level worse should be allowed without warning. If there are no solutions which don't make the situation worse, then it would be ideal to prompt the user with the best options for resolving the situation as well as possible.

PR #94, opened 2018-01-05
Add backtracking to max-sum, improve greedy solver; the solver is now exact

carlobaldassi:
With this change, the solver is now exact, at least for what concerns SAT/UNSAT discrimination, i.e. it should always find a solution if it exists, and fail only when it doesn't, producing a reasonable log trace in that case. (Ok to be perfectly honest the exactness claim relies on a conjecture that I haven't proven yet. But I think it should hold. It certainly does in all the cases that I tested.)

Optimality - finding the best solution among those that exist - is still based on a heuristic, unless the solution is found by the greedy solver, which is quite improved as well anyway - so much so that I had to explicitly add a new generator of nasty test instances just to trick it.

Of course, the problem being NP-hard in general, this means that in principle the solver could now take ages to decide whether a problem is satisfiable. From my (many) experiments, I don't think this will be a problem at all in practice. However, we could add a timer, for example, in case it turns out that I'm being too optimistic.

@StefanKarpinski: With this, I think that something like #52 could be addressed by simply trying first with the full constraints, then if the problem is unsat some second-level constraints could be relaxed, etc.

Also, this is based on top of #86, which is on top of #83, which is on top of #81. Things are starting to pile up... I really need some kind of feedback on #81.

### DIFF
--- a/bin/Pkg2/types.jl
+++ b/bin/Pkg2/types.jl
@@ -5,7 +5,6 @@ module Types
 export VersionInterval, VersionSet
 import Base: show, isempty, in, intersect, union!, union, ==, hash, copy, deepcopy_internal
 
-import Pkg3.equalto
 import Pkg3.iswindows
 
 struct VersionInterval
--- a/src/GraphType.jl
+++ b/src/GraphType.jl
@@ -6,7 +6,8 @@ using ..Types
 import ..Types.uuid_julia
 import Pkg3.equalto
 
-export Graph, ResolveLog, add_reqs!, add_fixed!, simplify_graph!, get_resolve_log, showlog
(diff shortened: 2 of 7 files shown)


KristofferC:
Sorry for being slow with stuff. Holidays and all that :). I'm not qualified at all to review the resolver stuff but if you have tried it out and it looks good then that is enough for me. I hope we can get the PRs from you merged in a timely manner now that the holidays are over.

StefanKarpinski:
Frankly, I don't think that anyone but Carlo is qualified to review this. Perhaps after we've got 1.0 out I can take some time to sit down an understand the resolver, but until then, @carlobaldassi, I think you should merge this as you see fit to do so. It sounds great to me from your description :)

PR #96, opened 2018-01-05
Fix errors when JULIA_ENV ends with .toml

vdayanand:
when JULIA_ENV is set with value which ends with `.toml`, instead of file `Project.toml`, directory Project.toml is created.


StefanKarpinski:
Looks good to me but @KristofferC should really take a look since he wrote this bit.

PR #99, opened 2018-01-08
fix deprecations

KristofferC:
Fixes some of the 0.7 deprecations (enough to make Pkg3 load and run).

However, I am seeing a problem on 0.7 that is not on 0.6. Running e.g. `pkg> up --fixed` on 0.7 with the Pkg3 `Project.toml` I get:

```
ERROR: AssertionError: gmsk1[j1] == (gmsk0[j0])'
Stacktrace:
 [1] check_consistency(::Pkg3.GraphType.Graph) at /Users/kristoffer/.julia/v0.7/Pkg3/src/GraphType.jl:480
 [2] (::getfield(Pkg3.GraphType, Symbol("##Graph#24#41")))(::Bool, ::Type, ::Dict{Base.Random.UUID,Set{VersionNumber}}, ::Dict{Base.Random.UUID,Dict{Pkg3.Types.VersionRange,Dict{String,Base.Random.UUID}}}, ::Dict{Base.Random.UUID,Dict{Pkg3.Types.VersionRange,Dict{String,Pkg3.Types.VersionSpec}}}, ::Dict{Base.Random.UUID,String}, ::Dict{Base.Random.UUID,Pkg3.Types.VersionSpec}, ::Dict{Base.Random.UUID,Pkg3.Types.Fixed}) at /Users/kristoffer/.julia/v0.7/Pkg3/src/GraphType.jl:333
 [3] Type at /Users/kristoffer/.julia/v0.7/Pkg3/src/GraphType.jl:242 [inlined]
 [4] deps_graph(::Pkg3.Types.EnvCache, ::Dict{Base.Random.UUID,String}, ::Dict{Base.Random.UUID,Pkg3.Types.VersionSpec}, ::Dict{Base.Random.UUID,Pkg3.Types.Fixed}) at /Users/kristoffer/.julia/v0.7/Pkg3/src/Operations.jl:191
 [5] resolve_versions!(::Pkg3.Types.EnvCache, ::Array{Pkg3.Types.PackageSpec,1}) at /Users/kristoffer/.julia/v0.7/Pkg3/src/Operations.jl:211
```

This is perhaps related to https://github.com/JuliaLang/Pkg3.jl/issues/45 because the julia version is now enforced again. The error message should probably be improved though @carlobaldassi.

I tried increasing the compatibility of Compat and SHA in the registry manually, but it didn't help.


carlobaldassi:
The problem is that the `'` syntax (eg. `bm'`) on a BitArray now produces a view (`Adjoint`) instead of a copy. I fixed it by using `transpose` explicitly (#101). The result is now a more reasonable error, as you were expecting.

KristofferC:
@carlobaldassi Hmm, I still hit this assert on 0.7. I had to add a `copy` to the Transpose to strip the `Tranpose` wrapper but even after adding that, I get the assert.

carlobaldassi:
It works for me. I cherry-picked #101 onto this branch, then changed all the `transpose(x)` in GraphType.jl  to `copy(x')` (I'll do a PR for this) and I don't see the assertion failure any more when calling `pkg> add JSON`.

KristofferC:
Alright, I'll try again, thanks!

KristofferC:
This should now have most deprecations fixed. The code loading still has to be integrated. If you are planning on doing that @StefanKarpinski, basing it on this branch might make sense.

StefanKarpinski:
This now allows me to load and use Pkg3 on current master. Haven't tested package building – do you know a package that has a build step that actually works on master?

src/GraphType.jl
```diff
@@ -301,7 +301,7 @@ mutable struct Graph
                 adjdict[p0][p1] = j1
 
                 bm = trues(spp[p1], spp[p0])
-                bmt = transpose(bm)
+                bmt = permutedims(bm)
```
carlobaldassi: This is much less efficient than `copy(bm')`.

StefanKarpinski: Why? It should be doing the same thing.

carlobaldassi: It calls the generic `permutedims!` method in "base/multidimensional.jl" rather than the specialized `transpose!` method in "stdlib/LinearAlgebra/src/bitarray.jl"

StefanKarpinski: The solution seems to be to add a specialized method for this operation then, no?

KristofferC: Oh sorry. I thought they would end up being the same thing. I will change it back then. I didn't feel like LinearAlgebra stuff would be needed here.

carlobaldassi: > The solution seems to be to add a specialized method for this operation then, no?

That can be done, I guess. We would need to bring back the transpose code from `LinearAlgebra` into `Base`. It would also mean that we would need to specialize `permitedims!`  for `BitArray`s with the `perm=(2,1)` argument value. (At this point, it seems weird not to call it `transpose` though...)

StefanKarpinski: Ok, maybe just revert this for now and open an issue to figure out the best approach to having a fallback to address the performance.

KristofferC: What is the absolute numbers in performance here though? It is a bit unfortunate to have Pkg3 depend on the Linear Algebra stdlib since presumably people will want to disable that one while still be able to use the package manager.

StefanKarpinski:
With this change Pkg3 loads on https://github.com/JuliaLang/julia/pull/25920. It also requires adding correct project files to SHA and Compat, which are these:
```toml
name = "SHA"
uuid = "ea8e919c-243c-51af-8825-aaa63cd721ce"

[deps]
Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
```
```toml
name = "Compat"
uuid = "34da2185-b29b-5c13-b0c7-acf172513d20"

[deps]
Base64 = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"
DelimitedFiles = "8bb1440f-4735-579b-a4ab-409b98df4dab"
IterativeEigensolvers = "de555fa4-b82f-55e7-8b71-53f60bbc027d"
Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
LinearAlgebra = "37e2e46d-f89d-539d-b4ee-838fcccc9c8e"
Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
SharedArrays = "1a1011a3-84de-559e-8e89-a11a2f7dc383"
SparseArrays = "2f01184e-e22b-5df5-ae63-d93ebab69eaf"
SuiteSparse = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"
Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"
```
That gets `pkg> st` working but `pkg> add` is broken, but getting Pkg3 loading on master is a good step in the right direction.

PR #100, opened 2018-01-08
fixes for cloning (+ help requested)

KristofferC:
I found a problem with cloning when checking out #83.

I looked at https://discourse.julialang.org/t/why-mypackage-is-forcing-other-package-downgrade-with-an-unbounded-require and created a fork (https://github.com/KristofferC/OdsIO.jl) of that repo.
When I changed the REQUIRE file to `julia 0.6` and used Pkg2 `clone` it correctly used the latest version of DataFrames.

However, when I cloned it in Pkg3, it was still getting a downgraded version of DataFrames. So I figured that compatibility info from the registry (for version 0.2.0 of OdsIO.jl) must somehow interfere. And yes, in #83, the identical version as the last registered version for the cloned package is used which will likely lead to problems when compatibility for the registered package with that version is added. Therefore I did the following change: https://github.com/JuliaLang/Pkg3.jl/compare/kc/clonerepo2-rebased...kc/no_deps_or_reqs_for_fixed?expand=1#diff-920ed338dcbd9d62489c6d12dc2fbb18R834 to put the version of the cloned package "above" the last registerd version, suspecting that that would fix the problem.

However, it did not.

Only when I stopped fixed packages from adding any compatibility or direct dependencies did the problem go away (the `uuid in keys(fixed) && continue`statements). Without that I still got:

```
 DataFrames [a93c6f00] log:
 ├─possible versions are: [0.1.0, 0.2.0-0.2.5, 0.3.0-0.3.16, 0.4.0-0.4.3, 0.5.0-0.5.12, 0.6.0-0.6.11, 0.7.0-0.7.8, 0.8.0-0.8.5, 0.9.0-0.9.1, 0.10.0-0.10.1, 0.11.0-0.11.2] or uninstalled
 ├─restricted to versions 0.8.0-9223372036854775807.9223372036854775807.9223372036854775806 by OdsIO [56b0d19f], leaving only versions [0.8.0-0.8.5, 0.9.0-0.9.1, 0.10.0-0.10.1, 0.11.0-0.11.2]
 ├─restricted by compatibility requirements with OdsIO [56b0d19f] to versions: [0.8.0-0.8.5, 0.9.0-0.9.1, 0.10.0-0.10.1]
 ├─versions reduced by equivalence to: [0.8.0, 0.8.3-0.8.5, 0.9.1, 0.10.1]
 └─set by the solver to its maximum version: 0.10.1
```

which indicates that compatibility for OdsIO is still holding DataFrames down.

@carlobaldassi Do you have an idea what is happening? It seems that the julia 0.5 requirement from 0.2.0 is still interfering, even though I put the cloned package to 0.2.0+.


KristofferC:
Or perhaps it is correct to not add the UUIDs for fixed packages since their dependencies have already been added to the wanted `UUID` vector...

carlobaldassi:
I'll have to look into this in detail. I suspect that versions "0.2.0+" and "0.2.0" are getting their dependencies "merged", in which case I suppose that the thing to do would be to remove all information coming from the registry for packages that are being cloned.

PR #101, opened 2018-01-08
Use `transpose` instead of `'` syntax

carlobaldassi:
Fixes a problem in julia 0.7

Also a minor fix to keep the julia name in `uuid_to_name`.

Ref #99.


KristofferC:
Great, thanks.

PR #105, opened 2018-01-11
Add newer Julia versions to loadmeta.jl

andyferris:
Not sure how necessary this is?


andyferris:
> those should be tree hashes not commit hashes, though

I copied what I thought v0.6.0 was doing, which seems to be a commit hash: https://github.com/JuliaLang/julia/commit/903644385b91ed8d95e5e3a5716c089dd1f1b08a

Is v0.6.0 incorrect?

StefanKarpinski:
I ended up doing this, forgetting that you already has a PR up.

PR #111, opened 2018-02-01
refactor parsing using enums and structs and use enums in more places

KristofferC:
* Refactors the parsing for REPLMode. Instead of each token being a `Tuple{Symbol, Vararg{Any}}` it is now just a `Token` which is defined as `const Token = Union{Command, Option, VersionRange, String}`. `Command` is a struct representing a command like `add`, `Option` is a struct representing an option like `--path=foo` or `--manifest`. There is no more `pkg`-token because it is not up to the lexer to decide if a word is a path or a package because it is context sensitive. Instead just lex those as a `String` and try to parse it as a package when it is expected in different commands. This makes the quotes not needed in paths and urls anymore.

Also uses enums instead of symbols in more places where there are a finite set of choices.

Also fixes not printing some stuff when using `status` in API mode.

Sorry for bundling some random stuff together...


KristofferC:
This got rekt.

PR #112, opened 2018-02-05
add support for archives in the registry

KristofferC:
this is done by allowing each version to declare a "archive-url" field pointing
to a tarball of the archive, and "archive-sha256" field with a sha256 hash
of the archive

if such an entry exists, we use that to download, otherwise, we fall back to the repo and use git.

PR for the registry at https://github.com/JuliaRegistries/Uncurated/pull/5.


StefanKarpinski:
Nice! I do think we probably want to think a bit about the structure of the "version properties" especially now that [dotted keys](https://github.com/toml-lang/toml#user-content-keys) are official in the TOML spec. E.g.:
```jl
tree.git.sha1 = "..."
tree.mercurial.sha1 = "..." # if we supported Mercurial for checkout

archive.url = "https://github.com/.../something.tar.gz"
archive.sha256 = "..."

StefanKarpinski:
The way forward here is that we're going to serve package from our own servers with predictable tarball locations instead of relying on putting this metadata in the registry. Until then we can continue to special case GitHub and GitLab based on knowing their specific tarball URL schemes.

PR #115, opened 2018-02-06
Significantly improve performance by not using broadcasting for very small arrays

KristofferC:
From profiling, a large bottleneck was found in creating the `Graph`. This is because broadcasting (while convenient) actually has a significant overhead when used on small arrays. This was frequently the case here.

Before the changes here, creating the `Graph` took 0.93 seconds when adding DifferentialEquations with a total time for `Pkg3.add` of 2.0 seconds.

After the changes, `Graph`takes 0.30 seconds and total time 1.36 seconds.

The changes to Operations.jl are for readability, was hard to me to keep the nested generators in my head (also this way we only do one loop).

### DIFF
--- a/src/GraphType.jl
+++ b/src/GraphType.jl
@@ -245,7 +245,7 @@ mutable struct Graph
         data = GraphData(versions, deps, compat, uuid_to_name, verbose = verbose)
         pkgs, np, spp, pdict, pvers, vdict, rlog = data.pkgs, data.np, data.spp, data.pdict, data.pvers, data.vdict, data.rlog
 
-        extended_deps = [[Dict{Int,BitVector}() for v0 = 1:(spp[p0]-1)] for p0 = 1:np]
+        extended_deps = [Vector{Dict{Int,BitVector}}(spp[p0]-1) for p0 = 1:np]
         for p0 = 1:np, v0 = 1:(spp[p0]-1)
             n2u = Dict{String,UUID}()
             vn = pvers[p0][v0]
@@ -275,7 +275,11 @@ mutable struct Graph
                 get!(req, pdict[uuid]) do; VersionSpec() end
             end
             # Translate the requirements into bit masks
-            req_msk = Dict(p1 => (pvers[p1][1:(spp[p1]-1)] .∈ vs) for (p1,vs) in req)
+            req_msk = Dict{Int,BitVector}()
+            for (p1, vs) in req
+                pv = pvers[p1]
+                req_msk[p1] = [pv[i] ∈ vs for i in 1:(spp[p1]-1)]
+            end
             extended_deps[p0][v0] = req_msk
         end
 
--- a/src/GraphType.jl
+++ b/src/GraphType.jl
@@ -147,10 +147,21 @@ mutable struct GraphData
 
         # equivalence classes (at the beginning each state represents just itself)
         eq_vn(v0, p0) = (v0 == spp[p0] ? nothing : pvers[p0][v0])
-        eq_classes = Dict(pkgs[p0] => Dict(eq_vn(v0,p0) => Set([eq_vn(v0,p0)]) for v0 = 1:spp[p0]) for p0 = 1:np)
+
+        # Hot code
+        eq_classes = Dict{UUID,Dict{InstState,Set{InstState}}}()
+        for p0 = 1:np
+            d = Dict{InstState, Set{InstState}}()
+            for v0 = 1:spp[p0]
+                let p0 = p0
+                    d[eq_vn(v0,p0)] = Set([eq_vn(v0,p0)])
+                end
+            end
+            eq_classes[pkgs[p0]] = d
+        end
 
         # the resolution log is actually initialized below
-        rlog = ResolveLog(uuid_to_name, verbose)
+        rlog = ResolveLog(uuid_to_name, false)
 
         data = new(pkgs, np, spp, pdict, pvers, vdict, uuid_to_name, pruned, eq_classes, rlog)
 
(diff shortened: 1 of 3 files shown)


src/GraphType.jl
```diff
@@ -275,7 +275,11 @@ mutable struct Graph
                 get!(req, pdict[uuid]) do; VersionSpec() end
             end
             # Translate the requirements into bit masks
-            req_msk = Dict(p1 => (pvers[p1][1:(spp[p1]-1)] .∈ vs) for (p1,vs) in req)
+            req_msk = Dict{Int,BitVector}()
+            for (p1, vs) in req
+                pv = pvers[p1]
+                req_msk[p1] = [pv[i] ∈ vs for i in 1:(spp[p1]-1)]
```
carlobaldassi: If this is performance-critical, it may be worth trying not to go through the construction of a `Vector{Bool}` and then converting it to a `BitArray`. This could be easily done with the generator syntax `BitArray(pv[i] ∈ vs for i in 1:(spp[p1]-1)`. That however is optimized for larger arrays probably, so another option would be using a standard `BitArray` constructor and then a loop.

src/Operations.jl
```diff
@@ -157,16 +164,16 @@ function deps_graph(env::EnvCache, uuid_to_name::Dict{UUID,String}, reqs::Requir
 
                 union!(all_versions_u, versions)
 
-                for (vr,dd) in deps_data
-                    all_deps_u_vr = get_or_make!(all_deps_u, vr)
+                for (ver, dd) in deps_data
```
carlobaldassi: Here and below, `vr` stood for "version range", so I would either revert these changes or spell it out.

src/GraphType.jl
KristofferC: I'll do some measurements, however, I think it was just the broadcast call that was the significant time thief.

KristofferC:
From more profiling I added some more optimizations. These typically fixes manifestations of https://github.com/JuliaLang/julia/issues/15276 and some cases where generator are slow. This, unfortunately, makes the code a bit uglier but I think it is worth it.

With this PR creating the `Graph` takes:

```
  0.104140 seconds (545.37 k allocations: 35.921 MiB, 22.12% gc time)
```

compared to before

```
  1.026278 seconds (2.72 M allocations: 168.556 MiB, 19.84% gc time)
```

when adding DifferentialEquations.

With #116, we are now at a total time of 0.46 seconds to `Pkg3.add("DifferentialEquations")` compared to ~3seconds two days ago.

src/Operations.jl
KristofferC: Reverted this.

src/GraphType.jl
```diff
@@ -147,10 +147,21 @@ mutable struct GraphData
...
+                let p0 = p0
+                    d[eq_vn(v0,p0)] = Set([eq_vn(v0,p0)])
+                end
+            end
+            eq_classes[pkgs[p0]] = d
+        end
 
         # the resolution log is actually initialized below
-        rlog = ResolveLog(uuid_to_name, verbose)
+        rlog = ResolveLog(uuid_to_name, false)
```
carlobaldassi: ?

src/GraphType.jl
```diff
@@ -242,10 +253,13 @@ mutable struct Graph
         extra_uuids = union(keys(reqs), keys(fixed), map(fx->keys(fx.requires), values(fixed))...)
         extra_uuids ⊆ keys(versions) || error("unknown UUID found in reqs/fixed") # TODO?
 
-        data = GraphData(versions, deps, compat, uuid_to_name, verbose = verbose)
+        data = GraphData(versions, deps, compat, uuid_to_name, verbose = verbose)::GraphData
```
carlobaldassi: That this would be needed is quite surprising to me. Any clue on the reason?

src/GraphType.jl
```diff
@@ -242,10 +253,13 @@ mutable struct Graph
...
         pkgs, np, spp, pdict, pvers, vdict, rlog = data.pkgs, data.np, data.spp, data.pdict, data.pvers, data.vdict, data.rlog
 
-        extended_deps = [[Dict{Int,BitVector}() for v0 = 1:(spp[p0]-1)] for p0 = 1:np]
+        local extended_deps
+        let spp = spp
+            extended_deps = [Vector{Dict{Int,BitVector}}(spp[p0]-1) for p0 = 1:np]::Vector{Vector{Dict{Int, BitVector}}}
+        end
```
carlobaldassi: What's the reason for using a `let` block here?

(also, extra space between `Int` and `BitVector`)

src/GraphType.jl
```diff
@@ -275,13 +289,25 @@ mutable struct Graph
...
             extended_deps[p0][v0] = req_msk
         end
 
         gadj = [Int[] for p0 = 1:np]
         gmsk = [BitMatrix[] for p0 = 1:np]
-        gconstr = [trues(spp[p0]) for p0 = 1:np]
+        local gconstr
+        let spp = spp
+            gconstr = [trues(spp[p0]) for p0 = 1:np]
+        end
```
carlobaldassi: Same here, what's the `let` block for?

src/GraphType.jl
KristofferC: https://github.com/JuliaLang/julia/issues/25918

src/GraphType.jl
KristofferC: Sorry, had to remove keyword arguments to use `code_warntype`, forgot to put this back.

src/GraphType.jl
KristofferC: https://github.com/JuliaLang/julia/issues/15276. Otherwise `spp` gets "tainted" and promoted to a `Core.Box` due to its usage in a closure (comprehension).

src/GraphType.jl
KristofferC: See https://github.com/JuliaLang/Pkg3.jl/pull/115#discussion_r166598165

src/GraphType.jl
KristofferC: See e.g. this comment: https://github.com/JuliaLang/julia/issues/15276#issuecomment-297596373

src/GraphType.jl
KristofferC: The type assertion is because inference couldn't figure out the type of `extended_deps` so it poisons other variables derived from this one.

carlobaldassi: Sorry, I see now that it's due to https://github.com/JuliaLang/julia/issues/15276. Maybe add a comment so that when that is eventually fixed we can revert?

carlobaldassi:
I also vote for performance over code aesthetics, the difference seems quite significant.

src/GraphType.jl
carlobaldassi: Ah I see thanks. Maybe also add a note in a comment pointing to that issue then?

KristofferC:
Added some comments explaining the reasons for some type asserts and `let`. Fixed the spurious space.

PR #130, opened 2018-02-13
delete unused function

KristofferC:
Not sure what the purpose of this function is and it is unused. Something you want to keep around @StefanKarpinski?


StefanKarpinski:
It was only used by the package loading code which is now in base and needs to be made interactive again once we've got Pkg3 moved into base. But that's why we have git, so fine to delete for now.

Issue #122, opened 2018-02-09
capturing Pkg3 output


stevengj:
It would be nice to make it easy to capture Pkg3 output to a stream if desired, instead of always printing to stdout/stderr.   This is especially useful for things like IJulia and IDE environments.

There seem to be some haphazard attempts at this in the source code, but it would be nice to see a more consistent use of an `io::IO` argument in all functions that do I/O, e.g. in the `Display` module.

By the way, I notice that there is actually a [`print_diff(io, x)` method](https://github.com/JuliaLang/Pkg3.jl/blob/86cc164033e0ad5c541588f00a8896078d162fc1/src/Display.jl#L101), but it appears to be buggy: it ignores its `io` argument and just prints to stdout.

KristofferC:
What's a good API to do this? Just a global variable that sets the output stream?

stevengj:
No globals, I'm thinking of just having an `io::IO` argument to `Pkg3.foo(...)` functions, which defaults to `STDOUT`.

(If there is also `STDERR` output, hopefully that can be written in terms of the new Base logging API, `@warn` etcetera, which has its own mechanisms to intercept output.)

StefanKarpinski:
A pretty large amount of options and state need to be passed around between all of these functions and that will only grow as we add more functionality to the interface. We should probably bite the bullet and have an type that wraps all of the options which can also have input, output and error streams in it.

KristofferC:
It seems a bit odd to have an IO argument to something like `Pkg.add`. It is a bit similar as the `@test` macros don't take an IO. Like the logging system where you configure the logger itseld, I feel this should be somehow be set to Pkg3 as a whole and not on individual functions.

StefanKarpinski:
Using the logging system for this does make sense.

KristofferC:
Just need some docs now.

fonsp:
I'd like to work on a PR that allows the syntax:
```julia
sprint(Pkg.status)
```

`sprint` calls the given function with a buffered `IO` as the first argument, but `Pkg.status` only works with a `Context` as the first argument. 

I could add a method for `status` that takes an `IO` object as the first argument, and turns it into a `Context` to pass to the original method. Is that a good approach?

fonsp:
I understand that `IO` as the first argument would be odd. For `Pkg.status`, however, I feel like it is appropriate, since the command has no side effects, other than printing to IO.

[...]

PR #131, opened 2018-02-13
thread a "PkgContext" through most of the Operations.

KristofferC:
This contains settings and state that needs to be passed around

Removes the `GLOBAL_SETTINGS` struct.

Based on the comment in #122:

> A pretty large amount of options and state need to be passed around between all of these functions and that will only grow as we add more functionality to the interface. We should probably bite the bullet and have an type that wraps all of the options which can also have input, output and error streams in it.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -12,48 +12,48 @@ using Pkg3.Types
 using Pkg3.TOML
 
 
-previewmode_info() = @info("In preview mode")
+preview_info() = @info("In preview mode")
 
 add(pkg::String; kwargs...)               = add([pkg]; kwargs...)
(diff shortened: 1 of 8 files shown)


KristofferC:
Yeah, I thought the same. Right now you can just use keyword arguments to each API method to set fields of the `PkgContext` though, like `Pkg3.add("JSON"; preview = true)` sets the `preview` field of the `PkgContext` so there is no need to create the context yourself.

StefanKarpinski:
Sure, but I have to see the type name in the code base itself. And it's not externally visible, so what's the harm in a shorter, more context-specific (see what I did there?) name?

KristofferC:
Yeah, that's why "I thought the same". The rest was just a remark.

KristofferC:
Changed it to just `Context`.

PR #133, opened 2018-02-14
do not resolve standard libraries, these are automatically satisfied

KristofferC:
Does this make sense to you @carlobaldassi.

Basically, these just gives the standard library a version (the same one as the current julia version, and then ignores any dependencies they have).


StefanKarpinski:
I'm going to merge this since it's an improvement over what we currently have in any case.

PR #139, opened 2018-02-16
allow v in VersionRange string, also fix return type of findfirst

KristofferC:


src/Types.jl
```diff
@@ -150,7 +150,7 @@ VersionRange(b::VersionBound=VersionBound()) = VersionRange(b, b)
 VersionRange(t::Integer...)                  = VersionRange(VersionBound(t...))
 VersionRange(v::VersionNumber)               = VersionRange(VersionBound(v))
 function VersionRange(s::AbstractString)
-    m = match(r"^\s*((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*)(?:\s*-\s*((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*))?\s*$", s)
+    m = match(r"^\s*[v]?((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*)(?:\s*-\s*[v]?((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*))?\s*$", s)
```
StefanKarpinski: I think that `[v]` can just be `v` – a character class containing a single letter is equivalent to the letter itself.

KristofferC: makes sense

PR #156, opened 2018-02-25
pkgsearch feature

chakravala:
As discussed in #43, with this you can do search in the pkg registry with various options. First
```Julia
import Pkg3
import Pkg3.pkgsearch
```
For example, find all packages that contain `Tree` in their name
```Julia
julia> Pkg3.pkgsearch("Tree",:name)
10-element Array{String,1}:
 "AbstractTrees"           
 "D3Trees"                 
 "DecisionTree"            
 "ExtremelyRandomizedTrees"
 "IntervalTrees"           
 "KDTrees"                 
 "PhyloTrees"              
 "RegionTrees"             
 "RigidBodyTreeInspector"  
 "TreeView"   
```
or find all packages that depend on `AbstractTrees` or contain it in their name:
```Julia
julia> pkgsearch("AbstractTrees",:name,:deps)
11-element Array{String,1}:
 "ASTInterpreter"    
 "AbstractTrees"     
 "CSTParser"         
 "Cascadia"          
 "DWARF"             
 "DataDepsGenerators"
 "ExprRules"         
 "Gallium"           
 "Gumbo"             
 "JuliaParser"       
 "TerminalUI"  
```
You can also search using the `:keywords` and `:desc` mode, but the current registry does not contain any data for that category yet.

It is also possible to match multiple strings for a single mode:
```Julia
julia> pkgsearch(:name,"Math","Fun")
26-element Array{String,1}:
 "ApproxFun"              
 "BasisFunctionExpansions"
 "DiffEqApproxFun"        
 "DocumentFunction"       
 "FunctionWrappers"       
 "FunctionalCollections"  
 "FunctionalData"         
 "FunctionalDataUtils"    
 "GLPKMathProgInterface"  
 "LossFunctions"          
 "LowLevelFloatFunctions" 
 "MathProgBase"           
 "MathToolkit"            
 "Mathematica"            
 "NaNMath"                
 "NumericFuns"            
 "ParameterizedFunctions" 
 "PenaltyFunctions"       
 "RationalFunctions"      
 "ResumableFunctions"     
 "ReusableFunctions"      
 "RollingFunctions"       
 "SpecialFunctions"       
 "StatsFuns"              
 "UnicodeFun"             
 "WebFuncs"   
```
Lastly, you can also do a search for any field by using `:any` mode or by doing
```Julia
pkgsearch(::String...)
```
Let me know what your thoughts are on this feature and how it can be improved.

This is only an initial go at it, there are still other ways this can be combined with Pkg3.

PR #169, opened 2018-03-02
add a command to create a new package, write some docs, add a framework for testing the REPL

KristofferC:
Very simple package generation so far.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,0 +1,66 @@
+# Pkg3.jl
+
+!!! warning
+    This documentation is a work in progress and the information in it might be or become outdated.
+
+Sections:
+
+```@contents
+Pages = [
+    "index.md"]
+```
+
+## Introduction
+
+Pkg3 is the package manager for Julia.
+
+## Getting Started
+
+The Pkg REPL-mode is entered using from the Julia REPL using the key `]`. To generate the files for a new project, use `pkg> generate`.
+
+```
+pkg> generate HelloWorld
+```
+
+This creates a new project `HelloWorld` with the following files
+
+```jl
+julia> cd("HelloWorld")
+shell> tree .
+.
+├── Project.toml
+└── src
+    └── HelloWorld.jl
+
+1 directory, 2 files
+```
+
+The `Project.toml` file contains the name of the package, its unique UUID, its version, the author and eventual dependencies:
+
+```toml
+name = "HelloWorld"
+uuid = "b4cd1eb8-1e24-11e8-3319-93036a3eb9f3"
+version = "0.1.0"
+author = ["Some One <someone@email.com>"]
+
+[deps]
+```
+
+The content of `src/HelloWorld.jl` is:
+
+```jl
+module HelloWorld
+
+greet() = print("Hello World!")
+
+end # module
+```
+
+We can now load the project and use it:
+
+```jl
+julia> import HelloWorld
+
+julia> HelloWorld.greet()
+Hello World!
+```
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -1,0 +1,73 @@
+function generate(path::String)
+    dir, pkg = dirname(path), basename(path)
+    isdir(path) && cmderror("$(abspath(path)) already exists")
+    printstyled("Generating"; color=:green, bold=true)
+    print(" project $pkg:\n")
+    project(pkg, dir)
+    entrypoint(pkg, dir)
+    return
+end
+
+function genfile(f::Function, pkg::String, dir::String, file::String)
+    path = joinpath(dir, pkg, file)
+    println(stdout, "    $path")
+    mkpath(dirname(path))
+    open(f, path, "w")
+    return
+end
+
+function project(pkg::String, dir::String)
+    name = email = nothing
+    gitname = LibGit2.getconfig("user.name", "")
+    isempty(gitname) || (name = gitname)
+    gitmail = LibGit2.getconfig("user.email", "")
+    isempty(gitmail) || (email = gitmail)
+
+    if name == nothing
+        for env in ["GIT_AUTHOR_NAME", "GIT_COMMITTER_NAME", "USER", "USERNAME", "NAME"]
+            name = get(ENV, env, nothing)
+            name != nothing && break
+        end
+    end
+
+    if name == nothing
+        cmderror("could not determine user, please set ", Sys.iswindows ? "USERNAME" : "USER",
+                 " environment variable")
+    end
+
+    if email == nothing
+        for env in ["GIT_AUTHOR_EMAIL", "GIT_COMMITTER_EMAIL", "EMAIL"];
+            email = get(ENV, env, nothing)
+            email != nothing && break
+        end
+    end
+
+    authorstr = "[\"$name " * (email == nothing ? "" : "<$email>") * "\"]"
+
+    genfile(pkg, dir, "Project.toml") do io
+        print(io,
+            """
+            name = "$pkg"
+            uuid = "$(UUIDs.uuid1())"
+            version = "0.1.0"
+            author = $authorstr
+
+            [deps]
+            """
+        )
+    end
+end
+
+function entrypoint(pkg::String, dir)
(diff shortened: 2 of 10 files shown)


docs/src/index.md
```diff
@@ -0,0 +1,68 @@
...
+## Introduction
+
+Pkg3 is the package manager for Julia.
+
+## Getting Started
+
+The Pkg REPL-mode is entered using from the Julia REPL using the key `]`. To start a new project, use `pkg> new`.
+
+```
+pkg> new HelloWorld
```
StefanKarpinski: Somehow `new` feels off – `gen`/`generate` maybe?

docs/src/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+This creates a new package `HelloWorld` with the following files
+
+```jl
+julia> cd("HelloWorld")
+shell> tree .
+.
+├── Project.toml
+└── src
+    └── HelloWorld.jl
```
StefanKarpinski: We should probably generate stubs for `test/runtests.jl` and `docs/index.md` as well since I find that the more you reduce the friction for these things the more likely people are to actually write them.

docs/src/index.md
```diff
@@ -0,0 +1,68 @@
...
+version = "0.1.0"
+author = ["Some One <someone@email.com>"]
+
+[deps]
+```
+
+The content of `src/HelloWorld.jl` is:
+
+```jl
+__precompile__(true)
```
StefanKarpinski: I feel like this has all the downsides of making `__precompile__(true)` the default without the upsides: people will just write code and if there are precompilation hazards then they will hit them without realizing it. But we still have to have this ugly thing at the top of the package entry point file. If we're going to generate this, we may as well default to having this be on. Ref: https://github.com/JuliaLang/julia/issues/26282.

src/generate.jl
```diff
@@ -0,0 +1,74 @@
+function new(pkg::String; minimal::Bool=true)
+    minimal == true || cmderror("only minimal package generation implemented")
```
StefanKarpinski: What do you imagine the difference between minimal and maximal generation to be?

test/repl.jl
```diff
@@ -0,0 +1,36 @@
+module TestREPL
```
StefanKarpinski: ❤️

docs/src/index.md
```diff
@@ -0,0 +1,68 @@
...
+
+## Getting Started
+
+The Pkg REPL-mode is entered using from the Julia REPL using the key `]`. To start a new project, use `pkg> new`.
+
+```
+pkg> new HelloWorld
+```
+
+This creates a new package `HelloWorld` with the following files
```
StefanKarpinski: I think we should design this to generate projects in general – both packages (reusable) and applications (standalone). From my wip glossary:

> **Project:** a source tree with a standard layout for a Julia project, including a `src` directory for the main body of Julia source code, a `test` directory for testing the project, `docs` for documentation files, and optionally `build` for build scripts and outputs.
> 
> **Application:** a project which provides standalone functionality that is not intended to be reused other Julia projects.
> 
> **Package:** a project which provides functionality that can be used by other Julia projects via `import X` or `using X`. A package will typically have a project file with a `uuid` entry giving its package UUID. This UUID is used to identify the package when other projects depend on it.

Most of the infrastructure is common, but this is a good opportunity to support both use cases with a single shared body of development tools.

src/generate.jl
```diff
@@ -0,0 +1,74 @@
...
+    # Find a name + email
+    name = email = nothing
+    dot_gitconfig = joinpath(homedir(), ".gitconfig")
+    if isfile(dot_gitconfig)
+        content = read(dot_gitconfig, String)
+        m = match(r"name = (.*?)\n", content)
+        m != nothing && (name = m.captures[1])
+        m = match(r"email = (.*?)\n", content)
+        m != nothing && (email = m.captures[1])
+    end
```
StefanKarpinski: Why parse `.gitconfig` for this? Why not use `libgit2` or `git` to get this info?

docs/src/index.md
KristofferC: Ok so `--application` or `--package` options? And an `application` does not get a UUID?

docs/src/index.md
KristofferC: True

src/generate.jl
KristofferC: I'll take it from LibGit2 instead.

docs/src/index.md
StefanKarpinski: No harm in generating one, I guess. Then if someone decides to convert their application to a package (which is pretty common), they're good to go. I think the main difference is actually that an application can set global configuration while a package cannot since it will be loaded by something else which controls global configuration.

docs/src/index.md
KristofferC: Ok but then `Test` needs to go into the `project` and it needs a `compatability` and we need to discuss test-only dependencies etc.

docs/src/index.md
KristofferC: `create`?

src/generate.jl
KristofferC: `.travis`, `appveyor`, `.codecov`, `test/runtests.jl`, `docs/make.jl`, git initialization etc...

KristofferC: I removed this kwarg for now.

docs/src/index.md
```diff
@@ -0,0 +1,66 @@
...
+    "index.md"]
+```
+
+## Introduction
+
+Pkg3 is the package manager for Julia.
+
+## Getting Started
+
+The Pkg REPL-mode is entered using from the Julia REPL using the key `]`. To start a new project, use `pkg> new`.
```
StefanKarpinski: There's still a `new` here.

docs/src/index.md
StefanKarpinski: Fair enough, we can wait until that's sorted out.

StefanKarpinski:
So I still prefer `generate` to `new` or `create` because it's very explicit about what this command does: it generates files for you with a particular layout.

KristofferC:
Sure, but what you want to do is to >create< a >new< project. Generating files is just a means to an end. Naming the command after what you want to do, not how it does it, seems better to me.

martinholters:
If I want to create a new project, I want Pkg3 to generate the boilerplate for me. Should we name the command after what I want to do or what I want Pkg3 to do?

KristofferC:
If someone wants to create a new package, I would assume they would reach for `help create` as the first thing, or `help new`, so that seems like good commands.

docs/src/index.md
ulysses4ever: Sorry for interjecting: wouldn't it be easier for beginners to do the opposite: 1) turn precompilation on be default (I'm generally in favor of this), 2) generate `__precompile__(false)` in a stub, possibly with the comment: "erase if confident" or similar.

chakravala:
Personally, I like `generate` since that was previously used in `Pkg.generate` and it seems appropriate because I hope to generate some new files. `new` could work as well, but somehow I feel like generate is a stronger and more decisive word for what I'd like to accomplish with the action.

It would be useful if we could have a basic functionality for generating packages working.

While working on the package search #156 I was wondering how exactly the `TOML` files will be generated. Specifically, I was wondering how the `keywords` and `desc`ription data fields in the `Project.toml` will be initialized. One possible way is to insert blank entries into the file for those
```
desc = "$pkg"
keywords = ["julia"]
license = "none"
name = "$pkg"
uuid = "$(UUIDs.uuid1())"
version = "0.1.0"
author = $authorstr

[deps]
```
This may not be it, but there needs to be some kind of initialization for this data, so that the user is encouraged to edit these values and fill them in with relevant information.

Another question I had: is it possible to synchronize the description and keywords values of the Project.toml with the github repositories description and keywords?

KristofferC:
Ok, went with `generate` because I don't have time to bike-shed :). This is very bare bone on purpose, mostly to get some docs at all up.

PR #171, opened 2018-03-03
WIP add tools for handling registries and use them to test Pkg3

KristofferC:
Quite a lot of WIP since the design has to come before tests... This PR put some things in place, at least.

This PR adds functionality for creating a registry and to register a package in the registry.
It adds tests that do this, and checks that the registered package can be added.

We can use this PR as a base make sure we have test for all different features we support in the registry, without having a registry in the while to test on.
We can also use it to test the workflow for package developers.

### DIFF
--- a/src/Registry.jl
+++ b/src/Registry.jl
@@ -1,0 +1,266 @@
+module Registry
+
+import UUIDs
+import LibGit2
+import ..TOML, ..Operations, ..API
+using ..Types
+
+function write_toml(f::Function, names::String...)
+    path = joinpath(names...) * ".toml"
+    mkpath(dirname(path))
+    open(path, "w") do io
+        f(io)
+    end
+end
+
+function create_registry(path; repo::Union{Nothing, String} = nothing, uuid = UUIDs.uuid1(), description = nothing)
+    isdir(path) && error("$(abspath(path)) already exists")
+    mkpath(path)
+    write_mainfile(path, uuid, repo, description)
+    LibGit2.with(LibGit2.init(path)) do repo
+        LibGit2.add!(repo, "*")
+        LibGit2.commit(repo, "initial commit for registry $(basename(path))")
+    end
+    return
+end
+
+function write_mainfile(path, uuid, repo, description)
+    d = Dict{String, Any}()
+    d["name"] = basename(path)
+    d["uuid"] = string(uuid)
+    if repo !== nothing
+        d["repo"]= repo
+    end
+    if description !== nothing
+        d["description"] = description
+    end
+    write_mainfile(path, d)
+end
+
+function write_mainfile(path::String, data::Dict)
+    open(joinpath(path, "Registry.toml"), "w") do io
+        println(io, "name = ", repr(data["name"]))
+        println(io, "uuid = ", repr(data["uuid"]))
+        if haskey(data, "repo")
+            println(io, "repo = ", repr(data["repo"]))
+        end
+        println(io)
+
+        if haskey(data, "description")
+            print(io, """
+            description = \"\"\"
+            $(data["description"])\"\"\"
+            """
+            )
+        end
+
+        println(io)
+        println(io, "[packages]")
+        if haskey(data, "packages")
+            for (uuid, data) in sort!(collect(data["packages"]), by=first)
+                println(io, uuid, " = { name = ", repr(data["name"]), ", path = ", repr(data["path"]), " }")
+            end
+        end
+    end
+end
+
+struct PackageReg
+    uuid::UUID
+    name::String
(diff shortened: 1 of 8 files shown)


KristofferC:
I made this on top of #169 because something like that is needed here.

andyferris:
Awesome, this is what I've been wanting! I want to create a registry and add all of our private packages. And not entirely by hand, preferably :)

The other thing I'd like to do is work with git repositories with multiple packages inside (perhaps even a "monorepo", so to speak). Regarding this PR I'd love to be able to create a new git repository, and then create multiple new packages inside it in subdirectories - I'm not sure if this is two separate steps, whether we expect to be able to do this without external use of `git init`, or whatever.

KristofferC:
> The other thing I'd like to do is work with git repositories with multiple packages inside (perhaps even a "monorepo", so to speak). Regarding this PR I'd love to be able to create a new git repository, and then create multiple new packages inside it in subdirectories - I'm not sure if this is two separate steps, whether we expect to be able to do this without external use of git init, or whatever.

Yes, I've been thinking about that too. The way I've envisioned it is a "base" project with relative `path` entries to all the packages in subdirectories.

Could you explain with a bit more detail how you envisioned the workflow to be.

o314:
Monorepo can be great, to speedup development. Stefan has posted about that.
Caveats: There are some really dirty codebase out there.

Inner repos in a master monorepo should not be spreaded everywhere at unlimited depth.
Worst case: there is some hop in the traversal of the packaging: /a/b/c where a is a package, b is not, c is a package too. These with hundreds of package will really be a mess.
I dont' know if is possible in Julia yet. It should not be.

Everything in a root subdir could be a good convention.

If you got time, there's some very good points there:
- Build-Level Components (slides); 2005; Merijn de Jonge [PDF](https://github.com/JuliaLang/Pkg3.jl/files/1800571/Build-level.Components.slides.pdf), [PPT](http://www.win.tue.nl/ipa/archive/springdays2005/Jonge.ppt)
- Decoupling Source Trees into Build-Level Components; 2005; merijn de Jonge [PDF](http://www2.win.tue.nl/ipa/archive/springdays2005/Jonge1.pdf)

KristofferC:
> Regarding this PR I'd love to be able to create a new git repository, and then create multiple new packages inside it in subdirectories

You can pretty much do this now by

```
pkg> generate BasePackage
shell> cd "BasePackage"
pkg> generate SubPackage1
pkg> generate SubPackage2
pkg> develop SubPackage1
pkg> develop SubPackage2

shell> tree .
.
├── Manifest.toml
├── Project.toml
├── SubPackage1
│   ├── Project.toml
│   └── src
│       └── SubPackage1.jl
├── SubPackage2
│   ├── Project.toml
│   └── src
│       └── SubPackage2.jl
└── src
    └── BasePackage.jl

5 directories, 7 files

pkg> st
Status "~/Documents/BasePackage/Project.toml"
 [3d06c3e4] SubPackage1 v0.1.0 ["SubPackage1"]
 [3e6e98ee] SubPackage2 v0.1.0 ["SubPackage2"]

shell> cat Manifest.toml
[[SubPackage1]]
path = "SubPackage1"
uuid = "3d06c3e4-25d7-11e8-12c9-9d4210033fc5"
version = "0.1.0"

[[SubPackage2]]
path = "SubPackage2"
uuid = "3e6e98ee-25d7-11e8-2482-9d413da045bf"
version = "0.1.0"
```

andyferris:
Ok - thank you - I’m impressed (as usual, @KristofferC)!

KristofferC:
Ok, I have added functionality so one can register + tag packages.

What is lacking is the compression scheme that is in the current registry.
Right now, it will just list all the deps and compat for each separate version.

@StefanKarpinski I am going to need your help here. Hopefully, it shouldn't be too hard to hook up your compression stuff just after the new version data has been put into `Compat.toml` and `Deps.toml`.

src/Registry.jl
```diff
@@ -0,0 +1,266 @@
...
+            else
+                name = registry_packages[string(uuid)]["name"]
+                new_compat[name] = v
+            end
+            new_deps[name] = uuid
+        end
+        deps_data[VersionRange(pkg.version)] = new_deps
+        compat_data[VersionRange(pkg.version)] = new_compat
+
+        # TODO: compression
```
KristofferC: Somewhere here is probably where the compression should go?

BenJWard:
@KristofferC What is the current status of this PR? Is the functionality in this PR available for people who might want to work with registires to test out? If so would it be useful to rebase, since the last commit was July 2018?

KristofferC:
Obsolete

PR #175, opened 2018-03-05
allow packages to be added from repos and using git revs

KristofferC:
```
pkg> add Example#master
pkg> add Example#43oi54h3254
pkg> add https://github.com/fredrikekre/ImportMacros.jl#branch
````

Packages tracking branches will be updated on major upgrades.

TODO: Accidentally added some submodules...


KristofferC:
AV 64 bit is on a 2 week old commit and is the only CI that fails, so I will blame that.

PR #177, opened 2018-03-09
show a progress bar when cloning git repos

KristofferC:
In the spirit of https://github.com/JuliaLang/Pkg3.jl/issues/89.

http://www.giphy.com/gifs/2rAET3llXUjR9pfUVO

Requires https://github.com/JuliaLang/julia/pull/26387

cc @jperla

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -1,5 +1,6 @@
 module GitTools
 
+using ..Pkg3
 import LibGit2
 using Printf
 
@@ -16,6 +17,8 @@ end
 function showprogress(io::IO, p::MiniProgressBar)
     perc = p.current / p.max * 100
     prev_perc = p.prev / p.max * 100
+    # Bail early if we are not updating the progress bar,
+    # Saves printing to the terminal
     if p.has_shown && !(perc - prev_perc > 0.1)
         return
     end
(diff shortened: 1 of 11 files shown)


src/GitTools.jl
```diff
@@ -0,0 +1,125 @@
...
+    received_objects::Cuint
+    local_objects::Cuint
+    total_deltas::Cuint
+    indexed_deltas::Cuint
+received_bytes::Csize_t
+end
+
+function transfer_progress(progress::Ptr{GitTransferProgress}, p::Any)
+    progress = unsafe_load(progress)
+    bar = unsafe_pointer_to_objref(p.transfer_progress)
```
KristofferC: I would have assumed that the payload sent to this function was supposed to be the `p.transfer_progress` instead of the whole struct.

omus: I forgot to update this code with my Pkg3.jl PR. This will need to be changed to:
```julia
transfer_progress_ptr = payload_unpack(p, :transfer_progress)
@assert transfer_progress_ptr != C_NULL
bar = unsafe_pointer_to_objref(transfer_progress_ptr)
```
At the moment if you use the tuple syntax for `RemoteCallbacks` it breaks apart the tuple and puts each of the payloads inside of a `RemotePayloads` container. I decided that having each callback unpack this object was the better solution instead of unwrapping it before calling the user's callback.

KristofferC: >  I decided that having each callback unpack this object was the better solution instead of unwrapping it before calling the user's callback.

Alright.

KristofferC: @omus 

> payload_unpack(p, :transfer_progress)

doesn't work because `payload_unpack` is an instance of `RemotePayloads` and not a pointer to it (like `payload_unpack` expects).

omus:
> I would have assumed that the payload sent to this function was supposed to be the `p.transfer_progress` instead of the whole struct (since those are given together to `RemoteCallbacks`).

I could change #26437 to handle this more transparently. The reason I didn't go in this direction is because it would involve an extra `cfunction` which would intercept the callback from libgit2, unpack the payload, then execute the original C call.

KristofferC:
In some sense that would be "cleaner" in that your callback get exactly the corresponding payload that you specified. On the other hand, maybe you want to look at the payloads for all the other callbacks? Doesn't matter too much to me :)

src/GitTools.jl
```diff
@@ -0,0 +1,127 @@
...
+    total_deltas::Cuint
+    indexed_deltas::Cuint
+received_bytes::Csize_t
+end
+
+function transfer_progress(progress::Ptr{GitTransferProgress}, p::Any)
+
+    progress = unsafe_load(progress)
+    @assert p.transfer_progress != C_NULL
+    bar = unsafe_pointer_to_objref(p.transfer_progress)
```
omus: If you define this function as `transfer_progress(progress::Ptr{GitTransferProgress}, bar::MiniProgressBar)` you can perform the unsafe operations in another method like:
```julia
function transfer_progress(progress::Ptr{GitTransferProgress}, p::LibGit2.RemotePayloads)
    transfer_progress(progress, unsafe_pointer_to_objref(p.transfer_progress))
end
```

KristofferC: What's the advantage?

omus: May make the code slightly cleaner

KristofferC:
I'll rebase this when nightlies are updated.

StefanKarpinski:
This is a pretty slick feature 😁

KristofferC:
Mac and Linux nightlies are up, tests are passing, things are awesome :tada:.

PR #186, opened 2018-03-11
some more basic docs

KristofferC:


KristofferC:
Mac failure is nightly not being updated to undef PR.

src/Display.jl
```diff
@@ -17,12 +17,14 @@ const colors = Dict(
 const color_dark = :light_black
 
 function git_file_stream(repo::LibGit2.GitRepo, spec::String; fakeit::Bool=false)::IO
-    blob = try LibGit2.GitBlob(repo, spec)
+    blob = try LibGit2.GitBlob, repo, spec
```
StefanKarpinski: I'm confused by this change.

KristofferC: should be changed back

PR #194, opened 2018-03-13
Add credential handling

omus:
Adds credential handling into Pkg3. Work was based on top of https://github.com/JuliaLang/Pkg3.jl/pull/177 and was used to validate the changes in https://github.com/JuliaLang/julia/pull/26437.

Requires https://github.com/JuliaLang/julia/pull/26437

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -46,7 +46,7 @@ function transfer_progress(progress::Ptr{LibGit2.GitTransferProgress}, p::Any)
     return Cint(0)
 end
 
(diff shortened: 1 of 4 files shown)


KristofferC:
@omus Is it ok to reuse the same `CachedCredentials` in many calls? If so, I could put it in what we call the `Context` which is like a container for a bunch of random stuff that has to propagate through everything.

KristofferC:
Included in https://github.com/JuliaLang/Pkg3.jl/pull/177. Passes tests on https://github.com/JuliaLang/julia/pull/26437

PR #209, opened 2018-03-16
testing/building projects

KristofferC:


KristofferC:
Travis mac fails because nightly not updated.

PR #215, opened 2018-03-19
Rework autocompletion

pfitzseb:
This significantly improves the autocompletion code by re-using `tokenize`. Also fixes https://github.com/JuliaLang/Pkg3.jl/issues/213 and works around https://github.com/JuliaLang/Pkg3.jl/issues/200 by only showing short-form completions (for commands and options) if you've typed absolutely nothing yet:
```
pkg>
+         ?          dev        gc         install    rm         test       upgrade
-         add        develop    generate   pin        search     uninstall
.         build      find       help       preview    st         up
/         checkout   free       init       remove     status     update
pkg> s
search status
```
This also means that typing `r\tab` completes to `remove` instead of showing a selector.

Let me know what you guys think!

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -758,22 +758,15 @@ end
 
 function complete_command(s, i1, i2)
     # only show short form commands when no input is given at all
-    if isempty(s)
-        cmp = filter(cmd -> startswith(cmd, s), all_commands_sorted)
-    else
-        cmp = filter(cmd -> startswith(cmd, s), long_commands)
-    end
+    cmp = filter(cmd -> startswith(cmd, s), isempty(s) ? all_commands_sorted : long_commands)
     return cmp, i1:i2, !isempty(cmp)
 end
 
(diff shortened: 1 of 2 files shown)


KristofferC:
Reusing as much of the REPL parser as possible is great because otherwise the one for the completions might become stale.

pfitzseb:
Test failure on MacOS seems unrelated?

KristofferC:
Yes, nightly not updated for Mac.

StefanKarpinski:
Oh, delightful! This is exactly what I had hoped for when I made the REPL mode 😁. A few people have (quite reasonably) questioned the need for as compared to just having a Julia API – but really stellare tab Pkg-specific tab completion support was one of the motivations.

KristofferC:
Would be good if we could add a few tests, for example `preview remo<TAB>` since there is some special logic for that.

KristofferC:
Pushed some more tests, will merge after CI.

PR #234, opened 2018-03-27
Compatibility in Pkg3

KristofferC:
Early draft to not do unnecessary work.

Compatability entry in Project.toml is simply:

```
[compatibility]
Example = "0.4"
```

The version string gets parsed to a range using some type of semver versioning. When we resolve, we intersect whatever version range we want to satisfy with the compatibility range.

TODO: 

- Respect compatibility in Project files in "checked out" packages.
- ~~Figure out how to solve the problem that upper bounds in `VersionRange` are inclusive where for the semver notation we wan't them exclusive.~~ Thx Stefan.
  Actually no:
    ```
    julia> in(v"0.5.1", Pkg3.Types.VersionSpec("0.4.1-0.5"))
    true
    ```
- Test it properly.


test/pkg.jl
```diff
@@ -15,6 +15,20 @@ include("utils.jl")
 
 const TEST_PKG = (name = "Example", uuid = UUID("7876af07-990d-54b4-ab0e-23690620f79a"))
 
+import Pkg3.Types.semver_spec
+@testset "semver notation" begin
+    @test semver_spec("^1.2.3") == Pkg3.Types.VersionSpec("1.2.3-2")
+    @test semver_spec("^1.2")   == Pkg3.Types.VersionSpec("1.2.0-2")
+    @test semver_spec("1")      == Pkg3.Types.VersionSpec("1.0.0-2")
```
StefanKarpinski: Is this implementing the Cargo-style semver_spec where "no prefix" == "^ prefix"? If so, it might be better to test it here with the caret for clarity and then separately test that the no prefix version is equal to the `^` version.

KristofferC: :+1:, done

test/pkg.jl
```diff
@@ -19,14 +19,17 @@ import Pkg3.Types.semver_spec
 @testset "semver notation" begin
     @test semver_spec("^1.2.3") == Pkg3.Types.VersionSpec("1.2.3-2")
     @test semver_spec("^1.2")   == Pkg3.Types.VersionSpec("1.2.0-2")
-    @test semver_spec("1")      == Pkg3.Types.VersionSpec("1.0.0-2")
+    @test semver_spec("^1")     == Pkg3.Types.VersionSpec("1.0.0-2")
     @test semver_spec("^0.2.3") == Pkg3.Types.VersionSpec("0.2.3-0.3")
     @test semver_spec("0.0.3")  == Pkg3.Types.VersionSpec("0.0.3-0.0.4")
     @test semver_spec("^0.0")   == Pkg3.Types.VersionSpec("0.0.0-0.1")
     @test semver_spec("0")      == Pkg3.Types.VersionSpec("0.0.0-1")
```
StefanKarpinski: Same here.

test/pkg.jl
```diff
@@ -19,14 +19,17 @@ import Pkg3.Types.semver_spec
 @testset "semver notation" begin
     @test semver_spec("^1.2.3") == Pkg3.Types.VersionSpec("1.2.3-2")
     @test semver_spec("^1.2")   == Pkg3.Types.VersionSpec("1.2.0-2")
-    @test semver_spec("1")      == Pkg3.Types.VersionSpec("1.0.0-2")
+    @test semver_spec("^1")     == Pkg3.Types.VersionSpec("1.0.0-2")
     @test semver_spec("^0.2.3") == Pkg3.Types.VersionSpec("0.2.3-0.3")
     @test semver_spec("0.0.3")  == Pkg3.Types.VersionSpec("0.0.3-0.0.4")
```
StefanKarpinski: and maybe here too, although in this case there's no difference between `^` and `=`.

src/Types.jl
```diff
@@ -1217,7 +999,7 @@ function write_env(ctx::Context; display_diff=true)
         if !ctx.preview
             mkpath(dirname(env.project_file))
             open(env.project_file, "w") do io
-                TOML.print(io, project, sorted=true)
+                TOML.print(io, project, sorted=true, by=project_key_order)
```
StefanKarpinski: Maybe sort by `key -> (project_key_order(key), key)`? So that unknown keys are ordered alphabetically.

src/versions.jl
```diff
@@ -0,0 +1,299 @@
...
+    print(io, ']')
+end
+Base.show(io::IO, s::VersionSpec) = print(io, "VersionSpec(\"", s, "\")")
+
+
+###################
+# Semver notation #
+###################
+
+const ver_reg = r"^([~^]?)?[v]?([0-9]+?)(?:\.([0-9]+?))?(?:\.([0-9]+?))?$"
```
StefanKarpinski: You can just write `v?` instead of `[v]?`, it's equivalent. Should this allow whitespace in some places? I guess we can start by being strict and not allowing it; we can always allow some if necessary.

src/versions.jl
```diff
@@ -0,0 +1,299 @@
...
+    ranges = VersionRange[]
+    for ver in split(s, ',')
+        push!(ranges, semver_interval(ver))
+    end
+    return VersionSpec(ranges)
+end
+
+function semver_interval(s::AbstractString)
+    if !contains(s, ver_reg)
+        error("invalid version specifier")
```
StefanKarpinski: Maybe print it here so the user can see what was invalid at least?

src/versions.jl
```diff
@@ -0,0 +1,299 @@
...
+    end
+    m = match(ver_reg, s)
+    @assert length(m.captures) == 4
+    n_significant = count(x -> x !== nothing, m.captures) - 1
+    typ, _major, _minor, _patch = m.captures
+    major =                           parse(Int, _major)
+    minor = (n_significant < 2) ? 0 : parse(Int, _minor)
+    patch = (n_significant < 3) ? 0 : parse(Int, _patch)
+    if n_significant == 3 && major == 0 && minor == 0 && patch == 0
+        error("invalid version 0.0.0")
```
StefanKarpinski: Again, print the actual version string too?

src/versions.jl
```diff
@@ -0,0 +1,299 @@
...
+    end
+    vertyp = (typ == "" || typ == "^") ? :caret : :tilde
```
StefanKarpinski: Might want a comment here indicating that this makes `^` the default type specifier similar to Cargo but dissimilar from NPM.

src/versions.jl
```diff
@@ -0,0 +1,299 @@
...
+    v0 = VersionBound((major, minor, patch))
+    v1 = nothing
```
StefanKarpinski: What is `v1` for? Isn't this dead code?

src/versions.jl
```diff
@@ -0,0 +1,299 @@
...
+
+###################
+# Semver notation #
+###################
+
+const ver_reg = r"^([~^]?)?[v]?([0-9]+?)(?:\.([0-9]+?))?(?:\.([0-9]+?))?$"
+
+function semver_spec(s::String)
+    ranges = VersionRange[]
+    for ver in split(s, ',')
```
StefanKarpinski: Comma-separated specs don't seem to be tested at all.

src/Operations.jl
```diff
@@ -955,21 +969,25 @@ function up(ctx::Context, pkgs::Vector{PackageSpec})
...
         info = manifest_info(ctx.env, pkg.uuid)
-        if haskey(info, "repo-url")
-            pkg.repo = Types.GitRepo(info["repo-url"], info["repo-rev"])
-            new = handle_repos_add!(ctx, [pkg]; upgrade_or_add = (level == UPLEVEL_MAJOR))
-            append!(new_git, new)
-        else
-            ver = VersionNumber(info["version"])
-            if level == UPLEVEL_FIXED
-                pkg.version = VersionNumber(info["version"])
+        if info != nothing
```
StefanKarpinski: Is checking this a bug-fix? I.e. was it possible previously to have `info == nothing` here?

src/Operations.jl
```diff
@@ -126,11 +129,12 @@ end
...
 function collect_project!(pkg::PackageSpec, path::String, fix_deps_map::Dict{UUID,Vector{PackageSpec}})
     project_file = joinpath(path, "Project.toml")
-    !isfile(project_file) && return false
     fix_deps_map[pkg.uuid] = valtype(fix_deps_map)()
+    !isfile(project_file) && return false
     project = read_project(project_file)
+    compat = get(project, "compatibility", Dict())
     for (deppkg_name, uuid) in project["deps"]
-        vspec = VersionSpec() # TODO: Update with compatibility from Project
+        vspec = haskey(compat, deppkg_name) ? Types.semver_spec(compat["deppkg_name"]) : VersionSpec()
```
StefanKarpinski: Should this be `compat[deppkg_name]` instead of `compat["deppkg_name"]`?

src/Operations.jl
KristofferC: Yes, if you do `up` without having a Manifest. But this is fixed on master.

PR #243, opened 2018-04-06
Avoid overwriting user-defined settings in REPL

carlobaldassi:
Without this my custom REPL settings from `.julia/config/startup.jl` get overwritten, because keybinding customization is done by calling `REPL.setup_interface` too, cf. https://docs.julialang.org/en/latest/stdlib/REPL/#Customizing-keybindings-1, and it runs before Pkg3 is loaded.


KristofferC:
Thanks!

PR #251, opened 2018-04-16
make Pkg3 test itself

KristofferC:
No longer rely on Pkg2 for anything. The fact that Pkg3 is a stdlib makes this a bit more complicated. Let's see if this work.


KristofferC:
Need to update the registry updater with "Sockets".

PR #253, opened 2018-04-17
update registries when adding

KristofferC:
Do you like this @StefanKarpinski? It seems people sometimes get confused when they add packages and not get the latest version because they haven't run `up` in a long time. Perhaps just updating the registry even when adding makes sense.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -77,7 +77,13 @@ function update_registry(ctx)
                         return
                     end
                     branch = LibGit2.headname(repo)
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Yes, I think we should be quite liberal with updating the registry since we're now so much more conservative about upgrading packages. We do probably need and offline mode so that people can do local package operations without having Julia keep trying to reach the Internet and failing. The offline mode would basically just assume that the Internet is unavailable but do whatever operations it can do without that.

PR #275, opened 2018-05-03
wip fix permission of files downloaded

KristofferC:
Check that this fails the new test.


KristofferC:
Ok, this fails on linux like expected, now we remove the commented out stuff and hope that succeed.

KristofferC:
Should be fixed in Base. Ill reopen this to workaround it if really needed.

PR #278, opened 2018-05-04
implement instantiate

KristofferC:
can be used to instantiate a manifest without going through the resolver
also, show warning in status output about non downloaded packages

Closes https://github.com/JuliaLang/Pkg3.jl/issues/208


KristofferC:
I also want us to store the location where we downloaded the package from in the manifest whereby the registry is not needed at all for instantiating (but can be used if the package has moved).

Issue #279, opened 2018-05-08
Cannot `develop Pkg3`


simonbyrne:
```
(v0.7) pkg> develop Pkg3
ERROR: The following package names could not be resolved:
 * Pkg3 (not found in project, manifest or registry)
Please specify by known `name=uuid`.
```

Even if this is not possible, it would be useful to have some instructions as to how I can use this repo instead of the one that comes with Base.

KristofferC:
Since Pkg3 is also an stdlib it is a little bit special.

One way to do it is just to clone the repo, remove the `uuid = ` field from the Project, go into the Pkg3 folder and then you can import it.

simonbyrne:
What's the plan for allowing users to update stdlib? I seem to recall this being the plan at some point, but I can't find any details.

StefanKarpinski:
We'd need one of the following options to work:

1. building multiple sysimgs with different stdlib versions and choosing one based on manifest contents;
2. stdlibs not built into the system image but precompiled with machine code so that loading and using them is really fast;
3. the ability to replace stdlibs in an already-running system if they haven't been used yet.

These are roughly in order of immediate feasibility (easiest to hardest).

KristofferC:
Instructions how to develop Pkg3 is now in the README. Discussion for how to update stdlib at all is a big one that also involves code loading so should probably be had in Base.

timholy:
:+1:. If you run a very recent `master` of Julia and are up-to-date on your packages, `Revise.track(Pkg)` should work now for the stdlib version. It will `eval` a diff between whatever you have on your filesystem and whatever was in your last git commit, so best to build julia after you make commits.

PR #280, opened 2018-05-08
use Base.download to get archives

simonbyrne:
Note: I haven't actually tested this (see #279).

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -369,9 +369,8 @@ function install_archive(
         if archive_url != nothing
             path = tempname() * randstring(6) * ".tar.gz"
             url_success = true
-            cmd = BinaryProvider.gen_download_cmd(archive_url, path);
             try
-                run(cmd, (devnull, devnull, devnull))
+                download(archive_url, path)
             catch e
                 e isa InterruptException && rethrow(e)
                 url_success = false


KristofferC:
What the reason for BinaryProvider not using `Base.download` @staticfloat

KristofferC:
This now prints

```
 Resolving package versions...
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   128    0   128    0     0   1809      0 --:--:-- --:--:-- --:--:--  1828
  0     0    0  2132    0     0  17149      0 --:--:-- --:--:-- --:--:-- 2082k
```

and we spawn multiple downloads at once so it will look pretty bad I think.

staticfloat:
Back in the day, it was because the powershell implementation was broken, and we couldn't update people's `Base` as easily as we could update `BinaryProvider`. Only three reasons nowadays:

(1) More flexible output handling

(2) Resume capabilities (e.g. for large downloads on flaky connections)

(3) Doesn't require the usage of the `which` utility.  (This will be fixed by https://github.com/JuliaLang/julia/pull/26559)

KristofferC:
Well, the docs of `download` says

> For production use or situations in which more options are needed, please use a package that provides the desired functionality instead.

I wanted to easily redirect the output, so that's what I did.

simonbyrne:
We should probably add a "suppress output" option to Base.download then.

staticfloat:
In `Pkg3`, I tend to agree, but `BinaryProvider` has pretty specific requirements in that we want to use the `OutputCollector` functionality which does things like logs to files, interleaves `stdout` and `stderr` on a per-line basis, colorizes output, etc..... We want all of that, and so we can't just use a `download()` function, we need something that provides a `Cmd` that we can then pass to `OutputCollector`, because the only way to get the `stdout` and `stderr` streams directly is to call `spawn()` directly.  So I don't think `BinaryProvider` will be able to use `Base.download()`.

simonbyrne:
I don't see how that changes my argument

staticfloat:
> I don't see how that changes my argument

If by that you mean "this change should still be merged regardless of what `BinaryProvider` does" then yes, I agree and that's what I was trying to communicate.

If by that you mean "My point still stands about only having one `download()` function, and all implementations should use the one implementation in `Base`, I believe my point above is addressing that pretty well; I don't want a `download()` function, I want a `gen_download_cmd()` function, and that doesn't seem like something that would be exported from `Base`, so we resort to writing our own.

simonbyrne:
My understanding (which might be wrong) is that BinaryProvider.jl is intended to become part of the stdlib, in which case we will end up with 2 "download" implementations in the standard Julia distribution.

I don't see why your requests (which are entirely reasonable) can't be accommodated in the standard Julia `download` function, wherever it might live.

That said, this might better be discussed as an issue in the julia repo itself.

simonbyrne:
re: silent download, there is an old PR https://github.com/JuliaLang/julia/pull/15930

KristofferC:
`Base.download` doesn't provide the flexibility we need right now. We can revisit when it does.

PR #281, opened 2018-05-08
Download archives by tree hash instead of tag

simonbyrne:
Since we don't do any validation of the archives, downloading based on tag presents a potential security hole whereby a compromised repository retags a version.

This should fix that by downloading the archive for the tree directly. Note that the documentation (https://developer.github.com/v3/repos/contents/#get-archive-link) says that it should be a valid git reference, but using hashes seems to work as well.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -357,7 +357,9 @@ function get_archive_url_for_version(url::String, ref)
     end
     return nothing
 end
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
It wouldn't hurt to verify the tree hash before continuing for defense in depth, but that's a separate PR.

fredrikekre:
Will this method work if the release was made from a branch that is deleted afterwards?

KristofferC:
It should, the content is still the same.

PR #290, opened 2018-05-17
add an option to test in a "virgin environment" as would be the case when e.g. running on CI

KristofferC:
This is what we discussed a bit @pfitzseb. Running e.g. `pkg> test --project JSON` will now ignore your manifest and run with freshly resolved dependencies, just like what would happen on CI.

TODO: Tests.

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -445,13 +445,16 @@ const helps = Dict(
     is modified.
     """, CMD_TEST => md"""
 
-        test [opts] pkg[=uuid] ...
+        test [-p|project]  [opts] pkg[=uuid] ...
+        test [-m|manifest] [opts] pkg[=uuid] ...
 
         opts: --coverage
 
     Run the tests for package `pkg`. This is done by running the file `test/runtests.jl`
-    in the package directory. The option `--coverage` can be used to run the tests with
-    coverage enabled.
+    in the package directory. If `project` is passed the versions in the current manifest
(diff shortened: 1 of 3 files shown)


src/REPLMode.jl
```diff
@@ -445,13 +445,16 @@ const helps = Dict(
...
-        test [opts] pkg[=uuid] ...
+        test [-p|project]  [opts] pkg[=uuid] ...
+        test [-m|manifest] [opts] pkg[=uuid] ...
 
         opts: --coverage
 
     Run the tests for package `pkg`. This is done by running the file `test/runtests.jl`
-    in the package directory. The option `--coverage` can be used to run the tests with
-    coverage enabled.
+    in the package directory. If `project` is passed the versions in the current manifest
```
pfitzseb: Maybe "If called with `--project` the versions..." instead (same for `mainfest` to `--manifest` below)?

pfitzseb:
I'm not sure if I like calling the options `--project`/`--manifest`, but I guess that's consistent with the rest of Pkg3. Apart from that point (which is just personal preference I guess) this looks good!

StefanKarpinski:
The `--project` option name is a bit unintuitive although it does make sense. Perhaps having a more intuitive alias would be good. Can be added whenever, of course.

IanButterworth:
What about this. It's not super common, so doesn't need to be short IMO
```
pkg> test --ignoremanifest JSON
```

PR #291, opened 2018-05-22
fix cache-repos not getting updated

KristofferC:


src/Operations.jl
```diff
@@ -423,14 +423,13 @@ function install_git(
         GitTools.clone(urls[1], repo_path; isbare=true, header = "[$uuid] $name from $(urls[1])", credentials=creds)
     end
     git_hash = LibGit2.GitHash(hash.bytes)
-    for i = 2:length(urls)
+    for url in urls
```
StefanKarpinski: Maybe skip the first if you just cloned it?

KristofferC: I'm not sure you get all the remotes that we give in `refspecs` when you clone?

StefanKarpinski:
It looks like this bug was in the original version: https://github.com/JuliaLang/Pkg3.jl/blob/a00d6bd7e9156f0a226daf4baebbb0f552742ae4/src/Operations.jl:

* The correct code path is when you clone the repo and then don't update from it;
* The incorrect code path is when you don't clone the repo and then don't update from it.

In the latter case you end up skipping the first (and often only) URL entirely. What about this patch, which preserves the first, correct, case and doesn't update from a just-cloned URL but should fix the second, incorrect, case:
```diff
diff --git a/src/Operations.jl b/src/Operations.jl
index 47cb0a75c..d76cadb13 100644
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -388,12 +388,14 @@ function install_git(
     clones_dir = joinpath(depots()[1], "clones")
     ispath(clones_dir) || mkpath(clones_dir)
     repo_path = joinpath(clones_dir, string(uuid))
+    just_cloned = false
     repo = ispath(repo_path) ? LibGit2.GitRepo(repo_path) : begin
+        just_cloned = true
         printpkgstyle(ctx, :Cloning, "[$uuid] $name from $(urls[1])")
         LibGit2.clone(urls[1], repo_path, isbare=true)
     end
     git_hash = LibGit2.GitHash(hash.bytes)
-    for i = 2:length(urls)
+    for i = (1 + just_cloned):length(urls)
         try with(LibGit2.GitObject, repo, git_hash) do g
             end
             break # object was found, we can stop
```

KristofferC:
As I tried to ask in https://github.com/JuliaLang/Pkg3.jl/pull/291#discussion_r189917043 I am not sure that cloning will give you everything that fetching with `refspecs = ["+refs/*:refs/remotes/cache/*"]` does so we might miss stuff.

And even if it does, we will immidiately break since `with(LibGit2.GitObject, repo, git_hash)` will succeed so it is not like we will end up doing an extra fetch.

PR #296, opened 2018-05-24
update Pkg with changes from Base

KristofferC:


KristofferC:
Somethings up with AV but passes on travis: https://travis-ci.org/JuliaLang/Pkg.jl/jobs/383083262.

PR #308, opened 2018-05-28
fix versionrange with old ranges of the type 0.1.0 0.1.0+

### LINKED ISSUE #293: adding Deprecations.jl fails with unsatisfiable requirement
```
(v0.7) pkg> add https://github.com/JuliaComputing/Deprecations.jl
...
ERROR: Unsatisfiable requirements detected for package CSTParser [00ebfdb7]:
 CSTParser [00ebfdb7] log:
 ├─possible versions are: [0.0.2-0.0.6, 0.1.0-0.1.1, 0.2.0-0.2.2, 0.3.0-0.3.5] or uninstalled
 └─restricted to versions ∅ by (unknown) [0487f878] — no versions left
   └─(unknown) [0487f878] log:
     ├─possible versions are: 0.0.0 or uninstalled
     └─(unknown) [0487f878] is fixed to version 0.0.0
```

Interesting part:
```
restricted to versions ∅ by (unknown) [0487f878]
```

Why does this happen? Wrong parsing of the REQUIRE file?

Also ` (unknown) [0487f878]` should be `(Deprecations)`.

KristofferC:
Fixes #293 

```
(Pkg) pkg> add https://github.com/JuliaComputing/Deprecations.jl
[ Info: Assigning UUID 5a972732-6268-11e8-288b-df1ad7efd8ef to Deprecations
  Updating registry at `~/.julia/registries/Uncurated`
  Updating git-repo `https://github.com/JuliaRegistries/Uncurated.git`
 Resolving package versions...

Downloaded Tokenize ────── v0.4.2
Downloaded AbstractTrees ─ v0.1.0
Downloaded CSTParser ───── v0.3.5
```

@carlobaldassi seems ok to you?

Problem was that when using old REQUIRE files things like `0.1.0 0.1.0+` gave empty version ranges.

### DIFF
--- a/src/Pkg2/types.jl
+++ b/src/Pkg2/types.jl
@@ -28,6 +28,8 @@ function Base.convert(::Type{VersionRange}, a::VersionInterval)
     lb = VersionBound(lower.major, lower.minor, lower.patch)
 
     vb = UInt32[upper.major, upper.minor, upper.patch]
+    # Consider e.g. "0.1.0 0.1.0+" to be the same as VerionRange("0.1.0")
+    # by making the "+" bump the patch version
     if typeof(upper.build) == Tuple{String} && vb[3] != typemax(UInt32)
         vb[3] += 1
     end
--- a/src/Pkg2/types.jl
+++ b/src/Pkg2/types.jl
@@ -28,6 +28,9 @@ function Base.convert(::Type{VersionRange}, a::VersionInterval)
     lb = VersionBound(lower.major, lower.minor, lower.patch)
 
     vb = UInt32[upper.major, upper.minor, upper.patch]
+    if typeof(upper.build) == Tuple{String} && vb[3] != typemax(UInt32)
+        vb[3] += 1
+    end
     i = 3
     while i > 0 && vb[i] == 0
         pop!(vb)


carlobaldassi:
Wouldn't this mean that the upper bound `0.1.0+` is translated into `0.1.0` which then would also include `0.1.0+` as valid? If that's not a concern, then it seems ok to me. (And if that's a concern, I suppose that `VersionRange` is just not equipped to deal with that...)

Apart from this: maybe add a comment? The code itself is kind of cryptic without knowing its purpose.

KristofferC:
This now gives:

```
julia> using Pkg: Types, Pkg2.Pkg2Types

julia> convert(VersionRange, VersionInterval(v"0.1.0", v"0.1.0+"))
VersionRange("0.1.0")
```

Added a comment.

carlobaldassi:
Yes, so the translation is necessarily imperfect:
```
julia> v"0.1.0+" ∈ VersionInterval(v"0.1.0", v"0.1.0+")
false

julia> v"0.1.0+" ∈ VersionRange("0.1.0")
true
```

But I suppose that this patch is better than the previous situation anyway, so I'd say to go for it.

KristofferC:
Yes, I think this is the best we can do and I don't think it will cause problems in practice.

Issue #302, opened 2018-05-25
Project verifier. 


KristofferC:
Since the project might need to be tweaked by hand it would be good to ha e a verify er that can spot errors in it so we don't error out in weird ways later in the code. 

Ref https://discourse.julialang.org/t/problems-after-switch-to-pkg3/11144/5?u=kristoffer.carlsson

PR #309, opened 2018-05-28
WIP: verify Project files after they are parsed

KristofferC:
WIP because this will throw error just by entering the REPL if the project file is invalid with a big scary stacktrace. Perhaps think a bit how to make this more user friendly.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/302

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -592,8 +592,41 @@ function write_env_usage(manifest_file::AbstractString)
     close(io)
 end
 
+const UUID_REGEX = r"^[0-9a-f]{8}(?:-[0-9a-f]{4}){3}-[0-9a-f]{12}$"
+
+function verify_project(p::Dict)
+    errors = String[]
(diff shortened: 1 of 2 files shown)


src/Types.jl
```diff
@@ -592,8 +592,41 @@ function write_env_usage(manifest_file::AbstractString)
...
+
+function verify_project(p::Dict)
+    errors = String[]
+    if haskey(p, "name") && typeof(p["name"]) != String
+        push!(errors, "expected `name` key to have a value of type String")
+    end
+    if haskey(p, "uuid")
+        uuid = p["uuid"]
+        if typeof(uuid) != String || !occursin(UUID_REGEX, uuid)
+            push!(errors, "expectet `uuid` key to be a valid UUID, got \"$uuid\"")
```
StefanKarpinski: "expected"

Issue #306, opened 2018-05-27
Tests fail when run through Pkg


staticfloat:
Hey there, I'm trying to update the `BinaryProvider` tests to work on 0.7.  I've noticed that there seems to be a difference between running `runtests.jl` and running `] test BinaryProvider`.  Looking at the [`sf/new_pkg` branch](https://github.com/JuliaPackaging/BinaryProvider.jl/tree/sf/new_pkg), running `julia runtests.jl` passes, but running `] test BinaryProvider`, it fails due to some problem with the code loading machinery:

```
ERROR: LoadError: ArgumentError: Module LibFoo not found in current path.
Run `Pkg.add("LibFoo")` to install the LibFoo package.
Stacktrace:
 [1] require(::Module, ::Symbol) at ./loading.jl:867
 [2] include at ./boot.jl:314 [inlined]
 [3] include_relative(::Module, ::String) at ./loading.jl:1070
 [4] include(::Module, ::String) at ./sysimg.jl:29
 [5] exec_options(::Base.JLOptions) at ./client.jl:264
 [6] _start() at ./client.jl:424
in expression starting at /home/sabae/.julia/dev/BinaryProvider/test/LibFoo.jl/test/runtests.jl:2
LibFoo.jl: Error During Test at /home/sabae/.julia/dev/BinaryProvider/test/runtests.jl:727
  Got exception ErrorException("failed process: Process(`/home/sabae/local/dist/julia-master/bin/julia -Cnative -J/home/sabae/local/dist/julia-master/lib/julia/sys.so --compile=yes --depwarn=yes --color=yes test/runtests.jl`, ProcessExited(1)) [1]") outside of a @test
  failed process: Process(`/home/sabae/local/dist/julia-master/bin/julia -Cnative -J/home/sabae/local/dist/julia-master/lib/julia/sys.so --compile=yes --depwarn=yes --color=yes test/runtests.jl`, ProcessExited(1)) [1]
  Stacktrace:
   [1] error(::String, ::Base.Process, ::String, ::Int64, ::String) at ./error.jl:42
   [2] pipeline_error at ./process.jl:712 [inlined]
   [3] #run#511(::Bool, ::Function, ::Cmd) at ./process.jl:670
   [4] run at ./process.jl:668 [inlined]
   [5] (::getfield(Main, Symbol("##71#77")))() at /home/sabae/.julia/dev/BinaryProvider/test/runtests.jl:781
[...]

PR #311, opened 2018-05-28
 fix using local projects when testing

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/306. cc @staticfloat

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -777,7 +777,7 @@ function with_dependencies_loadable_at_toplevel(f, mainctx::Context, pkg::Packag
         write_env(localctx, display_diff = false)
         will_resolve && build_versions(localctx, new)
         sep = Sys.iswindows() ? ';' : ':'
-        withenv(f, "JULIA_LOAD_PATH" => "$tmpdir$sep$(Types.stdlib_dir())")
+        withenv(f, "JULIA_LOAD_PATH" => "@|$tmpdir$sep$(Types.stdlib_dir())")
     end
 end
 
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -777,7 +777,7 @@ function with_dependencies_loadable_at_toplevel(f, mainctx::Context, pkg::Packag
         write_env(localctx, display_diff = false)
         will_resolve && build_versions(localctx, new)
         sep = Sys.iswindows() ? ';' : ':'
-        withenv(f, "JULIA_LOAD_PATH" => "@|$tmpdir$sep$(Types.stdlib_dir())")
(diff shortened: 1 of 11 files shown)


src/Operations.jl
```diff
@@ -777,7 +777,7 @@ function with_dependencies_loadable_at_toplevel(f, mainctx::Context, pkg::Packag
         write_env(localctx, display_diff = false)
         will_resolve && build_versions(localctx, new)
         sep = Sys.iswindows() ? ';' : ':'
-        withenv(f, "JULIA_LOAD_PATH" => "$tmpdir$sep$(Types.stdlib_dir())")
+        withenv(f, "JULIA_LOAD_PATH" => "@|$tmpdir$sep$(Types.stdlib_dir())")
```
StefanKarpinski: This is certainly the right fix to add the current env to the path (i.e. look in parent directories for something that looks like a project or a git repo). I'm not quite sure that I understand the scenario that @staticfloat is hitting or why this is the right fix. Can you explain a bit?

staticfloat: Yes; this is for the `BinaryProvider` tests; one of the things I wanted to test was that the whole `Pkg.*` machinery works, so I have a package called `LibFoo.jl` [embedded within `BinaryProvider/test`](https://github.com/JuliaPackaging/BinaryProvider.jl/tree/master/test/LibFoo.jl).  With `Pkg2`; getting `Pkg.build()` and `Pkg.test()` to be able to reach this package is pretty easy; [you can just set `JULIA_LOAD_PATH` and be on your way](https://github.com/JuliaPackaging/BinaryProvider.jl/blob/74f50756a0bd96e874fa82d36933cf867af0abe1/test/runtests.jl#L771-L782). 0.7 makes it a little more complicated because doing so will allow `LibFoo` to be loaded, but `LibFoo` will then not be able to load `Compat` or `BinaryProvider`.  What I ended up doing was adding a `Project.toml` for `LibFoo` as well, and that convinced `Pkg3` to expose `Compat` and `BinaryProvider` to `LibFoo.jl` (Which is such an amazingly simple and staightforward fix, I love it) but that only worked when running directly from the command line, e.g. `julia test/runtests.jl` from within the `BinaryProvider` directory would be able to pass all tests, including those of the nested `LibFoo.jl`.  Running `Pkg.test("BinaryProvider")` would fail and be unable to load `LibFoo.jl`.

staticfloat:
How can I test this locally?  Is there a way to override the `Pkg` that my local julia master checkout uses?

src/Operations.jl
KristofferC: Ah, I think I want `$sep` here.

staticfloat:
Confirmed that this fixes all my `BinaryProvider` problems.  Thanks @KristofferC!

PR #320, opened 2018-05-30
Some more tests for union! and isjoinable

kshyatt:
Couple missing branch paths here ... most of these types don't define a `==` so I went digging into their fields, because that seems better than adding new methods just for these tests.


test/pkg.jl
```diff
@@ -180,23 +202,6 @@ temp_pkg_dir() do project_path
...
-                    cd("FailBuildPkg") do
-                        write_build(pwd(), "error()")
-                        @test_throws CommandError Pkg.build()
-                    end
-                end
-            finally
-                popfirst!(LOAD_PATH)
-            end
-        end
-    end
```
StefanKarpinski: This PR shouldn't remove these test, right?

kshyatt: lol wtf happened there

PR #352, opened 2018-06-07
don't delete build.log file on success

### LINKED ISSUE #347: build.log file disappears?
Moved from https://github.com/JuliaLang/julia/issues/27463

In the process of updating packages for 0.7, I notice that building packages doesn't seem to leave the promised `build.log` file:
```jl
(v0.7) pkg> build ZMQ
  Building ZMQ → `~/.julia/dev/ZMQ/deps/build.log`
 Resolving package versions...

julia> isfile(homedir() * "/.julia/dev/ZMQ/deps/build.log")
false
```
By watching the directory in anothe terminal window, I see that an empty `build.log` exists while the build is occurring, but it never seems to contain anything, and it is apparently deleted when the build completes.  This doesn't seem very useful?

stevengj:
Closes #347.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -933,9 +933,7 @@ function build_versions(ctx::Context, uuids::Vector{UUID}; might_need_to_resolve
             ok = open(log_file, "w") do log
                 success(pipeline(cmd, stdout=log, stderr=log))
             end
-            if ok
-                Base.rm(log_file, force=true)
-            else
+            if !ok
                 @error("Error building `$name`; see log file for further info")
                 build_succeeded = false
             end


KristofferC:
I wonder if we should just print as before?

StefanKarpinski:
The log output is pretty verbose, I prefer just clicking on a file, but I guess printing would be fine too.

stevengj:
I would prefer leaving it in the file as well... often, when you build multiple packages, you are only interested in the build output from one of them, and seeing it all at once is confusing.   And most of the time (when all goes well) you aren't interested in seeing the build output at all.

KristofferC:
How would one currently diagnose build errors on CI?

fredrikekre:
Perhaps the default CI scripts could do something like `Pkg.build($Package, verbose = true)`?

KristofferC:
Or print the build output to stdout as well if not in isinteractive().

PR #356, opened 2018-06-08
Tilder vs Carrot example match to words

oxinabox:
I think the text was backwards?


StefanKarpinski:
Looks correct to me but @KristofferC should confirm that this is the right doc fix.

oxinabox:
This can be closed in favor of #348, yes?

PR #357, opened 2018-06-08
fix missing stdlibs; update bin/*.toml files

StefanKarpinski:
`OldPkg` and `Sockets` were missing from the list of stdlibs in the `bin` scripts. The `bin/sha1map.toml` and `bin/stdlib.toml` maps also hadn't been updated in a long time so this took a really long time to run from scratch. @KristofferC, maybe it would be good to push updates from the version of this script that you run regularly? Also, what do I need to do to get this change to be picked up there?


bin/gitmeta.jl
```diff
@@ -3,7 +3,7 @@
 import Pkg.TOML
 import LibGit2
 
-const STDLIBS = [
+const STDLIBS = sort!(by = lowercase, [
```
fredrikekre: TIL you can use kwargs like this, thats pretty neat.

StefanKarpinski: Yup, keyword and positional args can be interleaved, a nice aspect of keeping them distinct.

KristofferC:
Please squash merge or rebase merge. Makes it easier to cherry-pick.

StefanKarpinski:
This was squashed.

KristofferC:
Sorry, it was the `compat` rename PR.

StefanKarpinski:
Yeah, I'm not used to rebase being an option since it was turned off on JuliaLang/julia for so long.

PR #363, opened 2018-06-09
unmixup `^` and `~` in "Version specifier format"

JeffreySarnoff:
Some `^`s were called "tilde" and vice-versa.  One example in that section was fixed.


KristofferC:
Thanks.

Issue #362, opened 2018-06-08
malformed manifest file causes prompt to crash


StefanKarpinski:
Try deleting a dependency stanza from the manifest file and then hitting `]` get this:
```
julia> ┌ Error: Error in the keymap
│   exception =
│    KeyError: key "Compat" not found
│    Stacktrace:
│     [1] getindex(::Dict{String,Any}, ::String) at ./dict.jl:487
│     [2] read_manifest(::IOStream) at /Users/stefan/projects/julia/usr/share/julia/stdlib/v0.7/Pkg/src/Types.jl:615
[...]
```

PR #368, opened 2018-06-12
fix repl crashing with invalid env files

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/362

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -913,14 +913,21 @@ function completions(full, index)
 end
 
 function promptf()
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
We should also probably make it so that loading a broken file doesn't fail but defense in depth.

KristofferC:
We will crash before any command is run anyway, but it is a bit annoying to crash when scrolling through e.g. your REPL history so this is at least some improvement.

PR #370, opened 2018-06-13
make generate start packages at 1.0.0 instead of 0.1.0

KristofferC:
See e.g. https://docs.npmjs.com/getting-started/semantic-versioning#semver-for-publishers. SemVer before 1.0 is "under specified" and while we make the same interpretation as cargo when it comes to pre 1.0.0 versions, it is just better if things start at 1.0.0.

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -55,7 +55,7 @@ function project(pkg::String, dir::String; preview::Bool)
             authors = $authorstr
             name = "$pkg"
             uuid = "$(UUIDs.uuid1())"
-            version = "0.1.0"
+            version = "1.0.0"
 
             [deps]
             """


quinnj:
Sorry for the unrelated comment, but speaking of `generate`, why don't we generate a `test/runtests.jl` file as well?

KristofferC:
Because at the moment it was written we didn't have test-specific dependencies and I wanted to put `Test` in there. We do now so it will be updated soon to include it.

StefanKarpinski:
I would propose that we officially promote the following interpretation/extension of semver for pre-1.0 versions (which we are free to do since semver says they are open to do what one wants with):

* `0.0.1`: very, very unstable software still in the API sketching phase. There will only be API-breaking releases (`0.0.2`), no bug fix releases and no feature-only releases.

* `0.1.0`: not-fully-mature software. The API is still being worked out while features are being added so the primary focus of development is iterating on the API with breaking releases (`0.2.0`), but the software is stable enough to use and bug fix releases can be expected (`0.1.1`).

* `1.0.0`: mature software for with a full release lifecycle. The API is well-established and most development is adding features without breaking things (`1.1.0`). You can expect bug fix releases (`1.0.1`). If and when API-breaking changes need to be made, there will be a new major version release (`2.0.0`).

Given that interpretation, the question is which one of `0.0.1`, `0.1.0` or `1.0.0` is the best default choice for a newly registered package. I would argue that `0.1.0` is still the best default:

* Arguably `0.0.1` software shouldn't be registered at all. If we make it easy enough to let people install pre-release packages and work on the master version of something, I think we can generally discourage people from registering packages that aren't ready for a 0.1 release at all.

* I don't think there's anything wrong with publishing and expecting people to use `0.1.0` software as long as they understand that the API is likely to change. That's exactly what we've done with Julia itself, after all. Having a pre-1.0 version number helps communicate that expectation.

* It seems hypocritical to insist that all packages emerge with fully formed APIs that are ready for a `1.0.0` version number, which implicitly says that they are done figuring out the API and do not expect to change it for many years.

KristofferC:
> which implicitly says that they are done figuring out the API and do not expect to change it for many years.

PR #375, opened 2018-06-13
Backport and fix resolve

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/376


StefanKarpinski:
Thanks for the fix. I should have asked for your input that I was doing that right 😬

PR #380, opened 2018-06-13
 update docs, also make failed build packages print their output to console

KristofferC:


docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+  Updating registry at `~/.julia/registries/Uncurated`
+  Updating git-repo `https://github.com/JuliaRegistries/Uncurated.git`
+ Resolving package versions...
+  Updating `~/.julia/environments/v0.7/Project.toml`
+  [7876af07] + Example v0.5.1
+  Updating `~/.julia/environments/v0.7/Manifest.toml`
+  [7876af07] + Example v0.5.1
+  [8dfed614] + Test
+```
+
```
jrevels: In the very short time I've spent with the new `Pkg`, one of the biggest comprehension hurdles has been trying to understand/gain intuition for the differences/relationships between "projects", "environments", "registries" and "packages". For this reason, it might be helpful to have some explicit clarifying language immediately following this code example, e.g. 

"Here, we add the package `Example` to the current *project*.  In this example, we are using a fresh Julia installation, and this is our first time `add`ing a package using `Pkg`. By default, `Pkg` clones Julia's `Uncurated` *registry*, and uses this registry to look up packages requested for inclusion in the current *environment*." 

For all I know, this specific explanation/language isn't even correct, but you get my point 😛

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+Since this is the first time adding anything, the [default registry](https://github.com/JuliaRegistries/Uncurated) is cloned.
+The status update shows a short form of the package UUID to the left, then the package name, and the version.
+Since standard libraries (e.g. `Test`) are shipped with Julia, they do not have a version. The project status contains the packages
+you have added yourself, in this case, `Example`. The manifest status, in addition, includes the dependencies of explicitly
+added packages.
```
jrevels: Maybe show the output of `pkg> status` directly after `The project status contains...`.

Same for `The manifest status...`, but show the actual manifest status (is that just the content of the `Manifest.toml`)?

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
 To return to the `julia>` prompt, either press backspace when the input line is empty or press Ctrl+C.
 Help is available by calling `pkg> help`.
 
-To generate files for a new project, use `pkg> generate`.
+
+### Adding packages
+
+There are two ways of adding packages, either using the `add` command or the `dev` command.
+The most frequently used one is `add` and its usage is described first.
```
fredrikekre: Clarify that these are Pkg-REPL commands? Also, when the API is complete it would be nice with a table of them, for example:

| Operation | Pkg REPL | API |
|:- |:- |:- |
| Add a package | `add Foo` | `Pkg.add("Foo)` |

etc

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+
+#### Adding registered packages
+
+```
```
fredrikekre: Prepend something like "In the Pkg REPL packages can be added with the `add` command followed by the name of the package, for example:"

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+  Updating `~/.julia/environments/v0.7/Project.toml`
+  [7876af07] + Example v0.5.1
+  Updating `~/.julia/environments/v0.7/Manifest.toml`
+  [7876af07] + Example v0.5.1
+  [8dfed614] + Test
+```
+
+Since this is the first time adding anything, the [default registry](https://github.com/JuliaRegistries/Uncurated) is cloned.
+The status update shows a short form of the package UUID to the left, then the package name, and the version.
+Since standard libraries (e.g. `Test`) are shipped with Julia, they do not have a version. The project status contains the packages
```
fredrikekre: Can/Should we set the version to the same as Julia?

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+you have added yourself, in this case, `Example`. The manifest status, in addition, includes the dependencies of explicitly
```
fredrikekre: But `Test` was not added explicitly in this example?

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+After a package is added to the project, it can be loaded in julia:
+
+```
+julia> using Example
+
+julia> Example.hello("User")
+"Hello, User"
+```
+
+A specific version can be installed by giving by appending a version after a `@` symbol, e.g. `@v0.4`, to the package name:
```
fredrikekre: " by giving by appending"

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+  [7876af07] ~ Example v0.5.1+ #master [https://github.com/JuliaLang/Example.jl.git] ⇒ v0.5.1
+```
+
+
+#### Adding unregistered packages
+
+If a package is not in a registry, it can still be added by instead of the package name giving the URL to the repository to `add`.
+
+```
+(v0.7) pkg> add https://github.com/fredrikekre/ImportMacros.jl
```
fredrikekre: lol

KristofferC: <3

KristofferC: u famous

docs/src/index.md
StefanKarpinski: There is a glossary for this very reason: https://docs.julialang.org/en/latest/stdlib/Pkg/#Glossary-1

docs/src/index.md
```diff
@@ -193,7 +193,292 @@ Since we haven't created our own project yet, we are in the default project, loc
...
+This is done using the `init` command. Below we make a new directory and create a new project in it:
+
+```
+shell> mkdir MyProject
+
+shell> cd MyProject
+/Users/kristoffer/MyProject
+
+(v0.7) pkg> init
+Initialized project at /Users/kristoffer/MyProject/Project.toml
```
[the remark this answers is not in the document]
KristofferC: Merging `init` and `generate` into a common command seems to make sense to me. `create --package PackageName` or something to generate the package skeleton and just `create Project` to create a naked Project.toml in `Project`?

docs/src/index.md
jrevels: Right, I've read the glossary, but I'm not confident that I know how to use these words/concepts correctly just given their concise technical definitions. I'm sure many other users will feel the same.

This section (and the other sections here, since this seems to be a proper manual rather than straight API documentation) is an excellent place to demonstrate these words/concepts to use in context.

docs/src/index.md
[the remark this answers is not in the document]
staticfloat: Agreed; `create` is much clearer to me than `init` (I get confused with `OldPkg.init()`.

src/Operations.jl
```diff
@@ -937,8 +937,8 @@ function build_versions(ctx::Context, uuids::Vector{UUID}; might_need_to_resolve
                 success(pipeline(cmd, stdout=log, stderr=log))
             end
             if !ok
-                @error("Error building `$name`; see log file for further info")
-                build_succeeded = false
+                log_content = read(log_file, String)
+                @error("Error building `$name`: \n$log_content")
```
staticfloat: I would only print out the last X lines.  In BinaryBuilder, we print out the last 100 lines of the log if things go south. These logs can get pretty heavy, if someone is e.g. building TensorFlow from source, you don't want to spam the console with hundreds of thousands of lines of text; but you do want to store it so the savvy user can look into it.  I would still print out the location of the logfile and also tail it out to the user.

docs/src/index.md
Nosferican: Aren't `stdlib` suppose to be able to publish releases without waiting for next language release? Working like other packages?

docs/src/index.md
```diff
@@ -285,130 +570,90 @@ julia> HelloWorld.greet_alien()
...
 ```
 
-do update the dependencies and then precompile them.
+#### Test-specific dependencies
 
-## Preview mode
-
-If you just want to see the effects of running a command, but not change your state you can `preview` a command.
-For example:
+Sometimes one might to want to use some packages only at testing time but not enforce a dependency on them when the package is used.
```
Nosferican: Nice addition.

docs/src/index.md
```diff
@@ -454,7 +699,7 @@ Some examples are shown below.
 While the semver specification says that all versions with a major version of 0 are incompatible with each other, we have made that choice that
 a version given as `0.a.b` is considered compatible with `0.a.c` if `a != 0` and  `c >= b`.
 
-#### Tilde specifiers
+##### Tilde specifiers
```
Nosferican: Could there be an example of using Caret or Tilde specifiers in a `project.toml` to specify the dependency versions?

docs/src/index.md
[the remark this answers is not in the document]
StefanKarpinski: It does approximately what the old `init` did though, hence the name.

src/Operations.jl
KristofferC: Alright, makes sense.

docs/src/index.md
KristofferC: Yes, but this is a separate discussion from the PR.

docs/src/index.md
KristofferC: "*the dependencies* of explicitly added packages"

PR #393, opened 2018-06-17
Minor corrections to index.md

omus:


omus:
I'm not 100% sure about the "build" directory change. Maybe "build" is the new preferred name but since the rest of the documentation only referenced "deps" I thought it was probably a typo.

KristofferC:
Thanks

Issue #200, opened 2018-03-14
command spellings


StefanKarpinski:
I kind of threw in a bunch of different spellings of commands for kicks, including some symbols (`+`, `-`, `/`, `?`), but we should decide on which ones we actually want. Do we want to keep one short and one long for each command? Keep the symbols or ditch them?

KristofferC:
`?` for help seems ok but `+` and `-` are perhaps excessive since the commands themselves are short

StefanKarpinski:
Agree. There's still the issue of aliases like `rm` and `remove` which are fine most of the time but become needless clutter when tab completing for example. It seems silly when someone types `r<tab>` to list both `rm` and `remove` when they do the same thing.

pfitzseb:
Imho there's no real need for the short forms when autocomplete works.

StefanKarpinski:
I strongly prefer the shorter forms. If anything, I'd be more in favor of _only_ having the short forms.

KristofferC:
```
    "help"      => CMD_HELP,
    "?"         => CMD_HELP,
    "st"        => CMD_STATUS,
    "add"       => CMD_ADD,
    "rm"        => CMD_RM,
    "up"        => CMD_UP,
    "test"      => CMD_TEST,
    "gc"        => CMD_GC,
    "preview"   => CMD_PREVIEW,
    "init"      => CMD_INIT,
    "build"     => CMD_BUILD,
    "pin"       => CMD_PIN,
    "free"      => CMD_FREE,
    "develop"   => CMD_DEVELOP,
    "generate"  => CMD_GENERATE,
```
?

StefanKarpinski:
Yes, but `dev` and `gen` instead as well 😬

StefanKarpinski:
The only one that's a bit _too_ short is `st` for `status` but I use that alias all the time for git so it's hard to imagine having to type `status` all the time.

pfitzseb:
You *could* just write `s\tab` though :)

KristofferC:
> Yes, but `dev` and `gen` instead as well 

plzno

StefanKarpinski:
The only reason to have the short versions is for people who prefer typing a short name to tab completion (which e.g. I strongly do) and the main issue is double listing in the tab completion UI. Perhaps the solution is to only list long names when doing tab completion unless the only viable completions are abbreviations.

PR #398, opened 2018-06-18
clean up some REPL commands

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/200

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -28,20 +28,11 @@ const cmds = Dict(
     "?"         => CMD_HELP,
     "status"    => CMD_STATUS,
     "st"        => CMD_STATUS,
-    "."         => CMD_STATUS,
-    "search"    => CMD_SEARCH,
-    "find"      => CMD_SEARCH,
-    "/"         => CMD_SEARCH,
     "add"       => CMD_ADD,
-    "install"   => CMD_ADD,
-    "+"         => CMD_ADD,
     "rm"        => CMD_RM,
     "remove"    => CMD_RM,
-    "uninstall" => CMD_RM,
-    "-"         => CMD_RM,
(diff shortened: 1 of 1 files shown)


src/REPLMode.jl
```diff
@@ -28,28 +28,18 @@ const cmds = Dict(
...
     "add"       => CMD_ADD,
-    "install"   => CMD_ADD,
-    "+"         => CMD_ADD,
     "rm"        => CMD_RM,
     "remove"    => CMD_RM,
-    "uninstall" => CMD_RM,
-    "-"         => CMD_RM,
     "up"        => CMD_UP,
     "update"    => CMD_UP,
-    "upgrade"   => CMD_UP,
```
StefanKarpinski: Not sure if we should get rid of this one: the operation it's doing is an upgrade after all. I had proposed using the distinction between `update` and `upgrade` to differentiate getting a minor update versus doing a potentially-breaking `upgrade` but people didn't like the subtlety of the difference, which is why I made them aliases of each other, abbreviated to `up` for either one.

pfitzseb: Yep, the only potentially annyoing thing about keeping both is that right now `up\tab` shows both. We could canonicalize that to `update` though.

KristofferC: Sure it does an `upgrade` and `remove` does an `uninstall` but keeping a bunch of different verbs that are almost synonyms seems unnecessary. One short form and one long form should be enough.

StefanKarpinski: Fair enough.

staticfloat: Agreed; whenever I see both `update` and `upgrade` I get flashbacks to `brew` where they actually mean different things.

Issue #321, opened 2018-05-30
add doesn't support SSH clones


tkoolen:
```julia
(pkgtest) pkg> add git@github.com:jrevels/Cassette.jl.git
ERROR: ArgumentError: invalid version range: "github.com:jrevels/Cassette.jl.git"
Stacktrace:
 [1] Pkg.Types.VersionRange(::SubString{String}) at /Users/twan/code/julia/julia/usr/share/julia/stdlib/v0.7/Pkg/src/Types.jl:161
 [2] tokenize!(::Array{SubString{String},1}) at /Users/twan/code/julia/julia/usr/share/julia/stdlib/v0.7/Pkg/src/REPLMode.jl:214
 [3] tokenize(::String) at /Users/twan/code/julia/julia/usr/share/julia/stdlib/v0.7/Pkg/src/REPLMode.jl:177
 [4] #do_cmd#8(::Bool, ::Function, ::REPL.LineEditREPL, ::String) at /Users/twan/code/julia/julia/usr/share/julia/stdlib/v0.7/Pkg/src/REPLMode.jl:231
 [5] do_cmd at /Users/twan/code/julia/julia/usr/share/julia/stdlib/v0.7/Pkg/src/REPLMode.jl:230 [inlined]
 [6] (::getfield(Pkg.REPLMode, Symbol("##27#30")){REPL.LineEditREPL,REPL.LineEdit.Prompt})(::REPL.LineEdit.MIState, ::Base.GenericIOBuffer{Array{UInt8,1}}, ::Bool) at /Users/twan/code/julia/julia/usr/share/julia/stdlib/v0.7/Pkg/src/REPLMode.jl:948
 [7] top-level scope
```

I guess part of the problem is that the `@` conflicts with the `Pkg` version syntax.

KristofferC:
Yeah, the Pkg REPLMode has for a long time tokenized the input based on some chars, e.g. `@` and decides *at tokenization time* that what comes after `@` should be a version range. I think we need to restructure the REPL parsing code a bit for this to work.

KristofferC:
Note that as a workaround you can use `Pkg.add("git@github.com:jrevels/Cassette.jl.git")`:

```
julia> Pkg.add("git@github.com:jrevels/Cassette.jl.git")
   Cloning git-repo `git@github.com:jrevels/Cassette.jl.git`
Private key location for 'git@github.com' [/Users/kristoffer/.ssh/id_rsa]:
[...]

PR #400, opened 2018-06-18
support git@blabla in add

innerlee:
How to test this?

Fix https://github.com/JuliaLang/Pkg.jl/issues/321

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -405,6 +405,7 @@ const helps = Dict(
     pkg> add Example#master
     pkg> add Example#c37b675
     pkg> add https://github.com/JuliaLang/Example.jl#master
+    pkg> add git@github.com:JuliaLang/Example.jl.git
     pkg> add Example=7876af07-990d-54b4-ab0e-23690620f79a
     ```
     """, CMD_RM => md"""
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -167,7 +167,7 @@ end
 ################
 # REPL parsing #
 ################
-const lex_re = r"^[\?\./\+\-](?!\-) | [^@\#\s;]+\s*=\s*[^@\#\s;]+ | \#\s*[^@\#\s;]* | @\s*[^@\#\s;]* | [^@\#\s;]+|;"x
+const lex_re = r"^[\?\./\+\-](?!\-) | git@[^@\#\s;]+ | [^@\#\s;]+\s*=\s*[^@\#\s;]+ | \#\s*[^@\#\s;]* | @\s*[^@\#\s;]* | [^@\#\s;]+|;"x
 
 const Token = Union{Command, Option, VersionRange, String, Rev}
 
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -167,7 +167,7 @@ end
 ################
 # REPL parsing #
 ################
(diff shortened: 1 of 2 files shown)


KristofferC:
Could add a test to `repl.jl` that tests something like `add git@github.com:JuliaLang/Example.jl.git`.

pfitzseb:
It would probably be better to check that the preceding "word" is a valid package name and the `@` is followed by a valid version range, since you could have a package called `git`.

innerlee:
So I do not need to worry about the credentials, right?

src/REPLMode.jl
```diff
@@ -167,7 +167,7 @@ end
 ################
 # REPL parsing #
 ################
-const lex_re = r"^[\?\./\+\-](?!\-) | [^@\#\s;]+\s*=\s*[^@\#\s;]+ | \#\s*[^@\#\s;]* | @\s*[^@\#\s;]* | [^@\#\s;]+|;"x
+const lex_re = r"^[\?\./\+\-](?!\-) | git@[^@\#\s;]+ | [^@\#\s;]+\s*=\s*[^@\#\s;]+ | \#\s*[^@\#\s;]* | @\s*[^@\#\s;]* | [^@\#\s;]+|;"x
```
StefanKarpinski: That `git@` seems very specific, but I guess it gets the job done for now.

innerlee: How about [this regex](https://stackoverflow.com/a/22312124)
`((git|ssh|http(s)?)|(git@[\w\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)?`

innerlee:
Okay, so the credential thing is needed. Any workaround?

```bash
  [8dfed614] - Test 
   Cloning git-repo `git@github.com:JuliaLang/Example.jl.git`
Private key location for 'git@github.com':
No output has been received in the last 10m0s, this potentially indicates a stalled build or something wrong with the build itself.
Check the details on how to adjust your build configuration on: https://docs.travis-ci.com/user/common-build-problems/#Build-times-out-because-no-output-was-received
The build has been terminated
```

src/REPLMode.jl
```diff
@@ -167,7 +167,7 @@ end
 ################
 # REPL parsing #
 ################
-const lex_re = r"^[\?\./\+\-](?!\-) | [^@\#\s;]+\s*=\s*[^@\#\s;]+ | \#\s*[^@\#\s;]* | @\s*[^@\#\s;]* | [^@\#\s;]+|;"x
+const lex_re = r"^[\?\./\+\-](?!\-) | ((git|ssh|http(s)?)|(git@[\w\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)? | [^@\#\s;]+\s*=\s*[^@\#\s;]+ | \#\s*[^@\#\s;]* | @\s*[^@\#\s;]* | [^@\#\s;]+|;"x
```
KristofferC: This regex is becoming quite a beast :D. Oh well.

KristofferC:
We should probably change the REPL parsing a bit but this should be good enough for now. Thanks!

tkoolen:
Thank you!

src/REPLMode.jl
StefanKarpinski: It has an `x` modifier so it can be split across lines.

omus: Feel free to take inspiration from: https://github.com/JuliaLang/julia/blob/master/stdlib/LibGit2/src/utils.jl#L5-L26

PR #402, opened 2018-06-18
Update index.md

raminammour:
Exception for non git repositories dependencies (use `dev` not  `add`).

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -574,16 +574,29 @@ since we are inside the `HelloWorld` project directory):
 Both `Random` and `JSON` got added to the project’s `Project.toml` file, and the resulting dependencies got added to the `Manifest.toml` file.
 The resolver has installed each package with the highest possible version, while still respecting the compatibility that each package enforce on its dependencies.
 
-We can now use both `Random` and `JSON` in our project. Changing `src/HelloWorld.jl` to
+To add a dependency on a *local* package that is *not* a git repository, we can use the `develop` command:
+
+```
+(HelloWorld) pkg> develop /PATH/TO/LOCAL/PACKAGE/MyGreatPackage
+ Resolving package versions...
+  Updating "~/Documents/HelloWorld/Project.toml"
+  [f2b96e9e] + MyGreatPackage v0.1.0 /PATH/TO/LOCAL/PACKAGE/MyGreatPackage
+  Updating "~/Documents/HelloWorld/Manifest.toml"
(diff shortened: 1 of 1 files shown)


KristofferC:
This seems already described in the `Developing a package`-section?

raminammour:
Does that cover adding a local package as a dependency to a new package? 

Also remember that most of everyone (including myself) have just started using Pkg3, so all of our old local packages don't have any `*.toml` files. All statements like `using LocalPackage` (which work in the repl because of `LOAD_PATH`)  will fail inside a module.

Am I missing something here?

KristofferC:
I must say I am a bit confused about that description as well. I have no problems loading packages without Project files... @StefanKarpinski perhaps you can elaborate on what that means.

raminammour:
Just to be clear, the behavior I see on `0.7-alpha` (I don't know if this is still the same) is this:
1- Let `OldPkg` be an old package (not a git repo) without project file or uuid
2- `OldPkg` sits somewhere in `LOAD_PATH`
3- From REPL: `using OldPkg` -> works
4- From REPL:  
```
(v0.7) pkg> generate NewPackage
Generating project NewPackage:
    NewPackage/Project.toml
    NewPackage/src/NewPackage.jl
```
5-  cd("NewPackage/")
6- 
```
(NewPackage) pkg> develop /PATH/OldPkg
Assigning UUID 2b5dbb6a-7330-11e8-0a11-4f0db32bfeec to OldPkg
 Resolving package versions...
  Updating `Project.toml`
  [2b5dbb6a] + MyUtils v0.0.0 [`~/OldPkg`]
  Updating `Manifest.toml`
  [2b5dbb6a] + MyUtils v0.0.0 [`~/OldPkg`]
```
7- Inside `NewPkg` (new environment): `using NewPkg` works
8- Outside `NewPkg` (`v0.7` default environment): `using NewPkg` errors:
`ERROR: LoadError: ArgumentError: Module `OldPkg` not found in current path.`
9- Step 8 is a failure even if `OldPkg` is in `v0.7` environment

Note: If `OldPkg` is replaced with `NewPkg2` (built with `generate` step 8 works if `NewPkg` is in `LOAD_PATH`)

Is this behavior expected and/or documented?

KristofferC:
I think the problem is the `Assigning UUID 2b5dbb6a-7330-11e8-0a11-4f0db32bfeec to OldPkg` part.
Right now, we give a package a random UUID but that will not be the same for the second dev. If these would be consistent, I think `9-` should work.

KristofferC:
I made a PR that should probably fix point 9.

PR #392, opened 2018-06-15
WIP: add package converter (creating Project.toml from REQUIRE)

KristofferC:
As an example:

```
julia> Pkg.create_project("DataFrames")
[ Info: Registered package, using already given UUID: a93c6f00-e57d-5684-b7b6-d8193f3e46c0
[ Info: Added Project.toml to /Users/kristoffer/.julia/dev/DataFrames

shell> cat DataFrames/Project.toml
name = "DataFrames"
uuid = "a93c6f00-e57d-5684-b7b6-d8193f3e46c0"
version = "0.11.6"

[deps]
CategoricalArrays = "324d7699-5711-5eae-9e2f-1d82baa6b597"
CodecZlib = "944b1d66-785c-5afd-91f1-9de20f533193"
Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
DataStreams = "9a8bc11e-79be-5b39-94d7-1ccc349a1a85"
Missings = "e1d29d7a-bbdc-5cf2-9ac0-f12de2c33e28"
Reexport = "189a3867-3050-52da-a836-e630ba90ab69"
SortingAlgorithms = "a2af1166-a08f-5f64-846c-94a0d3cef48c"
StatsBase = "2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91"
TranscodingStreams = "3bb67fe8-82b1-5028-8e26-92a6c54297fa"
WeakRefStrings = "ea10d353-3f73-51f8-a26c-33c1cb351aa5"

[compat]
CategoricalArrays = "0.3.5"
CodecZlib = "0.4.0"
Compat = "0.59.0"
DataStreams = "0.3.0"
Missings = "0.2.3"
StatsBase = "0.11.0"
WeakRefStrings = "0.4.0"

[targets.test.deps]
DataStructures = "864edb3b-99cc-5e75-8d2d-829cb0a9cfe8"
LaTeXStrings = "b964fa9f-0449-5b57-a5c2-d3ea65f4040f"
```

The tricky thing is what to set the `compat` section to. The problem is that packages do not agree on what the minor version is used for pre 1.0. Many packages sees bumping the minor version as breaking but other do not.

So in this example `Compat = "0.59.0"` would give the range `[0.59.0, 0.60.0)` which would exclude a bunch of the recent Compat versions which is wrong.
[...]

PR #410, opened 2018-06-19
give packages without project uuids based on their name instead of a random uuid

KristofferC:
Should fix https://github.com/JuliaLang/Pkg.jl/pull/402#issuecomment-398176050 point 9.

Would be good with a test but I don't want to spend too much time adding test for things that are just for compatibility with REQUIRE.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -599,7 +599,8 @@ function parse_package!(ctx, pkg, project_path)
         if !is_registered
             # This is an unregistered old style package, give it a UUID and a version
             if !has_uuid(pkg)
-                pkg.uuid = uuid5("UNREG_PKG", pkg.name)
+                uuid_unreg_pkg = UUID(0xa9a2672e746f11e833ef119c5b888869)
+                pkg.uuid = uuid5(uid_unreg_pkg, pkg.name)
                 @info "Assigning UUID $(pkg.uuid) to $(pkg.name)"
             end
             pkg.version = v"0.0"
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -600,7 +600,7 @@ function parse_package!(ctx, pkg, project_path)
             # This is an unregistered old style package, give it a UUID and a version
             if !has_uuid(pkg)
                 uuid_unreg_pkg = UUID(0xa9a2672e746f11e833ef119c5b888869)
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Won't this lead to different people all getting the same UUID if they happen to pick the same name, which is not ok at all.

KristofferC:
> different people all getting the same UUID if they happen to pick the same name

Should be pointed out that this is not when creating new packages with `generate`. This is uppon adding (from a repo) or devving a package that neither have a Project file, nor is registered. This is purely a convenience thing while things are moving to Projects to allow people to incrementally add Projects to their packages. Before, the following situation would arise:

Local package A with project file:
 `dev local/Q ->  uuid_Q_1`
(we have package Q uuid `uuid_Q_1`).

Project B:
`dev local/Q -> uuid_Q_2`
(this time Q got uuid `uuid_Q_2`).

Project B:
`dev A`
 
So `A` depends on `Q` with `uuid_Q_1` but the one we have in our current manifest is `uuid_Q_2`.
Package `A` will thus fail to load from project `B`. This is because we gave a different UUID to the same package `Q`.
If we have no uuid (no project file, no registry) we assume that they are the same package (which is what the old Pkg does, and that is what is emulated here until things get project files).

StefanKarpinski:
What about generating a UUID based on the Mac address and realpath to the package then?

raminammour:
What about just doing supporting `generate` would have done for old existing local package (no `*.toml`, non git). And just issuing an error upon:
```
dev local/Q
Error: run this command on `local/Q` to generate uuid... and run dev local/Q again
```
I for one would want to fix all my old existing local packages once and for all :)

If there is a way to do this already without renaming folder/creating folder with same name/ copying, I would love to hear about it...

KristofferC:
The PR that adds this functionality is https://github.com/JuliaLang/Pkg.jl/pull/392 but there is still some stuff left with the version bounds on dependencies to work out how to convert in the best way.

KristofferC:
The previous package manager identified packages by name so that is what we have to do as well to use old packages smoothly. Again, this is just a convenience thing so people don't get inexplicable uuid inconcistencies when using their old packages. I really don't think this will cause any problems and if it does we can revisit.

StefanKarpinski:
Ok, carry on then.

PR #412, opened 2018-06-19
Add `start` command to REPL API

00vareladavid:
I figure most people issue a `cd` after `generate` when starting a new project (at least that's what makes sense to me, why would you create a project and then go do something else?). From a usability standpoint, it makes sense to provide a command to do both. Thoughts?

Note: this PR is just for reference, I don't know what your policy is and I didn't do extensive testing.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -2,7 +2,6 @@ desc = "The next-generation Julia package manager."
 keywords = ["package", "management"]
 license = "MIT"
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 
 [deps]
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -655,7 +655,7 @@ Testing...
 
 #### Test-specific dependencies
 
-Sometimes one might to want to use some packages only at testing time but not enforce a dependency on them when the package is used.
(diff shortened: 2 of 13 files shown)


KristofferC:
Hm, it seems a bit excessive to have a completely new command that only provides a `cd`. We could think about defaulting to cd into the new package but I don't think there should be a completely separate verb for it.

00vareladavid:
Best left as an issue for later

fredrikekre:
We could make `generate Foo` do an implicit `activate Foo` withouth `cd`ing to it. This would be similar to `init` (although there we create the project in `pwd()`).

StefanKarpinski:
It very much seems like these could be combined somehow...

00vareladavid:
That makes sense to me.
The workflow I was thinking of is: `generate Foo` followed by `add Example`.
My gut reaction was that the `add` would apply to the environment I most recently  handled (which would be `Foo`).

#418 is also relevant, they wanted to do a `generate Foo` followed by `import Foo`. Activating the environment would also make this workflow possible.

00vareladavid:
Actually, I though about #418 some more and the best fix for that seems to add a flag to `dev`. E.g. `dev --generate`. 

`generate Foo` could be used if the focus is `Foo`.

`dev --generate Foo` can be used if you wish to add a dependency which does not exist yet (the focus would remain on the current environment). `dev --generate Foo` would not `activate Foo` and so the defaults would reflect the intended workflow.

Issue #413, opened 2018-06-20
support quoted string arguments


StefanKarpinski:
A number of people have tried passing quoted URLs and paths to commands which doesn't work. It would be good if `pkg add "https://github.com/User/Repo.jl"` worked.

fredrikekre:
Fixed by #414 ?

PR #414, opened 2018-06-21
Add support for quoted args in REPL mode

00vareladavid:
For #413. All tests passing on my end.

Caveat: this change will unquote any "word": be it a flag, command, argument to a command, etc.. If only certain "words" should be unquoted, I can go ahead and do that by break up `lex_re`.

Also, don't know if I should add tests on this PR or separately (I'm thinking separate...)


KristofferC:
Would also need to add test for the new feature.

00vareladavid:
Ok, let me make the valid input concrete before I file another commit. Please let me know if I missed anything. 

### Valid Input
- `add "BinDeps" Compat`
- `add https://github.com/JuliaLang/BinDeps.jl "https://github.com/JuliaLang/Compat.jl"`
- `add "./BinDeps.jl"`
- `add ./Compat.jl`

### Questionable Input
- embedded space in arguments
    - URLs can not have embeded space
    - Package names can not have embedded space (because they correspond directly to julia variable names)
    - file paths CAN have embedded space (at least under Linux). they wouldn't have it in the basename (because it corresponds directly with the julia variable names), but it could be in the dirname I suppose
- no space between quoted arguments
    - this one is more arbitrary. I guess it depends on the final implementation whether supporting this feature is worth it. I would need someones explicit Yay/Nay on this
- allowing commands / flags to be wrapped in quotes. I see no harm in this.

00vareladavid:
Update: I searched through docs/source and PCRE.UNGREEDY would be the simplest solution, but seems to be no API for it yet. I also can't rely on the current lexer as it relies on whitespace as a delimiter. In light of this, the best solution seems a stack-based solution + escaping quotes ala Python (eg `add '/home/nur0n/pkg/weird"dir/BinDeps.jl'`).

StefanKarpinski:
I agree that the current regex lexer is reaching it's breaking point. Fortunately, this isn't urgent and we can rewrite the command parsing stuff better at some point in the future. Thanks for taking a crack at it!

00vareladavid:
Gotcha. I don't want to take up too much time. I redid the parser for quoted arguments and added some tests to go along with it. It should at least allow people to use quoted arguments and paths that contain weird characters. Feel free to close this PR if you think the project should go a different way. Thanks

Project.toml
```diff
@@ -2,7 +2,6 @@ desc = "The next-generation Julia package manager."
 keywords = ["package", "management"]
 license = "MIT"
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
StefanKarpinski: I think you committed this by accident.

00vareladavid: Thanks. I'll replace it

KristofferC:
I think we need to change up a bit how we do the parsing in the future but for now, this is good and just having extra tests is good in itself.

PR #421, opened 2018-06-23
improve develop err message

innerlee:


innerlee:
Note this change is not tested locally. I had a hard time to make the pkg system work... Maybe it's better to wait until the CI works again.

Issue #419, opened 2018-06-21
No replacement for Pkg.setprotocol! in 0.7


iroddis:
My work computer is behind an SSL-intercepting proxy, which causes Julia to complain about invalid SSL certs. I was able to get around this in 0.6 by changing the protocol Pkg used to init and add packages via the Pkg.setprotocol!("git"), but this option doesn't seem to be available in the version of Pkg used by default in 0.7. Could it, or something analogous, please be added?

PR #430, opened 2018-06-26
add setprotocol! to new package manager

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/419.

Need docs but will have to do a full pass of that later anyway.


rened:
@KristofferC
With this commit, the package manager asks for my github username / password for all private github `git@github.com/...` repos.

`up` works for me again after `setprotocol!(nothing)`, which seems to restore the previous behaviour.

May I suggest `nothing` as the default, and / or a more elaborate form along the lines of `mapprotocol = Dict("http" => "https", "git" => "git")` which allows per-protocol control?

rened:
Filed as issue to better keep track of this.

PR #438, opened 2018-06-27
generate help string name

simonbyrne:
The help says it's called "create"


KristofferC:
Yeah, that was a good name...

Thanks!

Issue #437, opened 2018-06-27
Adding from url without a branch seems to not record the default branch of the repo


KristofferC:
```
julia> Pkg.add("https://github.com/JuliaString/ModuleInterfaceTools.jl")
   Cloning git-repo `https://github.com/JuliaString/ModuleInterfaceTools.jl`
  Updating git-repo `https://github.com/JuliaString/ModuleInterfaceTools.jl`
 Resolving package versions...
 Installed StrTables ─ v0.2.4
  Updating `Project.toml`
  [5cb8414e] + ModuleInterfaceTools v0.1.4 # (https://github.com/JuliaString/ModuleInterfaceTools.jl)
```

The branch at `#` is empty.

KristofferC:
Fixed by #439

PR #439, opened 2018-06-27
Record rev in all cases

00vareladavid:
For #437. 
I think by that point, the rev we want to target is already determined, so we just have to store it.


KristofferC:
Excellent, thanks!

PR #441, opened 2018-06-28
add inequality specifiers

KristofferC:


StefanKarpinski:
I assume this is following the same format that NPM and Cargo use?

KristofferC:
Cargo yes, not sure about npm. Can also change it up if you have a better format in mind.

StefanKarpinski:
Cargo seems like the system to follow here, so let's stick with that.

PR #428, opened 2018-06-26
`Project.toml` is not always the project file

00vareladavid:
Small bug; `Project.toml` was hardcoded as the project file.
All tests passing locally.

PR #445, opened 2018-06-29
Modularize resolution of project/manifest file paths

00vareladavid:
I got some errors (I can try to recreate them, if needed) even after #428, so I went ahead and added some helper functions to get rid of the errors more generally. Also added some tests.


00vareladavid:
One more thing, the only place I couldn't get to was `bin/genstdlib.jl`. But assuming all stdlib packages use `Project.toml` is a safe assumption, I think?

KristofferC:
Thanks!

00vareladavid:
Welcome!

PR #450, opened 2018-06-29
Refactor `parse_package` and add unit tests

00vareladavid:
I don't know what the feelings are about this kind of refactoring during this stage of the project?

I noticed the logic could be simplified and made more uniform.
I also added unit tests to make sure everything is locked in.


KristofferC:
Simplifying code is always welcome. The fact that the API functions calls back into the REPL `parse_package` should really be fixed (it is the wrong direction of "responsibility") but this is good nonetheless.

Issue #403, opened 2018-06-18
Better error message when adding local path that is not a git repo


KristofferC:
A common mistake which is easy to detect and the current error message is LibGit2 non-friendliness so should be done.

KristofferC:
Closed by https://github.com/JuliaLang/Pkg.jl/issues/403

PR #451, opened 2018-07-01
Make sure repo.url points to a git repo before adding

00vareladavid:
What about something like this for #403?


KristofferC:
Yes, this looks good! Thanks again.

The CI errors should be fixed after #456

src/GitTools.jl
```diff
@@ -127,4 +127,34 @@ function fetch(repo::LibGit2.GitRepo, remoteurl=nothing; header=nothing, kwargs.
...
+                    err.code == LibGit2.Error.ENOTFOUND &&
+                    err.class == LibGit2.Error.Repository
+                Pkg.Types.cmderror("'$(url)' does not point to a valid git repository")
+            else
+                throw(err)
+            end
+        end
+    else
+        try
+            # try to fetch anything just to see if it throws an error
```
KristofferC: Won't the extra fetch give extra latency? Can't we put this inside `GitTools.fetch` and `GitTools.clone` to not do a redundant fetch?

00vareladavid: Ok, that would be a lot better.

KristofferC:
Thanks, great!

00vareladavid:
Welcome!

Issue #462, opened 2018-07-03
Add test to REPL prompt function


KristofferC:
So the prefix cant be deleted and tested still passing.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/481

PR #466, opened 2018-07-05
WIP: Add unit tests for `REPLMode.promptf`

00vareladavid:
For #462. Something like this should do right?


KristofferC:
Yes, but it would also be good to see that the prompt picks up the `name = ` if that exists in the Project. And that the caching system works by modifying that `name = ` in that project file and looking at the prompt function again (to make sure it updates).

00vareladavid:
Hmm, the tests seem to be failing at `startswith(pwd(), proj_dir)` at https://github.com/JuliaLang/Pkg.jl/blob/642f5a5ee4c691066d03bf0a9673913a35a5ebd7/src/REPLMode.jl#L1011, but I don't see why that constraint is enforced? Why is the working directory relevant here?

00vareladavid:
I have some further questions about project names if you don't mind: 
Running `activate` on a path that is just an empty directory continues silently. This behavior does not make much sense to me. It seems that projects can be named by an explicit 'name = ' in the Project.toml, or by just using the directory name if that is missing. Is this behavior supported because `REQUIRE` files are still supported?

KristofferC:
Sorry for forcing you to deal with the mess that is apple file system! :(

00vareladavid:
Where else would I find such interesting bugs? :wink:

KristofferC:
> , but I don't see why that constraint is enforced? Why is the working directory relevant here?

I think that was when the current directory automatically was the active project. Which is not the case anymore. It looks indeed wrong.

00vareladavid:
:laughing: Well, if you look at the last failed apple CI. I think that's where some of the errors where coming from. `pwd` returns 'private/var/...' while `dirname(proj_dir)` returns `/var/...`. I found online that `/var` is just a symlink to `/private/var`. You have to call `realpath` on both paths before using `startswith`.

KristofferC:
Don't spend way too much time on this. If it is a mess, we can just test the most basic.

00vareladavid:
Gotcha. Perhaps I did get a little carried away on this one. Apple is finally running, I'll try a bit more to get Win64 running. If not, I'll just leave as is

00vareladavid:
I think I pinned down the Win64 error too: `Base.active_project()` is returning `..Test2/Project.toml/Project.toml` instead of `..Test2/Project.toml` when `Project.toml` does not exist.

The relevant line is [here](https://ci.appveyor.com/project/KristofferC/pkg-jl/build/1.0.177/job/y60jww5f66sw2vpj#L1573).

All other systems are running fine now.

KristofferC:
Nice detective work. cc @StefanKarpinski, any idea why that would happen?

StefanKarpinski:
The function to look at is `Base.active_project` which should return `..Test2/Project.toml` in that case regardless of platform. Maybe there’s a Windows-specific big in one of the file check functions that `active_project` uses? That’s been a surprisingly tricky function to get right in all cases.

00vareladavid:
I couldn't easily isolate the error and I think it would unproductive to continue from this side. I'll try to add some tests from julia's side and hopefully I can weed it out there. I'll leave the failing tests, but comment them out (they are a bit of an edge case) to revisit them at a later point. I'll close this PR and submit one with a clean commit history.

PR #470, opened 2018-07-05
don't use startup.jl when precompiling, building and testing

### LINKED ISSUE #435: should test load startup.jl?
I have something like
```julia
if VERSION ≥ v"0.7-"
    @async begin
        sleep(0.1)
        @eval using Revise
    end
else
   ...
end
```

in my `startup.jl` so when I test packages, I get (innocuous) error messages like

```julia
ERROR (unhandled task failure): ArgumentError: Package Revise not found in current path:
 - Run `Pkg.add("Revise")` to install the Revise package.

Stacktrace:
 [1] require(::Module, ::Symbol) at ./loading.jl:857
 [2] eval at ./boot.jl:319 [inlined]
 [3] macro expansion at /home/tamas/.julia/config/startup.jl:10 [inlined]
 [4] (::getfield(Main, Symbol("##3#4")))() at ./task.jl:257
```

which, if I understand correctly, comes from the test running in a "clean" environment (just the `Project.toml`). But running with `--startup-file=no` may not be a good option either, as some machine-specific configuration could reside there.

Alternatively, can I somehow test for the julia process being run in a test?

fredrikekre:
don't use startup.jl when precompiling, building and testing unless the user has explicitly asked for it with
```
  --startup-file=yes
```

For testing I think not loading `startup.jl` is the correct choice since you don't want local things to affect testing. Not sure about building and precompiling, those are local things in some sense and perhaps the startup file should be run in those cases...

fix #435

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -504,7 +504,7 @@ function precompile(ctx::Context)
         printpkgstyle(ctx, :Precompiling, pkg * " [$i of $(length(needs_to_be_precompiled))]")
         run(pipeline(ignorestatus(```
         $(Base.julia_cmd()) -O$(Base.JLOptions().opt_level) --color=no --history-file=no
-        --startup-file=$(Base.JLOptions().startupfile != 2 ? "yes" : "no")
+        --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
         --compiled-modules="yes"
         --depwarn=no
         --eval $code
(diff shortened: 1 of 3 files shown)


KristofferC:
I also think this is the right default.

KristofferC:
We should have this in the docstrings for `Pkg.test` and `Pkg.build` when those are available but for now, perhaps we can jot something down about this in the manual on test / build?

PR #482, opened 2018-07-10
debug AV failure [travis skip]

KristofferC:


00vareladavid:
I'm starting to believe this error is not limited to windows. I had noticed before that my `/tmp` directory does not get cleaned up correctly (on Linux, of course). I ran the tests again and some directories that were created by `mktempdir` do not get properly unlinked. The difference on Linux is that there is no indication of an error from Julia. Hopefully this is useful info.

KristofferC:
That's useful to know. My best bet is that some LibGit2 object is not being closed properly and I have found many such cases but fixing them hasn't really improved the situation.

KristofferC:
Seems to pass when I remove the clean up of the temp direcotries.

KristofferC:
The `install_git` method is still not safe to not "leak" LibGit2 objects when errors occur.

StefanKarpinski:
Aside: https://github.com/JuliaLang/julia/issues/7721 would be really helpful for this kind of thing. All the try/catch and do-block refactoring for this kind of stuff is quite annoying.

KristofferC:
Extremely annoying.

KristofferC:
But note that we have `LibGit2.with` which kinda does this. The problem is that the git objects are created in different scopes and from different functions so it is sometimes hard to just encapsulate it with a `with`.

KristofferC:
@00vareladavid I am quite sure it is this test that makes the problem occur

https://github.com/JuliaLang/Pkg.jl/blob/2a0f311299a5bb6defeb6a19de52dbf658c10904/test/pkg.jl#L358-L378

Since I disabled that on Julia master, AV has never errored there.

00vareladavid:
Ok, I'll take a look at it, perhaps there is a way to rewrite it. From a initial glance, perhaps there is a reference to `Project.toml` which does not get cleaned up on the `mv` operation (or something along those lines...). Is it possible to disable it on this side until I take a look at it?

PR #486, opened 2018-07-10
make sure LibGit2 repo closes if we fail in install_git

KristofferC:


KristofferC:
Still getting warnings about not being able to delete folders on AV...... jeeez.

KristofferC:
But we didn't crash at least... Let's try run AV a few times.

PR #489, opened 2018-07-10
support for multiline input

fredrikekre:
This implements support for 
```
pkg> add Test
     rm Random
```
and similarly
```
julia> pkg"""
       add Test
       rm Random
       """
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@
 *.jl.cov
 *.jl.*.cov
 docs/build
+/Manifest.toml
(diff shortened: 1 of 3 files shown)


.gitignore
```diff
@@ -3,3 +3,4 @@
 *.jl.cov
 *.jl.*.cov
 docs/build
+Manifest.toml
```
StefanKarpinski: This should start with a `/`; otherwise it ignores files named `Manifest.toml` anywhere in the entire repo.

PR #490, opened 2018-07-10
Save dev'd path as given

### LINKED ISSUE #485: dev relative/path should put relative/path in Manifest.toml
Currently we have that
```
pkg> dev dir/Foo
```
puts `dir/Foo` in `Manifest.toml`, but 
```
pkg> dev ../Foo
```
puts the absolute path of `Foo` in `Manifest.toml`. I think it would be useful if whatever path you `dev` is put in `Manifets.toml`, and if you want an absolute path you have to dev an absolute path, e.g. 
```
pkg> dev /absolute/path/Foo
```

This would be useful for a structure like
```
Foo
├── Project.toml
├── docs
│   ├── Manifest.toml
│   └── Project.toml
└── src
    └── Foo.jl
```
where I want to be able to 
```
(docs) pkg> dev ../
```
to add `Foo` to `docs/Project.toml`.

fredrikekre:
Seems to work, but should try to add some tests etc

fix #485

### DIFF
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -229,6 +229,10 @@ temp_pkg_dir() do project_path; cd(project_path) do
                         pkg"develop ../SubModule2"
                         @test Pkg.installed()["SubModule1"] == v"0.1.0"
                         @test Pkg.installed()["SubModule2"] == v"0.1.0"
+                        # make sure paths to SubModule1 and SubModule2 are relative
+                        manifest = Pkg.Types.Context().env.manifest
+                        @test manifest["SubModule1"][1]["path"] == "SubModule1"
+                        @test manifest["SubModule2"][1]["path"] == "SubModule2"
                     end
                 end
                 cp("HelloWorld", joinpath(other_dir, "HelloWorld"))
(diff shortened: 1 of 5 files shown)


KristofferC:
image

test/repl.jl
```diff
@@ -236,8 +236,8 @@ temp_pkg_dir() do project_path; cd(project_path) do
                 with_current_env() do
                     # Check that these didnt generate absolute paths in the Manifest by copying
                     # to another directory
-                    @test Base.find_package("SubModule1") == joinpath(pwd(), "SubModule1", "src", "SubModule1.jl")
-                    @test Base.find_package("SubModule2") == joinpath(pwd(), "SubModule2", "src", "SubModule2.jl")
+                    # @test Base.find_package("SubModule1") == joinpath(pwd(), "SubModule1", "src", "SubModule1.jl")
+                    # @test Base.find_package("SubModule2") == joinpath(pwd(), "SubModule2", "src", "SubModule2.jl")
```
fredrikekre: A bit ironic that all tests except those pass tests.

fredrikekre:
Hmm. This was green before I pushed the one line of doc change. It still passes on Linux everytime.

fredrikekre:
Fixed the issue on AV. On macOS I get:
```
ERROR: LoadError: LoadError: path /var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/tmpGjiO8B/environments/v0.7/../../../../../../../../private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/tmpVWhTKJ/HelloWorld/Foo for package Foo no longer exists. Remove the package or `develop` it at a new path
```
which is a result of trying to merge paths
```
/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/tmpGjiO8B/environments/v0.7/
```
and
```
/private/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/tmpVWhTKJ/HelloWorld/Foo
```
Why is there a `private` in one of them here? Otherwise this would join nicely to
```
/var/folders/bb/n7t3rs157850byt_jfdcq9k80000gn/T/tmpVWhTKJ/HelloWorld/Foo
```

fredrikekre:
Ok, after expanding some symlinks before taking the relative path it looks like this is working now!

test/runtests.jl
```diff
@@ -1,6 +1,7 @@
 module PkgTests
 
 include("pkg.jl")
+include("repl.jl")
```
KristofferC: Please put back.

The reason is that the REPL tests require internet and they are also noisy and in Base, the `pkg` testgroup is special in that it run last.

fredrikekre:
Will have to adjust some things which was merged in  #517 before merging this.

fredrikekre:
I think this is also good to go now.

fredrikekre:
Fixed

PR #496, opened 2018-07-11
make installed only look in the project

KristofferC:
The return data structure `Dict{String, VersionRange}` for `Pkg.installed` is inherently flawed since it is using a package name as the key (and there can be multiple packages with the same key).
Package authors that are trying to replace `Pkg.installed()` from the old package manager, naturally find this function and start to use it.

We need to decide what to do here. Just remove it until a suitable replacement is found or alt, like in this PR, limit it to only those packages in the project file, which are known to be unique.

cc @vdayanand @StefanKarpinski

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -239,9 +239,8 @@ function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, kwargs...
     return
 end
 
-
-function installed(mode::PackageMode=PKGMODE_MANIFEST)
-    diffs = Display.status(Context(), mode, #=use_as_api=# true)
+function installed()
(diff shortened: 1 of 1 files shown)


vdayanand:
I was under the assumption that there can't be multiple package with same name within a given environment. Maybe a dumb question, but how does the package import work in such scenario?

KristofferC:
Packages with the same name are disambiguated by their UUID.

daschw:
So will there be a replacement for the current (Julia 0.6) `Pkg.installed` behavior or does this mean that e.g. Plots should depend on all its backends?

StefanKarpinski:
I think with this we can support `Pkg.installed` again but with the caveat that it will only find packages that are explicit dependencies of the main project. Otherwise the API needs to be a bit different.

PR #498, opened 2018-07-12
Refactor `REPLMode.parse_quotes`

00vareladavid:


src/REPLMode.jl
```diff
@@ -182,13 +185,12 @@ end
 function parse_quotes(cmd::String)
     in_doublequote = false
     in_singlequote = false
-    all_tokens::Array = []
+    qwords::Array{QuotedWord}= []
```
KristofferC: Typically you would write `qwords = QuotedWord[]`

00vareladavid: Ah OK, that makes sense. Still working on my 'idiomatic' Julia

StefanKarpinski: This makes `qwords` a typed local and assignment to it therefore attempts conversion to `Array{QuotedWord}` (and fails if the value cannot be converted). `[]` is equivalent to `Any[]`. Converting `Any[]` to `Array{QuotedWord}` gives you the same value as `QuotedWord[]` but in a very indirect fashion. The compiler could potentially eliminate all that work, but it's better to just construct `QuotedWord[]` directly.

00vareladavid: OK, I see what is going on now: `qwords = QuotedWord[]` should be preferred because it avoids an unnecessary call to `convert`. Thanks for the tip.

00vareladavid:
Included in #509

PR #499, opened 2018-07-12
Refactor argument order checking

00vareladavid:
I think it removes clutter from the core logic. Plus more tests!


KristofferC:
The `tempdir_util` shouldn't be needed because Base already does this.

00vareladavid:
@KristofferC Sorry about that, I tried testing it on this branch before you explained to me that the bug was more complicated than I thought.

PR #500, opened 2018-07-12
Refactor `REPLMode.do_generate!`

00vareladavid:
Small change. The tests should be useful though.


KristofferC:
Thanks, I've been thinking about moving `generate` out to a new version of https://github.com/JuliaLang/PkgDev.jl (with some WIP at https://github.com/KristofferC/PkgDev.jl/tree/kc/07).

But this is good to have in the meantime.

PR #501, opened 2018-07-13
Increase isolation of test utilities

00vareladavid:
Currently, anything which is `dev`ed by the tests will contaminate the user's home directory by writing to `~/.julia/dev`. I set up the main test utility to follow the temporary `DEPOT_PATH` instead.

This brings up a question for me: why is this not the default behavior?

One of the tests seemed to think it was: [here](https://github.com/JuliaLang/Pkg.jl/blob/1599d42e4dc9e961afdfa507f0263b69c33bb9a0/test/repl.jl#L46). If this is the intention, then a one-line change [here](https://github.com/JuliaLang/Pkg.jl/blob/1599d42e4dc9e961afdfa507f0263b69c33bb9a0/src/Pkg.jl#L10) should do the trick. FWIW, the docs make it seem like the only two alternatives are `~/.julia/dev` or explicitly specifying with `JULIA_PKG_DEVDIR`.

### DIFF
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -4,11 +4,13 @@ function temp_pkg_dir(fn::Function)
     local old_depot_path
     local old_home_project
     local old_active_project
+    local old_julia_pkg_devdir
     try
         old_load_path = copy(LOAD_PATH)
         old_depot_path = copy(DEPOT_PATH)
         old_home_project = Base.HOME_PROJECT[]
         old_active_project = Base.ACTIVE_PROJECT[]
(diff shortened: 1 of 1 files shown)


KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/473/files#diff-fa0f8d7a95cfa5fcc42e790accc462c5R10

Is the correct thing to do. I need to get that PR merged.

00vareladavid:
Ok, that change would do it. Thanks

PR #504, opened 2018-07-14
Use `mktempdir` shim (until bugfix is complete)

00vareladavid:
What if we use this until the LibGit stuff is sorted out? It should allow people to test changes on the CI without being bothered by the LibGit bug. Thankfully the only place (so far) that the bug manifests itself is when cleaning up temporary directories. I've tried my best not to inadvertently catch unrelated exceptions. Its kind of a hack, but its limited to test code and should only be temporary.

### DIFF
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -178,7 +178,7 @@ temp_pkg_dir() do project_path
         Pkg.add(TEST_PKG.name)
         old_v = Pkg.installed()[TEST_PKG.name]
         Pkg.rm(TEST_PKG.name)
-        mktempdir() do devdir
+        tempdir_util() do devdir
             withenv("JULIA_PKG_DEVDIR" => devdir) do
                 Pkg.REPLMode.pkgstr("develop $(TEST_PKG.name)")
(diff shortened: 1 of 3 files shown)


KristofferC:
The funny thing is that the Base version already does this... https://github.com/JuliaLang/julia/blob/master/base/file.jl#L581

00vareladavid:
I'm confused. The difference would be in the `@error` vs `@warn`, no?

KristofferC:
`@error` doesn't throw, it just prints a log message with the error heading.

00vareladavid:
That's wierd, [here](https://ci.appveyor.com/project/KristofferC/pkg-jl/build/1.0.245/job/ua4nd98o9sk2i61u#L430) it looks like the exception is going right through `mktempdir`.

KristofferC:
Yeah, but see https://github.com/JuliaLang/julia/issues/27982#issuecomment-403252072. It might be a different error but the stacktrace is just wrong... Not pleasant to debug.

00vareladavid:
Ok, thanks for explaining.

PR #513, opened 2018-07-18
Don't update registry when developing a local path

fredrikekre:
Well, that was easier than I thought... (fingers crossed for CI)

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -31,9 +31,6 @@ function add_or_develop(ctx::Context, pkgs::Vector{PackageSpec}; mode::Symbol, k
         cmderror("Trying to $mode julia as a package")
 
     ctx.preview && preview_info()
-    if !UPDATED_REGISTRY_THIS_SESSION[]
-        update_registry(ctx)
-    end
     if mode == :develop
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
What happens if you develop a local path and it causes some new package to be installed? We still want to update the registry if it hasn't be previously updated in that situation, right?

fredrikekre:
TBH didn't think too much about this, I was just getting really frustrated how useless the package manager was without an internet connection.

StefanKarpinski:
Maybe also an "offline" mode for not trying to connect to the internet at all?

fredrikekre:
Yea, it is probably better to tackle this with something like that. The tricky thing would be resolving I imagine, since only package versions that are already downloaded should be considered, no matter what the registry says.

StefanKarpinski:
Well, to be clear I think we should do this too—the package manager should only hit the internet when it needs to and to add a local package it doesn't need to.

KristofferC:
There is a bug now where it actually throws when it fails to update the registry. It is because we are throwing a different error than before so the code fails to catch it.

PR #517, opened 2018-07-21
Rework the meaning of `pkg> activate Foo`

fredrikekre:
Rework the meaning of `pkg> activate Foo` to do the following:

  1. If Foo is an existing path, activate that
  2. If Foo is a developed dependency of the current environment, activate that
  3. Activate the (non-existing) directory Foo

To sidestep 2. above, and activate a non-existing directory Foo even though there is a developed dependency Foo, use `pkg> activate ./Foo`.

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -324,7 +324,11 @@ function do_cmd!(tokens::Vector{Token}, repl)
             cmderror("misplaced token: ", token)
         end
     end
-    cmd.kind == CMD_ACTIVATE && return Base.invokelatest(do_activate!, tokens)
+
+    if cmd.kind == CMD_ACTIVATE
+        return Base.invokelatest(do_activate!, Base.active_project() === nothing ?
+            nothing : EnvCache(env_opt), tokens)
+    end
 
     ctx = Context(env = EnvCache(env_opt))
     if cmd.kind == CMD_PREVIEW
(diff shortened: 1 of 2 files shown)


src/REPLMode.jl
```diff
@@ -337,6 +336,7 @@ function do_cmd!(tokens::Vector{Token}, repl)
 
     # Using invokelatest to hide the functions from inference.
     # Otherwise it would try to infer everything here.
+    cmd.kind == CMD_ACTIVATE    ? Base.invokelatest(      do_activate!, ctx, tokens) :
```
KristofferC: A problem with putting this here is that it requires an active environment. Otherwise creating `ctx` will fail.

fredrikekre: Should be fixed now.

KristofferC:
Cool, good to go?

fredrikekre:
Yes, I think so!

src/REPLMode.jl
```diff
@@ -816,15 +820,27 @@ function do_resolve!(ctx::Context, tokens::Vector{Token})
...
-        return API.activate(abspath(token))
+        # `pkg> activate path` does the following
+        # 1. if path exists, activate that
+        # 2. if path exists in deps, and the dep is deved, activate that path
+        # 3. activate the non-existing directory (e.g. as in `pkg> activate .` for initing a new env)
+        if Types.isdir_windows_workaround(path)
+            API.activate(abspath(path))
+        elseif env !== nothing && haskey(env.project, "deps") && haskey(env.project["deps"], path) &&
+            haskey(env.manifest, path) && haskey(env.manifest[path][1], "path") &&
+            env.manifest[path][1]["path"] != nothing
```
fredrikekre: Is there a better way to make sure there is a deved path to activate than these 3 lines?

KristofferC: The `1` here is not correct in case there are multiple packages with the same name in the Manifest.
Get the `uuid` from `env.project["deps"][path]` and use `info = manifest_info(env, uuid)` instead.
Then look if `info["path"]` exist.

src/REPLMode.jl
```diff
@@ -816,15 +820,31 @@ function do_resolve!(ctx::Context, tokens::Vector{Token})
...
+        path = popfirst!(tokens)
+        if !isempty(tokens) || !(path isa String)
             cmderror("`activate` takes an optional path to the env to activate")
         end
-        return API.activate(abspath(token))
+        devpath = nothing
+        if env !== nothing && haskey(env.project["deps"], path)
+            uuid = UUID(env.project["deps"][path])
+            info = manifest_info(env, uuid)
+            devpath = info["path"]
```
fredrikekre: This will be 'nothing` if there is no `"path"` entry, correct?

KristofferC: No, `info` is a dict, so it will error.

fredrikekre: Ok, added a test for this code-path which was missing.

fredrikekre:
I think this was bad. It is too confusing that you can `activate Foo` only sometimes.

PR #519, opened 2018-07-21
[WIP] Implement --local/-l and --shared/-s options for

fredrikekre:
Implement `--local`/`-l` and `--shared`/`-s` options for determining where a `develop`ed package ends up. Defaults to `shared`, but the default can be controlled by setting `JULIA_PKG_DEVMODE` to `"shared"` or `"local"` respectively.


KristofferC:
If you have tried this locally it would be good to get in. Since we want to have an RC soon it is nice to get as much in as possible. For now, lets keep shared as the default to reduce confusion with changing defaults.

fredrikekre:
No computer access until tomorrow, but I can finish this up tomorrow evening. It would be helpful if we could decide on these in the meantime:
1. Should we keep `-l` and `-s` short forms? (Maybe remove them for now in case we need them for something else? They can always be implemented later. )
2. Name of the environment variable? If we inplement the same options for `pkg> generate` it does bot make sense to control that with `JULIA_PKG_DEVMODE` for example.
3. `--local` means we `dev` to `joinpath(dirname(path-to-active-project), "dev", pkgname) `?
4. For consistency, I propose we change the global default to `.julia/environments/v0.7/dev`. Sure, it is somewhat deep, but you can always symlink it to wherever you want, or set `JULIA_PKG_DEVDIR` to whatever you want (for example the current location).

StefanKarpinski:
(2) I think we really want to move away from configuring everything with environment variables, so let's get rid of that. We'll introduce a different configuration mechanism in the future—something like `~/.julia/Config.toml` where we merge config data from each depot in `DEPOT_PATH` and that determines your system configuration. (1) I'm less concerned about the short options; we can always change them later but then again, they're not necessary. (3) Yes, sounds good. (4) Yes, sounds good.

fredrikekre:
Should be good to go. Now you can 
```
pkg> develop --local Example
```
and that will place the `Example` clone in `joinpath(dirname(path-to-active-project), "dev", "Example")`.
I removed the environment variable, and removed the short options. Regarding (4) above: I think `~/.julia/dev` is a good default shared location. `.julia/environments/v0.7/dev` is where packages end up with `--local` when the default environment is active, so that does not seem like a good shared location.

PR #525, opened 2018-07-23
prevent a O(n^2) when collecting package names

KristofferC:
Since we didn't have a way to go from uuid -> name in the cache, we just brute forced through all the name -> uuid and collected the names for the uuid that matched. In e.g. `deps_graph` this is inside a loop over `uuids` which means this will scale quadratically with the size of the dependency graph. Introduce a direct uuid -> name cache to fix.


StefanKarpinski:
How come no merge? Did you decide to do this a different way?

KristofferC:
Sorry, should have written, included in https://github.com/JuliaLang/Pkg.jl/pull/527

PR #526, opened 2018-07-23
RFC: new targets design

StefanKarpinski:
Instead of writing
```toml
[deps]
Required = "<uuid>"

[targets.build.deps]
Compiler = "<uuid>"

[targets.test.deps]
Test = "<uuid>"
QuickCheck = "<uuid>"
```
with this design one would write:
```toml
[deps]
Compiler = "<uuid>"   # build-only
Required = "<uuid>"   # required
Test = "<uuid>"       # test-only 
QuickCheck = "<uuid>" # test-only

[targets]
build = ["Compiler"]
test = ["Test", "QuickCheck"]
```
In other words:
- all dependencies are listed with their UUIDs in the main `[deps]` section
- the `[targets]` section is just about defining subsets of these dependencies
- if you ignore `[targets]` you can resolve anything that any target might load
- the `main` target is everything that's not explicitly in any other target
- the dependencies of a target are the `main` targets plus the listed ones

Incidentally, this PR causes a lot of places where we were previously only looking at the main deps—i.e. those in `[deps]`—to now consider all deps because all of them are in the main `[deps]` section. As far as I can tell, this is mostly more correct and all of these places should previously have been merging target deps into the main deps. The only place that should really consider targets and subset the deps because of them is when resolving versions: only the main deps are actually required. Overall, I think this is a simplification and improvement but I want to get feedback.

I have a similar design for `[options]` and system-specific dependencies (expressed via options).

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1088,6 +1088,15 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
         @info "No changes"
         return
     end
+    deps_names = collect(keys(ctx.env.project["deps"]))
+    if haskey(ctx.env.project, "targets")
+        filter!(ctx.env.project["targets"]) do (target, deps)
+            !isempty(filter!(in(deps_names), deps))
+        end
+        if isempty(ctx.env.project["targets"])
+            delete!(ctx.env.project, "targets")
+        end
+    end
     # only keep reachable manifest entires
     prune_manifest(ctx.env)
     # update project & manifest
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -262,14 +262,18 @@ function deps_graph(ctx::Context, uuid_to_name::Dict{UUID,String}, reqs::Require
 end
 
 # Resolve a set of versions given package version specs
-function resolve_versions!(ctx::Context, pkgs::Vector{PackageSpec})::Dict{UUID,VersionNumber}
+function resolve_versions!(
+    ctx::Context,
+    pkgs::Vector{PackageSpec},
+    target::Union{Nothing, String} = nothing,
+)::Dict{UUID,VersionNumber}
     printpkgstyle(ctx, :Resolving, "package versions...")
     # anything not mentioned is fixed
     uuids = UUID[pkg.uuid for pkg in pkgs]
     uuid_to_name = Dict{UUID, String}(uuid => stdlib for (uuid, stdlib) in ctx.stdlibs)
     uuid_to_name[uuid_julia] = "julia"
 
-    for (name::String, uuidstr::String) in ctx.env.project["deps"]
+    for (name::String, uuidstr::String) in get_deps(ctx, target)
         uuid = UUID(uuidstr)
         uuid_to_name[uuid] = name
 
@@ -748,6 +752,11 @@ function with_dependencies_loadable_at_toplevel(f, mainctx::Context, pkg::Packag
     need_to_resolve = false
     is_project = Types.is_project(localctx.env, pkg)
 
+    target = nothing
+    if pkg.special_action == PKGSPEC_TESTED
+        target = "test"
+    end
+
     if is_project # testing the project itself
         # the project might have changes made to it so need to resolve
         need_to_resolve = true
@@ -757,7 +766,7 @@ function with_dependencies_loadable_at_toplevel(f, mainctx::Context, pkg::Packag
         localctx.env.pkg = nothing
         localctx.env.project["deps"][pkg.name] = string(pkg.uuid)
         localctx.env.manifest[pkg.name] = [Dict(
-            "deps" => mainctx.env.project["deps"],
+            "deps" => get_deps(mainctx, target),
             "uuid" => string(pkg.uuid),
             "path" => dirname(localctx.env.project_file),
             "version" => string(pkg.version)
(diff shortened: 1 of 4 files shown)


KristofferC:
Only take out "normal" deps in https://github.com/JuliaLang/Pkg.jl/blob/da925e698aa96417d9c6978ee46cd44fda064ace/src/Operations.jl#L141?

And here: https://github.com/JuliaLang/Pkg.jl/blob/da925e698aa96417d9c6978ee46cd44fda064ace/src/Operations.jl#L677?

Will look more in a while.

KristofferC:
If I make a project file with

```
[deps]
UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"

[targets]
test = ["Test"]
```

and resolve, I still get `Test` in the Manifest. That shouldn't happen, right?

StefanKarpinski:
Are you sure you were using the dev version of `Pkg` and had this directory activated? This is what I get in my manifest file when I try that:
```toml
[[UUIDs]]
uuid = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
```
I created the `Project.toml` file in `tmp` and ran
```sh
$ JULIA_LOAD_PATH=@:/Users/stefan/projects/Pkg3.jl: julia -q --project
julia> LOAD_PATH
4-element Array{String,1}:
 "@"
 "/Users/stefan/projects/Pkg3.jl"
 "@v#.#"
 "@stdlib"
```
Ultimately, I think we should resolve all the deps and put entries for them that have a `targets` list:
```
[[Test]]
targets = ["test"]
uuid = "8dfed614-e22c-5e08-85e1-65c5234f0b40"

[[UUIDs]]
uuid = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
```
And in this case, `require` should possibly take a `target` keyword, which, if supplied, causes the package to only be loaded if the corresponding stanza has no targets list or has the given target in its `targets` list. So for this manifest, `Test` would only be loadable from the `test` target while `UUIDs` would always be loadable. Technically, this would be a breaking change to code loading, but it wouldn't break any manifests we've ever generated, so I don't think that would be a real issue.

StefanKarpinski:
Added some tests to verify that behavior and some others.

fredrikekre:
I like this design, but it is slightly more difficult to get a quick overview of what packages will be installed when you add the package (i.e. what the `main` target includes). Does it make sense to allow putting `main` as a target?

mortenpi:
To echo @fredrikekre, I think this is a good design, but I am also a bit uneasy about the _"the `main` target is everything that's not explicitly in any other target"_ part. I don't feel that this behaviour is obvious enough (by just looking at the example in the OP I couldn't figure out which ones are the main deps, until I read the comment).
 
I can also think of a case where you remove one of the deps from a target, but you forget to update `deps`, and now you silently have an additional unnecessary main dependency.

One possible solution I can think of is to have a separate `deps` section for the dependencies used in the targets:

```toml
[deps]
Required = "<uuid>"   # required

[targetdeps]
Compiler = "<uuid>"   # build-only
Test = "<uuid>"       # test-only 
QuickCheck = "<uuid>" # test-only

[targets]
build = ["Compiler"]
test = ["Test", "QuickCheck"]
```

with the following rules:

* All targets have access to everything in `deps`
* `targets` can only refer to things in `targetdeps`
* Each `targetdep` must (should?) be referred to by at least one target
* You can't have a overlaps between `deps` and `targetdeps`

The benefit over just having a `main` target is that you don't have to list all the deps again. And `main` seems a bit special anyway, so you might not want to equate it with all the other targets.

KristofferC:
That would also keep the semantics the same with the previous version when it comes to code loading.

StefanKarpinski:
One of the good things about this change is that in the code most of the places where Pkg was just looking at deps, it should have been looking at all the target deps. Of course, we can fix that internally, but I feel like it’s a good default that if you ignore targets you get all of the possible dependencies for anything you might want to do with the project.

KristofferC:
Isn't it also the case that you can load recursive dependencies that happen to be in your test dependencies in your top-level package? It's in the `[deps]` section and in the Manifest so you can load it.

StefanKarpinski:
The presence of targets dependencies in `[deps]` causes non-main deps to be _identifiable_ but not loadable since they're not in the manifest since resolution doesn't resolve them (which is tested in this PR). It seems fine to be able to answer the question "what is `X` to this project?" but only be able to load `X` from targets that include it. I hear the concern about this making main and target deps harder to distinguish and I'm thinking on it, but I still think that informational comments might be best.

StefanKarpinski:
> It seems the opposite. Wherever we looked at `"deps"` we now change it to `get_deps()`.

Grep for `project["deps"]`—there are plenty of places where I looked to see if I should change them to `get_deps` and found that I actually shouldn't and the code was previously ignoring target deps when it shouldn't. For example, should `up` consider all deps or just the main ones? I would argue that it should consider all of them. Should `status` consider all deps or just the main ones? Again, it seems like showing the changes to all deps would be the right thing to do. In fact, the _only_ place where it seems beneficial to ignore target deps is when doing version resolution. Accordingly, the only place where `get_deps` is called is from inside of `resolve_versions!`—in only two places. All the other places in the code still look at `env.project["deps"]` and therefore consider all the deps.

KristofferC:
>  since they're not in the manifest since resolution doesn't resolve them (which is tested in this PR). 

You dependencies might put them in the Manifest though.

>  For example, should up consider all deps or just the main ones? I would argue that it should consider all of them. 

Only the main ones are in the Manifest, how can you even upgrade a non-installed test dependency?

> Should status consider all deps or just the main ones? 

It only shows deps in the Manifest so nothing will have changed here (except target dependencies that happen to be in the manifest from recursive dependencies will show in `status`. Confusing).

KristofferC:
All the comments above comes stems from the issue that from looking at only `[deps]` and the manifest (which code loading does), there is no difference between a target dependency that happens to be in the Manifest due to some of your dependencies pulling it in and a direct dependency, pulled in by the active project.

StefanKarpinski:
> You dependencies might put them in the Manifest though.

That's true, if we mark those entries with `targets` and don't allow loading those at the top level, that would fix that problem.

> Only the main ones are in the Manifest, how can you even upgrade a non-installed test dependency? 

In the future, I'm envisioning a single manifest file with target annotations that can be used for all the targets at once. Actual installation of target dependencies can be deferred until they're needed but we really want to resolve a single global set of versions that support the project and all its targets.

KristofferC:
> That's true, if we mark those entries with targets and don't allow loading those at the top level, that would fix that problem.

Isn't that exactly https://github.com/JuliaLang/Pkg.jl/pull/526#issuecomment-407378102?

KristofferC:
> but we really want to resolve a single global set of versions that support the project and all its targets.

Only for those that are active during normal runtime?

Anyway, as it is right now I think something here needs to change. Different pacakges getting shown in `status` and different updating behavior depending on the interaction between recursive dependencies and your target dependencies will be confusing. The target deps in the project need to be marked somehow and putting them under a separate header seems the easiest.

PR #527, opened 2018-07-23
Stop trying to be lazy

KristofferC:
It seems a bit silly that tab completion is allowed to fully TOML parse all the registries (and it does so very fast), while we are constructing this big regular expression to avoid parsing it for the resolver (where there is much more of other stuff going on). It also ties us to using a custom regex parser which might break if someone outputs TOML in another format.

We can also introduce a cache here so that we only need to reparse Registry.toml in case it is updated. This means that often it is only required to be parsed a single time during the session.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -842,13 +842,12 @@ function registries(; clone_default=true)::Vector{String}
     return [r for d in depots() for r in registries(d)]
 end
 
-
+# path -> (mtime, TOML Dict)
 const REGISTRY_CACHE = Dict{String, Tuple{Float64, Dict{String, Any}}}()
 
 function read_registry(reg_file)
     t = mtime(reg_file)
     if haskey(REGISTRY_CACHE, reg_file)
-        println("Cache hit")
         prev_t, registry = REGISTRY_CACHE[reg_file]
         t == prev_t && return registry
     end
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
This seems fine with me. What's the speed of parsing this these days? When I wrote the regex code, parsing the registry was pretty slow and the regex craziness was the easiest way (for me) to make it really fast.

KristofferC:
Parsing the `Registry.toml` takes ~11ms. Considering the first `Pkg.add` takes about 7 seconds and later adds ~0.5 seconds, I think this should be ok. It also isn't slower than what it replaces. While a single pass with the regex function is faster, it has to pass multiple times as more part of the dependency graph gets discovered since only those uuids that we are explicitly looking for is recorded. Also, it is nice not to have to reparse it everytime you do a tab completion.

src/Types.jl
```diff
@@ -257,43 +257,46 @@ mutable struct EnvCache
     # registered package info:
     uuids::Dict{String,Vector{UUID}}
     paths::Dict{UUID,Vector{String}}
+    names::Dict{UUID,Vector{String}}
```
KristofferC: It's a bit unfortunate to store things like this where each `Vector{String}` is in the vast majority of elements in `names` only one element.

Storing it something like

```
uuids_collection::Vector{UUID}
uuids::Dict{String, UnitRange{Int}}
```

and extracting `vew`s from that might be better.

Issue #529, opened 2018-07-24
status help text copypasta


pbouffard:
Copypasta in help text for `status` (last sentence is duplicated):

```
(v0.7) pkg> help status
  status
  status [-p|--project]
  status [-m|--manifest]

  Show the status of the current environment. By default, the full contents of the project file is
  summarized, showing what version each package is on and how it has changed since the last git commit
  (if in a git repo), as well as any changes to manifest packages not already listed. In --project mode,
  the status of the project file is summarized. In --project mode, the status of the project file is
  summarized.
```

https://github.com/JuliaLang/Pkg.jl/blob/c113826d9fffb4717daa3cbef3a97d97c1a8a67b/src/REPLMode.jl#L443

PR #530, opened 2018-07-24
Update copypasta line (#529)

pbouffard:
I'm note sure what goes in `???` but just to get the PR rolling.


pbouffard:
Apologies, probably doing this wrong, I'm used to a somewhat different PR process. But I learn fast. :)

PR #531, opened 2018-07-25
RFC: move extra deps into [extras] section

StefanKarpinski:
This is an alternative to defining the main deps as the ones that are not mentioned in any target in the [targets] section. Instead this puts extra dependencies in an [extras] section which has the same layout as [deps]. cc @KristofferC, @fredrikekre; this is similar to Fredrik's proposal in response to #526 but I didn't like the name `[targetdeps]` so I called it `[extras]` instead.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -689,13 +689,18 @@ Testing...
 
 #### Test-specific dependencies
 
-Sometimes one might want to use some packages only at testing time but not enforce a dependency on them when the package is used.
-This is possible by adding dependencies to a "test target" to the Project file. Here we add the `Test` standard library as a
-test-only dependency by adding the following to the Project file:
+Sometimes one might want to use some packages only at testing time but not
+enforce a dependency on them when the package is used. This is possible by
+adding `[extra]` dependencies and adding a a "test target" to the Project file.
+Here we add the `Test` standard library as a test-only dependency by adding the
+following to the Project file:
 
 ```
-[targets.test.deps]
+[extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+
+[targets]
+test = ["Test"]
 ```
 
 We can now use `Test` in the test script and we can see that it gets installed on testing:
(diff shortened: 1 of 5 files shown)


tpapp:
I like this structure for the `.toml` file very much: I can see which packages are needed for just loading the package, and it specifies the extras in one place. It also allows adding new `targets` seamlessly, should the need arise. I also think that `target = '*'` is a nice shorthand.

Is `target = nothing` equivalent to not having `target` in `[extras]`? What's the use case for specifying it explicitly?

fredrikekre:
So if I'm reading this correctly we will now have something like
```
[deps]
Example = <uuid>

[extras] 
Test = <uuid>
Compiler = <uuid>

[targets]
test = ["Test"]
build = ["Compiler"]
```
which looks better than before. I am just wondering what the advantage is with this design over the original one
```
[deps]
Example = <uuid>

[targets.test.deps]
Test = <uuid>

[targets.build.deps]
Compiler = <uuid>
```

KristofferC:
The argument is that if you e.g. the same dependency as a test target and a build target you don't want to list the uuid twice.
This design is better than before in that you list your name space in one place (name => uuid) and then the usage in the other places (just name).

KristofferC:
:+1:

Would be good to update the docs as well.

fredrikekre:
Yea, makes sense.

StefanKarpinski:
> I like this structure for the `.toml` file very much:

Great, I'm glad to hear it 👍

> Is `target = nothing` equivalent to not having target in `[extras]`?

It's just an internal way of asking for the main target, i.e. `target === nothing`, i.e. "no target".

StefanKarpinski:
Right, the argument is:
* Keep all the UUID mappings in only two places—`[deps]` and `[extras]`
* Make sure that names are consistently used across targets—it would be weird for one target to use a name to mean one thing and a different target to mean another thing
* We should require that the name mappings in `[deps]` and `[extras]` are disjoint so that can't introduce any conflicting names either
* The `[targets.test.deps]` structure is just kind of ugly and overly nested—it doesn't have the "feel" of a nice TOML schema
* This matches the way I want to do options where you write something like this:
  ```toml
  [options]
  graphics = ["Gtk", "Qt", "TclTk"]
  compiler = ["Clang", "GCC", "MSVC"]
  ```
  Having the syntax for targets and options be consistent will make these files easier to understand.

PR #535, opened 2018-07-26
don't allow loading packages from LOAD_PATH when testing or building

KristofferC:
also put deps in Manifest even for stdlibs and use them in the resolver

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -146,7 +146,7 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
     if haskey(project, "version")
         pkg.version = VersionNumber(project["version"])
     else
-        @warn "project file for $(pkg.name) is missing a `version` entry"
+        # @warn "project file for $(pkg.name) is missing a `version` entry"
         set_maximum_version_registry!(ctx.env, pkg)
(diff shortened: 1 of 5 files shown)


StefanKarpinski:
We should do this, but I don't believe it's ever passed tests...

KristofferC:
Sure, it did. It had a straight flush at https://github.com/JuliaLang/julia/pull/27572 when I said it was ready for review.

StefanKarpinski:
I don't recall ever seeing it in a passing, mergeable state. When it gets there again, please merge.

src/Operations.jl
```diff
@@ -146,7 +146,7 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
     if haskey(project, "version")
         pkg.version = VersionNumber(project["version"])
     else
-        @warn "project file for $(pkg.name) is missing a `version` entry"
+        # @warn "project file for $(pkg.name) is missing a `version` entry"
```
StefanKarpinski: Post-review: how come this is commented out now?

KristofferC: Sorry, it was spammy and made people too panicky about Project files.

PR #536, opened 2018-07-26
improve printing when running tagets

KristofferC:
We used to print the diff vs the current project when testing a dependency. This was typically always very spammy and non-interesting. This instead just shows the "absolute" (non-diff) dependencies that are used when testing.


src/Types.jl
```diff
@@ -337,6 +337,7 @@ Base.@kwdef mutable struct Context
     graph_verbose::Bool = false
     stdlibs::Dict{UUID,String} = gather_stdlib_uuids()
     # Remove next field when support for Pkg2 CI scripts is removed
+    currently_running_target::Bool =false
```
StefanKarpinski: Space after `=`.

PR #543, opened 2018-07-28
Make `Pkg.activate(path)` behave like `pkg> activate path`.

fredrikekre:
Just moves the logic from REPLMode to API and make the REPL mode call API.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -570,7 +570,28 @@ function status(ctx::Context, mode=PKGMODE_PROJECT)
 end
 
 function activate(path::Union{String,Nothing}=nothing)
+    if path !== nothing
+        devpath = nothing
+        env = Base.active_project() === nothing ? nothing : EnvCache()
+        if env !== nothing && haskey(env.project["deps"], path)
+            uuid = UUID(env.project["deps"][path])
+            info = manifest_info(env, uuid)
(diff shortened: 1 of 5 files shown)


test/api.jl
```diff
@@ -0,0 +1,39 @@
+module APITests
+
+using Pkg, Test
+
+@testset "Pkg.activate" begin
```
fredrikekre: This testset exactly mirrors the one for activate in repl.jl, but with using the API. A bit unfortunate, would be nice to tests equivalent things like this in the same place

KristofferC:
Yes this is good. Should be as little logic in the REPL functions as possible.

src/API.jl
```diff
@@ -557,7 +557,30 @@ function status(ctx::Context, mode=PKGMODE_PROJECT)
 end
 
 function activate(path::Union{String,Nothing}=nothing)
-    Base.ACTIVE_PROJECT[] = Base.load_path_expand(path)
+    env = Base.active_project() === nothing ? nothing : EnvCache()
+    return activate(env, path)
+end
+function activate(env::Union{EnvCache,Nothing}, path::Union{String,Nothing})
```
fredrikekre: Maybe we can call this `_activate`, or simply merge those two functions.

fredrikekre:
Tests pass locally before the precompile default julia commit.

PR #553, opened 2018-08-01
Fix parser (lexer)

### LINKED ISSUE #516: pkg> ?dev
It would be nice if this worked (in addition to `pkg> help dev`):
```
(v0.7) pkg> ?dev
ERROR: invalid command: "?dev"
```

I feel like it did at some point.

Edit: It is `pkg> ? dev` that works. Would still be nice to support this if possible, since you request help without the space in the regular julia REPL.

00vareladavid:
Should fix #516


KristofferC:
Can we add a test for #516?

00vareladavid:
Sure!

KristofferC:
Thanks

00vareladavid:
Welcome!

PR #559, opened 2018-08-02
Refactor `activate` to avoid complexity in the main method

00vareladavid:
More idiomatic too, I think

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -569,8 +569,7 @@ function status(ctx::Context, mode=PKGMODE_PROJECT)
     return
 end
 
-activate() = activate(nothing)
-activate(path::Nothing) = (Base.ACTIVE_PROJECT[] = Base.load_path_expand(path))
+activate() = (Base.ACTIVE_PROJECT[] = Base.load_path_expand(nothing))
 function activate(path::String)
     devpath = nothing
     env = Base.active_project() === nothing ? nothing : EnvCache()
--- a/src/API.jl
+++ b/src/API.jl
@@ -569,29 +569,28 @@ function status(ctx::Context, mode=PKGMODE_PROJECT)
     return
 end
 
-function activate(path::Union{String,Nothing}=nothing)
-    if path !== nothing
-        devpath = nothing
-        env = Base.active_project() === nothing ? nothing : EnvCache()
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -569,29 +569,28 @@ function status(ctx::Context, mode=PKGMODE_PROJECT)
...
-        # 3. activate the non-existing directory (e.g. as in `pkg> activate .` for initing a new env)
-        if Types.isdir_windows_workaround(path)
-            path = abspath(path)
-        elseif devpath !== nothing
-            path = abspath(devpath)
-        else
-            path = abspath(path)
-        end
+activate() = activate(nothing)
+activate(path::Nothing) = (Base.ACTIVE_PROJECT[] = Base.load_path_expand(path))
```
fredrikekre: I don't think this method needs to be exposed, perhaps just replace these two methods with
```
activate() = (Base.ACTIVE_PROJECT[] = Base.load_path_expand(nothing))
```
and update https://github.com/JuliaLang/Pkg.jl/blob/06c96968e2216fecbf873460db135aea711d0e28/src/REPLMode.jl#L606 to 
```
return API.activate() 
```
?

00vareladavid: Ok, that seems nicer to me too

src/API.jl
```diff
@@ -569,29 +569,27 @@ function status(ctx::Context, mode=PKGMODE_PROJECT)
...
-        # 2. if path exists in deps, and the dep is deved, activate that path (`devpath` above)
-        # 3. activate the non-existing directory (e.g. as in `pkg> activate .` for initing a new env)
-        if Types.isdir_windows_workaround(path)
-            path = abspath(path)
-        elseif devpath !== nothing
-            path = abspath(devpath)
-        else
-            path = abspath(path)
-        end
+activate() = (Base.ACTIVE_PROJECT[] = Base.load_path_expand(nothing))
```
Evizero: isn't  `Base.load_path_expand(nothing))` simply `nothing` ?

00vareladavid: Yes, I hadn't considered that. But given that `activate` is already coupled with `load_path_expand`, I think it makes sense to be explicit and avoid "magic values".

PR #561, opened 2018-08-02
Remove `mode` field from `PackageSpec`

00vareladavid:
It seems that the main use for this was to decide whether to use `manifest_resolve!` or `package_deps_resolve!`. I don't think it makes much sense to have the packages carry this around everywhere and have the calling function set the mode for each package. 

A new function, `pkg_resolve!`, handles this now.


KristofferC:
You need to update the doc strings for all the methods. Also, there is nothing now preventing you from mixing modes in the package list but this would prevent that.

00vareladavid:
I tried grepping through the code and it seems `PackageSpec` is the only affected docstring. But I definitely could have missed some!

KristofferC:
Since this has been documented now it would be breaking to remove so I think we can keep it as is for now.

00vareladavid:
ok, that's a fair point

PR #563, opened 2018-08-02
`depots()[1]` -> `depots1()`

fredrikekre:
When `DEPOT_PATH` is empty, this throws a better error message compared to a `BoundsError`.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -5,8 +5,14 @@ import REPL
 using REPL.TerminalMenus
 
 depots() = Base.DEPOT_PATH
-logdir() = joinpath(depots()[1], "logs")
-devdir() = get(ENV, "JULIA_PKG_DEVDIR", joinpath(depots()[1], "dev"))
+function depots1()
+    d = depots()
+    isempty(d) && cmderror("no depots found in DEPOT_PATH")
+    return d[1]
+end
+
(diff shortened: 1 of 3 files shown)


src/Pkg.jl
```diff
@@ -5,8 +5,14 @@ import REPL
 using REPL.TerminalMenus
 
 depots() = Base.DEPOT_PATH
-logdir() = joinpath(depots()[1], "logs")
-devdir() = get(ENV, "JULIA_PKG_DEVDIR", joinpath(depots()[1], "dev"))
+function depots1()
+    d = depots()
+    isempty(d) && cmderror("no depots found in DEPOT_PATH")
```
fredrikekre: maybe this should just be a regular error? Or I guess this will make more sense after #475

src/Pkg.jl
```diff
@@ -5,8 +5,14 @@ import REPL
 using REPL.TerminalMenus
 
 depots() = Base.DEPOT_PATH
-logdir() = joinpath(depots()[1], "logs")
-devdir() = get(ENV, "JULIA_PKG_DEVDIR", joinpath(depots()[1], "dev"))
+function depots1()
```
StefanKarpinski: This should really be renamed to `user_depot` since that's a much more meaningful name for it.

Issue #479, opened 2018-07-09
warn about invalid/unexpected sections in Project.toml


tpapp:
I mispelled `[targets.test.deps]` as `[target.test.deps]` (note lack of `s` after `target`); a warning would have helped me locate the problem quicker.

fredrikekre:
(You might not have misspelled, I did had the same problem, but that was because I copied from the docs; https://github.com/JuliaLang/Pkg.jl/pull/474)

tpapp:
That may have been the case, I don't remember exactly. But in any case, a warning would be useful for unrecognized sections.

Issue #212, opened 2018-03-17
Add a citation function


ChrisRackauckas:
It would be nice to have a way to grab a citation for a package. In many cases like DifferentialEquations.jl such a citation explanation [can be pretty long](http://juliadiffeq.org/citing.html), but we should have an easy way that everyone knows how to get to such information. For a default fallback, it could use [the JORS style](https://openresearchsoftware.metajnl.com/about/#q11)

Datseris:
I give +1 here.

Another way is to allow for packages that have published papers associated with them to simply pass their DOIs to this information. (or maybe Chris also meant this)

fredrikekre:
Couldn't this informaton be printed by e.g. a `DifferentialEquations.cite()` function? Why is it useful to access this via the package manager?

Datseris:
My guess is that it won't be global across different packages. Unless we all agree on using this `.cite()` format, the user will have to search for the function name for each package, which is pretty much the same as searching for the citing information in the documentation.

Another reason why I think this is a plus, is because Julia is really geared towards scientific computing, which means that having DOIs associated with softwares is not a far fetch.

ChrisRackauckas:
`cite(::Module)` could work too.

KristofferC:
Imo, we should just have a `[metadata]` table where people can put whatever they want, and it is completely ignored by the package manager. If people want to put a `cite` key in there, go ahead.

StefanKarpinski:
The main thing to do here is just to agree on a way to store and use the information across packages. Should it be recorded as a single text block that is just dumped out when someone asks to cite a package? Or should there be individual fields that get pieced together to form a citation entry with a particular format? IIRC, different publications and people prefer different citation formats. How does R deal with all of this? If the citation information is just a block of TeX, is there any reason not to just have a `cite.tex` file in a project and use that as the citation information?

The part of this that would make it relevant to the package manager is that you could get a block of citations for an entire project, e.g. `pkg> cite` and it prints the citations for all of the packages in your project file (or manifest) that has citation entries plus Julia itself. That way you can just run `pkg> cite` and paste the output into your references section (or do it automatically).

StefanKarpinski:
[...]

PR #568, opened 2018-08-03
[WIP] Increase integrity of project/manifest

00vareladavid:
A (small) step forward in making sure that Pkg gets clean input.

Should fix: #479

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -424,15 +424,41 @@ function write_env_usage(manifest_file::AbstractString)
     close(io)
 end
 
-project_toplevel_spec = [
-    "desc", "keywords", "license", "name", "authors", "uuid",
-    "deps", "extras", "targets",
+project_toplevel_specs = [
+    semver_spec("^0") => [
+        "desc", "keywords", "license", "name", "authors", "uuid", "metadata",
+        "deps", "extras", "targets",
+    ],
+    semver_spec("^0.5") => ["introduced0.5"],
+    semver_spec("^0.6") => ["introduced0.6"],
+    semver_spec("^1") => [
+        "desc", "keywords", "license", "name", "authors", "uuid", "metadata",
+        "deps", "extras", "targets", "newone",
+    ],
 ]
 
 function enforce_project!(project)
-    for key in keys(project)
-        key in project_toplevel_spec ||
-            cmderror("Unrecognized entry in project file: '$key'")
+    # grab Project.toml version
+    pkg_version = get(project, "pkg-version", nothing)
+    if pkg_version === nothing
+        @warn("no Pkg version found in `Project.toml`: assuming old version")
+        pkg_version = "0.1.0"
+    end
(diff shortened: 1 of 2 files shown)


src/Types.jl
```diff
@@ -424,13 +424,27 @@ function write_env_usage(manifest_file::AbstractString)
...
 
+project_toplevel_spec = [
+    "desc", "keywords", "license", "name", "authors", "uuid",
+    "deps", "extras", "targets",
+]
+
+function enforce_project!(project)
+    for key in keys(project)
+        key in project_toplevel_spec ||
+            cmderror("Unrecognized entry in project file: '$key'")
```
fredrikekre: Should this be enforced by Pkg? What if I want to add some other metadata field to my project file?

00vareladavid: Good point. Maybe warn and remove from dictionary?

00vareladavid: I would prefer input checks be as strong as possible. Maybe a toplevel toml field that users can shove anything they want into.

fredrikekre: What if we add another allowed field in the future? Then a newer project file would error on older Pkg versions.

00vareladavid: Another good point! If we want really strict input that would mean we would have to add a pkg-version field and map a pkg-version with some sort of spec for toml files. ...but that seems overzealous at this point :laughing: .

For now, how about: a toplevel `user`(?) field + warn about any unrecognized fields. If you feel really strongly against this check, then we can make no warnings the default allow users to opt-in with some '--strict-parse' flag.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/issues/212#issuecomment-373947965

src/Types.jl
StefanKarpinski: It’s fine for the package manager to be strict about the structure of things it knows about but it should just ignore anything that it doesn’t know about. This is really crucial for adding features in the future and continuing to work gracefully with older versions of the package manager and language in general. Ideally it should only verify that the structure confirms to what it requires to function correctly.

KristofferC: > This is really crucial for adding features in the future and continuing to work gracefully with older versions of the package manager and language in general.

If it ignores everything except what we have implemented today, then adding features will break existing file that uses the name of the feature since everything is in the same namespace. Enforcing that all user customization need to happen in its own namespace `metadata` solves that.

00vareladavid:
The latest commit is just a sketch, but I've tried to handle your concerns. At the very least, is seems prudent to begin tagging the toml files (even if we don't do anything with them right now) so that compatibility can be handled cleanly in the future.

The idea is to keep a history of features and use it to check the integrity of toml files.

The only thing this system wouldn't handle is reading a toml file created by a `Pkg` that has a higher major number. I think this is justified since the field names could be the same but can have differing semantics.

00vareladavid:
Closing in favor of #857

PR #579, opened 2018-08-03
Propagate track-allocation option to test process.

tkoolen:
Thanks to @fredrikekre for pointing out what to change. Not sure how to test this; I couldn't find how other options are being tested.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1310,6 +1310,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false)
             --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
             --check-bounds=yes
             --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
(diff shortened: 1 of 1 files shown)


fredrikekre:
> Not sure how to test this; I couldn't find how other options are being tested.

I don't think they are. Did you try this locally?

tkoolen:
I made sure that current tests pass and that `"--track-allocation=$(("none", "user", "all")[Base.JLOptions().malloc_log + 1])"` prints the correct thing when run from three REPL sessions with the three different options for `--track-allocation`. I then ran

```julia
import Pkg
Pkg.add("LoopThrottle") # random example package
Pkg.test("LoopThrottle")
```

from a REPL with `--track-allocation=user` and ensured that a `LoopThrottle.jl.mem` file was created in `import LoopThrottle; dirname(pathof(LoopThrottle))`.

I then tried the same with `Dates` and `track-allocation=all`, but didn't see any `*.jl.mem` files in `import Dates; dirname(pathof(Dates))`. But maybe testing stdlib packages like this isn't supposed to work anyway?

KristofferC:
Not sure with the stdlib but if it works with normal packages, good enough for me.

PR #580, opened 2018-08-03
allow using a path kword arg to PackageSpec

KristofferC:
Also fixes https://github.com/JuliaLang/Pkg.jl/issues/556. Seems this chunk of code was copy pasted and occurred in two different places @00vareladavid


00vareladavid:
Ah yes, that was my mistake. Sorry about that

PR #581, opened 2018-08-03
remove ability to give a git revision to devved packages

KristofferC:
Doing `dev Foo#branch1` and then `dev Foo#branch2` and then not get a change of branch is confusing. For now, just allow downloading the repo at the main branch and then one can just use git to get the repo in whatever state is desired.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -543,13 +543,10 @@ function handle_repos_develop!(ctx::Context, pkgs::AbstractVector{PackageSpec}, 
                 project_path = mktempdir()
                 cp(repo_path, project_path; force=true)
                 LibGit2.with(LibGit2.GitRepo(project_path)) do repo
-                    rev = pkg.repo.rev
-                    if isempty(rev)
-                        if LibGit2.isattached(repo)
-                            rev = LibGit2.branch(repo)
-                        else
-                            rev = string(LibGit2.GitHash(LibGit2.head(repo)))
-                        end
+                    if LibGit2.isattached(repo)
+                        rev = LibGit2.branch(repo)
+                    else
+                        rev = string(LibGit2.GitHash(LibGit2.head(repo)))
                     end
                     gitobject, isbranch = get_object_branch(repo, rev)
                     try
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -511,7 +511,9 @@ function handle_repos_develop!(ctx::Context, pkgs::AbstractVector{PackageSpec}, 
                     # Relative paths are given relative pwd() so we
                     # translate that to be relative the project instead.
                     # `realpath` is needed to expand symlinks before taking the relative path.
-                    pkg.path = relpath(realpath(abspath(pkg.repo.url)), realpath(dirname(ctx.env.project_file)))
+                    project_dir = dirname(ctx.env.project_file)
+                    isdir(project_dir) && (project_dir = realpath(project_dir))
+                    pkg.path = relpath(realpath(abspath(pkg.repo.url)), project_dir)
                 end
                 folder_already_downloaded = true
                 project_path = pkg.repo.url
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -350,13 +350,13 @@ If we try to `dev` a package at some branch that already exists at `~/.julia/dev
 For example:
 
 ```
-(v0.7) pkg> dev Example#master
+(v0.7) pkg> dev Example
   Updating git-repo `https://github.com/JuliaLang/Example.jl.git`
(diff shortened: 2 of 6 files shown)


KristofferC:
I don't really understand how this test error happens due to this PR

```
Generating project Foo:
    Foo/Project.toml
    Foo/src/Foo.jl
(pkg.repo).url = "Foo"
abspath((pkg.repo).url) = "/private/var/folders/64/76tk_g152sg6c6t0b4nkn1vw0000gn/T/tmppjyCtY/HelloWorld/Foo"
realpath(abspath((pkg.repo).url)) = "/private/var/folders/64/76tk_g152sg6c6t0b4nkn1vw0000gn/T/tmppjyCtY/HelloWorld/Foo"
(ctx.env).project_file = "/var/folders/64/76tk_g152sg6c6t0b4nkn1vw0000gn/T/tmpFTWtM6/environments/v1.0/Project.toml"
ERROR: SystemError: realpath: No such file or directory
Stacktrace:
 [1] #systemerror#39(::Nothing, ::Function, ::Symbol, ::Bool) at ./error.jl:106
 [2] systemerror at ./error.jl:106 [inlined]
 [3] realpath(::String) at ./path.jl:322
 [4] macro expansion at ./show.jl:556 [inlined]
```

fredrikekre:
I got that error too in some of my PRs, really strange...

fredrikekre:
Wait, did you add those `@show`s for this PR? They look like the ones I added when debugging https://github.com/JuliaLang/Pkg.jl/pull/490

KristofferC:
I added them locally.

00vareladavid:
It seems isolating that test allows it to pass... maybe its interacting with the one above it?

fredrikekre:
I think we can safely change https://github.com/JuliaLang/Pkg.jl/blob/d2b8419063861e2a2315686a0db332603b2871e9/src/Types.jl#L248 to
```julia
project_file = joinpath(realpath(env), Base.project_names[end]) 
```
and
https://github.com/JuliaLang/Pkg.jl/blob/d2b8419063861e2a2315686a0db332603b2871e9/src/Types.jl#L250-L251
to
```julia
project_file = endswith(env, ".toml") ? abspath(realpath(env)) : 
    abspath(realpath(env), Base.project_names[end]) 
```
?

src/Types.jl
```diff
@@ -511,7 +511,9 @@ function handle_repos_develop!(ctx::Context, pkgs::AbstractVector{PackageSpec},
                     # Relative paths are given relative pwd() so we
                     # translate that to be relative the project instead.
                     # `realpath` is needed to expand symlinks before taking the relative path.
-                    pkg.path = relpath(realpath(abspath(pkg.repo.url)), realpath(dirname(ctx.env.project_file)))
+                    project_dir = dirname(ctx.env.project_file)
+                    isdir(project_dir) && (project_dir = realpath(project_dir))
```
fredrikekre: This needs to call `realpath` on the part of `ctx.env.project_file` which exists, i.e. given
`/a/b/c/Project.toml`, if only `/a/b` exists, we need something like `joinpath(realpath(/a/b), c/Project.toml)`.

src/Types.jl
```diff
@@ -511,7 +511,9 @@ function handle_repos_develop!(ctx::Context, pkgs::AbstractVector{PackageSpec},
                     # Relative paths are given relative pwd() so we
```
fredrikekre: We can also change this, such that when deving a path you write the path relative the project file, but that will be very annoying I think.

src/Types.jl
fredrikekre: Basically we need a `safe_realpath`, somehting like
```
julia> function safe_realpath(path)
           ispath(path) && return realpath(path)
           a, b = splitdir(path)
           return joinpath(safe_realpath(a), b)
       end
safe_realpath (generic function with 1 method)

julia> str = "/home/fredrik/Foo/test/Project.toml"
"/home/fredrik/Foo/test/Project.toml"

julia> realpath(str)
ERROR: SystemError: realpath: No such file or directory
Stacktrace:
 [1] #systemerror#39(::Nothing, ::Function, ::Symbol, ::Bool) at ./error.jl:106
 [2] systemerror at ./error.jl:106 [inlined]
 [3] realpath(::String) at ./path.jl:322
 [4] top-level scope at none:0

julia> safe_realpath(str)
"/home/fredrik/Foo/test/Project.toml"
```

src/Types.jl
KristofferC: Yes

00vareladavid:
Well that was demoralizing... :sob: . Maybe I'm missing something: is it really `realpath` that is failing? Wouldn't that be a Julia bug?

fredrikekre:
:tada:

PR #584, opened 2018-08-04
WIP refactor parser

00vareladavid:
The removed check is redundant with the check [here](https://github.com/JuliaLang/Pkg.jl/blob/fb761ea6784436b9dfd16878767ef29cf5c73ef7/src/REPLMode.jl#L438)


KristofferC:
Why close?

00vareladavid:
I'm working on a set of related commits (refactoring the parser a bit). I probably should have just labeled this WIP instead of closing.

00vareladavid:
Actually, I'll just include it in #662 :sweat_smile:

PR #585, opened 2018-08-05
Use declarative format when disallowing `develop` revisions

00vareladavid:
Introduced in #581. The declarative system already supports this.

I definitely could have used better names for this:
- `ARG_RAW` : don't do any checking on the kinds of arguments
- `ARG_PKG`: only allow package identifiers (i.e. the `String`s which `parse_package` understands)
- `ARG_VERSION`: allow `Example` and `Example@0.1.0`, but not `Example#foo`
- `ARG_REV`: allow `Example` and `Example#foo`, but not `Example@0.1.0`
- `ARG_ALL`: allow any kind of package specification

Maybe specifying the negation would be more natural?
- `ARG_RAW` -> `ARG_RAW`
- `ARG_PKG` -> `ARG_PKG_BARE`
- `ARG_VERSION` -> `ARG_PKG_NO_REV`
- `ARG_REV` -> `ARG_PKG_NO_VERSION`
- `ARG_ALL` -> `ARG_PKG`

If so, I can change it

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -397,9 +397,6 @@ function package_args(args::Vector{Token}, spec::CommandSpec)::Vector{PackageSpe
         elseif arg isa VersionRange
             pkgs[end].version = arg
         elseif arg isa Rev
-            if spec.kind == CMD_DEVELOP
-                pkgerror("a git revision cannot be given to `develop`")
-            end
             pkg = pkgs[end]
             if pkg.repo == nothing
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
We generally always try to avoid double negatives, so if you have a choice, use a positive option with a default of true and allow opting for the negative by explicitly setting the option to false.

00vareladavid:
Ok, that makes sense. I'll change it

PR #588, opened 2018-08-07
Registry functionality with REPL and API support

### LINKED ISSUE #453: Command to add registry
It would be nice if there was a way to add registries from the `Pkg` REPL mode. As it is now, I have to clone the registry into `.julia/registries`. Perhaps the following syntax would work
```julia
(v0.7) pkg> add registry <path/repo here>
```

I took a look at the `Pkg` code, but unfortunately adding this sort of feature is way over my head.

fredrikekre:
This adds some registry functionality.

| REPL | API |
|:-----|:-----|
| `pkg> registry add General` | `Pkg.Registry.add("General")` |
| `pkg> registry add General=uuid` | `Pkg.Registry.add(RegistrySpec(name="General", uuid = uuid))` |
| `pkg> registry add www.registry.com` | `Pkg.Registry.add(RegistrySpec(url="www.registry.com")` |

and similar with `registry up`/`Pkg.Registry.up` and `registry rm`/`Pkg.Registry.rm`

This also differentiates registries based on uuid, instead of just name. This means that registries are now stored in `.julia/registries/name/uuidslug` instead of just `.julia/registries/name`.

### Demo
```
pkg> registry add General

pkg> registry add https://github.com/fredrikekre/Registry"
   Cloning registry from "https://github.com/fredrikekre/Registry"
     Added registry `Registry` to `~/.julia/registries/Registry`

pkg> registry st
Registry Status 
 [23338594] General (https://github.com/JuliaRegistries/General.git)
 [ae0cb698] Registry (https://github.com/fredrikekre/Registry.git)

pkg> registry up General
  Updating registry at `~/.julia/registries/General/SxLXF`

pkg> registry up
  Updating registry at `~/.julia/registries/General/SxLXF`
  Updating registry at `~/.julia/registries/Registry/xhgIZ`
  Updating git-repo `https://github.com/fredrikekre/Registry`

pkg> registry rm General
  Removing registry `General` from ~/.julia/registries/General/SxLXF

pkg> registry rm Registry
  Removing registry `Registry` from ~/.julia/registries/Registry/xhgIZ
```

fix #453

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -85,6 +85,7 @@ function normalize_url(url::AbstractString)
 end
 
 function clone(url, source_path; header=nothing, kwargs...)
+    @assert !isdir(source_path) || isempty(readdir(source_path))
     Pkg.Types.printpkgstyle(stdout, :Cloning, header == nothing ? "git-repo `$url`" : header)
     transfer_payload = MiniProgressBar(header = "Fetching:", color = Base.info_color())
     callbacks = LibGit2.Callbacks(
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -9,7 +9,7 @@ import REPL
 import REPL: LineEdit, REPLCompletions
 
 import ..devdir, ..Types.casesensitive_isdir, ..TOML
-using ..Types, ..Display, ..Operations, ..API
+using ..Types, ..Display, ..Operations, ..API, ..Registry
 
 #################
 # Git revisions #
@@ -81,7 +81,7 @@ end
                    CMD_STATUS, CMD_TEST, CMD_GC, CMD_BUILD, CMD_PIN,
                    CMD_FREE, CMD_GENERATE, CMD_RESOLVE, CMD_PRECOMPILE,
                    CMD_INSTANTIATE, CMD_ACTIVATE, CMD_PREVIEW,
-                   CMD_REGISTRY_ADD,
+                   CMD_REGISTRY_ADD, CMD_REGISTRY_RM, CMD_REGISTRY_UP, CMD_REGISTRY_STATUS,
                    )
 @enum(ArgClass, ARG_RAW, ARG_PKG, ARG_VERSION, ARG_REV, ARG_ALL)
 struct ArgSpec
@@ -145,9 +145,9 @@ function CommandSpecs(declarations::Vector{CommandDeclaration})::Dict{String,Com
     return specs
 end
 
-###################
-# Package parsing #
-###################
+############################
+# Package/registry parsing #
+############################
 let uuid = raw"(?i)[0-9a-z]{8}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{4}-[0-9a-z]{12}(?-i)",
     name = raw"(\w+)(?:\.jl)?"
     global const name_re = Regex("^$name\$")
@@ -177,6 +177,41 @@ function parse_package(word::AbstractString; add_or_develop=false)::PackageSpec
         pkgerror("`$word` cannot be parsed as a package")
     end
 end
+
+# Registries can be identified through: uuid, name, or name+uuid
+# when updating/removing. When adding we can accept a local path or url.
+function parse_registry(word::AbstractString; add=false)::RegistrySpec
+    word = replace(word, "~" => homedir())
+    registry = RegistrySpec()
+    if add && Types.isdir_windows_workaround(word) # TODO: Should be casesensitive_isdir
+        if isdir(joinpath(word, ".git")) # add path as url and clone it from there
+            registry.url = abspath(word)
+        else # put the path
+            registry.path = abspath(word)
+        end
+    elseif occursin(uuid_re, word)
+        registry.uuid = UUID(word)
+    elseif occursin(name_re, word)
+        registry.name = String(match(name_re, word).captures[1])
+    elseif occursin(name_uuid_re, word)
+        m = match(name_uuid_re, word)
+        registry.name = String(m.captures[1])
+        registry.uuid = UUID(m.captures[2])
+    elseif add
+        # Guess it is a url then
+        registry.url = String(word)
+    else
+        pkgerror("`$word` cannot be parsed as a registry")
+    end
+    return registry
+end
+
+function parse_registry(raw_args::Vector{String}; add=false)
+    regs = RegistrySpec[]
+    foreach(x -> push!(regs, parse_registry(x; add=add)), raw_args)
+    return regs
+end
+
 
 ################
 # REPL parsing #
@@ -347,7 +382,7 @@ end
 ##############
 const PackageIdentifier = String
 const PackageToken = Union{PackageIdentifier, VersionRange, Rev}
-const PkgArguments = Union{Vector{String}, Vector{PackageSpec}}
+const PkgArguments = Union{Vector{String}, Vector{PackageSpec}, Vector{RegistrySpec}}
 const APIOptions = Dict{Symbol, Any}
 struct PkgCommand
     spec::CommandSpec
@@ -574,9 +609,6 @@ function do_test!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
     API.test(Context!(ctx), args; collect(api_opts)...)
 end
 
-function do_registry_add!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
-end
-
 do_precompile!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions) =
     API.precompile(Context!(ctx))
 
@@ -632,6 +664,27 @@ function do_develop!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
     API.add_or_develop(Context!(ctx), args; collect(api_opts)...)
 end
 
+# registry commands
+function do_registry_add!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
+    Registry.add(Context!(ctx), args)
+end
+
+function do_registry_rm!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
+    Registry.rm(Context!(ctx), args)
+end
+
+function do_registry_up!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
+    if isempty(args)
+        return Registry.update(Context!(ctx))
+    else
+        return Registry.update(Context!(ctx), args)
+    end
+end
+
+function do_registry_status!(#=ctx::APIOptions,=# args::PkgArguments, api_opts::APIOptions)
+    return Registry.status()
+end
+
 ######################
 # REPL mode creation #
 ######################
@@ -681,13 +734,13 @@ end
 function complete_remote_package(partial)
     cmp = String[]
     julia_version = VERSION
-    for reg in Types.registries(;clone_default=false)
-        data = Types.read_registry(joinpath(reg, "Registry.toml"))
+    for reg in Types.collect_registries()
+        data = Types.read_registry(joinpath(reg.path, "Registry.toml"))
         for (uuid, pkginfo) in data["packages"]
             name = pkginfo["name"]
             if startswith(name, partial)
                 compat_data = Operations.load_package_data_raw(
-                    VersionSpec, joinpath(reg, pkginfo["path"], "Compat.toml"))
+                    VersionSpec, joinpath(reg.path, pkginfo["path"], "Compat.toml"))
                 supported_julia_versions = VersionSpec(VersionRange[])
                 for (ver_range, compats) in compat_data
                     for (compat, v) in compats
@@ -956,17 +1009,6 @@ end
 # SPEC #
 ########
 command_declarations = [
-#=
-"registry" => CommandDeclaration[
-[
-    :kind => CMD_REGISTRY_ADD,
-    :name => "add",
-    :handler => do_registry_add!,
-    :arg_count => 1 => Inf,
-    :description => "Currently just a placeholder for a future command",
-],
-], #registry
-=#
 "package" => CommandDeclaration[
 [   :kind => CMD_TEST,
     :name => "test",
@@ -1155,7 +1197,6 @@ A pinned package has the symbol `⚲` next to its version in the status list.
     :completions => complete_installed_packages,
     :description => "run the build script for packages",
     :help => md"""
-
     build [-v|verbose] pkg[=uuid] ...
 
 Run the build script in `deps/build.jl` for each package in `pkg` and all of their dependencies in depth-first recursive order.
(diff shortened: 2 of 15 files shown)


fredrikekre:
🎉

src/GitTools.jl
```diff
@@ -78,6 +78,7 @@ function normalize_url(url::AbstractString)
 end
 
 function clone(url, source_path; header=nothing, kwargs...)
+    @assert !isdir(source_path) || isempty(readdir(source_path))
```
fredrikekre: I removed my `Pkg` dev dir by accident when cloning to `pwd()` so ...

src/REPLMode.jl
```diff
@@ -1203,6 +1238,41 @@ is modified.
     """,
 ),
 ], #package
+
+["registry"] => CommandDeclaration[
+(
+    CMD_REGISTRY_ADD,
+    ["add"],
+    do_registry_add!,
+    (ARG_RAW, []),
```
fredrikekre: @00vareladavid I could use your help here; currently I put `ARG_RAW` here, and manually call `parse_registry` in the `do_` functions, but I guess I should instead have something like `ARG_REGISTRY` and handle that earlier in the pipeline, such that the `do_` functions obtain `RegistrySpec`s instead of the raw argument? Not sure how to do this, help appreciated.

fredrikekre:
We could potentially have the registry API as `Pkg.add` and dispatch on `PackageSpec` and `RegistrySpec`, but would be ambiguous in the `String` case. I also like the symmetry with `pkg> registry add Foo` and `Pkg.Registry.add("Foo")`. Messing with registries will probably be much less common than working with packages, so I think it is OK that it is a bit more verbose.

src/GitTools.jl
KristofferC: yeah, good idea

src/REPLMode.jl
00vareladavid: Sure! I mentioned [here](https://github.com/JuliaLang/Pkg.jl/pull/585#issuecomment-410911201) that this part of the spec needs more work. I was planning on including including it with some other changes, but I'll get it done first instead. It should support easily adding new tokens.

00vareladavid: Hopefully #595 will do it. `CMD_REGISTRY_ADD` would use `(ARG_RAW, [])` -> `([], parse_registry, [:add => true])`. The other ones would be  `([], parse_registry, [])` I think

fredrikekre:
Ok this is not longer WIP, please review :)

src/REPLMode.jl
```diff
@@ -1203,6 +1238,41 @@ is modified.
...
+    [],
+    nothing,
+),
+(
+    CMD_REGISTRY_RM,
+    ["rm"],
```
00vareladavid: it may be a good idea to include a `remove` alias to mirror the `package` REPL API?

src/Types.jl
```diff
@@ -892,52 +893,251 @@ function ensure_resolved(env::EnvCache,
...
+# Return `RegistrySpec`s of each registry in a depot
+function collect_registries(depot::String)
     d = joinpath(depot, "registries")
-    ispath(d) || return String[]
-    regs = filter!(readdir(d)) do r
-        isfile(joinpath(d, r, "Registry.toml"))
+    regs = RegistrySpec[]
+    ispath(d) || return regs
+    for name in readdir(d)
+        for slug in readdir(joinpath(d, name))
```
00vareladavid: I think this is fine for now, but it will error out if `joinpath(d, name)` is not a directory right? I'm starting to believe we will need to validate depots sometime soon so we can avoid erroring out if assumptions like these don't hold (which shouldn't be the case, but it's easy for users to poke around and mess stuff up).

src/Types.jl
```diff
@@ -892,52 +893,251 @@ function ensure_resolved(env::EnvCache,
...
+            file = joinpath(d, name, slug, "Registry.toml")
+            if isfile(file)
+                registry = read_registry(file)
+                # verify_registry(registry)
+                spec = RegistrySpec(name = registry["name"],
+                                    uuid = UUID(registry["uuid"]),
+                                    url = haskey(registry, "repo") ? registry["repo"] : nothing,
```
00vareladavid: I think `get(registry, "repo", nothing)` is more idiomatic

src/Types.jl
```diff
@@ -892,52 +893,251 @@ function ensure_resolved(env::EnvCache,
...
+# Registries #
+##############
+
+mutable struct RegistrySpec
+    name::Union{String,Nothing}
+    uuid::Union{UUID,Nothing}
+    url::Union{String,Nothing}
+    # the path field can be a local source when adding a registry
+    # otherwise it is the path where the registry is installed
+    path::Union{String,Nothing}
```
00vareladavid: I'm confused about the purpose of `path` here: is it being used for two different things?

src/Types.jl
```diff
@@ -892,52 +893,251 @@ function ensure_resolved(env::EnvCache,
...
+        end
+    end
+    return RegistrySpec[r for d in depots() for r in collect_registries(d)]
+end
+
+function populate_known_registries_with_urls!(registries::Vector{RegistrySpec})
+    known_registries = DEFAULT_REGISTRIES # TODO: Some way to add stuff here?
+    for reg in registries, known in known_registries
+        if reg.uuid !== nothing
+            if reg.uuid === known.uuid
```
00vareladavid: `===` -> `==`?

00vareladavid: Also, I think it would be better to use `findfirst` so that you don't have to iterate over `known_registries` in the case that `reg.uuid === nothing`.

src/Types.jl
```diff
@@ -892,52 +893,251 @@ function ensure_resolved(env::EnvCache,
...
+                reg.url = known.url
+            end
+        elseif reg.name !== nothing
+            if reg.name == known.name
+                named_regs = filter(r -> r.name == reg.name, known_registries)
+                if !all(r -> r.uuid == first(named_regs).uuid, named_regs)
+                    pkgerror("multiple registries with name `$(needle.name)`, please specify with uuid.")
```
00vareladavid: `needle`->`reg`?

src/REPLMode.jl
fredrikekre: Yes, didn't actually know about that alias!

src/Types.jl
```diff
@@ -892,52 +893,251 @@ function ensure_resolved(env::EnvCache,
...
     end
-    return [r for d in depots() for r in registries(d)]
+    return nothing
 end
 
 # path -> (mtime, TOML Dict)
 const REGISTRY_CACHE = Dict{String, Tuple{Float64, Dict{String, Any}}}()
 
-function read_registry(reg_file)
+function read_registry(reg_file, cache=true)
```
00vareladavid: I think `cache` as a kwarg would make calls to `read_registry` more readable.

src/Types.jl
fredrikekre: Yea, it can play the role of the path to where we add it from (if adding from a path) but most often this is just the path where we read the registry from. I am not sure we should allow adding registries from paths, if they are not git repos though.

src/Types.jl
fredrikekre:
```
julia> a = UUID("bdb5d00a-7a24-44d2-be8d-ca1835c73734");

julia> b = UUID("bdb5d00a-7a24-44d2-be8d-ca1835c73734");

julia> a === b
true
```

> Also, I think it would be better to use findfirst

You mean `findfirst(x -> reg.uuid == x.uuid, known_registries)` ?

fredrikekre:
bors try

fredrikekre:
bors try

rapus95:
just out of curiosity what about supplying some sort of reset or at least check function which simply deletes and redownloads the registry. (or checks the registry to be identical to the remote registry) 
Use case is that if you somehow managed to get into a dirty state to ensure that everything is setup as it should be

src/Types.jl
00vareladavid: Ah, I didn't know `UUID` behaved like that, thanks.

And yes, that's what I meant by `findfirst`. It is a detail, so I think it's fine as is too.

fredrikekre:
bors try

DilumAluthge:
@fredrikekre What's the status of this PR? It looks like CI is passing. Is there anything else blocking before you can merge this?

fredrikekre:
There is no rush with this, can be merged just before 1.1. Not merging makes it easier to make bug fixes for 1.0.1 etc.

DilumAluthge:
Sounds good, thanks!

fredrikekre:
bors try

fredrikekre:
bors try

src/REPLMode.jl
```diff
@@ -714,6 +741,28 @@ function do_develop!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
     API.add_or_develop(Context!(ctx), args; collect(api_opts)...)
 end
 
+# registry commands
+function do_registry_add!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions)
+    # TODO: Should accept a --depot option ?
```
fredrikekre: From recent discussions I take this as a no? Instead we assume that `depots1()` is a writeable user depot and put it there?

KristofferC: I think so, yes.

fredrikekre:
bors try

fredrikekre:
bors try

fredrikekre:
bors try

fredrikekre:
Someone wants to review again? Otherwise good to merge?

bors try

src/Types.jl
```diff
@@ -898,63 +899,249 @@ function ensure_resolved(env::EnvCache,
...
+
+clone_or_cp_registries(regs::Vector{RegistrySpec}, depot::String=depots1()) =
+    clone_or_cp_registries(Context(), regs, depot)
+function clone_or_cp_registries(ctx::Context, regs::Vector{RegistrySpec}, depot::String=depots1())
+    populate_known_registries_with_urls!(regs)
+    # TODO Bail early on preview
+    for reg in regs
+        # clone to tmpdir first
+        tmp = mktempdir()
+        if reg.path !== nothing # copy from local source
```
00vareladavid: It might make sense to error out in the case where both `reg.path` and `reg.url` are `!== nothing`. Something like "Ambiguous RegistrySpec". Or maybe feedback that one was chosen over the other.

fredrikekre:
bors try

StefanKarpinski:
What is the purpose of the slugs in registry paths?

StefanKarpinski:
Yes, I don't feel like the motivations for using UUIDs to identify packages apply to registries. I could be convinced otherwise, but the case hasn't really been made.

DilumAluthge:
It seems like a lot of registries have very generic names. For example, the
general registry is simply called “General.” Fredrik has a registry that is
simply called “Registry.”

So suppose I create my own registry and also call it “Registry.” Is it now
impossible for me to use both my “Registry” and Fredrik’s “Registry.”

Given that registries seem to often be named very generically, I definitely
think that users will run into the problem of wanting to use two different
registries that have the same name.

StefanKarpinski:
How about we use `hostname/name` then instead?

DilumAluthge:
I think that would work great!

src/Types.jl
```diff
@@ -900,63 +901,260 @@ function ensure_resolved(env::EnvCache,
...
+        ispath(dirname(regpath)) || mkpath(dirname(regpath))
+        if isdir_windows_workaround(regpath)
+            existing_registry = read_registry(joinpath(regpath, "Registry.toml"))
+            if registry["uuid"] == existing_registry["uuid"]
+                @info("registry `$(registry["name"])` already exist in `$(Base.contractuser(regpath))`.")
+            else
+                throw(PkgError("registry `$(registry["name"])=\"$(registry["uuid"])\"` conflicts with " *
+                    "existing registry `$(existing_registry["name"])=\"$(existing_registry["uuid"])\"`. " *
+                    "To install it you can clone it manually into " *
+                    "`$(Base.contractuser(joinpath(depot, "registries", registry["name"]*"_2")))` or similar."))
```
fredrikekre: I have to say that it feels a bit shitty to leave this so close to working. This is literally the *only* problem with adding another registry with the same name. I am going to change this to a warning and then we clone it into `registries/name-uuid` in case of conflicts.

StefanKarpinski: Ok! Would you rename the existing one to use the same scheme then? I would avoid using the whole UUID since long paths cause problems on Windows at least—use the slug instead.

fredrikekre: We can rename it too I guess.
Just adding a note to self; when a user adds a third registry with the same name it is not enough to just check the path, we need instead to check that there is no registry with the same name in the depot.

fredrikekre: Of course it is 🙂 Just annoying that there is one single spot in the code that can not cope with multiple ones like this.

00vareladavid: FWIW I still vote for just erroring out and keeping it simple. I think filesystem-as-a-database is a nice default

00vareladavid: Since we are relying on GitHub so much already... Perhaps `registries/UserName/RegName` is enough to avoid name collisions?

StefanKarpinski: Not a bad idea—we could support GitLab and others in a similar fashion.

fredrikekre: I think I will leave it like this for now then.

> Perhaps `registries/UserName/RegName` is enough to avoid name collisions?

Yea I guess, but that does not solve the problem raised in the other thread (https://github.com/JuliaLang/Pkg.jl/issues/812#issuecomment-429187763 and https://github.com/JuliaLang/Pkg.jl/issues/812#issuecomment-429484241) where the user wanted to "just make a clone".

StefanKarpinski: I like this plan, seems simpler and easier to work with than slugs or whatever else.

fredrikekre:
I'm done here.
bors try

fredrikekre:
bors try

fredrikekre:
This also cuts down runtime of the test from 300s to 220s locally, basically since it implements #858.

fredrikekre:
bors try

fredrikekre:
A nice sideeffect of this is that you can update registries in lower depots by explicitly calling
```
pkg> registry up Registry
```

test/registry.jl
```diff
@@ -0,0 +1,241 @@
...
+            uuid = "$(pkg_uuids[i])"
+            repo = "https://github.com/JuliaLang/Example.jl.git"
+            """)
+        write(joinpath(regpath, "Example", "Versions.toml"), """
+            ["0.5.1"]
+            git-tree-sha1 = "8eb7b4d4ca487caade9ba3e85932e28ce6d6e1f8"
+            """)
+        write(joinpath(regpath, "Example", "Deps.toml"), """
+            ["0.5"]
+            julia = "0.6-1.0"
```
GunnarFarneback, line 36: This is a copy of `Compat.toml`, not what should go in `Deps.toml`.

PR #594, opened 2018-08-07
add path completions to add/dev in repl

### LINKED ISSUE #408: REPL tabcompletion for paths
Would be nice to have when you are doing something like `dev ../..` etc.

pfitzseb:
These are only triggered if there's a path separator in the string to be completed, which I think is a pretty sensible behaviour.
If you want to test this locally, call `Pkg.REPLMode.repl_init(Base.active_repl)` after `import Pkg`.

Fixes #408

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -697,9 +697,20 @@ function complete_package(s, i1, i2, lastcommand, project_opt)
     if lastcommand in [CMD_STATUS, CMD_RM, CMD_UP, CMD_TEST, CMD_BUILD, CMD_FREE, CMD_PIN]
         return complete_installed_package(s, i1, i2, project_opt)
     elseif lastcommand in [CMD_ADD, CMD_DEVELOP]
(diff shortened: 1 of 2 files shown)


pfitzseb:
Travis failure is obviously unrelated, but tests pass locally on 0.7 for me. Should we start testing Pkg on 0.7 and `nightly`?

fredrikekre:
Why not always complete paths?

pfitzseb:
Because package completions are usually more relevant, no? Though I guess we could also concat both.

fredrikekre:
Right, but it is not often you have a folder with a name of a package anyway I think.

KristofferC:
bors +r

KristofferC:
Seems like the test needs to be special cased a bit on Windows.

KristofferC:
bors try

pfitzseb:
bors try

KristofferC:
Please squash and merge when this is ready

bors delegate+

PR #595, opened 2018-08-07
Declare an argument parser for each command

00vareladavid:
This changes the command declarations.

For example, the `pin` command: 
`(ARG_VERSION, [])` -> `([], parse_pkg, [:valid => [VersionRange]]),`

The first field lists the number of expected arguments **after** they have been parsed (empty means don't count).
The next field specifies the parser to be applied on that command's arguments.
The final field allows a command to key into a parser.


src/REPLMode.jl
```diff
@@ -901,7 +893,7 @@ command_declarations = [
     CMD_REGISTRY_ADD,
     ["add"],
     do_registry_add!,
-    (ARG_PKG, []),
+    ([], nothing, []),
```
fredrikekre: We can also use `identity` as a parser and change `Union{Function,Nothing}` with judt `Function`.

00vareladavid:
Updates for 1.0 from @fredrikekre's #588

KristofferC:
Argh, we should have squashed this.

PR #599, opened 2018-08-08
Remove `--env` meta option

### LINKED ISSUE #549: `parse_env` should be something else?
It is [here](https://github.com/JuliaLang/Pkg.jl/blob/e1627a6a5dd7d93aafae545c58d9b584a618ade7/src/REPLMode.jl#L84) now.
It was [here](https://github.com/JuliaLang/Pkg.jl/blob/0621d8651e4753922ca0ba5fb0785adc15574597/src/REPLMode.jl#L321) before #509. But I can't find any reference to it?

### LINKED ISSUE #590: Remove `--env` stuff
I think we have settled to use `activate`.

00vareladavid:
Should close #590 and close #549

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -79,9 +79,8 @@ function parse_option(word::AbstractString)::Option
     return Option(option_name, option_arg)
 end
 
-meta_option_declarations = OptionDeclaration[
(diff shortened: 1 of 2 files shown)


KristofferC:
Since this is completely broken on 1.0 it should be fine to backport.

KristofferC:
bors try

PR #618, opened 2018-08-11
better info messages for activating new env

fredrikekre:


Evizero:
much better

src/API.jl
```diff
@@ -611,7 +611,7 @@ function activate(path::String; shared::Bool=false)
         # unless the shared environment already exists, place it in the first depots
         if !isdir(fullpath)
             fullpath = joinpath(Pkg.envdir(Pkg.depots1()), path)
-            @info("new shared environment \"$path\" will be placed at $fullpath")
+            @info("activating new shared environment \"$path\" at $fullpath.")
```
KristofferC: Even though we use `--shared` I don't think we should call it `shared environment`. It's just a normal environment, no?

KristofferC:
Looks like CI need to be fixed.

Evizero:
Oh yea, I added tests for the log messages, which will need to be updated

PR #626, opened 2018-08-12
Update from Base and recommit changes in Pkg.jl on top of it.

KristofferC:
The Pkg.jl repo and Base Pkg has slightly diverged. This merges the history by adding a commit that simply sets the file to the version of Base at 1.0 and then recommits the changes in the repo on top of it.


src/Types.jl
```diff
@@ -908,8 +914,19 @@ end
 function registries(; clone_default=true)::Vector{String}
     isempty(depots()) && return String[]
     user_regs = abspath(depots1(), "registries")
+    # TODO: delete the following let block in Julia 1.0
```
fredrikekre: rm?

src/resolve/FieldValues.jl
```diff
@@ -72,9 +72,9 @@ Base.copy(a::FieldValue) = FieldValue(a.l0, copy(a.l1), copy(a.l2), a.l3)
 # some hard constraint is being violated
 validmax(a::FieldValue) = a.l0 >= 0
 
-# like usual indmax, but favors the highest indices
+# like usual argmax, but favors the highest indices
 # in case of a tie
-function Base.indmax(f::Field)
+function Base.argmax(f::Field)
```
fredrikekre: Maybe no reason to extend this since it is only used internally

test/pkg.jl
```diff
@@ -345,6 +346,8 @@ temp_pkg_dir() do project_path
     end
 end
 
+# This test seems to error on AppVeyor so disable for now
```
fredrikekre: Still true?

KristofferC: Yes

src/Types.jl
KristofferC: Can removing it in a while.

Issue #636, opened 2018-08-13
`develop` branch checkout fails


pbouffard:
The help text for `develop` implies that a it should be possible to check out a branch:

```
 Example

  pkg> develop Example
  pkg> develop Example#master
  pkg> develop Example#c37b675
  pkg> develop https://github.com/JuliaLang/Example.jl#master
  pkg> develop --local Example
```
... but attempting to do so results in an error:

```
(v1.0) pkg> develop https://github.com/noahstier/PackageCompiler.jl.git#SyntaxFix
ERROR: a git revision cannot be given to `develop`
```

KristofferC:
Yes the docs are outdated.

PR #639, opened 2018-08-13
update docs for develop

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/636


src/REPLMode.jl
```diff
@@ -1040,9 +1040,7 @@ This operation is undone by `free`.
 *Example*
 ```jl
 pkg> develop Example
-pkg> develop Example#master
-pkg> develop Example#c37b675
-pkg> develop https://github.com/JuliaLang/Example.jl#master
+pkg> develop https://github.com/JuliaLang/Example.jl
```
fredrikekre: maybe add a path example?

PR #644, opened 2018-08-13
add changelog

KristofferC:


fredrikekre:
Seems to be backported to release-1.0 already. Removing the label.

Issue #606, opened 2018-08-09
Freeing a package tracking a repo doesn't collect compat from registry


omus:

omus:
cc: @KristofferC

KristofferC:
Do you remember the reproducer for this? I fail to reproduce it locally now.

omus:
Umm, from what I can remember I was freeing a developed package and I it complained that one of it's dependencies was un-resolvable. I'll re-open this if it occurs again.

PR #646, opened 2018-08-13
remove unnecessary version handling when dealing with repos

KristofferC:
This is completely the wrong place to assign versions to packages, that takes place just before calling the resolver.

Only have crappy mobile internet so we'll see what CI says, didn't run locally.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/645
Fixes https://github.com/JuliaLang/Pkg.jl/issues/606

Tests for this are quite hard to make because the code is so stupid.


KristofferC:
bors try

KristofferC:
Well, it is a little bit more complicated than this. Will fix tomorrow.

KristofferC:
bors try

KristofferC:
bors try

KristofferC:
bors try

KristofferC:
bors try

PR #657, opened 2018-08-15
Fix a bug where the path was stored as relative

fredrikekre:
Fix a bug where the path was stored as relative because `devdir()` happened to be in the project. The path should only be relative if the user used `--local`.


fredrikekre:
bors try

PR #661, opened 2018-08-15
`help` should not error when no docs available

00vareladavid:
Use case is `pkg> help add remove up` when e.g. `remove` does not have docs. Currently the whole command fails. I think it would be better if you just get a warning and get help for the commands which docs are available.


00vareladavid:
bors try

fredrikekre:
I guess a better solution is to add docs? :smile:

00vareladavid:
Yes, perhaps we can do
```
if spec.help === nothing
    @shame_devs("Really!? No docs?")
    continue
end
```

00vareladavid:
Actually, I think the suggestion in #663 would be better. Forcing the spec to include help would obviate this PR.

KristofferC:
Closing in favour of #663

Issue #663, opened 2018-08-15
Each REPL command should store a short form help.


KristofferC:
We would use these to build up the full help for `?`. That would make it obvious what commands we have failed to document, or in fact, would make it impossible to fail to short form document a command if we require that it exists.

00vareladavid:
I was thinking along the same lines @KristofferC . Breaking up the [help doc](https://github.com/JuliaLang/Pkg.jl/blob/ec413a28fca07c64c466523b737ba979ccc87269/src/REPLMode.jl#L1306) and shoving the pieces in a  `description` field in `CommandSpec`.

Or did you mean something else?

pfitzseb:
Only slightly related, but I would *love* to have a nice autocompletion frontend like this:
![image](https://user-images.githubusercontent.com/6735977/44206246-7ca2b780-a159-11e8-9eea-f057785e4d7b.png)

Interactive autocompletions should go into Base though (and I hope I'll actually find the time to work on that).

00vareladavid:
Fixed?

KristofferC:
Indeed

PR #671, opened 2018-08-18
Add a `description` field to `CommandSpec`

00vareladavid:
For #663

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -109,6 +109,7 @@ const CommandDeclaration = Tuple{CommandKind,
                                        Vector{Pair{Symbol, Any}}, # parser keys
                                        }, # arguments
                                  Vector{OptionDeclaration}, # options
+                                 String, #description
                                  Union{Nothing, Markdown.MD}, #help
                                  }
 struct CommandSpec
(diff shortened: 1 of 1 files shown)


00vareladavid:
bors try

src/REPLMode.jl
```diff
@@ -1302,7 +1323,7 @@ end
 # TODO remove this
 command_specs = super_specs["package"]
 
-const help = md"""
+help = md"""
```
KristofferC: Why remove `const`? `const`is about the constness of the binding but the object can still be mutated.

src/REPLMode.jl
```diff
@@ -971,6 +974,7 @@ command_declarations = [
     do_registry_add!,
     (1=>Inf, identity, []),
     [],
+    "Currently just a placeholder for a future command",
```
KristofferC: Will this be shown in the help?

00vareladavid: @KristofferC Sure!
@fredrikekre :+1:

src/REPLMode.jl
00vareladavid: Thanks! I didn't know that

00vareladavid:
The changed test is because `registry` is disabled

PR #680, opened 2018-08-22
work around strange "Unreachable reached" when the close function for a repo fires when updating registries

KristofferC:
No idea why this helps but at least it doesn't crash anymore...


KristofferC:
bors try

PR #682, opened 2018-08-22
User feedback: disambiguate between path and name

### LINKED ISSUE #597: Adding / devving from path that could be a package can sometimes be confusing
If you have an `Example` folder and do `add Example` it might be worth infoing or something that we are going to be add based on the folder `Example` that we found.

00vareladavid:
Should resolve #597


src/REPLMode.jl
```diff
@@ -170,9 +170,10 @@ end
 
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
-function parse_package(word::AbstractString; add_or_develop=false)::PackageSpec
+function parse_package(word::AbstractString; add_or_develop::Bool=false)::PackageSpec
     word = replace(word, "~" => homedir())
     if add_or_develop && casesensitive_isdir(word)
+        @info "Resolving package specifier `$word` as a directory at `$(abspath(word))`."
```
fredrikekre: `Resolving` -> `resolving` and `abspath(word)` -> `Base.contractuser(abspath(word))`?

00vareladavid: good call :+1:

src/REPLMode.jl
```diff
@@ -170,9 +170,10 @@ end
 
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
-function parse_package(word::AbstractString; add_or_develop=false)::PackageSpec
+function parse_package(word::AbstractString; add_or_develop::Bool=false)::PackageSpec
     word = replace(word, "~" => homedir())
     if add_or_develop && casesensitive_isdir(word)
+        @info "resolving package specifier `$word` as a directory at `$(Base.contractuser(abspath(word)))`."
```
KristofferC: I only think we need to do this in case `occursin(uuid_re, word)`. If you are doing `add /foo/bar` you don't need info about it being a path?

00vareladavid: Ah, good point.

But I assume the case where users type in a UUID unlikely, even more unlikely is that they have a directory with that UUID in their `cwd`.

What about obviating the issue and requiring users to `pkg> add --path Example`?

fredrikekre: I guess Kristoffer meant `uuid_name` (or whatever the regex is called)

> `pkg> add --path Example`

No, this would be annoying. We have `Pkg.add(PackageSpec(name = ))` and `Pkg.add(PackageSpec(path = ))` in these cases. And it will basically never happen that you have a folder with the same name as a package you want to add.

00vareladavid: > And it will basically never happen that you have a folder with the same name as a package you want to add.
That's the whole point of #597 

I agree that these are edge cases. But I also think that automatic behavior can be confusing to users who are not familiar with the interface.

I think you make a good point that `--path` will be annoying. What about only warning when we resolve to a directory and the original string does not contain a path separator?

KristofferC: Yeah I meant the path matches a possible package (perhaps we should only look at registered packages).

00vareladavid: > perhaps we should only look at registered packages

I'm not sure what you mean

KristofferC: I mean if you write `add Example` and you happen to have a folder called `Example` in your pwd you will add that package by path, but maybe you just forgot you had that folder.

src/REPLMode.jl
```diff
@@ -170,9 +170,12 @@ end
 
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
-function parse_package(word::AbstractString; add_or_develop=false)::PackageSpec
+function parse_package(word::AbstractString; add_or_develop::Bool=false)::PackageSpec
     word = replace(word, "~" => homedir())
     if add_or_develop && casesensitive_isdir(word)
+        if (Base.Sys.iswindows() && !occursin("\\", word)) || !occursin("/", word)
```
fredrikekre: Maybe 
```julia
if !occursin(Base.Filesystem.path_separator_re, word)
    ...
```

00vareladavid: Thanks!

00vareladavid:
bors try

PR #683, opened 2018-08-23
Do not create a stdlib dict with every `Context`

00vareladavid:
The stdlib name->UUID mapping should be pretty stable over time, I can't think of a good reason to create one on every operation.

Also created a dict where the names are the keys so the resolver should be a little faster. If this PR gets turned down, I think we should at least keep that since I suspect most people type in the name of the library and not the UUID, or maybe thats just me... :laughing: .

Also, more tests. I don't think the UUIDs of the stdlibs should change over time, right?


00vareladavid:
bors try

00vareladavid:
Crap, harder than I thought, as always... I'll try again later

PR #685, opened 2018-08-23
cap julia version for probably-0.7-incompatible packages

StefanKarpinski:
This is a heuristic so definitely imperfect, but follows this reasoning about each versions julia compatibility version set:

* has interval only containing 0.7+ versions => 0.7 compatible
* supports an julia ≤ 0.6 & doesn't use Compat => 0.7 incompatible
* claims to support 0.7+ & tagged on July 14 or later => 0.7 compatible
* passes PkgEval => 0.7 compatible
* otherwise => 0.7 incompatible


KristofferC:
Is it possible to rebase this on top of https://github.com/JuliaLang/Pkg.jl/pull/554 (which has some other fixes, for example the 1.1 julia dummy release which is needed) or merge them somehow.

The heruistic seems reasonable to me. Would be interesting to see the diff in the registry this produces.

StefanKarpinski:
Sure, I’ll rebase and can post the registry diff as a gist—it’s quite large. I just ran the update script against an up-to-date copy of the metadata repo.

StefanKarpinski:
Rebasing master of Pkg onto that branch is causing a ton of conflicts that are totally unrelated to this PR. I guess I should just cherry-pick this change on top of that branch?

KristofferC:
Yeah, the branch is very outdated so easiest to cherry-pick the commit from that branch here (which is an updated branch). Either your commit first or the one in that PR, whichever is easiest.

KristofferC:
Let me try then

KristofferC:
Aha, you pushed an update to the shamaps to master. Ok, we can just ignore the maps in that PR then.

martinholters:
Maybe also
* has explicit upper bound => just trust the given bound

Or is that implicitly done right now? (I didn't check the code.)

StefanKarpinski:
Updated with @martinholters's suggestion. Here's a gist of the registry diff:

https://gist.github.com/StefanKarpinski/b9df8cc7eb9eca8f02a73e6a9d08a888

StefanKarpinski:
I also cherry-picked @KristofferC's update commit into there before the substantive changes here.

KristofferC:
I'll deploy this, run it manually and open the changes as a PR so we can look at it with GitHub diff interface before committing to it.

KristofferC:
This caps https://github.com/JuliaGraphics/Measures.jl which passes tests on 1.0 and is a dependency of Plots. Hmm.

KristofferC:
Dynamically loading it doesn't have an advantage because any package that starts to pass tests will have a new tag and thus won't be capped.

bin/loadmeta.jl
```diff
@@ -174,6 +174,7 @@ const date_cutoff = 1531526400 # July 14, 2018
 const all_vers = julia_versions()
 const old_vers = julia_versions(v -> v < v"0.7")
 const meta_dir = Pkg.Pkg2.dir("METADATA")
+const passing = readlines("bin/passing.txt")
```
KristofferC: `joinpath(@__DIR__, "passing.txt")`.

KristofferC:
Arguably, this should be capped to 0.7 upper bound instead of 0.6 since 0.7 should have deprecations in place?

KristofferC:
Things like WriteVTK now get capped which does pass test because, going along the dependency chain, TranscodingStreams fails tests which is a dependency, so WriteVTK gets skipped.

Also, this seems to cap PyPlot. That doesn't seem right?

StefanKarpinski:
So you think anything that’s not failing should not be capped? It seems to me that only packages that are actually passing should be exempt from capping logic. If a package is skipped we don’t know if it should be capped or not the heuristic should apply.

KristofferC:
Yeah, I'm just saying it is unfortunate that we cap working packages. Of course, it is a heuristic so that might happen, but we should try to keep that to a minimum.

StefanKarpinski:
There’s going to be some initial issues—they can be fixed by putting explicit correct caps on the packages.

KristofferC:
Could we run a pkgeval that just test all packages without doing the while "skipped" part. Many packages fail because their test code does something quirky but work fine with normal usage. They then block a bunch of working palette from getting tested.

StefanKarpinski:
@Keno could you do a PkgEval run like this so that we can get things capped?

Keno:
sure

StefanKarpinski:
@Keno, did that run ever finish? @KristofferC, are you back from your move yet? Can we get this done before I leave on vacation tomorrow?

KristofferC:
It depends whether we have think the heuristic is good enough, having the result of a full PkgEval would definitely help.

I do think we need to be a bit cautious, making working packages uninstallable and breaking people's projects (no longer resolves) is not nice. I wonder if there is a need for some more publicity with a discourse post that states the intention, what packages will be ko linger installable etc.

StefanKarpinski:
I don't know how much better we can make this heuristic. The only way this situation is going to get improved is by putting caps in place and then having people fix the upper bounds in METADATA. It's annoying and will be painful for a week or so, but after that we will have accurate upper bounds.

KristofferC:
It seems like making a post outlining the purpose and the effects before doing it might be a more "safe" route.

Keno:
The PkgEval run is here, though it looks like it didn't quite do all of them: https://gist.github.com/Keno/a705ced1e9ef70772140842e652e7a32. Will see if I messed something up in the config.

Keno:
I've updated the gist with a corrected PkgEval run.

StefanKarpinski:
Ok, I don’t have time to do anything about this before my vacation but I leave it in @Keno and @KristofferC’s capable hands :)

StefanKarpinski:
So, I guess this isn’t happening?

KristofferC:
I've been moving and been on vacation until 4 hours ago so not sure how anything would have magically happened here.

KristofferC:
Cap list: https://gist.github.com/KristofferC/6ef88542178e143fa9dc840fc1e13749
New PR: https://github.com/JuliaRegistries/General/pull/16

KristofferC:
BinaryBuilder likely shouldn't be capped.
Graphics likely shouldn't be capped.

Also, some of these pass on 0.7 and fail on 1.0. It doesn't make much sense to me to test them on 1.0 and cap them to 0.6. Shouldn't we run PkgEval at 0.7 instead?

KristofferC:
Perhaps not cap a package if it is a dependency to a package that passes tests?

StefanKarpinski:
So what's the path forward here? Can I get a list of packages I should whitelist?

KristofferC:
The path forward here is to have a heuristic that doesn't recursively make most package uninstallable (like capping BinaryProvider would do).

I am guessing BinaryProvider gets capped because it has `julia 0.7-rc3` in its REQUIRE and the `v < v"0.7"` check becomes `true`?

Then we need to generate a list like https://gist.github.com/KristofferC/6ef88542178e143fa9dc840fc1e13749 except also include recursively what packages are no longer installable and see that that list is not completely crazy.

StefanKarpinski:
@Keno, can you help us get such a list. Given a list, I can put it in here and we can finally cap.

KristofferC:
Hm, I think you missunderstood what list I meant (since you are asking Keno for help).

The list is just meant as a check that our cappening isn't effectively capping the whole ecosystem. The current heuristic does so by capping BinaryBuilder due to the likely "bug" that I mentioned in the previous post with the RC in REQUIRE.

So what we need to go ahead with this is a list of packages that are no longer installable by using `Pkg.add` (either because they are getting capped or a dependency of them is getting capped), manually look at that list and see that it is ok.

StefanKarpinski:
If I have a list of the packages that we believe to already be work, then I whitelist them and make sure their latest version is considered compatible. That's already how this works: there's an explicit whitelist that I was given in JSON format. There's no more tweaking of this heuristic that can really be done.

martinholters:
But are there any packages that do not get capped themselves, but become uninstallable because any of their dependencies get capped? If there are many, that might indicate a serious problem. If there are few, that might warrant a little more manual whitelisting.

KristofferC:
> But are there any packages that do not get capped themselves, but become uninstallable because any of their dependencies get capped? If there are many, that might indicate a serious problem. 

Yes, we need to have such a list to evaluate the full impact this would hav.

StefanKarpinski:
I just spent several hours trying to understand why BinaryProvider gets capped. But it does not actually get capped. Old versions of BinaryProvider get capped, but newer ones don't.

StefanKarpinski:
Now the dependencies of passing packages are also marked as passing, which affected a whopping grand total of three packages.

PR #696, opened 2018-08-23
Cache stdlib dict

00vareladavid:
I noticed we where reading through the stdlib directory when creating each `Context` object. Unless I'm missing something, this is unnecessary (and expensive!). Caching the dict makes creating a `Context` object roughly an order of magnitude faster.

I also hope we can completely remove the `stdlibs` field from `Context`. For now, I just left as is, and run a `Base.deepcopy` on the cached dict.

I did make two steps towards removing the `stdlibs` field: rewrite `stdlib_resolve` to not depend on a `Context` object, and replace `keys(ctx.stdlibs)` with `stdlib_uuids`.

Finally, tests for the `stdlib_resolve!` rewrite. They should be useful even if there is disagreement over the PR.

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -773,15 +773,6 @@ function complete_remote_package(partial)
     return cmp
 end
 
-const STDLIB_NAMES = Ref{Vector{String}}()
-function stdlib_names()
-    if !isassigned(STDLIB_NAMES)
-        STDLIB_NAMES[] = filter!(x->isdir(joinpath(Types.stdlib_dir(), x)),
-                                 readdir(Types.stdlib_dir()))
-    end
-    return STDLIB_NAMES[]
-end
-
 function canonical_names()
     names = String[]
     for (super, specs) in pairs(super_specs)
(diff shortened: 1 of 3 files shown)


00vareladavid:
bors try

StefanKarpinski:
From Pkg Triage: the build-time stdlib list can be wrong relative to the run-time stdlib list, so @KristofferC suggests that it would be better to cache this on first read and then reuse the cached copy on subsequent calls.

fredrikekre:
Something like https://github.com/JuliaLang/Pkg.jl/pull/844

00vareladavid:
Thanks @fredrikekre! I remembered seeing a PR like this but I couldn't find it.

00vareladavid:
@StefanKarpinski #844 is for the REPL  completions, which just needs the names of standard libraries. This PR is for more internal stuff, which requires a `Dict{UUID,String}`. If anything, this PR should replace most of the code in #844.

StefanKarpinski:
Cool, carry on!

00vareladavid:
When rebasing, I felt that I was a little heavy handed with the original PR. I think its better if we cache only the most expensive part: reaching out to the filesystem and parsing the project files.

I also removed some REPL code so that we only have one source of truth for stdlib info.

StefanKarpinski:
Bors r+

00vareladavid:
bors try

PR #706, opened 2018-09-02
Rewrite `JuliaProject.toml` test to avoid AppVeyor error

### LINKED ISSUE #705: JuliaProject.toml still valid?
The [manual](https://julialang.github.io/Pkg.jl/latest/#Glossary-1) mentions it as an alternative for `Project.toml`, but my impression from reading the code is that the latter is hardcoded pretty much everywhere, and the [tests for it](https://github.com/JuliaLang/Pkg.jl/blob/d123b48e63b5a5e09cb224c6385a3f512627e4ac/test/pkg.jl#L385) are commented out.

00vareladavid:
Basically it was the `mv` in these tests that threw the "unlink" error. Instead of renaming the files this way, I copy them to a new directory and rename them during the copy. I still have no idea why the `mv` throws an error in the first place...

Fixes #705

### DIFF
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -354,8 +354,6 @@ temp_pkg_dir() do project_path
     end
 end
 
-# This test seems to error on AppVeyor so disable for now
-#=
 temp_pkg_dir() do project_path
     @testset "valid project file names" begin
         extract_uuid(toml_path) = begin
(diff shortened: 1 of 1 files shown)


00vareladavid:
bors try

00vareladavid:
bors try

KristofferC:
Great thanks!

bors delegate+

PR #712, opened 2018-09-04
Expand ~ in path for activate develop and add

fredrikekre:
Expand `~` in `path` for `activate path`, `Pkg.activate(path)`, `add path`, `develop path` and `PackageSpec(path = path)`.

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -171,9 +171,8 @@ end
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
 function parse_package(word::AbstractString; add_or_develop=false)::PackageSpec
-    word = replace(word, "~" => homedir())
-    if add_or_develop && casesensitive_isdir(word)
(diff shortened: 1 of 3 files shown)


fredrikekre:
> I'm not sure this should be done for anything else than the REPL commands.

Agreed. Thats what I had first, and went back to that now.

bors try

src/REPLMode.jl
```diff
@@ -171,9 +171,8 @@ end
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
 function parse_package(word::AbstractString; add_or_develop=false)::PackageSpec
-    word = replace(word, "~" => homedir())
```
fredrikekre: I wonder if we should keep this instead of using `expanduser`, to make this work on Windows too...

00vareladavid: I think `expanduser` is the right choice here. It seems that tilde has a distinct, special meaning in Windows (temporary files). 

I'm all for replacing `replace` :grin:

PR #720, opened 2018-09-05
Allow trailing backslash in regex

### LINKED ISSUE #719: adding local package doesn't work if it has a trailing backslash
Easy enough to remove the backslash, but it may not be obvious to others this is needed.
```julia
(v0.7) pkg> add c:\MyLocalModule\
   Cloning git-repo `c:\MyLocalModule\`
  Updating git-repo `c:\MyLocalModule\`
ERROR: cannot determine package name from URL or path: c:\MyLocalModule\

(v0.7) pkg> add c:\MyLocalModule
   Cloning git-repo `c:\MyLocalModule`
  Updating git-repo `c:\MyLocalModule`
[ Info: Assigning UUID b2b55072-4cfa-5819-82e2-19cc2e95909c to MyLocalModule
```

If I'm correct there is no match here:
https://github.com/JuliaLang/Pkg.jl/blob/7254e17c783c479b30aed933e29d5efb7d6c4762/src/Types.jl#L713
So this regex would need to support a trailing backslash:
https://github.com/JuliaLang/Pkg.jl/blob/7254e17c783c479b30aed933e29d5efb7d6c4762/src/Types.jl#L485

visr:
Also escapes the forward slash at the beginning, although this appears to not have been an issue.
Fixes #719

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -482,7 +482,7 @@ end
 
 
 const refspecs = ["+refs/*:refs/remotes/cache/*"]
-const reg_pkg = r"(?:^|[/\\])(\w+?)(?:\.jl)?(?:\.git)?(?:\/)?$"
+const reg_pkg = r"(?:^|[\/\\])(\w+?)(?:\.jl)?(?:\.git)?(?:[\/\\])?$"
 
 # Windows sometimes throw on `isdir`...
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Why do we want to allow the trailing backslash? Is this for paths on Windows?

PR #724, opened 2018-09-06
Don't mutate argument passed to Pkg.add

tkf:
Currently, `Pkg.add` modifies the argument `pkg::Vector{PackageSpec}` passed to it in-place even though the name does not end with `!`:

```julia
julia> VERSION
v"1.0.0"

julia> packages = [PackageSpec("JSON")]
1-element Array{Pkg.Types.PackageSpec,1}:
 PackageSpec(name=JSON)

julia> Pkg.add(packages)
 Resolving package versions...
# ...

julia> packages
158-element Array{Pkg.Types.PackageSpec,1}:
 PackageSpec(name=JSON, uuid=682c06a0-de6a-54ab-a142-c8b1cf79cde6, v=v"0.19.0")
 PackageSpec(name=Documenter, uuid=e30172f5-a6a5-5a46-863b-614d45cd2de4, v=v"0.19.6")
# and so on ...
```

I propose to `deepcopy` it before passing it to the destructive internal function.  Note that this PR adds a function `Pkg.API.add_or_develop!`.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -29,6 +29,7 @@ add_or_develop(pkgs::Vector{String}; kwargs...)            = add_or_develop([che
 add_or_develop(pkgs::Vector{PackageSpec}; kwargs...)       = add_or_develop(Context(), pkgs; kwargs...)
 
 function add_or_develop(ctx::Context, pkgs::Vector{PackageSpec}; mode::Symbol, shared::Bool=true, kwargs...)
+    pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     Context!(ctx; kwargs...)
 
     # All developed packages should go through handle_repos_develop so just give them an empty repo
@@ -73,6 +74,7 @@ rm(pkgs::Vector{String}; kwargs...)            = rm([PackageSpec(pkg) for pkg in
 rm(pkgs::Vector{PackageSpec}; kwargs...)       = rm(Context(), pkgs; kwargs...)
 
 function rm(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT, kwargs...)
+    pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     for pkg in pkgs
         # TODO only overwrite pkg.mode if default value ?
         pkg.mode = mode
@@ -166,6 +168,7 @@ up(pkgs::Vector{PackageSpec}; kwargs...)       = up(Context(), pkgs; kwargs...)
 
 function up(ctx::Context, pkgs::Vector{PackageSpec};
             level::UpgradeLevel=UPLEVEL_MAJOR, mode::PackageMode=PKGMODE_PROJECT, do_update_registry=true, kwargs...)
+    pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     for pkg in pkgs
         # TODO only override if they are not already set
         pkg.mode = mode
@@ -205,6 +208,7 @@ pin(pkgs::Vector{String}; kwargs...)            = pin([PackageSpec(pkg) for pkg 
 pin(pkgs::Vector{PackageSpec}; kwargs...)       = pin(Context(), pkgs; kwargs...)
 
 function pin(ctx::Context, pkgs::Vector{PackageSpec}; kwargs...)
+    pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     Context!(ctx; kwargs...)
     ctx.preview && preview_info()
     project_deps_resolve!(ctx.env, pkgs)
@@ -219,6 +223,7 @@ free(pkgs::Vector{String}; kwargs...)            = free([PackageSpec(pkg) for pk
 free(pkgs::Vector{PackageSpec}; kwargs...)       = free(Context(), pkgs; kwargs...)
 
 function free(ctx::Context, pkgs::Vector{PackageSpec}; kwargs...)
+    pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     Context!(ctx; kwargs...)
     ctx.preview && preview_info()
     registry_resolve!(ctx.env, pkgs)
@@ -250,6 +255,7 @@ test(pkgs::Vector{String}; kwargs...)             = test([PackageSpec(pkg) for p
 test(pkgs::Vector{PackageSpec}; kwargs...)        = test(Context(), pkgs; kwargs...)
 
 function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, kwargs...)
+    pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     Context!(ctx; kwargs...)
     ctx.preview && preview_info()
     if isempty(pkgs)
@@ -423,6 +429,7 @@ build(pkg::Array{Union{}, 1}) = build(PackageSpec[])
 build(pkg::PackageSpec) = build([pkg])
 build(pkgs::Vector{PackageSpec}) = build(Context(), pkgs)
 function build(ctx::Context, pkgs::Vector{PackageSpec}; kwargs...)
+    pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     Context!(ctx; kwargs...)
 
     ctx.preview && preview_info()
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -505,6 +505,16 @@ end
     @test Pkg.Types.pathrepr(path) == "`@stdlib/Test`"
 end
 
+
+temp_pkg_dir() do project_path
+    @testset "Pkg.add should not mutate" begin
+        package_names = ["JSON"]
+        packages = PackageSpec.(package_names)
+        Pkg.add(packages)
+        @test [p.name for p in packages] == package_names
+    end
+end
+
 include("repl.jl")
 include("api.jl")
 


KristofferC:
Good catch,

I think this might be true for almost all functions though. Perhaps just adding a `deepcopy` as the first line to the API functions is an easier way to do this?

tkf:
Are you think about something like this?

```diff
diff --git a/src/API.jl b/src/API.jl
index b95f4d3..54a7fb0 100644
--- a/src/API.jl
+++ b/src/API.jl
@@ -29,6 +29,7 @@ add_or_develop(pkgs::Vector{String}; kwargs...)            = add_or_develop([che
 add_or_develop(pkgs::Vector{PackageSpec}; kwargs...)       = add_or_develop(Context(), pkgs; kwargs...)

 function add_or_develop(ctx::Context, pkgs::Vector{PackageSpec}; mode::Symbol, shared::Bool=true, kwargs...)
+    pkgs = deepcopy(pkgs)
     Context!(ctx; kwargs...)

     # All developed packages should go through handle_repos_develop so just give them an empty repo
```

I thought about it, but it felt weird since `ctx` is mutated in-place while `pkgs` is not.

tkf:
Though I guess I can treat `add_or_develop(ctx, ...)` as something like `rand(rng, ...)`.

src/API.jl
```diff
@@ -26,9 +26,9 @@ end
 
 add_or_develop(pkg::Union{String, PackageSpec}; kwargs...) = add_or_develop([pkg]; kwargs...)
 add_or_develop(pkgs::Vector{String}; kwargs...)            = add_or_develop([check_package_name(pkg) for pkg in pkgs]; kwargs...)
-add_or_develop(pkgs::Vector{PackageSpec}; kwargs...)       = add_or_develop(Context(), pkgs; kwargs...)
+add_or_develop(pkgs::Vector{PackageSpec}; kwargs...)       = add_or_develop!(Context(), deepcopy(pkgs); kwargs...)
```
StefanKarpinski: I believe this only needs to `copy` not `deepcopy`.

fredrikekre:
Can I suggest the followin alterative patch:
```diff
diff --git a/src/Types.jl b/src/Types.jl
index dec502d..c2e3cfa 100644
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -153,6 +153,8 @@ mutable struct PackageSpec
     special_action::PackageSpecialAction # If the package is currently being pinned, freed etc
     repo::Union{Nothing,GitRepo}
 end
+PackageSpec(pkg::PackageSpec) =
+    PackageSpec(pkg.name, pkg.uuid, pkg.version, pkg.mode, pkg.path, pkg.special_action, pkg.repo)
 PackageSpec(name::AbstractString, uuid::UUID, version::VersionTypes,
             mode::PackageMode=PKGMODE_PROJECT, path=nothing, special_action=PKGSPEC_NOTHING,
             repo=nothing) = PackageSpec(String(name), uuid, version, mode, path, special_action, repo)
fredrik@fredrik-laptop:~/dev/Pkg$ git diff
diff --git a/src/API.jl b/src/API.jl
index b95f4d3..6212b00 100644
--- a/src/API.jl
+++ b/src/API.jl
@@ -25,8 +25,8 @@ function check_package_name(x::String)
 end
 
 add_or_develop(pkg::Union{String, PackageSpec}; kwargs...) = add_or_develop([pkg]; kwargs...)
-add_or_develop(pkgs::Vector{String}; kwargs...)            = add_or_develop([check_package_name(pkg) for pkg in pkgs]; kwargs...)
-add_or_develop(pkgs::Vector{PackageSpec}; kwargs...)       = add_or_develop(Context(), pkgs; kwargs...)
+add_or_develop(pkgs::Vector{<:Union{String, PackageSpec}}; kwargs...) =
+    add_or_develop(Context(), PackageSpec.(pkgs); kwargs...)
 
 function add_or_develop(ctx::Context, pkgs::Vector{PackageSpec}; mode::Symbol, shared::Bool=true, kwargs...)
     Context!(ctx; kwargs...)
@@ -69,8 +69,8 @@ add(args...; kwargs...) = add_or_develop(args...; mode = :add, kwargs...)
 develop(args...; shared=true, kwargs...) = add_or_develop(args...; mode = :develop, shared = shared, kwargs...)
 
 rm(pkg::Union{String, PackageSpec}; kwargs...) = rm([pkg]; kwargs...)
-rm(pkgs::Vector{String}; kwargs...)            = rm([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
-rm(pkgs::Vector{PackageSpec}; kwargs...)       = rm(Context(), pkgs; kwargs...)
+rm(pkgs::pkgs::Vector{<:Union{String, PackageSpec}}; kwargs...) =
+    rm(Context(), PackageSpec.(pkgs); kwargs...)
 
 function rm(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT, kwargs...)
     for pkg in pkgs
@@ -161,8 +161,8 @@ end
 up(ctx::Context; kwargs...)                    = up(ctx, PackageSpec[]; kwargs...)
 up(; kwargs...)                                = up(PackageSpec[]; kwargs...)
 up(pkg::Union{String, PackageSpec}; kwargs...) = up([pkg]; kwargs...)
-up(pkgs::Vector{String}; kwargs...)            = up([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
-up(pkgs::Vector{PackageSpec}; kwargs...)       = up(Context(), pkgs; kwargs...)
+up(pkgs::Vector{<:Union{String, PackageSpec}}; kwargs...) =
+    up(Context(), PackageSpec.(pkgs); kwargs...)
 
 function up(ctx::Context, pkgs::Vector{PackageSpec};
             level::UpgradeLevel=UPLEVEL_MAJOR, mode::PackageMode=PKGMODE_PROJECT, do_update_registry=true, kwargs...)
@@ -201,8 +201,8 @@ resolve(ctx::Context=Context()) =
     up(ctx, level=UPLEVEL_FIXED, mode=PKGMODE_MANIFEST, do_update_registry=false)
 
 pin(pkg::Union{String, PackageSpec}; kwargs...) = pin([pkg]; kwargs...)
-pin(pkgs::Vector{String}; kwargs...)            = pin([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
-pin(pkgs::Vector{PackageSpec}; kwargs...)       = pin(Context(), pkgs; kwargs...)
+pin(pkgs::Vector{<:Union{String, PackageSpec}}; kwargs...) =
+    pin(Context(), PackageSpec.(pkgs); kwargs...)
 
 function pin(ctx::Context, pkgs::Vector{PackageSpec}; kwargs...)
     Context!(ctx; kwargs...)
@@ -215,8 +215,8 @@ end
 
 
 free(pkg::Union{String, PackageSpec}; kwargs...) = free([pkg]; kwargs...)
-free(pkgs::Vector{String}; kwargs...)            = free([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
-free(pkgs::Vector{PackageSpec}; kwargs...)       = free(Context(), pkgs; kwargs...)
+free(pkgs::Vector{<:Union{String, PackageSpec}}; kwargs...) =
+    free(Context(), PackageSpec.(pkgs); kwargs...)
 
 function free(ctx::Context, pkgs::Vector{PackageSpec}; kwargs...)
     Context!(ctx; kwargs...)
@@ -246,8 +246,8 @@ end
 
 test(;kwargs...)                                  = test(PackageSpec[]; kwargs...)
 test(pkg::Union{String, PackageSpec}; kwargs...)  = test([pkg]; kwargs...)
-test(pkgs::Vector{String}; kwargs...)             = test([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
-test(pkgs::Vector{PackageSpec}; kwargs...)        = test(Context(), pkgs; kwargs...)
+test(pkgs::Vector{<:Union{String, PackageSpec}}; kwargs...) =
+    test(Context(), PackageSpec.(pkgs); kwargs...)
 
 function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, kwargs...)
     Context!(ctx; kwargs...)
diff --git a/src/Types.jl b/src/Types.jl
index dec502d..c2e3cfa 100644
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -153,6 +153,8 @@ mutable struct PackageSpec
     special_action::PackageSpecialAction # If the package is currently being pinned, freed etc
     repo::Union{Nothing,GitRepo}
 end
+PackageSpec(pkg::PackageSpec) =
+    PackageSpec(pkg.name, pkg.uuid, pkg.version, pkg.mode, pkg.path, pkg.special_action, pkg.repo)
 PackageSpec(name::AbstractString, uuid::UUID, version::VersionTypes,
             mode::PackageMode=PKGMODE_PROJECT, path=nothing, special_action=PKGSPEC_NOTHING,
             repo=nothing) = PackageSpec(String(name), uuid, version, mode, path, special_action, repo)
```

tkf:
@fredrikekre I guess

```julia
PackageSpec(pkg::PackageSpec) =
    PackageSpec(pkg.name, pkg.uuid, pkg.version, pkg.mode, pkg.path, pkg.special_action, pkg.repo)
```

can simply be

```julia
PackageSpec(pkg::PackageSpec) = copy(pkg)
```

?  I just prefer avoiding a long list of positional arguments (because, e.g., it's easier to refactor `PackageSpec` struct).

However, I still think we need a `deepcopy` like this

```julia
PackageSpec(pkg::PackageSpec) = deepcopy(pkg)
```

because when it is created by `PackageSpec(url="...")` then `.repo` could be mutated in-place.

Also, do we want rename all the mutating functions (`add_or_develop` -> `add_or_develop!`, `rm` -> `rm!`, ...)?  For example, the right hand side `rm` of

```julia
rm(pkgs::pkgs::Vector{<:Union{String, PackageSpec}}; kwargs...) =
    rm(Context(), PackageSpec.(pkgs); kwargs...)
```

would be `rm!`.

Other than that, I like the idea of merging two method definitions.

KristofferC:
Instead of renaming functions, can we please just do the simplest and throw a `deepcopy` on all `pkgs` the first thing we do in the `add_or_develop`, `update` etc.

tkf:
Well, this PR is all about conforming to the expectation of non-bang function. So, I thought making them to not mutate `ctx` as well is also in the scope of this PR. You may say the call signature `add(ctx, pkgs)` is not documented so it's OK to mutate `ctx` but I think hiding this signature gives more reason to add mutating internal functions such as `add_or_develop!` and `rm!` so that `add(ctx, pkgs)` becomes impossible to invoke.

But then you may say those points don't matter in practice and... I guess I agree with that :)

Anyway, I just insert `deepcopy` everywhere and rebased the patch. Let me know if there is more to change.

fredrikekre:
bors try

PR #726, opened 2018-09-07
Implement option to yank/blacklist a release.

### LINKED ISSUE #355: Provide a mechanism to mark a release as broken
`cargo` has the option to "yank" a  release (https://doc.rust-lang.org/cargo/reference/publishing.html#cargo-yank) effectively marking it as broken. I think we need this for our package manager too. Translated to julia; yanking a release would mean that the broken version would not be considered an option for the resolver, but if you have a yanked release in a `Manifest.toml` you can still `instantiate` it to install the broken version that way. With the new version specification it is a bit harder to upper bound in case of broken releases (and if package maintainers follow SemVer you should never have to upper bound). Yanking a release would be a way to enforce SemVer on a higher level, since that would presumably be up to the registry maintainer to do. So instead of

> PackageA v1.2 is broken, so I upperbounded PackageB to PackageA v1.1

we should have 

> PackageA v1.2 is broken, so that release was "yanked"

which seems a lot better.

In practice I think this would need an entry in the registry which marks a release as yanked, and then the resolver filters out those when collecting possible versions to install.

fredrikekre:
Implement support for yanking versions in a package registry.

---

Very WIP, but this allows us to blacklist releases on the registry level. They can not be installed, but they can still be used, and you can `instantiate` a `Project.toml` + `Manifest.toml` with it, so it should not break working code. Here https://github.com/fredrikekre/Registry/blob/97919d39c95a3d917ea3b69e78aad68c7c02bf7c/I/ImportMacros/Versions.toml I have blacklisted version 0.2.0 of ImportMacros, and consequently:
```
(v1) pkg> add ImportMacros
  Updating registry at `~/.julia/registries/Registry`
  Updating git-repo `https://github.com/fredrikekre/Registry.git`
 Resolving package versions...
  [13b54a02] + ImportMacros v0.1.0
  Updating `~/.julia/environments/v1/Manifest.toml`
  [13b54a02] + ImportMacros v0.1.0
```
installs version 0.1.0. But we can `instantiate` a `Manifest.toml` with version 0.2.0 in it:
```
(v1) pkg> st
    Status `~/.julia/environments/v1/Project.toml`
→ [13b54a02] ImportMacros v0.2.0
┌ Warning: Some packages (indicated with a red arrow) are not downloaded, use `instantiate` to instantiate the current environment
└ @ Pkg.Display ~/dev/Pkg/src/Display.jl:190

(v1) pkg> instantiate
  Updating registry at `~/.julia/registries/Registry`
  Updating git-repo `https://github.com/fredrikekre/Registry.git`
 Installed ImportMacros ─ v0.2.0
```

fix #355

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -23,9 +23,12 @@ function find_installed(name::String, uuid::UUID, sha1::SHA1)
     return abspath(depots1(), "packages", name, slug_default)
 end
 
-function load_versions(path::String)
+function load_versions(path::String; include_yanked = false)
     toml = parse_toml(path, "Versions.toml")
-    return Dict{VersionNumber, SHA1}(VersionNumber(ver) => SHA1(info["git-tree-sha1"]) for (ver, info) in toml)
+    d = Dict{VersionNumber, SHA1}(
+            VersionNumber(ver) => SHA1(info["git-tree-sha1"]) for (ver, info) in toml
+                if !get(info, "yanked", false) || include_yanked)
+    return d
 end
 
 function load_package_data(f::Base.Callable, path::String, versions)
@@ -77,7 +80,7 @@ end
 function set_maximum_version_registry!(env::EnvCache, pkg::PackageSpec)
     pkgversions = Set{VersionNumber}()
     for path in registered_paths(env, pkg.uuid)
-        pathvers = keys(load_versions(path))
+        pathvers = keys(load_versions(path; include_yanked = false))
         union!(pkgversions, pathvers)
     end
     if length(pkgversions) == 0
@@ -262,7 +265,7 @@ function deps_graph(ctx::Context, uuid_to_name::Dict{UUID,String}, reqs::Require
                 end
             else
                 for path in registered_paths(ctx.env, uuid)
-                    version_info = load_versions(path)
+                    version_info = load_versions(path; include_yanked = false)
                     versions = sort!(collect(keys(version_info)))
                     deps_data = load_package_data_raw(UUID, joinpath(path, "Deps.toml"))
                     compat_data = load_package_data_raw(VersionSpec, joinpath(path, "Compat.toml"))
@@ -409,7 +412,7 @@ function version_data!(ctx::Context, pkgs::Vector{PackageSpec})
             end
             repo = info["repo"]
             repo in clones[uuid] || push!(clones[uuid], repo)
-            vers = load_versions(path)
+            vers = load_versions(path; include_yanked = true)
             if haskey(vers, ver)
                 h = vers[ver]
                 if haskey(hashes, uuid)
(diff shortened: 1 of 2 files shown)


fredrikekre:
Some questions:
Is `Versions.toml` a good place for this? Feels like it, but perhaps it is better to habr it formatted like
```toml
["0.1.0"]
git-tree-sha1 = "..."

["0.2.0"]
git-tree-sha1 = "..."
yanked = true
```

2. What should happen on `up`? Currently `ImportMacros` will be downgraded to 0.1.0.

00vareladavid:
`up`'s behavior seems reasonable to me. What would happen on `add ImportMacros@0.2.0`?

KristofferC:
Exactly the same as if the yanked version was not in the registry?

fredrikekre:
What should happen if `A` depends on `B@v0.2` and `B@v0.2` is yanked?

KristofferC:
Probably wouldn't allow yanking that release in that case.

fredrikekre:
So that should be checked on registry CI then.

fredrikekre:
Triage: In favor, but recorded in the registry as https://github.com/JuliaLang/Pkg.jl/pull/726#issuecomment-419445279 rather than https://github.com/fredrikekre/Registry/commit/97919d39c95a3d917ea3b69e78aad68c7c02bf7c

fredrikekre:
bors try

fredrikekre:
Should be good to go now, but would be good if someone could review the calls to `load_versions` and see if I got the `include_yanked = (true|false)` correctly.

fredrikekre:
bors try

src/Operations.jl
```diff
@@ -77,7 +80,7 @@ end
 function set_maximum_version_registry!(env::EnvCache, pkg::PackageSpec)
     pkgversions = Set{VersionNumber}()
     for path in registered_paths(env, pkg.uuid)
-        pathvers = keys(load_versions(path))
+        pathvers = keys(load_versions(path; include_yanked = true)) # ??
```
KristofferC: Probably `false` here.

fredrikekre: Ok, what does this do? So we can add a test

StefanKarpinski:
I wonder whether we should have this `include_yanked` option at all? Do we need it for anything?

src/Operations.jl
KristofferC: It is for backward compatibility with 0.6 stuff so not sure we need to add a test really. It just sets the version to the maximum one in the registry, only used when no Project file is present.

fredrikekre:
> Do we need it for anything?

We need to collect yanked versions when instantiating a Manifest (I think).

fredrikekre:
... but maybe we can go look directly for the tree hash instead of first collecting all versions

StefanKarpinski:
If it complicates things not having it, keeping is fine, I was just wondering about it.

fredrikekre:
bors try

fredrikekre:
> If it complicates things not having it, keeping is fine, I was just wondering about it.

Let's leave it there for now, can be revisited.

bors r+

PR #737, opened 2018-09-12
Only complete directories on `add`/`dev`

### LINKED ISSUE #736: Only complete directories on `add`/`dev`?
`pkg> add RE` will suggest "README.md". We could even take it a step further and check that the directories have a project file before suggesting

00vareladavid:
Resolves #736


00vareladavid:
bors try

src/REPLMode.jl
```diff
@@ -763,7 +763,7 @@ end
 
 function complete_local_path(s, i1, i2)
     cmp = REPL.REPLCompletions.complete_path(s, i2)
-    [REPL.REPLCompletions.completion_text(p) for p in cmp[1]], cmp[2], !isempty(cmp[1])
+    filter(isdir, [REPL.REPLCompletions.completion_text(p) for p in cmp[1]]), cmp[2], !isempty(cmp[1])
```
KristofferC: Not that it matters but can use `filter!` here. Also, perhaps this could be split into two statements, it is becoming a bit too nested imo (just add a variable name to the list and then `filter!` it).

Issue #615, opened 2018-08-11
Add a FAQ


KristofferC:
It would be good to have a FAQ. If people have some questions they feel are common they could ask them here and we could collect the answers into a FAQ.

StefanKarpinski:
Nice! Great start, I will have to spend some time filling things out here.

KristofferC:
FAQs are bad: https://thegood.com/insights/faq-pages/

PR #743, opened 2018-09-13
Add quickstart to docs

### LINKED ISSUE #689: Improvements to the Pkg docs
@mauro3 commented on [Thu May 24 2018](https://github.com/JuliaLang/julia/issues/27245)

Just reading through the new Pkg docs and I have a few suggestions:

- Title should be more descriptive, maybe as before "Packages" or "Packages (Pkg)".
- I think there should be a "Quick start" section right at the top.  Whilst the "Introduction" section is a nice long read, for someone wanting to just install a package, that is not very helpful.
- Then, whilst I understand the reason to have packages separate for each "project", for simple REPL work the old behavior of a global packages is better.  I suspect this is possible with the new Pkg, but it is not clear to me how.  I suggest to put this into the "Quick start".
- Maybe rename the "Getting started" to "Getting started with a project".

00vareladavid:
Should hopefully resolve #689

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -1,5 +1,85 @@
 # Pkg
 
+## Quickstart
+
+From the Julia REPL, press `]`, to enter the Pkg REPL.
+You should see the following prompt:
+
+```
+(v1.0) pkg>
+```
+
+To add a package, use the `add` command
+
+```
+(v1.0) pkg> add Plots
+```
+
+Multiple packages may be specified:
+
+```
+(v1.0) pkg> add Plots JuMP DataFrames
+```
+
+To remove a package, use the `rm` command.
+`rm` can also accept multiple packages:
+
+```
+(v1.0) pkg> rm JuMP DataFrames
+```
+
+So far, we have specified by names in a registry.
+If you want to add a package which is not in a registry,
+you can specify the location directly:
+
+```
+(v1.0) pkg> add https://github.com/JuliaPlots/Plots
+```
+
+To remove this package, use `rm` and specify the package by name (not URL!):
+
+```
+(v1.0) pkg> rm Plots
+```
+
+This should cover most use cases for simple package management: adding and removing dependencies.
+But say you are working on a project and you encounter a bug in one of your dependencies!
+How would you access the source? `Pkg` can help you out with `develop`:
+
+```
+(v1.0) pkg> develop --local Example
+```
+
+The `Example` package is now cloned to the `dev` subdirectory of your project directory.
+You can edit `Example`'s source and any changes you make will be visible to your project.
+
+Once upstream `Example` has been patched, you can stop tracking the local clone.
+Do this with a `free` command:
+
+```
+(v1.0) pkg> free Example
+```
+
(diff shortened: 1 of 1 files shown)


fredrikekre:
Great! I started on something similar last night (https://github.com/JuliaLang/Pkg.jl/tree/fe/doc-overhaul) but didn't get so far (https://github.com/JuliaLang/Pkg.jl/blob/fe/doc-overhaul/docs/src/getting-started.md).

00vareladavid:
Nice. I'm thinking its not a bad idea to heavily copy from [cargo's documentation](https://doc.rust-lang.org/cargo/) layout. I'm also fond of [nix's quickstart](https://nixos.org/nix/manual/#chap-quick-start).

00vareladavid:
Is it possible to "subtree" from Julia's docs? What I mean is to not shove everything into a single file. It seems your branch does that. I think it's a good idea.

fredrikekre:
Yea, thats what I am doing on that branch. But that will only deploy to https://julialang.github.io/Pkg.jl/latest/. I think that is fine, and only keep something like a quick guide with reference to the "real" docs in the Julia manual.

00vareladavid:
That sounds perfect

00vareladavid:
Do you have a specific organization in mind? Again, I think I could help with certain subbranches

00vareladavid:
Also #615 links to a good starting point for a FAQ: https://docs.google.com/document/d/1pdfao38w8fgSnEvf1JkF4Dz2EipWJK1TmaNCUXAlRVc/edit

fredrikekre:
> Do you have a specific organization in mind?

Not really, but this is a draft based on the current docs: https://github.com/JuliaLang/Pkg.jl/blob/066050812d9a1f362722c847d06bf563c100a1be/docs/make.jl#L10-L19

> I think cargo would be good to emulate. I could help with certain subbranches

Yea, I like the cargo docs.

00vareladavid:
I'm thinking roughly:

- quickstart
    - I just want access to `Plots` dammit!
- guide
    - basic commands (`add`, `rm`, `activate`, `status`, `help`)
    - recommended basic workflows (how would I do this? -> sequence of commands)
- reference (a cleaned up version of current docs)
- faq (use some things from #615 + things which come up on discourse)
    - include some less common workflows

For the guide, we could stick to using the REPL.
In the reference, it would be cool to have a mapping between REPL commands and the equivalent API expression.

fredrikekre:
Yea, something like that. Let me clean up that branch, and then we can fill in the blanks.

00vareladavid:
I have a rough [guide](https://github.com/00vareladavid/Pkg.jl/tree/00/add/docs) ready. Still needs a pass for clarity/correctness/etc but I think it should be a good starting point. The essential idea is to provide a set of increasingly complex motivating examples which we can use to describe how Pkg works.

StefanKarpinski:
It would be great to dust this off and merge it wherever is appropriate.

fredrikekre:
I cherry-picked this commit into  #803. Thanks!

Issue #745, opened 2018-09-15
Be explicit about empty environment


00vareladavid:
Running `status` on an empty environment currently looks like this:
```
(MyPackage) pkg> st
Project MyPackage v0.1.0
    Status `/tmp/MyPackage/Project.toml
```

What about:
```
(MyPackage) pkg> st
Project MyPackage v0.1.0
    Status `/tmp/MyPackage/Project.toml
    Empty environment.
```

It doesn't matter to me much what the message is. Just that we are explicit. This is mainly to help new users, but it doesn't hurt experienced users either.

StefanKarpinski:
Yes, seems like a good change to me.

00vareladavid:
Solved by #748

PR #748, opened 2018-09-16
Add a message for empty project deps

adamslc:
See #745. Example of new behavior:
```julia
julia> Pkg.status()
    Status `~/Documents/projects/Pkg/test/TestPkg/Project.toml`
 [empty environment]

(TestPkg) pkg> st
    Status `~/Documents/projects/Pkg/test/TestPkg/Project.toml`
 [empty environment]
```

I didn't see any testing of `Pkg` output, so I haven't added any tests. Point me in the right direction if you would like a test.

### DIFF
--- a/src/Display.jl
+++ b/src/Display.jl
@@ -186,6 +186,10 @@ function print_diff(io::IO, ctx::Context, diff::Vector{DiffEntry}, status=false)
         printstyled(io, " [$(string(x.uuid)[1:8])]"; color = color_dark)
         printstyled(io, "$v $(x.name) $vstr\n"; color = colors[verb])
     end
+    if isempty(diff)
+        str = ctx.env.git === nothing ? "  (empty environment)\n" : "  (no changes since last commit)\n"
+        printstyled(str, color = color_dark)
+    end
     if status && some_packages_not_downloaded
         @warn "Some packages (indicated with a red arrow) are not downloaded, use `instantiate` to instantiate the current environment"
     end


fredrikekre:
Wont this show the wrong thing for git repos?

KristofferC:
Yeah, good point.

adamslc:
What is the purpose of this block:
```julia
    if env.git != nothing
        LibGit2.with(LibGit2.GitRepo(env.git)) do repo
            git_path = LibGit2.path(repo)
            project_path = relpath(env.project_file, git_path)
            manifest_path = relpath(env.manifest_file, git_path)
            project₀ = read_project(git_file_stream(repo, "HEAD:$project_path", fakeit=true))
            manifest₀ = read_manifest(git_file_stream(repo, "HEAD:$manifest_path", fakeit=true))
        end
    end
```

Actually, I'm not really clear on what the `status` command displays w.r.t. git repos. If I create a new branch, and add a dependency, then `status` doesn't display the dep, but I can still `import` it.

KristofferC:
It shows the diff of the project compared to previous commit I think.

fredrikekre:
> It shows the diff of the project compared to previous commit I think.

I find this quite confusing though. Can we just print the aboslute status instead?

StefanKarpinski:
Showing the diff versus the last commit is very useful imo. That’s generally what I want to know that I’ve changed.

fredrikekre:
Will squash later
bors r-

PR #764, opened 2018-09-19
add bin/loop.sh script for updating General registry

StefanKarpinski:
@KristofferC used to run this under his account arctic (Julia Computing cluster); I'm now running it under my account on anubis (another JC machine).


StefanKarpinski:
I'm now running this again after rebasing on top of the update script fixes in https://github.com/JuliaLang/Pkg.jl/pull/769. This comments out the "cap" logic, but if this goes well for a while it should be merged. Then once we get updated PkgEval results from @keno, we can uncomment the cappening and hopefully be good to go with our caps.

StefanKarpinski:
Also adds the long-awaited pkgeval results.

PR #765, opened 2018-09-19
Remove OS-specific shell commands from Pkg docs

ChrisRackauckas:
It's a little disconcerting that if someone wants to run the Pkg docs but are using Windows they get an error.

```
shell> mkdir MyPackage
ERROR: IOError: could not spawn `mkdir MyPackage`: no such file or directory (ENOENT)
Stacktrace:
 [1] _jl_spawn(::String, ::Array{String,1}, ::Cmd, ::Tuple{RawFD,RawFD,RawFD}) at .\process.jl:367
 [2] (::getfield(Base, Symbol("##495#496")){Cmd})(::Tuple{RawFD,RawFD,RawFD}) at .\process.jl:509
 [3] setup_stdio(::getfield(Base, Symbol("##495#496")){Cmd}, ::Tuple{RawFD,RawFD,RawFD}) at .\process.jl:490
 [4] #_spawn#494(::Nothing, ::Function, ::Cmd, ::Tuple{RawFD,RawFD,RawFD}) at .\process.jl:508
 [5] _spawn at .\process.jl:504 [inlined]
 [6] #run#505(::Bool, ::Function, ::Cmd) at .\process.jl:652
 [7] run at .\process.jl:651 [inlined]
 [8] repl_cmd(::Cmd, ::REPL.Terminals.TTYTerminal) at .\client.jl:77
 [9] top-level scope at none:0
```

Instead of using shell commands, these docs should just be using Julia code so that it's not OS-specific since right now it assumes that the reader is using a shell that matches standard Linux.

### DIFF
--- a/docs/src/index.md
+++ b/docs/src/index.md
@@ -180,7 +180,7 @@ managed by system administrators.
 
 The Pkg REPL-mode is entered from the Julia REPL using the key `]`.
 
-```
+```julia-repl
 (v1.0) pkg>
 ```
 
@@ -206,7 +206,7 @@ The most frequently used one is `add` and its usage is described first.
 
 In the Pkg REPL packages can be added with the `add` command followed by the name of the package, for example:
 
-```
+```julia-repl
 (v1.0) pkg> add Example
    Cloning default registries into /Users/kristoffer/.julia/registries
    Cloning registry General from "https://github.com/JuliaRegistries/General.git"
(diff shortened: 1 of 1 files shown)


mortenpi:
The code blocks could have `julia-repl` as the language, to get a bit of highlighting for the `julia> ...` lines. I.e.

````
```julia-repl
julia> foo()
...
````

musm:
this looks pretty good, anything else to do here?

musm:
@ChrisRackauckas  looks like the docs got shuffled around, so this'll need a rebase

StefanKarpinski:
The Example example is not ideal now since it's so verbose but that's not your doing. I think we should merge this and open an issue about making the Example example less verbose.

StefanKarpinski:
Didn't realize there was a conflict here since the docs got moved around. Sorry, @ChrisRackauckas 😞

PR #780, opened 2018-09-26
Use `environment` in error message

goretkin:
moved from https://github.com/JuliaLang/julia/pull/29378


StefanKarpinski:
This is not correct i’m afraid, see the section on environments. A project has an associated environment but not all environments are projects. In this case it must be a project that is active, it cannot be anything else. It is a project that is modified by pkg actions.

goretkin:
Thanks for the information. I ventured a guess, and the guess was wrong :)

StefanKarpinski:
We could change it to "project environment" but that may not help clarify the matter.

PR #785, opened 2018-09-30
Make package files read-only for Pkg.add

### LINKED ISSUE #32: Detect modification of packages that are supposed to be immutable
If a user accidentally modifies a package (for example by using `@edit`), the package could easily break and I don't think there is a way to detect it.

Perhaps there should be some option of doing an integrity check that checks that the tree hashes in the saved packages actually are what they are supposed to be?

timholy:
Closes #32.

Test fail for me locally with a JSON error, but hopefully that's just because an existing JSON installation or something. Tests passed at https://github.com/JuliaLang/julia/pull/29439, for whatever that is worth.

Error:
```julia
ERROR: LoadError: LoadError: LoadError: Unsatisfiable requirements detected for package JSON [682c06a0]:
 JSON [682c06a0] log:
 ├─possible versions are: [0.1.0-0.1.1, 0.2.0-0.2.4, 0.3.0-0.3.9, 0.4.0-0.4.6, 0.5.0-0.5.4, 0.6.0-0.6.1, 0.7.0-0.7.1, 0.8.0-0.8.3, 0.9.0-0.9.1, 0.10.0, 0.11.0, 0.12.0, 0.13.0, 0.14.0, 0.15.0-0.15.2, 0.16.0-0.16.4, 0.17.1-0.17.2, 0.18.0, 0.19.0] or uninstalled
 ├─restricted to versions 0.16.0-0.16 by an explicit requirement, leaving only versions 0.16.0-0.16.4
 └─restricted by julia compatibility requirements to versions: [0.18.0, 0.19.0] or uninstalled — no versions left
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -61,7 +61,7 @@ function add_or_develop(ctx::Context, pkgs::Vector{PackageSpec}; mode::Symbol, s
     any(pkg -> Types.collides_with_project(ctx.env, pkg), pkgs) &&
         pkgerror("Cannot $mode package with the same name or uuid as the project")
 
-    Operations.add_or_develop(ctx, pkgs; new_git=new_git)
+    Operations.add_or_develop(ctx, pkgs; new_git=new_git, mode=mode)
     ctx.preview && preview_info()
     return
 end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -59,6 +59,17 @@ function load_package_data_raw(T::Type, path::String)
     return data
 end
 
+function set_readonly(path)
+    for (root, dirs, files) in walkdir(path)
+        for file in files
+            filepath = joinpath(root, file)
+            fmode = filemode(filepath)
+            chmod(filepath, fmode & (typemax(fmode) ⊻ 0o222))
+        end
+    end
+    return nothing
+end
+
 #######################################
 # Dependency gathering and resolution #
 #######################################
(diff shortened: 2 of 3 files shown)


timholy:
Ah, I see other PRs (e.g., #778) are failing for the same error, so I assume this isn't specific to this PR.

KristofferC:
Yes, I will fix

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/786

KristofferC:
Do you know if `@edit` will continue to work to read the source, but then just fail when one tries to save?

StefanKarpinski:
It seems to me that this should make files and directories read-only but that it should skip the `deps` directory in the case that `deps/build.jl` exists.

timholy:
> Do you know if @edit will continue to work to read the source, but then just fail when one tries to save?

Correct. Some editors like VScode will "helpfully" offer to remove the protection, though:
image

> It seems to me that this should make files and directories read-only but that it should skip the deps directory in the case that deps/build.jl exists.

Easily done. But why would one rewrite `deps/build.jl`? Doesn't one usually write build-step data to something like `deps/deps.jl`, and that file is not checked in?

StefanKarpinski:
> But why would one rewrite `deps/build.jl`?

One wouldn't, but the presence of `deps/build.jl` means that one will typically want to write something to the `deps` directory, usually `deps/deps.jl`. So I guess the rule could be that all existing files should be read-only no matter what and the `deps` directory should be writeable if `deps/build.jl` exists?

timholy:
In this PR all *directories* are writable, but the *files* are not.

StefanKarpinski:
Right, so I guess what I'm asking for is something more restrictive. But I guess we can go ahead with this change first and see how it goes.

timholy:
Some packages install data (e.g., [TestImages](https://juliaimages.github.io/TestImages.jl/)). Others might create some other kinds of files and then modify them (e.g., at least some of the early CUDA packages---and for all I know modern ones---would write `.ptx` files in the package directory and might change them depending on compute capability constraints.)

StefanKarpinski:
Ok, I guess let’s go with this for now. Eventually, I’d love to provide alternatives to all the ways packages write stuff into their source trees, but for now, this is good.

KristofferC:
Lets' try this

bors r+

Issue #795, opened 2018-10-04
Package files are sometimes truncated


ararslan:
Often we (Invenia) have run into an issue where, when installing packages on 0.7 and 1.0, random files in a particular package will be truncated in random places. Most often we've seen this with [AWSSDK](https://github.com/JuliaCloud/AWSSDK.jl), which has a lot of very large files, though the issue is not limited to that package. We noticed this because we've gotten an error on `using AWSSDK` that reports that, for example, a docstring or module is unterminated.

Some data points:

* When we set the default [concurrency](https://github.com/JuliaLang/Pkg.jl/blob/8e91d902458b9dbeab76fa7194ee58d1a6137334/src/Types.jl#L358) in `Context` to 1 instead of 8, the problem disappears.

* We know that the files are intact after the tarballs are extracted.

* When we replace the [`mv`](https://github.com/JuliaLang/Pkg.jl/blob/master/src/Operations.jl#L486) from the extracted tarball location to the package tree location with Julia's `cp`, the problem persists, but if we replace it with a ```run(`cp ...`)```, the problem goes away.

* We have only observed this on Debian Stretch runners on our internal GitLab CI, on both 32- and 64-bit systems, but more often on 32-bit. We have _not_ seen it on our Mac CI or on Amazon Linux.

* We've seen one instance of it happening locally to @rofinn on Elementary OS (based on Ubuntu which is in turn based on Debian) when installing GitHub.jl.

* The [`@async`s](https://github.com/JuliaLang/Pkg.jl/blob/master/src/Operations.jl#L577-L584) in `apply_versions` are never `@sync`ed, and the [`Channel`s](https://github.com/JuliaLang/Pkg.jl/blob/master/src/Operations.jl#L575-L576) are never closed.

* We use multiple registries: the public General registry and a private one for internal packages.

StefanKarpinski:
How large are the files? Any chance that https://github.com/JuliaLang/julia/issues/14574 is related?

ararslan:
The smallest file we've seen truncated was supposed to be 108,741 bytes, and was truncated at 65,536 bytes. Likely unrelated to that issue given that it's well under 2 GB.

KristofferC:
> When we replace the mv from the extracted tarball location to the package tree location with Julia's cp, the problem persists, but if we replace it with a run(`cp ...`), the problem goes away.

Seems like the most damning to me.

[...]

PR #797, opened 2018-10-05
Add an environment variable to control concurrency

ararslan:
This adds an environment variable, tentatively called `JULIA_PKG_CONCURRENCY`, which allows the user to control the number of concurrent downloads that Pkg uses when installing packages, using the previous default of 8 if the variable is not specified.

Setting the number of concurrent downloads to 1 using this variable provides a temporary workaround for #795.

Currently this PR does not document the variable. Suggestions for where that documentation should live would be welcome.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -355,6 +355,8 @@ Base.@kwdef mutable struct Context
     preview::Bool = false
     use_libgit2_for_all_downloads::Bool = false
     use_only_tarballs_for_downloads::Bool = false
+    # NOTE: The JULIA_PKG_CONCURRENCY environment variable is likely to be removed in
+    # the future. It currently stands as an unofficial workaround for issue #795.
     num_concurrent_downloads::Int = get(ENV, "JULIA_PKG_CONCURRENCY", 8)
     graph_verbose::Bool = false
     stdlibs::Dict{UUID,String} = gather_stdlib_uuids()
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -357,7 +357,7 @@ Base.@kwdef mutable struct Context
     use_only_tarballs_for_downloads::Bool = false
     # NOTE: The JULIA_PKG_CONCURRENCY environment variable is likely to be removed in
     # the future. It currently stands as an unofficial workaround for issue #795.
-    num_concurrent_downloads::Int = get(ENV, "JULIA_PKG_CONCURRENCY", 8)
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Seems reasonable to me, although of course, we also want to fix #795.

KristofferC:
Ugh, more environment variables. :(

One of the reasons for the context stuff was so that you could (and can) do

```
Pkg.add("Foo"; num_concurrent_downloads=1)
```

I wish I would just have done the work and had it read from an actual config file...

ararslan:
> we also want to fix #795.

Yeah for sure. But @rofinn has hit this locally with packages that have large files, and we determined it's likely because his home directory is encrypted, which (we think) means that a move command has to be a copy and delete rather than a rename. Understandably he'd prefer not to mess with the location of the temp directory.

> Ugh, more environment variables. :(

Yeah, not ideal. Not sure what else to suggest for the moment though given the above.

StefanKarpinski:
As long as we don't make this one official or document it, this seems fine. But yes, we need a better non-environment variable way to configure things globally.

KristofferC:
Can you put in a comment saying that this is likely to be removed in the future.

rofinn:
I'd also be happy with changing the default Context in my juliarc file if that's easier.

src/Types.jl
```diff
@@ -355,7 +355,9 @@ Base.@kwdef mutable struct Context
     preview::Bool = false
     use_libgit2_for_all_downloads::Bool = false
     use_only_tarballs_for_downloads::Bool = false
-    num_concurrent_downloads::Int = 8
+    # NOTE: The JULIA_PKG_CONCURRENCY environment variable is likely to be removed in
+    # the future. It currently stands as an unofficial workaround for issue #795.
+    num_concurrent_downloads::Int = get(ENV, "JULIA_PKG_CONCURRENCY", 8)
```
fredrikekre: Don't you need to parse the return value as an `Int`?

ararslan:
Comment addressed, test added

fredrikekre:
bors try

ararslan:
What needs to happen for bors to report a status?

fredrikekre:
It did: https://github.com/JuliaLang/Pkg.jl/pull/797#issuecomment-427586999

ararslan:
I mean in terms of PR status, like what's reported by Travis. In the CI checks, bors says "Expected — Waiting for status to be reported," and is a required check.

fredrikekre:
It is just waiting for commands.

fredrikekre:
Backport? I should have squashed this.

PR #799, opened 2018-10-05
WIP/RFH: Download packages from GitLab as tarballs

ararslan:
When a URL is recognized to be GitLab, we can use their API to download a tarball like we do with GitHub. This avoids having to download entire Git repos for GitLab-hosted packages.

Note however that self-hosted GitLab instances don't necessarily have to have "gitlab" in their URLs, but those that do typically start with "gitlab," so it provides a reasonable heuristic. For testing purposes, I've made a registry and an example package in my personal GitLab account on GitLab's public server: https://gitlab.com/ararslan/GitLabExampleRegistry and https://gitlab.com/ararslan/GitLabExample.jl.

~~As it stands, this PR doesn't test that a tarball is actually being used as opposed to a Git repo. Any suggestions for how best to do that would be welcome.~~ I _think_ what I have now should do it.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -620,9 +620,10 @@ end
 """
     setprotocol!(proto::Union{Nothing, AbstractString}=nothing)
 
-Set the protocol used to access GitHub-hosted packages when `add`ing a url or `develop`ing a package.
+Set the protocol used to access packages hosted on GitHub or GitLab when `add`ing a url or
+`develop`ing a package.
 Defaults to delegating the choice to the package developer (`proto == nothing`).
-Other choices for `proto` are `"https` or `git`.
+Other choices for `proto` are `"https"` or `"git"`.
 """
 setprotocol!(proto::Union{Nothing, AbstractString}=nothing) = GitTools.setprotocol!(proto)
 
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -68,19 +68,30 @@ end
 
 const GITHUB_REGEX =
     r"^(?:git@|git://|https://(?:[\w\.\+\-]+@)?)github.com[:/](([^/].+)/(.+?))(?:\.git)?$"i
+const GITLAB_REGEX =
+    r"^(?:git@|git://|https://(?:[\w\.\+\-]+@)?)gitlab([^:/]+)[:/](([^/].+)/(.+?))(?:\.git)?$"i
 const GIT_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
 
 setprotocol!(proto::Union{Nothing, AbstractString}=nothing) = GIT_PROTOCOL[] = proto
 
 # TODO: extend this to more urls
 function normalize_url(url::AbstractString)
-    m = match(GITHUB_REGEX, url)
-    if m === nothing || GIT_PROTOCOL[] === nothing
+    gh = match(GITHUB_REGEX, url)
+    gl = match(GITLAB_REGEX, url)
+    if (gh === nothing && gl === nothing) || GIT_PROTOCOL[] === nothing
         url
     elseif GIT_PROTOCOL[] == "ssh"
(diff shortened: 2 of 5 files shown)


StefanKarpinski:
Nice!

ararslan:
Well, not _too_ nice, as it doesn't seem to work properly yet 😛

ararslan:
Hm, in local testing, the modified functions do seem to correctly identify and deal with GitLab URLs, but Pkg is still downloading them as Git repos. It appears (though I'm not certain) that it's because the Git protocol is defaulting to `nothing`, which means that `normalize_url` uses the URL verbatim and thus won't append ".git," so it won't get properly recognized by the regular expression in `get_archive_url_for_version`. I tried manually specifying .git in the given URL to test this theory but that also fails, so I think there's a piece here I'm missing.

omus:
Looking at the failure it looks like when you specify the `url` as part of the `PackageSpec` then the provided version is ignored:
```julia
julia> Pkg.add(PackageSpec(name="Example", url="https://github.com/JuliaLang/Example.jl", version=v"0.3.0"))
   Cloning git-repo `https://github.com/JuliaLang/Example.jl`
  Updating git-repo `https://github.com/JuliaLang/Example.jl`
 Resolving package versions...
  Updating `~/.julia/environments/v1.0/Project.toml`
  [7876af07] + Example v0.5.1+ #master (https://github.com/JuliaLang/Example.jl)
  Updating `~/.julia/environments/v1.0/Manifest.toml`
  [7876af07] + Example v0.5.1+ #master (https://github.com/JuliaLang/Example.jl)
```

ararslan:
I've opened an issue for that: https://github.com/JuliaLang/Pkg.jl/issues/872

ararslan:
bors try

ararslan:
I made a public registry (https://gitlab.com/ararslan/GitLabExampleRegistry) and a package (https://gitlab.com/ararslan/GitLabExample.jl) for GitLab testing. The registry was created manually so I can't guarantee that isn't the cause of the failure.

ararslan:
bors try

ararslan:
bors try

fredrikekre:
bors try

fredrikekre:
bors try

fredrikekre:
Ok, so apparently powershell + GitLab is not a happy combination and we run into https://stackoverflow.com/questions/25596564/percent-encoded-slash-is-decoded-before-the-request-dispatch

Quoting @staticfloat :
> Easy solution: get rid of the `/` and `.` names in the GitLab repository name.  :stuck_out_tongue:
Worse solution: monkey-patch powershell via the “Workaround 1” in that SO post
but that is likely .NET version specific
which _sucks_
Best solution: complain to GitLab and get them to fix it

fredrikekre:
bors try

StefanKarpinski:
> I can not find anything about downloading a tarball from a tree-sha from GitLab.

It may not be possible through the GitLab API.

fredrikekre:
Exactly, I don't think it is.

StefanKarpinski:
Is there any path forward on this or are we pretty much stuck because of the GitLab API?

ararslan:
I'm out of ideas for how to get around the limitations of their API, so we may just be stuck, unfortunately.

fredrikekre:
Seems like a good feature request.

KristofferC:
Feels this can be closed via PkgServer.

StefanKarpinski:
Given that we don't use PowerShell for downloads anymore, it's possible that this might just work now. Worth dusting off and trying again as a fallback download mechanism for GitLab hosted code?

fredrikekre:
> it's possible that this might just work now.

That wasn't the problem in the end. The GitLab API just doesn't support requesting tree hashes.

StefanKarpinski:
Ah, ok. No working around that then.

PR #803, opened 2018-10-09
Misc doc updates + quickstart

fredrikekre:


docs/src/index.md
```diff
@@ -1,6 +1,17 @@
 # **1.** Introduction
 
-Pkg is the standard package manager for Julia 1.0 and newer. Unlike traditional
+Welcome to the documentation for Pkg, Julias package manager.
```
NHDaly: s/Julias/Julia's/

docs/src/index.md
```diff
@@ -1,6 +1,17 @@
...
+The documentation covers many things, for example managing package
+installations, developing packages, working with package registries and more.
+
+Throughout the manual the REPL interface to Pkg is used in the examples.
+There is also a functional API interface, which is prefered when not working
+interactively. This API is documented in the [API Reference](@ref) section.
+
+## Background and Design
+
+Pkg is a complete rewrite of Julias old package manager, released together
```
NHDaly: s/Julias/Julia's/

fredrikekre:
bors try

docs/Manifest.toml
```diff
@@ -0,0 +1,101 @@
...
+uuid = "56ddb016-857b-54e1-b83d-db4d58db5568"
+
+[[Markdown]]
+deps = ["Base64"]
+uuid = "d6f4376e-aef5-505a-96c1-9c027394607a"
+
+[[Pkg]]
+deps = ["Dates", "LibGit2", "Markdown", "Printf", "REPL", "Random", "SHA", "UUIDs"]
+uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+[[Pkg]]
```
StefanKarpinski: You seem to accidentally have two copies of `Pkg` in here.

fredrikekre: Well, one is the Pkg which `DocStringExtensions` depend on (the stdlib one) and the other one is the development version of Pkg so this is correct.

StefanKarpinski: I assume you'll change that before merging?

fredrikekre: Change what?

fredrikekre: How? The `DocStringExtension` deps is from the registry, and we can't change that, and we need a different uuid for `Pkg` to actually use the development version. I don't see the problem here.

StefanKarpinski: > I don't see the problem here.

Checking a bogus, wrong UUID into the git repo? They should both use the development version in that case and we should have a mechanism for replacing the sysimg copy.

fredrikekre: > in that case and we should have a mechanism for replacing the sysimg copy.

Yea that would be great, but we don't have this and hence we actively change the uuid of Pkg when testing/building docs: https://github.com/JuliaLang/Pkg.jl/blob/cb37ef1ad3d1eebc912b83ddcb414aa3d67631ca/.travis.yml#L34. This is nothing new introduced in this PR.

We are also no using DocStringExtensions in the Pkg docs, so it will not matter that there are different versions.

PR #814, opened 2018-10-10
code to (de)compress registry TOML files

StefanKarpinski:
@KristofferC, does this help you with https://github.com/JuliaLang/PkgDev.jl/pull/144?


StefanKarpinski:
I have confirmed that the compressed `Deps.toml` and `Compat.toml` files that this new code generates represent equal data to what the old code generates albeit sorted differently.

PR #818, opened 2018-10-11
Refactor command declarations

00vareladavid:
This replaces the old tuples with keyword vectors.
 
Advantages:
- delegates work to Julia's keyword argument mechanism
- more readable (IMO)
- order doesn't matter
- allow for defaults (e.g. if a command doesn't take flags, just omit the `:option_spec` keyword pair)

Note that this change should only affect the declarative interface and does not change the internal "Spec" structs (other than one small tweak)


00vareladavid:
bors try

PR #826, opened 2018-10-13
Refactor option and argument parsing

00vareladavid:
Changes:
- move option parsing to `parse`: it is purely syntactical and does not depend on the command specs
- remove redundant option check
- fold `enforce_argument_order` into `parse_pkg`: `enforce_argument_order` was needlessly abstract
- break up `arg_spec` into `arg_count` and `arg_parser`.
    - Omitting `arg_parser` means "don't do any special parsing on the arguments" (i.e. the "parser" is just `identity`).
    - Omitting `arg_count` means "this command does not take arguments".
- added some docstrings and comments

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -182,7 +182,7 @@ end
 ################
 mutable struct Statement
     command::Union{Nothing,CommandSpec}
-    options::Vector{String}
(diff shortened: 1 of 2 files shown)


00vareladavid:
bors try

PR #603, opened 2018-08-08
Fix `preview` and update completions

00vareladavid:
Should resolve #600

PR #839, opened 2018-10-18
Refactor completions

00vareladavid:
Changes:
- Introduce `completions` field to command declarations. This field declares a function `f : options x string_to_complete -> possible completions`. The intent is to isolate the mechanical (and edge case ridden!) parts from the meaningful.
- share logic between the regular parser and completions parser (this shared logic is now in`core_parse`)
- remove `CompletionCache` object. It introduced quite a bit of state for no noticeable speed benefit

Fixes:
- only complete canonical names for commands (i.e. not the short form) (I didn't know about #200, when I did #603)
- in the same spirit, only complete canonical form for options (e.g. `--manifest` but not `-m`)

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -97,10 +97,10 @@ struct CommandSpec
     handler::Union{Nothing,Function}
     argument_spec::ArgSpec
     option_specs::Dict{String, OptionSpec}
+    completions::Union{Nothing,Function}
     description::String
     help::Union{Nothing, Markdown.MD}
 end
-command_specs = Dict{String,CommandSpec}() # TODO remove this ?
 
 function SuperSpecs(compound_commands)::Dict{String,Dict{String,CommandSpec}}
     super_specs = Dict()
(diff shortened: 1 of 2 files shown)


00vareladavid:
bors try

00vareladavid:
Everything seems to be in order.. other than local path completions on Windows :expressionless:
All I do is [append `Base.Filesystem.path_separator`](https://github.com/JuliaLang/Pkg.jl/pull/839/files#diff-b36bd2f80c73b62d483ebe62164aac17R685). But the [tests](https://github.com/JuliaLang/Pkg.jl/pull/839/files#diff-564700bd12410fbbaf97099209c3259cR437) seem to expect two separators? Anyone know why?

00vareladavid:
bors try

00vareladavid:
bors try

KristofferC:
I think @pfitzseb wrote the original tests for this?

00vareladavid:
Upon further reflection, iterative completion of local paths is a compelling enough use case to justify additional complexity.

For everything else, this PR makes is easy to add completion functionality while being blissfully unaware of any completion internals. For example:
```
function complete_installed_packages(options, partial)
    mode = get(options, :mode, nothing)
    pkgs = mode == PKGMODE_MANIFEST ? API.__installed() : API.__installed(PKGMODE_PROJECT)
    return collect(keys(filter(p->p[1] in stdlib_names() || p[2] !== nothing, pkgs)))
end
```
Just look at the options and return a list of strings, that's all!

PR #548, opened 2018-07-31
fix unneeded use of pwd in REPL prompt function

KristofferC:
Likely left when pwd had some importance to the project being used.

Fixes #547

PR #848, opened 2018-10-22
Expand symlinks in find_project_file.

fredrikekre:
I have symlinked `environments/v1.0` and`environments/v1.1` to `environments/v1`. #548 removed a call to `realpath` in `promptf` with the result that the symlink was shown in the prompt.

### Before #548
Note the expanded symlink in the prompt, but not for the project file.
```
(v1) pkg> st
    Status `~/.julia/environments/v1.0/Project.toml`
```

### After #548
None of the links expanded
```
(v1.0) pkg> st
    Status `~/.julia/environments/v1.0/Project.toml`
```
### After this PR
This PR restores the pre-#548 behavior (which I liked better) and additionally expands the symlink for the displayed path, the result with this PR is thus
```
(v1) pkg> st
    Status `~/.julia/environments/v1/Project.toml`
```

Opinions?

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -256,7 +256,7 @@ function find_project_file(env::Union{Nothing,String}=nothing)
     @assert project_file isa String &&
         (isfile(project_file) || !ispath(project_file) ||
          isdir(project_file) && isempty(readdir(project_file)))
-     return project_file
+    return safe_realpath(project_file)
 end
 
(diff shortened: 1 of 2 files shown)


KristofferC:
Makes sense to me 👍

StefanKarpinski:
Me too.

fredrikekre:
bors try

PR #857, opened 2018-10-27
Structured Project data

00vareladavid:
Data flow:
`read_project`: file -> raw Dict 
`Project`: raw Dict -> Project struct
[Pkg Operations]
`destructure`: Project struct -> raw Dict
`write_env`: raw Dict -> file

A small step in the right direction I hope. It should at least be clear where we expect `String` data v typed data. And it should reduce the number of overall data conversions (e.g. no `UUID(uuid)` all over the place).

I left the `compat` as a `Dict{String,String}` (and not `Dict{String,VersionSpec}`) because `String` -> `VersionSpec` is a many-to-one mapping (I think).

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -7,6 +7,7 @@ using Random
 using Dates
 import LibGit2
 import REPL
+import Base.string
 using REPL.TerminalMenus
 
 using ..TOML
@@ -260,6 +261,20 @@ function find_project_file(env::Union{Nothing,String}=nothing)
     return safe_realpath(project_file)
 end
 
+Base.@kwdef mutable struct Project
+    raw::Dict{String,Any} = Dict{String,Any}()
+    # Fields
+    name::Union{String, Nothing} = nothing
+    uuid::Union{UUID, Nothing} = nothing
+    version::Union{VersionTypes, Nothing} = nothing
+    manifest::Union{String, Nothing} = nothing
+    # Sections
+    deps::Dict{String,UUID} = Dict{String,UUID}()
+    extras::Dict{String,UUID} = Dict{String,UUID}()
+    targets::Dict{String,Vector{String}} = Dict{String,Vector{String}}()
+    compat::Dict{String,String} = Dict{String,String}()# TODO Dict{String, VersionSpec}
+end
+
 mutable struct EnvCache
     # environment info:
     env::Union{Nothing,String}
@@ -273,7 +288,7 @@ mutable struct EnvCache
     pkg::Union{PackageSpec, Nothing}
 
     # cache of metadata:
-    project::Dict
+    project::Project
     manifest::Dict
 
     # registered package info:
@@ -286,24 +301,25 @@ function EnvCache(env::Union{Nothing,String}=nothing)
     project_file = find_project_file(env)
     project_dir = dirname(project_file)
     git = ispath(joinpath(project_dir, ".git")) ? project_dir : nothing
-
+    # read project file
     project = read_project(project_file)
-    if any(haskey.((project,), ["name", "uuid", "version"]))
+    # initiaze project package
+    if any(x -> x !== nothing, [project.name, project.uuid, project.version])
         project_package = PackageSpec(
-            get(project, "name", ""),
-            UUID(get(project, "uuid", 0)),
-            VersionNumber(get(project, "version", "0.0")),
+            something(project.name, ""),
+            something(project.uuid, UUID(0)),
+            something(project.version, VersionNumber("0.0")),
         )
     else
         project_package = nothing
     end
-    # determine manifest_file name
-    dir = abspath(dirname(project_file))
-    manifest_file = haskey(project, "manifest") ?
-        abspath(project["manifest"]) :
+    # determine manifest file
+    dir = abspath(project_dir)
+    manifest_file = project.manifest !== nothing ?
+        abspath(project.manifest) :
         manifestfile_path(dir)
     # use default name if still not determined
-    (manifest_file === nothing) && (manifest_file = joinpath(dir, "Manifest.toml"))
+    manifest_file = something(manifest_file, joinpath(dir, "Manifest.toml"))
     write_env_usage(manifest_file)
     manifest = read_manifest(manifest_file)
     uuids = Dict{String,Vector{UUID}}()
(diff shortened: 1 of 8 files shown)


00vareladavid:
This [test](https://github.com/JuliaLang/Pkg.jl/blob/9a53f24a0b00b04e5e901b8609437b5ff9cb12f8/test/pkg.jl#L199) is failing locally. I don't know why the a `dev`ed version must be strictly greater than the `add`ed version. Couldn't they be equal? They are equal in my local tests..

KristofferC:
Cool! Any performance numbers (e.g. doing something like `Pkg.add("DifferentialEquations")` over and over)?

00vareladavid:
I think so!

With regards to performance:
```
@time begin
    for i in 1:20
        Pkg.add("HackerNews")
    end
end
```
- this branch: `7.121146 seconds (13.32 M allocations: 736.612 MiB, 7.18% gc time)`
- master: `6.900136 seconds (13.28 M allocations: 732.869 MiB, 7.05% gc time)`

Pretty sure that's because I create new dicts though. If I mutate the values of the toml dicts in place (in `Project`), I suspect this branch would be faster than master.

Wouldn't downloading the registry + resolving overshadow these numbers though?

Or were you asking for something different? :sweat_smile:

00vareladavid:
bors try

00vareladavid:
bors try

KristofferC:
Regarding performance, I just wanted to see that nothing crazy happened :)

00vareladavid:
Linux/mac tests pass, but I'm hitting a reproducible libuv error on Windows, again. The error is triggered by the test utilities and not the test themselves.

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

fredrikekre:
I kinda liked that you can index the project with the keys from the project file directly (e.g. `project["deps"]` etc). Does `Project` need to be a struct like this or could it just be a Dict that is verified for consistency?

Also, this seems to throw away everything from `Project.toml` that is note used by Pkg? Should the raw dict that we read from file be stored in `Project` too?

KristofferC:
We want to emit the same TOML info that we read which includes things not currently used by Pkg.

00vareladavid:
It could be either a Dict or a struct. I like the struct because it clearly reflects what fields we expect from `Project.toml`. The struct constructor also clearly shows the default values in the case that `Project.toml` does not specify an expected field. What would be the advantage to using a dict?

The expected fields of the raw dict will be updated in `destructure`, right before we update `Project.toml`. The current plan is to re-read the `Project.toml` file right before we write back (I left it as a TODO in the code).

I was concerned we might end up with a stale cache if we cache the raw dict (e.g some other process updates a field that we don't care about). If that is not a valid (or likely) concern, then caching the raw dict would be a less expensive way to do it.

src/Types.jl
```diff
@@ -259,6 +260,31 @@ function find_project_file(env::Union{Nothing,String}=nothing)
      return project_file
 end
 
+mutable struct Project
```
StefanKarpinski: What about having an `other::Dict{String,Any}` field to carry around anything else that might be in the file?

StefanKarpinski:
As I've commented inline, what about having an `other::Dict{String,Any}` field in the project file and then merging that back into the the dict during destructure? That allows for whatever top-level entries people want as well as other sections too. They just get carried around as TOML data and ignored.

KristofferC:
It seems odd to me to reread the Project file upon writing it back. The output should be the read Project file with the changes made to it.

StefanKarpinski:
> It seems odd to me to reread the Project file upon writing it back.

Indeed, that does seem weird. On the other hand, if something gets edited between the first read and the write, then it is less likely to get clobbered... which is maybe a good thing? The amount of time is unlikely to be significant, of course.

00vareladavid:
Caching the raw dictionary it is then. FWIW I did that originally until I became concerned of clobbering the other fields. I do agree that it seems unlikely to occur. We can always change it back if it happens to be a problem in practice.

00vareladavid:
Any advice on what to do about the libuv error? Other tests just catch the error and ignore it. But for some reason, the error fully propagates on this specific test. I'm perplexed about what the difference is.

The simplest workaround would be to not clean up the test directory for the failing test. It should still test the correct thing, its just not ideal.

src/Types.jl
00vareladavid: Done

KristofferC:
Ugh, not sure what is best. This type of permission error with the windows filesystem in combination with it being extremely hard to find the problem from Julia is quite annoying.

src/Types.jl
```diff
@@ -259,6 +260,31 @@ function find_project_file(env::Union{Nothing,String}=nothing)
...
+                    nothing,
+                    nothing,
+                    nothing,
+                    nothing,
+                    Dict{String,UUID}(),
+                    Dict{String,UUID}(),
+                    Dict{String,Vector{String}}(),
+                    Dict{String,String}(),
+                    )
+end
```
omus: There is a flaw here. There are 9 fields in this struct and the constructor passes in 10 values. Seems to be an extra `nothing`.

00vareladavid: Thanks!

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
Progress!

Finally got to the correct error message, using the debug code by @omus.
The trace makes me think this should be handled upstream?

StefanKarpinski:
Did you mean to close this?

00vareladavid:
Nope, sorry about that. Still trying to debug :grimacing:

StefanKarpinski:
Good—I'm glad you haven't given up in frustration. Is there anything someone else can do to help track this issue down?

00vareladavid:
Thanks! I appreciate the support.

I think my only option now is to find a minimum change that will trigger the error so that I can work around it. I don't think more people will necessarily help at this point, although I'm open to high level workaround suggestions.

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
Make me wish that `rm` had a:
```
@static if Sys.iswindows()
   force && Base.GC.gc()
end
```

00vareladavid:
bors try

00vareladavid:
bors try

src/Types.jl
```diff
@@ -690,6 +758,9 @@ function handle_repos_add!(ctx::Context, pkgs::AbstractVector{PackageSpec};
             if !folder_already_downloaded
                 version_path = Pkg.Operations.find_installed(pkg.name, pkg.uuid, pkg.repo.git_tree_sha1)
                 mkpath(version_path)
+                @static if Sys.iswindows()
```
00vareladavid: This was enough to work around the filesystem error

KristofferC: Urgh, do we really need it in the Pkg code itself? Can't we put it in the test code? If it really is needed I guess we can keep it here.

00vareladavid: I think it needs to be here since it is the `mv` call inside of Pkg that triggers the error. Is there a way to work around that from outside of Pkg?

FWIW I'm not thrilled about it either.

00vareladavid:
I believe this is ready to merge. Are there any further concerns?

00vareladavid:
bors try

StefanKarpinski:
Yay—very excited to have this merged! Nice work, @00vareladavid.

00vareladavid:
Thanks! Glad to have it merged too

src/Types.jl
```diff
@@ -1351,13 +1422,37 @@ function project_key_order(key::String)
     return 8
 end
 
+string(x::Vector{String}) = x
```
StefanKarpinski: Ok, I should have caught this the first time around but this is some pretty bad type piracy.

Cross ref: https://github.com/JuliaLang/julia/issues/30945#issuecomment-460065670

Still exists in Pkg, needs to be removed.

PR #864, opened 2018-10-29
a couple of fixes and updates

StefanKarpinski:


StefanKarpinski:
This is already deployed so it should eliminate bogus dependencies on stdlibs from non-Julia files in package repos. There's some chance it will break some of them but we'll have to see.

Issue #865, opened 2018-10-30
Allow coverage tests to be executed with option `inline=no`


ronisbr:
Hi!

It turns out that Julia reports very wrong coverage information in some cases if the data is generated without `--inline=no` (see JuliaCI/Coverage.jl#187). Hence, I think it is necessary to provide a way to execute the package test with this option. Otherwise, at least for my case, the coverage information generated at Travis for Coveralls and Codecov is really wrong.

tkoolen:
While this would be good to have anyway, do note that https://github.com/JuliaLang/julia/issues/28192#issuecomment-418277362 reports that `--inline=no` is not a solution to the code coverage problems.

fingolfin:
@tkoolen it is not clear from me that the comment you link to is accurate. In particular, it's not clear what exactly they "tried" to come to that conclusion. After all, the reason @ronisbr has filed this is issue is that simply passing `--inline-no` to julia is not enough, a change to `Pkg.test` is apparently needed.

fingolfin:
PR #866 attempts to resolve this.

ronisbr:
In my case (ReferenceFrameRotations.jl), in which the tests and the functions are somewhat simple, I can say that `inline=no` improved a lot the coverage results.

00vareladavid:
Resolved by #870

PR #866, opened 2018-10-30
Turn of inlining when computing coverage (WIP)

fingolfin:
It can also be explicitly turned on or off via the new keyword argument
`inline` for Pkg.test.

This is meant to resolve issue #865. I marked it as WIP as clearly this new option should be documented; but before spending time on that, I wanted to see if this approach was deemed acceptable at all.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1280,7 +1280,7 @@ function free(ctx::Context, pkgs::Vector{PackageSpec})
     need_to_resolve && build_versions(ctx, new)
 end
 
-function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false)
+function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, inline=true)
     # See if we can find the test files for all packages
     missing_runtests = String[]
     testfiles        = String[]
@@ -1333,6 +1333,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false)
             --color=$(Base.have_color ? "yes" : "no")
             --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
             --check-bounds=yes
+            --inline=$(inline ? "yes" : "no")
             --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
             --track-allocation=$(("none", "user", "all")[Base.JLOptions().malloc_log + 1])
             --eval $code
--- a/src/API.jl
+++ b/src/API.jl
@@ -254,7 +254,7 @@ test(pkg::Union{String, PackageSpec}; kwargs...)  = test([pkg]; kwargs...)
 test(pkgs::Vector{String}; kwargs...)             = test([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
 test(pkgs::Vector{PackageSpec}; kwargs...)        = test(Context(), pkgs; kwargs...)
 
-function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, kwargs...)
+function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, inline=nothing, kwargs...)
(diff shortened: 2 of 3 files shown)


fingolfin:
Note that I made it so that now any coverage test by default turns off inlining, thus most packages should get the "fix" automatically. Drawback of this is that those tests also now may run a lot slower, which might cause new problems. So one may disagree whether this is a good idea or not. I can of course change the defaults if that's deemed safer, but then tons of projects may need to explicitly turn off inlining for their coverage tests.

Another argument to be made is whether the name `inline` for the option is good. Perhaps it should be `can_inline` to match `Base.JLOptions().can_inline` (but then shouldn't `coverage` be `code_coverage` as well). Anyway, I don't much care either way, and will use whatever name you folks prefer :)

src/Operations.jl
```diff
@@ -1333,6 +1333,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false)
             --color=$(Base.have_color ? "yes" : "no")
             --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
             --check-bounds=yes
+            --inline=$(inline ? "yes" : "no")
```
tkoolen: Base this on `Base.JLOptions().can_inline`, similar to other options?

fingolfin: I thought about doing that, too. One drawback of that would be that all packages affected by the problem at hand would have to be updated to explicitly add `--inline=no` to their test script(s), and of course also the default Julia scripts for Travis CI and AppVeyor would need to be adapted. I.e. far more people would have to change far more code.

That said, of course I can change it to do that *shrug*

fredrikekre:
It feels like we shouldn't have to fix this Julia bug (?) here in Pkg.

KristofferC:
Did coverage work ok in 0.6 with inlining off?

fingolfin:
@fredrikekre We can't fix a Julia bug (?) in Pkg anyway, we can merely work around it. So, consider this PR as a bandaid, to help people get working coverage reports *now*, as a bandaid until when (if?) Julia actually gets fixed / improved in this regard. But proper coverage tracking in compiled involving optimizations and inlining is a really difficult problem, so I would not expect a quick solution (although of course I'd love to be proven wrong); and the "official" solution may very well involve the suggestion to turn off inlining...

KristofferC:
This would turn off inlining for everyone using the default CI script by default which is not really acceptable. I think, for now, the best thing is to just propagate the `Base.JLOptions().can_inline` option to the test worker. It would be good to add a blurb to the documentation about this.

fingolfin:
@KristofferC to clarify: this would turn off inlining for *the part computing coverage results*. I'd argue that computing coverage results with inlining on largely makes no sense (at least with the current state of it in Julia), as the result is completely bonkers. Which is after all what motivated this PR in the first place. So I'd argue to the contrary that no disabling it is not really acceptable...

Anyway, if that's what the Julia pros decide is the way to go, so be it, I don't mind; after all, I know how to get things working for my projects either way :-).

KristofferC:
My point is that coverage is on by default in the CI scripts.

fingolfin:
Ahhhh, OK -- indeed! Then I guess I have to concur with your assessment (and there just is no good way out of this that doesn't require a lot of package authors to make changes)

ararslan:
>  I think, for now, the best thing is to just propagate the `Base.JLOptions().can_inline` option to the test worker.

See https://github.com/JuliaLang/julia/pull/29858

fingolfin:
While I agree that changing this in Julia might be better, I wonder if adding this in here wouldn't get this change much quicker into the hands of users? At least in my imagination, I'd expect `Pkg.jl` it to be easier to make a new bugfix release of `Pkg.jl` than of Julia itself. Then again, that might be an incorrect assessment?

fingolfin:
Err, and with "this", I actually meant PR #870, not this PR here. Apologies for the confusion.

PR #870, opened 2018-10-31
Transfer can_inline to subprocess

fingolfin:
This is meant as an alternative to PR #866 for resolving issue #865

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1333,6 +1333,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false)
             --color=$(Base.have_color ? "yes" : "no")
             --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
             --check-bounds=yes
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
@KristofferC is going to reword the doc a bit and then we'll merge this.

diegozea:
@tkelman @ninjin @staticfloat @simonbyrne Should the default Travis CI script for Julia use this `inline=no` when `codecov` or `coveralls` is set to `true` ? I'm using the default scripts and I'm seeing the coverage drop reported in https://github.com/JuliaCI/Coverage.jl/issues/187

KristofferC:
Only if we make `coverage=false` the default because `inline=no` is typically too slow to be usable in my experience.

StefanKarpinski:
Please don't ping so many people—everyone who watches this repo will already get this.

diegozea:
I'm sorry. I've just used the list suggested by [Travis CI docs](https://docs.travis-ci.com/user/languages/julia/) without checking the list of watchers.

StefanKarpinski:
Hah, ok, you did follow the instructions exactly. Those instructions should be updated!

PR #873, opened 2018-11-01
Replace trivial uses of rethrow(exc) with rethrow()

c42f:
In all these cases we're merely rethrowing the existing exception, so it
seems simpler and clearer just to use `rethrow()` without arguments.

Companion to https://github.com/JuliaLang/julia/pull/29794

### DIFF
--- a/bin/gitmeta.jl
+++ b/bin/gitmeta.jl
@@ -85,7 +85,7 @@ function gitmeta(pkgs::Dict{String,Package})
             if !updated
                 try LibGit2.GitObject(repo, git_commit_hash)
                 catch err
-                    err isa LibGit2.GitError && err.code == LibGit2.Error.ENOTFOUND || rethrow(err)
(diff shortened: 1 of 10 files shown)


StefanKarpinski:
bors try

StefanKarpinski:
Travis Mac failure before actual tests began to run:
```
0.28s$ hdiutil mount -readonly -mountpoint juliamnt julia.dmg
hdiutil: mount failed - image not recognized

The command "hdiutil mount -readonly -mountpoint juliamnt julia.dmg" failed and exited with 1 during .
```
https://travis-ci.org/JuliaLang/Pkg.jl/jobs/449364543#L75

Anyone know what's up with that?

KristofferC:
@StefanKarpinski Mac buildbot hasn't run in a month. Nightly gets deleted.

StefanKarpinski:
Should we disable the test then so that bors can work properly?

KristofferC:
Cutting the gordian knot ;)

Issue #876, opened 2018-11-02
Pkg sometimes removes [targets]


fredrikekre:
Sometimes the
```
[targets]
test = ["Test"]
```
section is removed from the `Project.toml` file. I have not been able to understand why, and I have always thought that "Well, I guess I accidentally removed it then...", but it keeps happening and it has driven me crazy! I HAVE NOT REMOVED IT!!

I have finally found a way to reproduce it:
1. Start with a clean clone of Pkg or `git clean -fdx` notably to remove the Manifest.toml file.
2. Remove the `uuid = ` section in Pkg
3. Run `julia --project` in the Pkg directory
4. Run `using Pkg` (this seems to be important ...)
5. `pkg> add Example`
6. `pkg> rm Example`

and then you have the following diff for the `Project.toml` file:

```diff
diff --git a/Project.toml b/Project.toml
index ed4c5a3..006b005 100644
--- a/Project.toml
+++ b/Project.toml
@@ -16,6 +16,3 @@ UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
-
-[targets]
-test = ["Test"]
```

PR #879, opened 2018-11-02
Debugging #857

00vareladavid:
I rolled back some inessential changes to see if the error persists.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -147,19 +147,17 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
     fix_deps_map[pkg.uuid] = valtype(fix_deps_map)()
     (project_file === nothing) && return false
     project = read_package(project_file)
-    compat = get(project, "compat", Dict())
-    if haskey(compat, "julia")
-        if !(VERSION in Types.semver_spec(compat["julia"]))
-            @warn("julia version requirement for package $(pkg.name) not satisfied")
-        end
+    compat = project.compat
+    if haskey(compat, "julia") && !(VERSION in Types.semver_spec(compat["julia"]))
+        @warn("julia version requirement for package $(pkg.name) not satisfied")
     end
-    for (deppkg_name, uuid) in project["deps"]
+    for (deppkg_name, uuid) in project.deps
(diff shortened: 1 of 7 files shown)


00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
:triumph:

00vareladavid:
Well, I'm tempted to just leave it at this. This retains most of the benefits of #857, except that Project is represented by a Dict and not a struct.

Thoughts?

src/Operations.jl
```diff
@@ -1159,21 +1156,22 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
...
-        end
-        if isempty(ctx.env.project["targets"])
-            delete!(ctx.env.project, "targets")
-        end
+    filter!(ctx.env.project["targets"]) do (target, deps)
+        !isempty(filter!(in(deps_names), deps))
+    end
+    #= WHY delete?
+    if isempty(ctx.env.project["targets"])
+        delete!(ctx.env.project, "targets")
```
fredrikekre: Oh, is this #876?

fredrikekre: Although I don't know why it would be empty.

00vareladavid: I don't think so; the "targets" section wouldn't be empty in that case, right?

KristofferC: I think this can just be deleted.

KristofferC: I think I was thinking you could `rm` a Test dependency and when the target section got empty we should remove the section.

fredrikekre: Maybe something accidentally removed it from targets, and then this line removed the targets completely.

StefanKarpinski:
I rather liked having the project as a struct. It means that you know exactly what the meaningful data in a project file is and the type system guarantees that it will have the expected types. I know one can always verify it first and then just assume that the types are correct later but it feels better to have the type of the project object itself encode this knowledge instead of just having it be implicit.

PR #880, opened 2018-11-02
Ignore leading ']' in REPL mode

christopher-dG:
I find that I often make this mistake since Pkg is the only sticky REPL mode (that I know of), and I figure this isn't very intrusive for people who don't mess this up in the first place.

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -926,8 +926,9 @@ function create_mode(repl, main)
     end
 
     repl_keymap = Dict()
+    repl_keymap[']'] = (s, o...) -> isempty(s) || LineEdit.edit_insert(s, ']')
     if shell_mode != nothing
-        repl_keymap[';'] = function (s,o...)
+        repl_keymap[';'] = function (s, o...)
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
I approved this but @KristofferC, @fredrikekre and @00vareladavid don't like it since it's inconsistent with how other modes work—none of them ignore pressing the activation key a second time.

thchr:
I wanted to bring up a possible alternative for what could happen when pressing `]` in the Pkg REPL: namely, exiting back to the standard REPL.
I bring this up because I've seen newcomers instinctively reach for `]` to exit the Pkg REPL, and because I think the symmetry makes sense, especially for a sticky REPL mode (and, as an admittedly small benefit, reduces the need to "traverse" back and forth between `]` and `backspace` for quick one-off-commands in the Pkg REPL).

The same could in principle be done for the Shell REPL via `;`, which also could not have any other interpretation (as far as I know?).
On the other hand, it cannot be done for the Help REPL, since `??` now refers to extended help, so there's a fair consistency argument to be made against this idea on that basis.

Issue #575, opened 2018-08-03
Add "pkg> status SomePackage"


rened:
In large projects with dozens of packages it would be helpful to only show info related to one package e.g. only

```
pkg> st FileIO
  [5789e2e9] ~ FileIO v0.9.1+ #44b3fd8 (git@github.com:contextflow/FileIO.jl.git) ⇒ v1.0.0+ #44b3fd8 (git@github.com:contextflow/FileIO.jl.git)
```

Perhaps also show the packages with are only installed as dependecies because of that particular package (not sure about this, though).

KristofferC:
Dup of https://github.com/JuliaLang/Pkg.jl/issues/381I think.

fredrikekre:
Not necessarily, that issue seems to be about showing metadata like license etc, while this is about showing status about the package as installed? (Basically just a filtered `pkg> status` call IIUC)

jtackm:
As a further suggestion, it would be great to also allow getting information for multiple packages, with syntax similar to ```]add```:

```julia
pkg> st LightGraphs MetaGraphs SimpleWeightedGraphs
```

cossio:
Should this be closed?

PR #881, opened 2018-11-02
pkg> status SomePackage

crstnbr:
I gave #575 a try. There is also this related [discourse thread](https://discourse.julialang.org/t/pkg-repl-mode-filter-status-list/17087).

This PR adds `pkg> status [pkg]`, where `pkg` can be one or multiple package names. It will actually just filter the full `status` list for the given package names.

It's my first Pkg.jl PR. Looking forward to comments!

WIP because I (at least) have to add some tests.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -209,7 +209,7 @@ end
 
 installed() = __installed(PKGMODE_PROJECT)
 function __installed(mode::PackageMode=PKGMODE_MANIFEST)
-    diffs = Display.status(Context(), mode, #=use_as_api=# true)
+    diffs = Display.status(Context(), PackageSpec[], mode=mode, use_as_api=true)
     version_status = Dict{String, Union{VersionNumber,Nothing}}()
     diffs == nothing && return version_status
     for entry in diffs
@@ -500,11 +500,17 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing, kwarg
 end
 
 
-status(mode=PKGMODE_PROJECT) = status(Context(), mode)
-function status(ctx::Context, mode=PKGMODE_PROJECT)
-    Pkg.Display.status(ctx, mode)
-    return
+@deprecate status(mode::PackageMode) status(mode=mode)
+
+status(; mode=PKGMODE_PROJECT) = status(PackageSpec[]; mode=mode)
+status(pkg::Union{String,PackageSpec}; mode=PKGMODE_PROJECT) = status([pkg]; mode=mode)
+status(pkgs::Vector{String}; mode=PKGMODE_PROJECT) = status([check_package_name(pkg) for pkg in pkgs]; mode=mode)
+status(pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT) = status(Context(), pkgs; mode=mode)
+function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
+    Pkg.Display.status(ctx, pkgs, mode=mode)
+    return nothing
 end
+
 
 activate() = (Base.ACTIVE_PROJECT[] = nothing)
 function activate(path::String; shared::Bool=false)
--- a/src/Display.jl
+++ b/src/Display.jl
@@ -29,11 +29,23 @@ function git_file_stream(repo::LibGit2.GitRepo, spec::String; fakeit::Bool=false
     return iob
 end
 
-function status(ctx::Context, mode::PackageMode, use_as_api=false)
+function status(ctx::Context, args::Vector{PackageSpec}=PackageSpec[];
+                mode::PackageMode=PKGMODE_PROJECT, use_as_api=false)
     env = ctx.env
     project₀ = project₁ = env.project
     manifest₀ = manifest₁ = env.manifest
     diff = nothing
+
+    pkgfilter = (diff) -> begin
+        for pkg in args
+            if has_uuid(pkg) ? pkg.uuid == diff.uuid : pkg.name == diff.name
+                return true
+            end
+        end
+        return false
+    end
+    filter_pkgs = length(args) > 0
+
     if !use_as_api
         pkg = ctx.env.pkg
         if pkg !== nothing
@@ -55,6 +67,7 @@ function status(ctx::Context, mode::PackageMode, use_as_api=false)
         m₀ = filter_manifest(in_project(project₀["deps"]), manifest₀)
         m₁ = filter_manifest(in_project(project₁["deps"]), manifest₁)
         diff = manifest_diff(ctx, m₀, m₁)
+        filter_pkgs && filter!(pkgfilter, diff)
         if !use_as_api
             printpkgstyle(ctx, :Status, pathrepr(env.project_file), #=ignore_indent=# true)
             print_diff(ctx, diff, #=status=# true)
@@ -62,6 +75,7 @@ function status(ctx::Context, mode::PackageMode, use_as_api=false)
     end
     if mode == PKGMODE_MANIFEST
         diff = manifest_diff(ctx, manifest₀, manifest₁)
+        filter_pkgs && filter!(pkgfilter, diff)
         if !use_as_api
             printpkgstyle(ctx, :Status, pathrepr(env.manifest_file), #=ignore_indent=# true)
             print_diff(ctx, diff, #=status=# true)
@@ -71,6 +85,7 @@ function status(ctx::Context, mode::PackageMode, use_as_api=false)
         m₀ = filter_manifest(p, manifest₀)
         m₁ = filter_manifest(p, manifest₁)
         c_diff = filter!(x->x.old != x.new, manifest_diff(ctx, m₀, m₁))
+        filter_pkgs && filter!(pkgfilter, c_diff)
         if !isempty(c_diff)
             if !use_as_api
                 printpkgstyle(ctx, :Status, pathrepr(env.manifest_file), #=ignore_indent=# true)
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -601,7 +601,7 @@ end
 
 # TODO set default Display.status keyword: mode = PKGMODE_COMBINED
 do_status!(ctx::APIOptions, args::PkgArguments, api_opts::APIOptions) =
-    Display.status(Context!(ctx), get(api_opts, :mode, PKGMODE_COMBINED))
+    Display.status(Context!(ctx), args, mode=get(api_opts, :mode, PKGMODE_COMBINED))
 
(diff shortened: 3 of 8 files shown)


StefanKarpinski:
Nice! How about `pkg> status [pkg...]` to indicate that you can list more than one?

StefanKarpinski:
Excellent!

crstnbr:
There are two things I'm not sure about
* I couldn't find any `status` related tests. Is it possible to test this?
* `using Pkg; Pkg.status()` currently only takes one optional argument `mode` to basically specify `--project` or `--manifest`. Should `Pkg.status("SomePackage")` also be possible?

fredrikekre:
> using Pkg; Pkg.status() currently only takes one optional argument mode to basically specify --project or --manifest. Should Pkg.status("SomePackage") also be possible?

Yes.

src/REPLMode.jl
```diff
@@ -1264,6 +1264,7 @@ The `startup.jl` file is disabled during precompilation unless julia is started
     :name => "status",
     :short_name => "st",
     :handler => do_status!,
+    :arg_count => 0 => Inf,
```
fredrikekre: Should add ` :arg_parser => (x -> parse_pkg(x)),` here, which will parse the arguments as packages.

src/REPLMode.jl
```diff
@@ -1275,13 +1276,15 @@ The `startup.jl` file is disabled during precompilation unless julia is started
     status
     status [-p|--project]
     status [-m|--manifest]
+    status [pkg...]
```
fredrikekre: Maybe write these as
```
status [pkgs...]
status [-p|--project] [pkgs...]
status [-m|--manifest] [pkgs...]
```
?

src/REPLMode.jl
```diff
@@ -1275,13 +1276,15 @@ The `startup.jl` file is disabled during precompilation unless julia is started
...
 
 Show the status of the current environment. By default, the full contents of
 the project file is summarized, showing what version each package is on and
 how it has changed since the last git commit (if in a git repo), as well as
 any changes to manifest packages not already listed. In `--project` mode, the
 status of the project file is summarized. In `--manifest` mode the output also
-includes the dependencies of explicitly added packages.
+includes the dependencies of explicitly added packages. Package names may be 
+given as arguments to limit the status output to those packages.
```
fredrikekre: Maybe: "If there are any packages listed as arguments the output will be limited to those packages." or something?

src/Display.jl
```diff
@@ -29,11 +29,12 @@ function git_file_stream(repo::LibGit2.GitRepo, spec::String; fakeit::Bool=false
     return iob
 end
 
-function status(ctx::Context, mode::PackageMode, use_as_api=false)
+function status(ctx::Context, mode::PackageMode, args...; use_as_api=false)
```
fredrikekre: make `args` the second argument here

crstnbr: I don't understand. That's not possible is it?

```
julia> f(x..., y) = @show x
ERROR: syntax: invalid ... on non-final argument
```

fredrikekre: dont splat it :)

src/API.jl
```diff
@@ -568,9 +568,10 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing, kwarg
 end
 
 
-status(mode=PKGMODE_PROJECT) = status(Context(), mode)
-function status(ctx::Context, mode=PKGMODE_PROJECT)
-    Pkg.Display.status(ctx, mode)
+status(mode::PackageMode=PKGMODE_PROJECT, pkgs::String...) = status(Context(), mode, pkgs...)
+status(pkgs::String...) = status(PKGMODE_PROJECT, pkgs...)
+function status(ctx::Context, mode::PackageMode=PKGMODE_PROJECT, pkgs::String...)
+    Pkg.Display.status(ctx, Pkg.REPLMode.parse_pkg(String[pkgs...]), mode)
```
crstnbr: Is there a better way here?

crstnbr:
I believe to have included your remarks.

src/API.jl
fredrikekre: Yes, the parsing is just for the REPL, where we parse the input string. This API function should follow the same pattern as the other API functions and be defined something like
```
status(pkg::Union{String,PackageSpec}) = status([pkg])
status(pkgs::Vector{String}) = status([check_package_name(pkg) for pkg in pkgs])
status(pkgs::Vector{PackageSpec}) = status(Context(), pkgs)
function status(ctx::Context, pkgs::Vector{PackageSpec})
    # implementation
end
```

src/Display.jl
```diff
@@ -29,11 +29,13 @@ function git_file_stream(repo::LibGit2.GitRepo, spec::String; fakeit::Bool=false
     return iob
 end
 
-function status(ctx::Context, mode::PackageMode, use_as_api=false)
+function status(ctx::Context, args::Vector{PackageSpec}, mode::PackageMode; use_as_api=false)
     env = ctx.env
     project₀ = project₁ = env.project
     manifest₀ = manifest₁ = env.manifest
     diff = nothing
+    pkgs = getproperty.(args, :name)
```
fredrikekre: If the uuid is given in the packagespec we should filter by that, otherwise name.

src/API.jl
00vareladavid: In addition, REPL Mode flags map to API keyword arguments.
```
status(pkg::Union{String,PackageSpec}; mode=PKGMODE_PROJECT) = status([pkg]; mode=mode)
status(pkgs::Vector{String}; mode=PKGMODE_PROJECT) = status([check_package_name(pkg) for pkg in pkgs]; mode=mode)
status(pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT) = status(Context(), pkgs; mode=mode)
function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
    # implementation
end
```

crstnbr: Alright, I'll make the changes.
(Note, however, that `mode` wasn't a keyword argument before this PR either.)

fredrikekre:
Looking good, maybe you can add some calls to these functions in the test just to make sure they don't error? Don't have to test the output since we don't do that anywhere else, but we should add output testing in the future.

src/API.jl
```diff
@@ -568,12 +568,17 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing, kwarg
...
-    Pkg.Display.status(ctx, mode)
-    return
+@deprecate status(mode::PackageMode) status(mode=mode)
+
+status(; mode=PKGMODE_PROJECT) = status(PackageSpec[]; mode=mode)
+status(pkg::Union{String,PackageSpec}; mode=PKGMODE_PROJECT) = status([pkg]; mode=mode)
+status(pkgs::Vector{String}; mode=PKGMODE_PROJECT) = status([check_package_name(pkg) for pkg in pkgs]; mode=mode)
+status(pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT) = status(Context(), pkgs; mode=mode)
+function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
+    Pkg.Display.status(ctx, pkgs, mode=mode)
```
fredrikekre: This should `return nothing` as before.

test/api.jl
```diff
@@ -42,4 +42,14 @@ include("utils.jl")
     end
 end
 
+@testset "Pkg.status" begin
+    Pkg.status();
+    @test_deprecated Pkg.status(Pkg.Types.PKGMODE_COMBINED)
+    Pkg.status("Random")
+    Pkg.status(["Random", "Dates"])
+    Pkg.status(Pkg.API.check_package_name.(["Random", "Dates"]))
```
fredrikekre: Just `PackageSpec.(["Random", "Dates"])` should work.

test/api.jl
```diff
@@ -42,4 +42,14 @@ include("utils.jl")
     end
 end
 
+@testset "Pkg.status" begin
```
fredrikekre: Maybe we should add the packages we call status on too (`Random` and `Dates`, and maybe swap one of them to `Example` to use a "real" package)? Also, wrap in `temp_pkg_dir() do project_path ... end` as the example above, otherwise this poisons the default env.

test/api.jl
```diff
@@ -42,4 +42,14 @@ include("utils.jl")
...
+    Pkg.status();
+    @test_deprecated Pkg.status(Pkg.Types.PKGMODE_COMBINED)
+    Pkg.status("Random")
+    Pkg.status(["Random", "Dates"])
+    Pkg.status(Pkg.API.check_package_name.(["Random", "Dates"]))
+    Pkg.status(; mode=Pkg.Types.PKGMODE_COMBINED)
+    Pkg.status("Random"; mode=Pkg.Types.PKGMODE_COMBINED)
+end
```
fredrikekre: Maybe also test
```
Pkg.status(PackageSpec(name = "Random", uuid = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"))
```

test/repl.jl
```diff
@@ -909,6 +909,14 @@ end
     end end end
 end
 
+@testset "status" begin
+    pkg"status"
+    pkg"status --project"
+    pkg"status --manifest"
+    pkg"status Random"
```
fredrikekre: Lets also test
```
pkg"status status Random=9a3f8284-a2c9-5f02-9a11-845980a1fd5c
pkg"status status 9a3f8284-a2c9-5f02-9a11-845980a1fd5c
```
just to make sure these work.

test/repl.jl
```diff
@@ -909,6 +909,14 @@ end
     end end end
 end
 
+@testset "status" begin
```
fredrikekre: Wrap this in `temp_pkg_dir` too, as above.

fredrikekre:
I sandboxed the tests and put in a note in the CHANGELOG, hope you don't mind!

bors try

fredrikekre:
bors try

fredrikekre:
third time's the charm
bors try

crstnbr:
> I sandboxed the tests and put in a note in the CHANGELOG, hope you don't mind!

Not at all. It's great that we'll have this in 1.1 soon.

fredrikekre:
`pkg> status --manifest X` should probably also include the dependencies of `X`, otherwise the manifest option to status becomes kind of useless. I also think it makes sense, then `pkg> status X` shows the exact environment that `X` will be run at, which I think is very useful when e.g. reporting a bug to package `X`.

StefanKarpinski:
100% agree with @fredrikekre's analysis of `status -m X` 👍

PR #884, opened 2018-11-05
 fix #711: multiple registries with the same package

fredrikekre:
Don't prompt the user for a choice when the same package name is found in multiple registries if they share the same uuid.

fix #711

PR #887, opened 2018-11-06
filter out unreachable registries when collecting.

fredrikekre:
This filters out identical registries when collecting them. A registry is unreachable if there exist another registry with the same uuid higher up in DEPOT_PATH.

WIP mostly because I don't know how to print this to make sense, some alternatives unreachable ones are greyed out, but not visible here):

1.
```
pkg> registry st
Registry Status 
  [23338594] General (https://github.com/JuliaRegistries/General.git)
             installed in [`~/.julia/registries/General`]
 Unreachable registries:
  [23338594] General (https://github.com/JuliaRegistries/General.git)
             installed in [`~/sandbox/registries/General`]
```
2. 
```
pkg> registry st
Registry Status 
  [23338594] General (https://github.com/JuliaRegistries/General.git)
 Unreachable registries:
  [23338594] General (https://github.com/JuliaRegistries/General.git)
```
3. 
```
pkg> registry st
Registry Status 
  [23338594] General (https://github.com/JuliaRegistries/General.git)
 Unreachable registries:
  [23338594] General (https://github.com/JuliaRegistries/General.git)
             installed in [`~/sandbox/registries/General`]
```

IMO 1. looks kinda messy, but 2. does not give any information about why it is unreachable. 3. is some kind of combination

### DIFF
--- a/src/Registry.jl
+++ b/src/Registry.jl
@@ -73,13 +73,12 @@ Display information about available registries.
 """
 function status()
     regs = Types.collect_registries()
-    regs = unique(r -> r.uuid, regs) # Maybe not?
     Types.printpkgstyle(stdout, Symbol("Registry Status"), "")
     if isempty(regs)
         println("  (no registries found)")
     else
         for reg in regs
-            printstyled(" [$(string(reg.uuid)[1:8])]"; color = :light_black)
+            printstyled("  [$(string(reg.uuid)[1:8])]"; color = :light_black)
             print(" $(reg.name)")
             reg.url === nothing || print(" ($(reg.url))")
             println()
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -931,7 +931,7 @@ function clone_default_registries()
     end
 end
 
-# Return `RegistrySpec`s of each registry in a depot
+# Return `RegistrySpec`s of each unique registry in a depot, sorted by name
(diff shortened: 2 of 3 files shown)


fredrikekre:
Eh, forget the first post; they are supposed to be identical so who cares if we print out the duplicates or not.

fredrikekre:
The test I added in #884 test the fix here, so I added a better test for #884 in this PR.
bors try

fredrikekre:
bors try

StefanKarpinski:
The original intention was to just union over all of them. So any version that's in any copy of a registry counts. Only in cases where there's a conflict about what a version means or what it's compatible with should it matter that there's multiple copies.

StefanKarpinski:
Pkg triage discussion concluded that we should go ahead with something like this. Unioning versions across registries makes sense; the only case where there may be a mismatch is compatibility claims, which are _not_ immutable. For compatibility claims, it would make sense to use the claims from the first registry in depot path which has that version. Note, however that since the same version of the same package can appear in different registries, not just in different copes of the same registry, there's a more general precedence issue. Currently, I think we just give registries in the same depot precedence based on their alphabetical ordering.

StefanKarpinski:
From triage this week: current behavior seems to fine and roughly match what we want except in some corner cases.

Issue #477, opened 2018-07-06
Generate in non-empty folder?


IainNZ:
Would it be possible to add support for generate in a nonempty folder? It seems like the only real issue arises if there is already a Project.toml, but in interactive mode it could be a prompt to override it?

ellisvalentiner:
I'm also interested in this.

@IainNZ I could see this working a few different ways and I'm curious what you (and others) are thinking.

1. Pkg doesn't export functions for manually generating the Project and Manifest files, but obviously already has the ability to do so (see https://github.com/JuliaLang/Pkg.jl/blob/master/src/generate.jl#L26). Pkg could export these functions and allow the user to use them directly when the directory is nonempty. This might be the least amount of work and doesn't change the behavior of `generate`.
2. Alternatively `generate` could prompt the user when the directory is nonempty and/or contains the Project or Manifest files.

I also suppose these aren't mutually exclusive options and that both could be implemented.

00vareladavid:
Same reason for closing as #889. When it comes to `generate`, https://github.com/invenia/PkgTemplates.jl is a lot more feature rich

PR #889, opened 2018-11-06
Generate in existing and non-empty directories

ellisvalentiner:
This is my attempt at #477 – allowing `generate <path>` with an existing and potential non-empty path.

If the path exists and already contains either a Project.toml and/or Manifest.toml file then `generate` does not proceed. However if the path exists and does not contain either of those file it will generate them. If there is already a file `src/<package name>.jl` then it skips generating that also.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,6 @@ keywords = ["package", "management"]
 license = "MIT"
 name = "Pkg"
 version = "1.0.0"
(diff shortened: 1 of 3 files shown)


00vareladavid:
Same as #1213, I think we will keep `generate` as barebones as possible and direct people to [PkgTemplates](https://github.com/invenia/PkgTemplates.jl)

PR #890, opened 2018-11-07
print manifest usage in one shot

KristofferC:


StefanKarpinski:
See also https://github.com/JuliaLang/Pkg.jl/pull/891 (seems a bit more readable).

KristofferC:
Readable but failable ;)

StefanKarpinski:
Weird, I tested locally but I guess it must have been using the built-in Pkg instead of the custom one.

PR #894, opened 2018-11-10
Structured Manifest Data

00vareladavid:
In the spirit of #857, structured manifest data.
As discussed earlier, a Manifest should really be `Dict{UUID, PackageEntry}`.

Analogous to Project data, `Manifest!` takes in a raw TOML dict and returns a `Manifest` data structure.
`destructure` works in the reverse direction.

I have tried to keep the changes minimal, except where the new data structure provides a clearly better alternative (e.g. `manifest_by_uuid` is no longer needed).

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -95,8 +95,8 @@ function collect_fixed!(ctx::Context, pkgs::Vector{PackageSpec}, uuid_to_name::D
     uuid_to_pkg = Dict{UUID,PackageSpec}()
     for pkg in pkgs
         local path
-        info = manifest_info(ctx.env, pkg.uuid)
-        if pkg.special_action == PKGSPEC_FREED && !haskey(info, "pinned")
+        entry = manifest_info(ctx.env, pkg.uuid)
+        if pkg.special_action == PKGSPEC_FREED && !entry.pinned
             continue
         elseif pkg.special_action == PKGSPEC_DEVELOPED
             @assert pkg.path !== nothing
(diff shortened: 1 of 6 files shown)


src/Types.jl
```diff
@@ -275,6 +275,18 @@ Base.@kwdef mutable struct Project
...
+Base.@kwdef mutable struct PackageEntry
+    name::Union{String, Nothing} = nothing
+    version::Union{String, Nothing} = nothing
+    path::Union{String, Nothing} = nothing
+    git_tree_sha::Union{SHA1, Nothing} = nothing
+    pinned::Bool = false
+    repo_url::Union{String, Nothing} = nothing
+    repo_rev::Union{String, Nothing} = nothing
+    deps::Dict{String,UUID} = Dict{String,UUID}()
+end
```
StefanKarpinski: Open question: should we have an `other::Dict{String,Any}` field here to preserve other TOML entries? Argument in favor is forwards compatibility: if we introduce fields that are meaningful to future versions of Julia/Pkg we would generally want older versions of Pkg to ignore and preserve those fields.

00vareladavid:
bors try

src/Operations.jl
```diff
@@ -757,18 +745,20 @@ function update_manifest(ctx::Context, pkg::PackageSpec, hash::Union{SHA1, Nothi
             end
             for dep_pkg in dep_pkgs
                 dep_pkg.name == "julia" && continue
-                deps[dep_pkg.name] = string(dep_pkg.uuid)
+                deps[dep_pkg.uuid] = dep_pkg.name
```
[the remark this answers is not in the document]
00vareladavid: No, that was accidental.

src/Types.jl
```diff
@@ -527,30 +551,53 @@ function read_package(f::String)
...
 
-function read_manifest(io::IO)
-    manifest = TOML.parse(io)
-    for (name, infos) in manifest, info in infos
-        haskey(info, "deps") || continue
-        info["deps"] isa AbstractVector || continue
-        for dep in info["deps"]
-            length(manifest[dep]) == 1 ||
-                error("ambiguious dependency for $name: $dep")
+function Manifest!(raw::Dict)::Manifest
```
StefanKarpinski: Why the `!` at the end of the name?

00vareladavid: :+1: No good reason

00vareladavid:
bors try

00vareladavid:
Good to merge from my side.

Simple benchmark
```
@time for i in 1:10
    Pkg.add("Example")
    Pkg.rm("Example")
end
```

- structured_manifest: `5.905365 seconds (12.31 M allocations: 603.799 MiB, 12.30% gc time)`
- upstream/master: `5.573539 seconds (12.40 M allocations: 608.124 MiB, 10.11% gc time)`

PR #897, opened 2018-11-12
Use permutedims instead of LinearAlgebra.transpose (via Base), fix #896.

### LINKED ISSUE #896: Pkg requires LinearAlgebra implicitly via Base.transpose
From using `Pkg` without loading `LinearAlgebra`:
```
ERROR: MethodError: no method matching transpose(::BitArray{2})
```

fredrikekre:
bors try

### DIFF
--- a/src/GraphType.jl
+++ b/src/GraphType.jl
@@ -329,7 +329,7 @@ mutable struct Graph
                 adjdict[p0][p1] = j1
 
                 bm = trues(spp[p1], spp[p0])
-                bmt = copy(transpose(bm))
+                bmt = permutedims(bm)
 
                 push!(gmsk[p0], bm)
                 push!(gmsk[p1], bmt)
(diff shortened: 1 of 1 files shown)


KristofferC:
Would be good with just a simple benchmark here e.g. `@time add Plots` a few times.

fredrikekre:
I timed `Pkg.add("Plots")` without needing to download anything (which should put as much relative time as possible in e.g. resolve) and there are no difference:

**master**
```
  0.121418 seconds
  0.103115 seconds
  0.132629 seconds
  0.102283 seconds
```
**PR**
```
  0.122515 seconds
  0.109008 seconds
  0.111462 seconds
  0.141039 seconds
```

and compare to a situation where we also need to download this is even less of a problem.

fredrikekre:
If this becomes a real problem we should move the optimized `transpose(::BitArray)` to `permutedims` and have `transpose` call that.
bors r+

PR #902, opened 2018-11-14
add two missing GC.@preserve calls

StefanKarpinski:
I believe these are necessary. @vtjnash, can you take a quick look and verify?

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -529,11 +529,13 @@ function install_git(
         tree isa LibGit2.GitTree ||
             error("$name: git object $(string(hash)) should be a tree, not $(typeof(tree))")
         mkpath(version_path)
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
bors try

StefanKarpinski:
Well, I have no idea what this error means:
```
ERROR: LoadError: LoadError: syntax: ssavalue with no def
...
in expression starting at /home/travis/build/JuliaLang/Pkg.jl/src/Types.jl:678
```

StefanKarpinski:
bors try

KristofferC:
Does just copy pasting this function into its own file reproduce. Could file a bug upstream then.

StefanKarpinski:
It does; I'll do that.

StefanKarpinski:
bors try

PR #903, opened 2018-11-14
fix unclosed file open call

StefanKarpinski:
@00vareladavid: this may have been the cause of all the Windows problems since this opened a project file and then never closed it. I'm going to bors try it and if it passes, can we undo the various hacks to call GC and see if this fixes the underlying issue?


StefanKarpinski:
bors try

fredrikekre:
closing in favor of #905

PR #905, opened 2018-11-14
Define read_project for IO types, fix #904.

### LINKED ISSUE #904: MethodError: no method matching read_project(::Base.GenericIOBuffer{Array{UInt8,1}})
```
(Pkg) pkg> st
Project Pkg v1.0.0
ERROR: MethodError: no method matching read_project(::Base.GenericIOBuffer{Array{UInt8,1}})
Closest candidates are:
  read_project(::String) at /home/fredrik/dev/Pkg/src/Types.jl:506
```
likely caused by https://github.com/JuliaLang/Pkg.jl/pull/857

fredrikekre:
Define read_project for IO types, fix #904.
Remove unnecessary GC.gc() call on Windows.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -502,16 +502,25 @@ function Project(raw::Dict)
     return project
 end
 
-function read_project(filename::String)
-    !isfile(filename) && return Project()
-    raw = nothing
+function _read_project(io::IO)
+    raw, err = nothing, nothing
     try
-        raw = TOML.parse(open(filename))
+        raw = TOML.parse(io)
     catch err
         err isa TOML.ParserError || rethrow()
-        pkgerror("Could not parse project file at `$filename`: $(err.msg)")
     end
-    return Project(raw)
+    return raw, err
+end
+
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Looks like it supersedes https://github.com/JuliaLang/Pkg.jl/pull/903 as well. Feel free to close that in favor of this PR.

00vareladavid:
Thanks for taking care of this!

Perhaps also remove the [gc call](https://github.com/JuliaLang/Pkg.jl/blob/03244bf4d0be113a4373d6393f49d70a0640ddac/src/Types.jl#L772) to verify that this indeed was the source of the Windows issues?

fredrikekre:
bors try

fredrikekre:
16 + 17 min on AV here. Maybe the slowdown was due to the open file handle?

StefanKarpinski:
> Maybe the slowdown was due to the open file handle?

It seems unlikely that open file handles would have that effect. Maybe all the extra GC calls?

Issue #892, opened 2018-11-10
]add checks github when package is already installed


Orbots:
This is a real problem in an offline environment.
```
julia> LOAD_PATH
4-element Array{String,1}:
 "@"                                                
 "@v#.#"                                            
 "@stdlib/dd/dept/software/3ps/julia/julia_packages"
 "."                                                

julia> using UnicodePlots
[ Info: Recompiling stale cache file /dd/dept/software/3ps/julia/julia_packages/compiled/v1.0/UnicodePlots/Ctj9q.ji for UnicodePlots [b8865327-cd53-5732-bb35-84acbb429228]

julia> UnicodePlots.lineplot(sin,1:10)
           ┌────────────────────────────────────────┐
         1 │⠀⠀⣀⣀⢄⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⢀⠔⢣⠀⠀⠀⠀⠀⠀⠀⠀│ sin(x)
           │⠉⠉⠀⠀⠈⡆⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⠀⡔⠁⠀⠀⢣⠀⠀⠀⠀⠀⠀⠀│
....

(Foo) pkg> add UnicodePlots
  Updating registry at `/dd/dept/software/3ps/julia/julia_packages/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`                  
_************* WAITS FOR A LONG TIME!!!! ****************_
┌ Warning: Some registries failed to update:
│     — /dd/dept/software/3ps/julia/julia_packages/registries/General — failed to fetch from repo
└ @ Pkg.API /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:157
 Resolving package versions...
  Updating `/dd/dept/software/users/mewert/swdevl/3ps/julia/1.0.2/Foo/Project.toml`
  [b8865327] + UnicodePlots v0.3.1
  Updating `/dd/dept/software/users/mewert/swdevl/3ps/julia/1.0.2/Foo/Manifest.toml`
  [864edb3b] + DataStructures v0.14.0
  [e1d29d7a] + Missings v0.3.1
....

julia> DEPOT_PATH
2-element Array{String,1}:
[...]

PR #906, opened 2018-11-15
Only clone default registries if there are no registries installed.

fredrikekre:
Possible "fix" for #892; if you don't want registries to update don't put them in `DEPOT_PATH[1]`.

bors try


fredrikekre:
bors try

PR #908, opened 2018-11-15
Normalize Gitlab urls

morris25:
Allows gitlab urls to be normalized by a separate protocol.

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -66,35 +66,24 @@ function transfer_progress(progress::Ptr{LibGit2.TransferProgress}, p::Any)
 end
 
 const GIT_REGEX =
-    r"^(?:(?:git|ssh|https)(?:://|@))?(?:[\w\.\+\-:]+@)?([^/:]+)[:/](.+?)(?:\.git)?$"i
-const GITHUB_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
-const GITLAB_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
-const FALLBACK_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
+    r"^(?:(?<proto>git|ssh|https)://)?(?:[\w\.\+\-:]+@)?(?<hostname>.+?)(?(<proto>)/|:)(?<path>.+?)(?:\.git)?$"
+const GIT_PROTOCOL = Dict{AbstractString, Union{Nothing, AbstractString}}()
 
-function setprotocol!(proto::Union{Nothing, AbstractString}=nothing, site::Symbol=:github)
-    if site == :github
-        GITHUB_PROTOCOL[] = proto
-    elseif site == :gitlab
-        GITLAB_PROTOCOL[] = proto
-    else
-        FALLBACK_PROTOCOL[] = proto
-    end
+function setprotocol!(
+    proto::Union{Nothing, AbstractString}=nothing,
+    site::AbstractString="github.com"
+)
+    GIT_PROTOCOL[site] = proto
 end
 
-# TODO: extend this to more urls
 function normalize_url(url::AbstractString)
     m = match(GIT_REGEX, url)
     m === nothing && return url
 
-    host = m.captures[1]
-    path = "$(m.captures[2]).git"
-    proto = if occursin("github", host)
-        GITHUB_PROTOCOL[]
-    elseif occursin("gitlab", host)
-        GITLAB_PROTOCOL[]
-    else
-        FALLBACK_PROTOCOL[]
-    end
+    host = m[:hostname]
+    path = "$(m[:path]).git"
+
+    proto = get(GIT_PROTOCOL, host, nothing)
 
     if proto === nothing
         url
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -67,21 +67,19 @@ end
 
 const GIT_REGEX =
     r"^(?:(?<proto>git|ssh|https)://)?(?:[\w\.\+\-:]+@)?(?<hostname>.+?)(?(<proto>)/|:)(?<path>.+?)(?:\.git)?$"
-const GIT_PROTOCOL = Dict{String, Union{Nothing, String}}()
+const GIT_PROTOCOLS = Dict{String, Union{Nothing, String}}()
+const GIT_USERS = Dict{String, Union{Nothing, String}}()
 
-@deprecate setprotocol!(proto::Nothing=nothing) setprotocol!("github.com", proto) false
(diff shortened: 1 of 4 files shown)


src/GitTools.jl
```diff
@@ -68,20 +68,45 @@ end
 
 const GITHUB_REGEX =
     r"^(?:git@|git://|https://(?:[\w\.\+\-]+@)?)github.com[:/](([^/].+)/(.+?))(?:\.git)?$"i
-const GIT_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
+const GITLAB_REGEX =
+    r"^(?:git@|git://|https://(?:[\w\.\+\-]+@)?)gitlab\.([^/:]+)[:/]((?:[^/].+)/(?:.+?))(?:\.git)?$"i
```
omus: Probably the approach should be to support any hostname. A more generalized regex can be seen [here](
https://github.com/JuliaLang/julia/blob/604046c1e1063d393261c07fc66a4f825b244697/stdlib/LibGit2/src/utils.jl#L9-L29).

omus: We don't need to go that with the regex though.

src/GitTools.jl
```diff
@@ -65,22 +65,43 @@ function transfer_progress(progress::Ptr{LibGit2.TransferProgress}, p::Any)
     return Cint(0)
 end
 
+const GIT_REGEX =
+    r"^(?:(?:git|ssh|https)(?:://|@))?(?:[\w\.\+\-:]+@)?([^/:]+)[:/](.+?)(?:\.git)?$"i
```
omus: I think it makes more sense to use named groups at this point. Also it is weird that `git@` is captured by the scheme part of the regex. I would recommend this:
```julia
r"^(?:(?<scheme>git|ssh|https)://)?(?:[\w\.\+\-:]+@)?(?<hostname>[^/:]+)(?(<scheme>)/|:)(?<path>.+?)(?:\.git)?$"
```

src/GitTools.jl
```diff
@@ -65,22 +65,43 @@ function transfer_progress(progress::Ptr{LibGit2.TransferProgress}, p::Any)
...
+const GITHUB_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
+const GITLAB_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
+const FALLBACK_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
```
omus: Can't we just use a single `GIT_PROTOCOL` here?

morris25: We could, but people might need to use different protocols for different sites. This at least leaves them the option to do that.

src/GitTools.jl
```diff
@@ -66,35 +66,24 @@ function transfer_progress(progress::Ptr{LibGit2.TransferProgress}, p::Any)
...
-        GITHUB_PROTOCOL[]
-    elseif occursin("gitlab", host)
-        GITLAB_PROTOCOL[]
-    else
-        FALLBACK_PROTOCOL[]
-    end
+    host = m[:hostname]
+    path = "$(m[:path]).git"
+
+    proto = get(GIT_PROTOCOL, host, nothing)
```
StefanKarpinski: Probably unnecessary but maybe `lowercase(host)`? Domain names are defined to be case insensitive.

src/GitTools.jl
```diff
@@ -66,35 +66,24 @@ function transfer_progress(progress::Ptr{LibGit2.TransferProgress}, p::Any)
 end
 
 const GIT_REGEX =
-    r"^(?:(?:git|ssh|https)(?:://|@))?(?:[\w\.\+\-:]+@)?([^/:]+)[:/](.+?)(?:\.git)?$"i
-const GITHUB_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
-const GITLAB_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
-const FALLBACK_PROTOCOL = Ref{Union{String, Nothing}}(nothing)
+    r"^(?:(?<proto>git|ssh|https)://)?(?:[\w\.\+\-:]+@)?(?<hostname>.+?)(?(<proto>)/|:)(?<path>.+?)(?:\.git)?$"
+const GIT_PROTOCOL = Dict{AbstractString, Union{Nothing, AbstractString}}()
```
StefanKarpinski: Just go with `String` instead of `AbstractString` here. The `Dict` data structure should convert for you.

src/GitTools.jl
```diff
@@ -66,35 +66,24 @@ function transfer_progress(progress::Ptr{LibGit2.TransferProgress}, p::Any)
...
-function setprotocol!(proto::Union{Nothing, AbstractString}=nothing, site::Symbol=:github)
-    if site == :github
-        GITHUB_PROTOCOL[] = proto
-    elseif site == :gitlab
-        GITLAB_PROTOCOL[] = proto
-    else
-        FALLBACK_PROTOCOL[] = proto
-    end
+function setprotocol!(
+    proto::Union{Nothing, AbstractString}=nothing,
```
StefanKarpinski: Is the idea here that `setprotocol!()` unsets the protocol? It's a little weird that it defaults to unsetting the protocol for `github.com`. It seems like `setprotocol!(hostname, [protocol])` might be more natural then? That way omitting the protocol to unset it works the same for every hostname. Also, given that all this does is stick a value into a dict, maybe just expose the dict directly?

morris25: I definitely like the `setprotocol!(hostname, [protocol])` signature better. I feel like keeping `GitTools.setprotocol!` is a bit neater than just exposing the dict.

omus:
bors try

fredrikekre:
Would be great to get rid of the deprecations.

StefanKarpinski:
Ok, so I feel bad because I think I may have suggested the current API in this PR, but it seems like using keyword arguments might be better. In other words something like this signature:
```jl
setprotocol!(;
    protocol::Union{Nothing, AbstractString} = nothing,
    domain::Union{Nothing, AbstractString} = nothing,
    user::Union{Nothing, AbstractString} = (protocol == "ssh" ? "git" : nothing),
)
```
That doesn't quite feel orthogonal either though. Argh, this is a bit tricky...

Also, did we have `setprotocol!` in Julia 1.0?

ararslan:
> Also, did we have `setprotocol!` in Julia 1.0?

Yes

morris25:
@StefanKarpinski Is there anything I should do?

StefanKarpinski:
I think this looks good. Hopefully we'll use a centralized server with a consistent URL scheme instead in the future so that this becomes unnecessary. We'll discuss on the next Pkg triage call and make a decision on this.

StefanKarpinski:
Crap, this should have been squashed.

KristofferC:
:( the negative with Bors

PR #909, opened 2018-11-16
Properly escape TOML fields in `generate`

00vareladavid:
Someone on slack hit a call to `hex` which was supposed to be deprecated.
Root cause of the issue was that `generate` did not properly escape the `authors` field.

I think just piping through `TOML.print` should do the trick.

Note I couldn't figure out how to keep the empty `deps` field when generating a project, so I just avoid printing it. Is it just there for aesthetic purposes? We don't care to preserve it when `rm`ing all dependencies...

### DIFF
--- a/ext/TOML/src/parser.jl
+++ b/ext/TOML/src/parser.jl
@@ -585,7 +585,8 @@ function escape(p::Parser, st::Int, multiline::Bool)
                 NONE()
             end
         else
-            error(p, st, position(p), "unknown string escape: `$("\\x"*hex(ch,2))`")
+            escape_str = "\\x"*string(UInt32(ch), base=16, pad=2)
+            error(p, st, position(p), "unknown string escape: `$escape_str`")
             NONE()
         end
     end
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -46,20 +46,16 @@ function project(pkg::String, dir::String; preview::Bool)
         end
     end
 
-    authorstr = "[\"$name " * (email == nothing ? "" : "<$email>") * "\"]"
+    authors = ["$name " * (email == nothing ? "" : "<$email>")]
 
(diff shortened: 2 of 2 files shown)


00vareladavid:
bors try

ext/TOML/src/parser.jl
```diff
@@ -585,7 +585,7 @@ function escape(p::Parser, st::Int, multiline::Bool)
                 NONE()
             end
         else
-            error(p, st, position(p), "unknown string escape: `$("\\x"*hex(ch,2))`")
+            error(p, st, position(p), "unknown string escape: `$("\\x"*string(UInt32(ch), base=16, pad=2))`")
```
StefanKarpinski: Maybe assign to a variable on the previous line? This has gotten a bit verbose for splicing inline.

src/generate.jl
```diff
@@ -46,20 +46,16 @@ function project(pkg::String, dir::String; preview::Bool)
...
-
-            [deps]
-            """
-        )
+        toml = Dict("authors" => authors,
+                    "name" => pkg,
+                    "uuid" => string(uuid),
+                    "version" => "0.1.0",
+                    )
+        TOML.print(io, toml, sorted=true, by=key -> (Types.project_key_order(key), key))
```
StefanKarpinski: Is this a related change?

00vareladavid: Yes, the reason the user hit the bug in the first place was that we were not properly escaping the authors field on `generate`. I think their git name had a backslash in it, which is not properly escaped if we just splice it in directly.

00vareladavid: I can make them separate commits if necessary

StefanKarpinski: No, that's cool, makes sense. I was just wondering about the connection.

PR #914, opened 2018-11-20
Refactor `Types.handle_repos_develop`

### LINKED ISSUE #668: Don't clone the repo if we know the name in develop
In `handle_repose_develop!` we can check already here: https://github.com/JuliaLang/Pkg.jl/blob/ec413a28fca07c64c466523b737ba979ccc87269/src/Types.jl#L534 if the `pkg.name` field is set and look if the directory already exists. i.e. if we did `pkg> dev Example` and we already have `Example` in `devdir`.

00vareladavid:
Thoughts? Some light refactoring + dividing into functions for readability + a few comments.

This would close #668.

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -84,6 +84,9 @@ function normalize_url(url::AbstractString)
     end
 end
 
+ensure_clone(target_path, url; kwargs...) =
+    ispath(target_path) ? LibGit2.GitRepo(target_path) : GitTools.clone(url, target_path; kwargs...)
+
 function clone(url, source_path; header=nothing, kwargs...)
     @assert !isdir(source_path) || isempty(readdir(source_path))
     url = normalize_url(url)
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -504,9 +504,9 @@ function install_git(
             clones_dir = joinpath(depots1(), "clones")
             ispath(clones_dir) || mkpath(clones_dir)
             repo_path = joinpath(clones_dir, string(uuid))
-            repo = ispath(repo_path) ? LibGit2.GitRepo(repo_path) : begin
-                GitTools.clone(urls[1], repo_path; isbare=true, header = "[$uuid] $name from $(urls[1])", credentials=creds)
-            end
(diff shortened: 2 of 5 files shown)


src/GitTools.jl
```diff
@@ -84,6 +84,9 @@ function normalize_url(url::AbstractString)
     end
 end
 
+ensure_clone(target_path, url; kwargs...) =
```
fredrikekre: shouldn't the kwargs be used here?

src/Operations.jl
```diff
@@ -504,9 +504,9 @@ function install_git(
             clones_dir = joinpath(depots1(), "clones")
             ispath(clones_dir) || mkpath(clones_dir)
             repo_path = joinpath(clones_dir, string(uuid))
-            repo = ispath(repo_path) ? LibGit2.GitRepo(repo_path) : begin
-                GitTools.clone(urls[1], repo_path; isbare=true, header = "[$uuid] $name from $(urls[1])", credentials=creds)
-            end
+            repo = GitTools.ensure_clone(repo_path, urls[1]; isbare=true,
+                                         header = "[$uuid] $name from $(urls[1])",
+                                         credentials=creds)
```
fredrikekre: the kwargs here are unused but ensure_clone.

src/GitTools.jl
00vareladavid: Yes. I forgot to splice it back in

00vareladavid:
I broke `handle_repos_develop` down further. `handle_repos_develop` is organized in terms of how packages are specified: by local path, by name, or by url. All the git complexity is now in `Types.fresh_clone`, I took special care in not messing with any of this logic.

Factoring in this way also prevents `Pkg` from reaching out to the network until absolutely necessary. Previously, we could pull down a git clone, parse its project file, and find out that target directory already exists!  

I also added more tests for `develop`'s API.

00vareladavid:
bors try

00vareladavid:
I would appreciate if someone could take a quick look :grin: this is supposed to be more readable, after all

PR #917, opened 2018-11-21
Add cut_body_after to bors.toml.

fredrikekre:
`cut_body_after` lets us control what ends up in the merge commit message, whats before `---` will be in there, and whats after will be removed.

---

This should thus not go in the commit message, but maybe bors uses the toml file on master so this might not work for this PR.

bors try


fredrikekre:
`---` seems like a sensible choice here, since that also creates a visible markdown line on github.

Issue #620, opened 2018-08-11
`pkg> git` command that just calls git?


StefanKarpinski:
We keep getting issues like https://github.com/JuliaLang/Pkg.jl/issues/610 and indeed, I often find myself accidentally calling `pkg> git diff` and such. You can, of course, `cd` into your active project and use `shell> git ...` to do various git operations in the active project but we perhaps we should just make `pkg> git ...` do that for you.

KristofferC:
Hmm, how is #610 related to this?

StefanKarpinski:
> Oh, but of course, @fredrikekre just reminded me that the copy in ~/.julia/packages isn't a full git clone. Womp womp.

If we just allowed calling `git` inside of `pkg>` then I think that kind of issue would be alleviated since it would be a full git client.

KristofferC:
I'm still confused. #610 was about someone wanting to dev a package while being offline, which was closed because that is fundamentally impossible since we havent downloaded the git repo.

How does a git command help with that?

StefanKarpinski:
Ok, forget the connection to #610. It's common to want to do git commands inside of the current active project. What about a `pkg> git ...` command that just call `git` in the right place?

KristofferC:
Tbh, I rather not add the extra complexity of this when one can almost literally just pres `;` and start running `git`.

PR #923, opened 2018-11-25
Add support for a `git` interface in the REPL

00vareladavid:
#620 seems like a neat idea. I think it makes sense to access `git` directly since Pkg relies on `git` so heavily.

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -80,7 +80,7 @@ end
 @enum(CommandKind, CMD_HELP, CMD_RM, CMD_ADD, CMD_DEVELOP, CMD_UP,
                    CMD_STATUS, CMD_TEST, CMD_GC, CMD_BUILD, CMD_PIN,
                    CMD_FREE, CMD_GENERATE, CMD_RESOLVE, CMD_PRECOMPILE,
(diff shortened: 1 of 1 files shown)


KristofferC:
I think this is a bit of a feature creep. Shell mode with `git` should work well enough imo.

00vareladavid:
The main issue with shell mode is that it does not automatically track the active environment. A user has to keep adjusting the current working directory in order for git to operate on the correct project. This is annoying when developing interactively.

I agree that it would be a feature creep for the API, but this is purely for interactive usage. This comes down to a judgement of how friendly the Pkg REPL should be.

00vareladavid:
FWIW I'm not completely sold on this idea yet. But it's nice to have an implementation to get a feel for how this affects your workflow :grin:

00vareladavid:
:thinking: What about a pkg-shell mode? Hitting some key would switch you to regular shell-mode, except that we wrap in a `cd(dirname(Base.active_project()))`? I think this would prevent any kind of scope creep since we just offload the work to shell-mode (or whatever its called..) and it would retain a big improvement in interactive usage

00vareladavid:
Upon more reflection, you guys are probably right. There are more pressing issues to deal with anyway

PR #926, opened 2018-11-27
Fix activate in the case that the active env is corrupted

00vareladavid:
This fixes `activate` when the currently active environment is corrupted.

- In the case where we give `activate` a path, we should just set `Base.ACTIVE_PROJECT[]` without doing any extra work.
- In the case where we give `activate` the name of a `develop`ed dependency, we have to be careful when calling `EnvCache` because it will fail if the current environment contains a corrupted project/manifest file.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -493,26 +493,37 @@ end
 
 
 activate() = (Base.ACTIVE_PROJECT[] = nothing)
+function _activate_dep(dep_name::AbstractString)
+    Base.active_project() === nothing && return
+    env = nothing
+    try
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
> We could even prevent users from activateing corrupted environments in the first place.

I really like this idea.

KristofferC:
Yeah, I think verifying and aborting makes sense if the env is corrupted.

StefanKarpinski:
What about the initial active env? When do we verify that and what do we do if it's corrupt?

PR #927, opened 2018-11-27
Expand tilde in local path completions

### LINKED ISSUE #882: Expand ~ in Pkg REPL completions
The Julia REPL replaces `~` with `home/user` in tab-completions. Would be nice to do this in the Pkg REPL too for `activate`, `dev` and `add`.

Looks like it would be enough to handle it here: https://github.com/JuliaLang/Pkg.jl/blob/f65ef8fb53498bca0c8a63c479d610cb9d996beb/src/REPLMode.jl#L690

Reference for the Julia REPL implementation https://github.com/JuliaLang/julia/blob/099979931ce1e666e15ba090cf257bd4fbae8ede/stdlib/REPL/src/REPLCompletions.jl#L271

00vareladavid:
Should resolve #882.

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -740,6 +740,10 @@ function complete_activate(options, partial, i1, i2)
 end
 
 function complete_local_dir(s, i1, i2)
+    if !isempty(s) && s[1] == '~'
+        return String[expanduser(s)], i1:i2, true
+    end
+
     cmp = REPL.REPLCompletions.complete_path(s, i2)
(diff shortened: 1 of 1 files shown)


KristofferC:
The reason we need to expand it is because the path completion we use from `REPLCompletions` need an expanded path? Can't we just expand before we ask for path completion (while leaving the `~` in the REPL)?

StefanKarpinski:
Triage: since we want `~` paths to be valid input, we should not expand them during tab completion.

fredrikekre:
This seems to work. We can revisit to not print the expansion in the Pkg REPL later #247 
bors r+

PR #932, opened 2018-11-27
minor changes to suport registrator work

StefanKarpinski:


StefanKarpinski:
bors try

ext/TOML/src/TOML.jl
```diff
@@ -3,6 +3,8 @@
 module TOML
     using Dates
 
+    const TYPE = Union{AbstractDict,AbstractArray,AbstractString,DateTime}
```
StefanKarpinski: @00vareladavid: add `Bool` here; also maybe move it into the `print.jl` file so it's close to the other related code. This kind of seems like a bad place to have put this definition in hind sight.

ext/TOML/src/print.jl
```diff
@@ -18,31 +18,28 @@ function printkey(io::IO, keys::Vector{String})
...
-        Base.print(io, value)
     end
+    Base.print(io, "]")
 end
+printvalue(io::IO, value::AbstractDict; sorted=false) =
+    _print(io, value, sorted=sorted)
+printvalue(io::IO, value::DateTime; sorted=false) =
+    Base.print(io, Dates.format(value, "YYYY-mm-ddTHH:MM:SS.sssZ"))
+printvalue(io::IO, value; sorted=false) =
+    Base.print(io, "\"$(escape_string(string(value)))\"")
```
StefanKarpinski: @00vareladavid: add method for `Bool` here that doesn't escape the output.

Issue #872, opened 2018-10-31
Version is ignored when URL is specified in PackageSpec


ararslan:
Example:
```julia
julia> using Pkg

julia> Pkg.add(PackageSpec(name="Example", url="https://github.com/JuliaLang/Example.jl", version=v"0.3.0"))
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
   Cloning git-repo `https://github.com/JuliaLang/Example.jl`
  Updating git-repo `https://github.com/JuliaLang/Example.jl`
 Resolving package versions...
  Updating `~/.julia/environments/v1.0/Project.toml`
  [7876af07] + Example v0.5.1+ #master (https://github.com/JuliaLang/Example.jl)
  Updating `~/.julia/environments/v1.0/Manifest.toml`
  [7876af07] + Example v0.5.1+ #master (https://github.com/JuliaLang/Example.jl)

(v1.0) pkg> status
    Status `~/.julia/environments/v1.0/Project.toml`
...
  [7876af07] Example v0.5.1+ #master (https://github.com/JuliaLang/Example.jl)
...
```
Noticed by @omus in https://github.com/JuliaLang/Pkg.jl/pull/799#issuecomment-434823318.

KristofferC:
What should happen?

ararslan:
The version is provided, shouldn't it install that version?

KristofferC:
Since versions only make sense for registered packages, the fact that you are providing an URL likely means that is what you want to use. It would make sense to warn or error when a version keyword is given and the same time as the URL though.

ararslan:
> Since versions only make sense for registered packages, the fact that you are providing an URL likely means that is what you want to use.

If the URL is a git repo and the version is a git tag, would it not make sense to just check out that tag after cloning? You're still using the given URL, but you're also making use of the other information provided. That seems like sensible behavior to me.

fredrikekre:
> If the URL is a git repo and the version is a git tag, would it not make sense to just check out that tag after cloning?

Isn't this what `Pkg.add(PackageSpec(..., rev = "v0.3.0"))` is for?

ararslan:
That is a weird distinction...

KristofferC:
A version is not simply a git tag. Then it would be mutable, anyone can rename / delete git tags. A version is identified with a content hash.


There are two options.

1.) Error / warn.
[...]

PR #936, opened 2018-11-29
Refactor types

00vareladavid:
1. Refactor types for consistency across codebase (e.g. sometimes we do `isempty(name)` and other times we do `name === nothing`, `PackageEntry` should have a `GitRepo` field, ...)
2. Split `PackageSpec` constructor into an internal constructor and a user-facing constructor (inspired by #872). It really is two different use cases: one is "create this object and set these fields", the other needs to do hand holding and error checking.

It was difficult to do one without the other, so that's why I included them together.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -11,6 +11,7 @@ import LibGit2
 import ..depots, ..depots1, ..logdir, ..devdir
 import ..Operations, ..Display, ..GitTools, ..Pkg, ..UPDATED_REGISTRY_THIS_SESSION
 using ..Types, ..TOML
+using Pkg.Types: VersionTypes
 
 
 preview_info() = printstyled("───── Preview mode ─────\n"; color=Base.info_color(), bold=true)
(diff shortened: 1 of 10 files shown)


00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
All tests pass and I double checked manually. Any objections?

00vareladavid:
bors try

src/API.jl
```diff
@@ -541,4 +537,24 @@ end
 
 setprotocol!(proto::Union{Nothing, AbstractString}=nothing) = GitTools.setprotocol!(proto)
 
+# API constructor
+function Package(;name::Union{Nothing,AbstractString} = nothing,
```
fredrikekre: This should be `PackageSpec`.

rofinn:
FWIW, this is a pretty breaking change because the exported `PackageSpec` is now an alias to a function rather than a type, so any code that dispatched on a `PackageSpec` will now need to be fully qualified. I think simply modifying the existing `Pkg.Types` constructors would have been easier. Alternatively, if the logic is overly complicated for a constructor then maybe consider a regular function `packagespec`?

PR #938, opened 2018-11-29
better error message when we cant find uuid in manifest

KristofferC:
This should not really happen but it seems to happen anyway, so at least give a better error message for it.


fredrikekre:
Hmm, I think we need the same check at https://github.com/JuliaLang/Pkg.jl/blob/193e494c22f7ce8b5956829286c6212133edcbf1/src/Operations.jl#L1022-L1023 where we still get an error like this.

PR #939, opened 2018-11-29
prune manifest before instantiating

KristofferC:


StefanKarpinski:
It's a bit weird to have a manifest in an unpruned state though... how did it happen in the first place?

KristofferC:
People are doing strange things and get weird errors. 

https://discourse.julialang.org/t/travis-doesnt-tell-me-where-an-error-occurs/18160/4

and discussion in slack #pkg-usage today.

Perhaps just prune in the beginning of this function?

fredrikekre:
Yea, lets prune directly

KristofferC:
Alternatively, this should be caught in the manifest loader and immediately fail some verification check?

00vareladavid:
I'm working on more thorough checking for both the manifest and project files. Project checking is almost done, and I'll do manifest after that. In case that makes a difference...

fredrikekre:
> fail some verification check?

This seems not very user friendly. How do you get out of such a problematic Manifest? Can't we just have a warning that some packages are orphaned and will not be instantiated?

00vareladavid:
perhaps have the option to run some recovery code? :thinking:

00vareladavid:
Agreed. Maybe we can just `@info` (or somehow log) in the validation phase that we detected some weird state and that we are recovering.

We could potentially lose data if we do recover automatically though. Should that be a issue?

fredrikekre:
For this particular case it would just ignore some parts of the manifest, and not actually modify the files, but it would be nice to somehow let Pkg correct the files, such that you don't get the warning the next time you instantiate the same manifest.

KristofferC:
`instantiate` does not write back the pruned data so no data loss there at least.

I think we should just not accept malformed input. Who knows what was intended and what a "recovery" should be.

00vareladavid:
That seems like the best option. Users should be using the API to create environments

StefanKarpinski:
Triage conclusion: the job of `instantiate` is just to install all the things in the manifest and build them in a dependency compatible order; it should not try to sync up or even look at project file in any way.

KristofferC:
Done

StefanKarpinski:
I thought this was the opposite of what we decided but sure 🤷‍♂️

KristofferC:
Ergh, I misremembered then :/. I don't think it will matter though.

StefanKarpinski:
Yeah, this is fine.

PR #940, opened 2018-11-29
[WIP] Stricter project file validation

00vareladavid:
Almost done, but still a WIP. Just throwing it up early in case there is feedback.

I'm planning to do something similar to the manifest and perhaps also checking for consistency between project and manifest files.


00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
I double checked this PR and I'm fairly content with it. The tests are reassuring.

Note that these checks focus on typing the string values and on checking the graph structure of the manifest. I think this is sufficient for a first approximation, but it might be beneficial to do checks at the node level in the near future. (e.g. if `entry.path !== nothing` then we don't expect the tree-hash field to be populated.)

PR #941, opened 2018-11-29
Two small logical fixes.

fredrikekre:


src/Types.jl
```diff
@@ -1363,7 +1363,7 @@ function find_registered!(env::EnvCache,
         end
     end
     # if there's still nothing to look for, return early
-    isempty(names) && isempty(uuids) && return
+    !force && isempty(names) && isempty(uuids) && return
```
fredrikekre: I am not 100% sure on this one, but the same check happens on L1344 above, and I think the `force` option is meant to go past this here and read the registry?

src/Types.jl
```diff
@@ -1478,7 +1478,7 @@ end
 # Find package by UUID in the manifest file
 # TODO this needs a better name now?
 function manifest_info(env::EnvCache, uuid::UUID)::Union{PackageEntry,Nothing}
-    uuid in values(env.uuids) || find_registered!(env, [uuid])
+    any(uuids -> uuid in uuids, values(env.uuids)) || find_registered!(env, [uuid])
```
fredrikekre: This was never true before, since `eltype(values(env.uuids))` is `Vector{UUID}` which is never equal to a `UUID`.

fredrikekre:
bors try

src/Types.jl
StefanKarpinski: It seems like pulling out the `force` option entirely might be best.

src/Types.jl
StefanKarpinski: I think that `env.uuids` was previously a `Dict{String,UUID}` but it's not anymore so this is now never true.

src/Types.jl
```diff
@@ -1390,11 +1389,11 @@ function find_registered!(env::EnvCache,
...
 find_registered!(env::EnvCache, uuids::Vector{UUID}; force::Bool=false)::Nothing =
-    find_registered!(env, String[], uuids, force=force)
+    find_registered!(env, String[], uuids)
 
-# Lookup all packages in project & manifest files
-find_registered!(env::EnvCache)::Nothing =
-    find_registered!(env, String[], UUID[], force=true)
+# # Lookup all packages in project & manifest files
+# find_registered!(env::EnvCache)::Nothing =
+#     find_registered!(env, String[], UUID[])
```
StefanKarpinski: Just delete, don't comment out.

fredrikekre: Yea, was just trying it out, but seems CI is happy so will clean up.

PR #947, opened 2018-12-03
Add basedocs for the base julia manual.

fredrikekre:
This snippet will be included in Julia's manual, essentially just the getting started section, and a pointer to the full manual.


fredrikekre:
After this we can bump Pkg in Julia base, just taking the head of master, and then people can try it out with julia master, and then we update Julia base to the actual release 1.1 Pkg commit.

PR #952, opened 2018-12-06
Add Julia 1.0.1 and 1.0.2 as recognized versions

ararslan:
Otherwise using these specific versions when specifying bounds in METADATA doesn't work and you end up with 1.1.

Issue identified (incredibly quickly, I might add) by @KristofferC

### DIFF
--- a/bin/loadmeta.jl
+++ b/bin/loadmeta.jl
@@ -95,6 +95,8 @@ const JULIA_VERSIONS = Dict(
     v"0.6.2"  => Version("d386e40c17d43b79fc89d3e579fc04547241787c"),
     v"0.7.0"  => Version("a4cb80f3edcf8cea00bd9660e3b65f544f41462f"),
     v"1.0.0"  => Version("5d4eaca0c9fa3d555c79dbacdccb9169fdf64b65"),
+    v"1.0.1"  => Version("0d713926f85dfa3e4e0962215b909b8e47e94f48"),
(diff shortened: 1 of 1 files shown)


KristofferC:
I vaguely remember something about the .toml files having to be updated manually, because of the line 

https://github.com/JuliaLang/Pkg.jl/blob/a17e84b1705d6f9e5c2489734b39423a1cc7ca1f/bin/gitmeta.jl#L64

Pkg won't do it itself. Perhaps @StefanKarpinski knows.

fredrikekre:
I also had to add a check: `pkg == "julia" && continue`. This PR + 01619c751c2affe4ddfb5f29ed8c4be652d97d0b seems to work and generate a reasonable registry diff.

StefanKarpinski:
> I also had to add a check: `pkg == "julia" && continue`. This PR + [01619c7](https://github.com/JuliaLang/Pkg.jl/commit/01619c751c2affe4ddfb5f29ed8c4be652d97d0b) seems to work and generate a reasonable registry diff.

I'm pretty sure that part is what caused the bad caps on Julia versions.

fredrikekre:
Ok, sorry. I got some key errors for package `julia` and thought that `julia` somehow was ending up as a package in the registry.

StefanKarpinski:
No problem, it's certainly not your fault. I was hurried because of the breakage, should have reviewed more carefully. I've added the SHA1 maps manually in https://github.com/JuliaLang/Pkg.jl/pull/959.

PR #953, opened 2018-12-06
Fetch all tagged Julia versions during General deployment

ararslan:
This way we don't need to keep a manual list, plus it ensures that the list stays up to date. This would make #952 unnecessary.


bin/update.jl
```diff
@@ -3,6 +3,13 @@
 
 pushfirst!(LOAD_PATH, joinpath(@__DIR__, "..", "ext"))
 
+# TODO: use Sys.STDLIBDIR instead once implemented
```
fredrikekre: `Sys.STDLIB` should do exactly this?

bin/loadmeta.jl
```diff
@@ -61,42 +61,17 @@ function load_versions(dir::String)
...
+for line in eachline(`git -C $juliadir ls-remote --tags origin`)
+    global JULIA_VERSIONS
+    # Lines are in the form 'COMMITSHA\trefs/tags/TAG'
+    sha, ref = split(line, '\t')
+    _, _, tag = split(ref, '/')
+    occursin(Base.VERSION_REGEX, tag) || continue
+    ver = VersionNumber(tag)
+    JULIA_VERSIONS[ver] = Version(sha)
+end
+JULIA_VERSIONS[v"1.1.0"] = Version("6ef1d76c50a39c0ce68b4e42948a1499c1551415") # dummy commit, not actual 1.1
```
StefanKarpinski: If you do this first then when the real version is released, the correct tree hash will replace it.

ararslan:
Comments addressed. Is this looking better now?

StefanKarpinski:
Looks pretty good, I'll have to check it out and try it on the system where the sync is running and make sure the results look sane.

bin/loadmeta.jl
```diff
@@ -61,42 +61,19 @@ function load_versions(dir::String)
...
+    global JULIA_VERSIONS
+    # Lines are in the form 'COMMITSHA\trefs/tags/TAG'
+    sha, ref = split(line, '\t')
+    _, _, tag = split(ref, '/')
+    occursin(Base.VERSION_REGEX, tag) || continue
+    ver = VersionNumber(tag)
+    JULIA_VERSIONS[ver] = Version(sha)
+end
+if !haskey(JULIA_VERSIONS, v"1.1.0") # Keep this up to date with the most recent dev version
+    JULIA_VERSIONS[v"1.1.0"] = Version("6ef1d76c50a39c0ce68b4e42948a1499c1551415") # dummy commit, not actual 1.1
```
StefanKarpinski: Why not just put this before the `for` loop? Then it will just get overwritten once a real `1.1.0` exists.

ararslan: Good idea

StefanKarpinski:
For me this causes `julia bin/update.jl` to fail:
```jl
ERROR: LoadError: LoadError: BoundsError: attempt to access 0-element Array{SubString{String},1} at index [3]
Stacktrace:
 [1] getindex(::Array{SubString{String},1}, ::Int64) at ./array.jl:731
 [2] top-level scope at /home/stefankarpinski/Pkg.jl/bin/genstdlib.jl:24
 [3] include at ./boot.jl:317 [inlined]
 [4] include_relative(::Module, ::String) at ./loading.jl:1038
 [5] include(::Module, ::String) at ./sysimg.jl:29
 [6] include(::String) at ./client.jl:388
 [7] top-level scope at none:0
 [8] include at ./boot.jl:317 [inlined]
 [9] include_relative(::Module, ::String) at ./loading.jl:1038
 [10] include(::Module, ::String) at ./sysimg.jl:29
 [11] exec_options(::Base.JLOptions) at ./client.jl:229
 [12] _start() at ./client.jl:421
in expression starting at /home/stefankarpinski/Pkg.jl/bin/genstdlib.jl:12
in expression starting at /home/stefankarpinski/Pkg.jl/bin/update.jl:12
```
I haven't had time to investigate why. Are you able to run `bin/update.jl` locally? I'll push recent git metadata so that it doesn't require cloning too much stuff.

KristofferC:
See https://github.com/JuliaLang/Pkg.jl/pull/952#issuecomment-445037816

ararslan:
I've seen that failure happen when `juliadir` was not a Git repo. I'll try it locally and get back to you.

StefanKarpinski:
I guess it might need to update my Julia on that machine. I'll try that and report back.

ararslan:
The code assumes that it's running from a source build of Julia and that `Sys.STDLIB` points to the stdlib directory in the source clone. I just tried it in a Docker image and ran into that error because I was using the official Julia binaries.

StefanKarpinski:
I'm running it using a source build of Julia but it's a bit old. Tried rebuilding but that didn't work so I deleted it and recloned julia and am rebuilding from scratch. Unfortunately, registries are not updating while I'm doing this 😂

StefanKarpinski:
Nope, same error on master Julia, freshly built from source.

ararslan:
Hm, okay. I'm not sure what the problem is then. I've definitely seen that error when using a trivial modification of the code in this repo (not with these changes) to deploy a private registry.

StefanKarpinski:
... and now it's broken on master too. Great.

ararslan:
Pkg not being in the repo has caused problems for private registry deployment. Does this change help master?
```diff
--- a/bin/genstdlib.jl
+++ b/bin/genstdlib.jl
@@ -21,7 +21,11 @@ for pkg in readdir(stdlibdir)
     isfile(project_file) || continue
     project = TOML.parsefile(project_file)
     stdlib_uuids[pkg] = project["uuid"]
-    version_file = joinpath(stdlibdir, "$pkg.version")
+    if isfile(joinpath(juliadir, "stdlib", "$pkg.version"))
+        version_file = joinpath(juliadir, "stdlib", "$pkg.version")
+    else
+        version_file = joinpath(stdlibdir, "$pkg.version")
+    end
     if isfile(version_file)
         r = Regex("^\\s*$(pkg)_SHA1\\s*=\\s*(\\S+)\\s*\$", "im")
         m = match(r, read(version_file, String))
```

fredrikekre:
The problem is that from Julia 1.0.2 we symlink each stdlib and not the full stdlib directory. Here is a patch that I think works:
```diff
diff --git a/bin/genstdlib.jl b/bin/genstdlib.jl
index 64c417a..67808f7 100644
--- a/bin/genstdlib.jl
+++ b/bin/genstdlib.jl
@@ -14,14 +14,15 @@ for pkg in readdir(Sys.STDLIB)
     isfile(project_file) || continue
     project = TOML.parsefile(project_file)
     stdlib_uuids[pkg] = project["uuid"]
-    version_file = joinpath(Sys.STDLIB, "$pkg.version")
+    version_file = joinpath(Sys.STDLIB, "..", "..", "..", "..", "..", "stdlib", "$pkg.version")
     if isfile(version_file)
         r = Regex("^\\s*$(pkg)_SHA1\\s*=\\s*(\\S+)\\s*\$", "im")
         m = match(r, read(version_file, String))
         m === nothing && error("expected PKG_SHA1 in $version_file")
         stdlib_trees[pkg] = m.captures[1]
     else
-        stdlib_trees[pkg] = split(readchomp(`git -C $juliadir ls-tree HEAD -- stdlib/$pkg`))[3]
+        dir = dirname(realpath(joinpath(Sys.STDLIB, pkg)))
+        stdlib_trees[pkg] = split(readchomp(`git -C $dir ls-tree HEAD -- $pkg`))[3]
     end
     stdlib_deps[pkg] = String[]
     haskey(project, "deps") || continue
diff --git a/bin/loadmeta.jl b/bin/loadmeta.jl
index a3b5299..b64ac18 100644
--- a/bin/loadmeta.jl
+++ b/bin/loadmeta.jl
@@ -64,7 +64,7 @@ end
 const JULIA_VERSIONS = Dict{VersionNumber,Version}()
 # Note: This is a dummy commit that will get overwritten by the loop below once 1.1 is released
 JULIA_VERSIONS[v"1.1.0"] = Version("6ef1d76c50a39c0ce68b4e42948a1499c1551415")
-for line in eachline(`git -C $juliadir ls-remote --tags origin`)
+for line in eachline(`git -C $(dirname(dirname(Sys.STDLIB))) ls-remote --tags origin`)
     global JULIA_VERSIONS
     # Lines are in the form 'COMMITSHA\trefs/tags/TAG'
     sha, ref = split(line, '\t')
diff --git a/bin/update.jl b/bin/update.jl
index 1b55f31..17c04d3 100644
--- a/bin/update.jl
+++ b/bin/update.jl
@@ -3,9 +3,6 @@
 
 pushfirst!(LOAD_PATH, joinpath(@__DIR__, "..", "ext"))
 
-# Used by both loadmeta and genstdlib, assumed to be a git repo
-juliadir = dirname(Sys.STDLIB)
-
 include("loadmeta.jl")
 include("utils.jl")
 include("gitmeta.jl")
```

StefanKarpinski:
Unfortunately, the setup that ran the sync script is now in a broken state and none of these patches apply. I've tried checking out various combinations of Pkg and Julia and now can't get this back in a working state. I don't really have time to work on it until Monday, so unfortunately there will not be any new versions of any Julia packages until I get around to fixing this on Monday.

ararslan:
Is it the same error noted above?

StefanKarpinski:
@fredrikekre, what branch is that patch supposed to apply to?

fredrikekre:
Oh, this comment was supposed to be on this PR: https://github.com/JuliaLang/Pkg.jl/pull/952#issuecomment-445529880

StefanKarpinski:
I fixed this so that the update script works and is running regularly again.

bors r+

fredrikekre:
squash? (bors won't do it :( )

StefanKarpinski:
How do I cancel the merge part?

fredrikekre:
Seems like some packages got capped? E.g. JSON v0.18 can't be installed, hence the test failure.

fredrikekre:
Also fails julia master builds

fredrikekre:
should not be needed anymore

PR #954, opened 2018-12-07
fix #915: status -m Package prints status of Package's deps

### LINKED ISSUE #915: status -m Package
Opening an issue to track this: https://github.com/JuliaLang/Pkg.jl/pull/881#issuecomment-436942479 which would be nice to have with the 1.1 release.
cc @crstnbr since you wrote the original PR.

fredrikekre:
Print the status of pkg's deps in manifest/combined mode, i.e.

    pkg> st -m pkg

fix #915


fredrikekre:
bors try

PR #959, opened 2018-12-10
just changes needed to deal with STDLIB changes

StefanKarpinski:


StefanKarpinski:
bors try

StefanKarpinski:
bors try

bin/gitmeta.jl
```diff
@@ -72,7 +72,7 @@ function gitmeta(pkgs::Dict{String,Package})
         for (ver, v) in sort!(collect(p.versions), by=first, rev=true)
             haskey(d[uuid], v.sha1) &&
             ((v"0.7" ∉ v.requires["julia"].versions &&
-             v"1.0" ∉ v.requires["julia"].versions) ||
+              v"1.0" ∉ v.requires["julia"].versions) ||
```
ararslan: Does this (and line 108) need to handle versions other than 0.7.0 and 1.0.0, or is that already being handled e.g. by truncating out patches?

ararslan: Never mind, answered my own question: https://github.com/JuliaLang/Pkg.jl/blob/master/bin/loadmeta.jl#L199. Sorry for the noise.

StefanKarpinski: All the stuff with specific versions is pretty nasty and getting out of control as we have more and more post-0.6 versions. We really need to switch over the whole process...

ararslan: We could probably do `any(ver->ver >= v"0.7", v.requires["julia"].versions)` here and on line 108, which would automate those bits a little more.

StefanKarpinski: Yeah, that would be great if you want to include that in a PR.

ararslan: Can do

PR #960, opened 2018-12-10
Functionality for generating remote URLs from filepaths

oxinabox:
In 0.6 julia had this cool feature,
(that I mostly used from IJulia)
where you could get a hyperlink to a method defination.

Code for that was here
https://github.com/JuliaLang/julia/blob/80820acf3ec3cd4314210f667f7ec63d775a1d88/base/methodshow.jl#L207

It still works in 1.0, for packages that have been `dev`ed.
But it was most useful to me for packages that I wasn't developing,
but wanted to check the source code for.

This code here works on packages that are not checked out.
I've not yet added an `include` for it as it is pretty WIP.
(I can't work out how to test it without mocking...)

I think it belong in `Pkg`, because it uses so many things that are not part of the `Pkg` public API,
that putting it elsewhere is asking for it to break.
It is also basically very useful.

I'ld really appreciate code review, and recommendations on how to test it right now.

### DIFF
--- a/src/file_urls.jl
+++ b/src/file_urls.jl
@@ -1,0 +1,102 @@
+# TODO remove these imports once this code is actually linked installed
+using Pkg
+using LibGit2
+
+
+
+
+
+"""
+    get_url
+
+Makes a best attempt at generating a URL from version control,
+Falls back to providing a local file URL.
+"""
+function get_url(_file, _module)
+    try
+        return giturl_dev(_file)
+    catch err1
+        @debug "Handling as dev'd package failed. Falling back." exception=err1
+        try
+            return giturl_release(_file, _module)
+        catch err2
+            @debug "Handling as added package failed. Falling back. Again." exception=err2
+            return Base.fileurl(_file)
+        end
+    end
+end
+
+
+function get_url(method::Method)
+    _file, _line = functionloc(method)
+    _module = method.module
+    url = get_url(_file, _module)
+    if !startswith(url, "file:")
+        url *= string("#L", _line)
+    end
+    return url
+end
+
+
+
+function giturl_dev(_file)
+    _dir = dirname(_file)
+    LibGit2.with(LibGit2.GitRepoExt(_dir)) do repo
+        LibGit2.with(LibGit2.GitConfig(repo)) do cfg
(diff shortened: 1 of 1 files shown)


KristofferC:
So from Pkg you need the functionality to go from UUID -> Url right?

Also, what is the plan for tying this together with `Base.url`?

oxinabox:
> So from Pkg you need the functionality to go from UUID -> Url right?

yes.


> Also, what is the plan for tying this together with Base.url?

I guess we use the same hack that is currently there, 
which tries and loads LibGit2, but instead make it try and load Pkg ?

Current code:
https://github.com/JuliaLang/julia/blob/80820acf3ec3cd4314210f667f7ec63d775a1d88/base/methodshow.jl#L223

oxinabox:
I actually have more general schemse for it, such as making a Logger that includes clickable link's to where the `@warn` etc was thrown.

00vareladavid:
#989 introduces `Operations.load_urls` which is essentially `load_urls : UUID -> URL`. This is only for registered packages. Would wrapping `load_urls` be sufficient for this functionality?

oxinabox:
Maybe.
 - Does it get the URL for the current tagged release, that is installed?
 - Does it get the SHA tree based URL for the current commit that is `dev`ed ?

00vareladavid:
So I finally got around to working on the introspection interface I mentioned a few weeks back. Perhaps you can take a look at #1086 and see if that is sufficient? It does support the URL thing (this seems generally useful), but I'm not sure what information you need from `dev`ed packages.

oxinabox:
I will add looking at #1086 t o me TODO thanks!


>  but I'm not sure what information you need from `dev`ed packages.

A `dev`ed  Pkg should return a URL like
https://github.com/JuliaLang/Pkg.jl/blob/13a84c054d01b0a2ecd7c5e7542050f89498ca3e/src/API.jl

00vareladavid:
@oxinabox I think this can be closed. `Pkg.dependencies` returns pretty much all the information that `Pkg` has easy access to. Anything else can be determined from outside of Pkg just as easily. Feel free to reopen if you disagree

oxinabox:
Thanks @00vareladavid

PR #970, opened 2018-12-20
Fix repo handling

### LINKED ISSUE #614: Pkg add with url overwrites existing files
Reprduce with `add url` twice.

### LINKED ISSUE #951: `Pkg.add` with explicit URL ignores pin
```
pkg> add Example@0.4.0
pkg> pin Example
pkg> add https://github.com/JuliaLang/Example.jl
```

`pin` is not respected in this case.

00vareladavid:
I started off just trying to break down `handle_repos_add` (it was a beast!) but the refactoring allowed two bugs to be fixed along the way.

Fixes #951. In the first stage (`resolve_repo_add!`) we make sure certain fields of the `PackageSpec` are filled out (`name`, `uuid`, `repo.url`, `repo.rev`). This means we can have more consistent logic for the rest of the operation. In particular, once we have the UUID, it is simple to check if the package is pinned.

Fixes #614. I added a test which simply makes sure that the `mtime` for the source is the same after an `add` by URL.

In terms of refactoring, moving all git-heavy operations into their own functions helps to clarify the `add`-specific logic. Also, `handle_repos_add!` was being called from many places. It turns out those calls just needed some more limited functionality which is now in `instantiate_pkg_repo!`.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -41,7 +41,7 @@ function add_or_develop(ctx::Context, pkgs::Vector{PackageSpec}; mode::Symbol, s
     ctx.preview && preview_info()
     new_git = mode == :develop ?
         handle_repos_develop!(ctx, pkgs, shared) :
-        handle_repos_add!(ctx, pkgs; upgrade_or_add=true)
+        handle_repos_add!(ctx, pkgs)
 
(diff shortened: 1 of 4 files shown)


00vareladavid:
bors try

PR #971, opened 2018-12-22
Free should resolve against the manifest

00vareladavid:
Resolving against the manifest first ensures that a package's UUID is the intended one


00vareladavid:
bors try

PR #976, opened 2018-12-26
Backports for 1.1.2

fredrikekre:
bors try


00vareladavid:
For future reference, how would this be pushed upstream? On this side we just bump the version in `Project.toml` and tag it using the GitHub AP, correct?

fredrikekre:
Yea, tag a release and bump the commit in JuliaLang/julia like https://github.com/JuliaLang/julia/pull/30521

00vareladavid:
Ok, thanks for explaining. I appreciate it.

One more question: how do you determine the checksums?

PR #982, opened 2019-01-03
 bug fixes to bin/gitmeta.jl

StefanKarpinski:
Fixes https://github.com/JuliaRegistries/General/issues/23.


bin/gitmeta.jl
```diff
@@ -105,7 +106,9 @@ function gitmeta(pkgs::Dict{String,Package})
             git_tree_hash = string(LibGit2.GitHash(git_tree))
             d[uuid][v.sha1] = git_tree_hash
             # scan for stdlib dependencies
-            (v"0.7" in v.requires["julia"].versions || v"1.0" in v.requires["julia"].versions) || continue
+            v"0.7" in v.requires["julia"].versions ||
```
KristofferC: Maybe this should be updated to `any(v -> v >= v"0.7", v.requires["julia"].versions)` or something?

StefanKarpinski: Sure, that could work too. Just wanted to get it fixed quickly since it was blocking packages working.

PR #986, opened 2019-01-11
Pkg REPL: sort commands within each super-spec

fredrikekre:
Sort commands within each super-spec instead of globally when displaying help.

---

This will display:
```
activate
[...]
update
registry-add
[...]
registry-update
```
instead of 
```
activate
[...]
preview
registry-add
[...]
registry-update
remove
[...]
update
```
i.e. group all `package` commands together first, and then all `registry` commands.

bors try

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -779,16 +779,16 @@ end
 
 function canonical_names()
     names = String[]
+    # add "package" commands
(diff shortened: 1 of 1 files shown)


fredrikekre:
bors try

fredrikekre:
bors try

PR #991, opened 2019-01-14
doc: fix numbering of chapters, fix #944.

### LINKED ISSUE #944: Numbering for chapters in docs
I noticed there was no chapter/section 8, it went from 7 to 9 in the devs version of the docs.

fredrikekre:
bors r+

### DIFF
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -1,4 +1,4 @@
-# [**10.** API Reference](@id API-Reference)
+# [**9.** API Reference](@id API-Reference)
 
 This section describes the function interface, or "API mode"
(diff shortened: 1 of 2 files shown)


fredrikekre:
> Seems good but why are we numbering chapters manually like this?

Because there is no other way with Documenter currently. We have few enough chapters that I think this is sustainable.

> Do we need numbers?

Not sure we "need" it, but it looks much better IMO. Compare the structure and "general feeling" of [Pkg](https://julialang.github.io/Pkg.jl/v1/) vs [Julias manual](https://docs.julialang.org/en/v1/).

StefanKarpinski:
Fair enough. The number is quite nice.

PR #998, opened 2019-01-17
doc: include docstrings for Pkg REPL commands in the manual.

fredrikekre:


00vareladavid:
Nice! I agree with the way the commands were separated i.e. `activate`, `status`, `resolve`, ... should be subcommands of `env`(or something along those lines). Hopefully this makes clear that there are 3 main objects that the package manager manages: packages, environment, registries.

Issue #629, opened 2018-08-12
clone is not entirely replaced by add


timholy:
If you use `Pkg.clone` you get a "legacy CI script reasons" warning, but it's not so simple.
I had to do this:
```
Pkg.add(Pkg.REPLMode.parse_package("https://github.com/timholy/HeaderREPLs.jl"; add_or_develop=true));
```
If you pass a URL to `Pkg.add` directly you end up triggering
```julia
julia> Pkg.API.check_package_name("https://github.com/timholy/HeaderREPLs.jl")
ERROR: https://github.com/timholy/HeaderREPLs.jl is not a valid packagename
Stacktrace:
 [1] pkgerror(::String) at /home/tim/src/julia-1.0/usr/share/julia/stdlib/v1.0/Pkg/src/Types.jl:120
 [2] check_package_name(::String) at /home/tim/src/julia-1.0/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:22
```

KristofferC:
What you have do to is

```
using Pkg
Pkg.add(PackageSpec(url = "https://github.com/timholy/HeaderREPLs.jl"))
```

There is some help about it

```
help?> Pkg.add
  Pkg.add(pkg::Union{String, Vector{String})
  Pkg.add(pkg::Union{PackageSpec, Vector{PackageSpec}})

  Add a package to the current project. This package will be available using the import and using keywords in the Julia REPL and if the current project is a package, also inside that
  package.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  Pkg.add("Example") # Add a package from registry
  Pkg.add(PackageSpec(name="Example", version="0.3")) # Specify version
  Pkg.add(PackageSpec(url="https://github.com/JuliaLang/Example.jl", rev="master")) # From url
  Pkg.add(PackageSpec(url="/remote/mycompany/juliapackages/OurPackage"))` # From path (has to be a gitrepo)

  See also PackageSpec.
```

but indeed, the deprecation warning should be improved.

00vareladavid:
Is there a reason to not use the REPL given your use case?

Affie:
Is there an equivalent function for `Pkg.clone(url::String, name::String)`
My use case is bitbucket only allows lower case names in the url. Therefore, I specify the name explicitly.
eg. `Pkg.clone("..../mypackage.jl.git", "MyPackage")`
Currently Pkg.add uses `mypackage` for the package name instead of `MyPackage`.

[...]

PR #1000, opened 2019-01-17
Added hint to use PackageSpec if a url- or path-like string is passed to add/develop functions

BioTurboNick:
Addresses #629 in part

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -24,7 +24,7 @@ function find_installed(name::String, uuid::UUID, sha1::SHA1)
 end
 
 function load_versions(path::String; include_yanked = false)
-    toml = parse_toml(path, "Versions.toml")
+    toml = parse_toml(path, "Versions.toml"; fakeit=true)
     d = Dict{VersionNumber, SHA1}(
             VersionNumber(ver) => SHA1(info["git-tree-sha1"]) for (ver, info) in toml
                 if !get(info, "yanked", false) || include_yanked)


src/API.jl
```diff
@@ -18,16 +18,34 @@ preview_info() = printstyled("───── Preview mode ─────\n"; c
 
 include("generate.jl")
 
+islikelyurl(x) = any(occursin.(['\\','/'], x))
```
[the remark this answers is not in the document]
BioTurboNick: Thank you, though I was intending to be permissive for local paths, just as the url and path keyword arguments are interchangeable for PackageSpec().

src/Pkg.jl
```diff
@@ -226,10 +229,10 @@ If `pkg` is given as a local path, the package at that path will be tracked.
 Pkg.develop("Example")
 
 # By url
-Pkg.develop(PackageSpec(url="https://github.com/JuliaLang/Compat.jl"))
+Pkg.develop(url="https://github.com/JuliaLang/Compat.jl")
 
 # By path
-Pkg.develop(PackageSpec(path="MyJuliaPackages/Package.jl"))
+Pkg.develop(url="MyJuliaPackages/Package.jl")
```
StefanKarpinski: This should be a `path` keyword, not a `url` keyword.

StefanKarpinski:
To attempt to answer my own question, my impression is that you wanted `Pkg.add(string)` to "do what I mean" (DWIM) in the fashion that `pkg> add ...` does. However, the API should be explicit and precise about intention rather than DWIMy. That's why having a `islikelyurl` function in here is concerning. We shouldn't be guessing what is or is not a URL or a path in API calls, it should be fully explicit in the way the API is called.

BioTurboNick:
@StefanKarpinski - I think your suggestion makes sense. Still figuring out the best use of the dispatch system. I can make that change.

With respect to the other changes:
- The "islikelyurl" is only used to decide whether to show the hint or not when the normal add("packagepathorurl") fails. I can place it inside `check_package_name` function to ensure that it is clear that it should not be used elsewhere.
- I found that trying to use "PackageSpec()" in some places was failing but "Package()" was working. Those can be reverted if something was just funky on my system.
- The `rm` change is something I tried but only partially reverted; I'll fix that so there is no change from current.
- The `show` change is a bug I discovered and fixed. I can move that into a separate branch and PR.

BioTurboNick:
The one issue with your proposal is that `kwargs...` is currently passed through to `Context!()`, and I don't know what misbehavior might occur if it was directed to `PackageSpec()` instead. Seems like it would be safer to capture particular keyword arguments and pass the rest along?

00vareladavid:
I like the change in error message! I think its strictly more helpful. I think perhaps you misread, Stefan? `islikeurl` does not affect the "happy path": it is only called when we have already decided to error out.

I am wary of changing the API without more time to consider the implications. I'm ok with the current verbosity of the API: users should really be using the REPL mode if they want a terse syntax.

I am against any API changes at this point and definitely for more useful error messages. Perhaps the error message could be even more explicit, along the lines of: "X is not a valid package name. X does appear to be a URL, did you mean `Pkg.add(PackgeSpec(;url=X))`?" This seems to be sufficient to get users out of the confusing spot you experienced @BioTurboNick?

src/API.jl
```diff
@@ -20,14 +20,21 @@ include("generate.jl")
...
-         pkgerror("$x is not a valid packagename")
+        islikelyurl(x) = any(occursin.(['\\','/'], x))
+        message = "$x is not a valid packagename."
+        if islikelyurl(x)
+            message *= " Did you mean url= or path= \"$x\"?"
+        end
+        pkgerror(message)
     end
-    return PackageSpec(x)
+    return Package(x)
```
00vareladavid: As described in the comments, `API.Package` is only meant as the API constructor. So this should be `PackageSpec(;name=x)`

BioTurboNick:
I'd be willing to go for the hint as you suggest, @00vareladavid, at least for now :-).

How do you see people interacting with Pkg with unregistered packages from scripts and notebooks, which don't have a normal prompt? Options as of now are Pkg.add(PackageSpec(url="url")) or the much-less-discoverable pkg"add url", yes?

Above you said PackageSpec() was meant to be internal though, so should regular users be constructing PackageSpec() objects? Or perhaps I just misunderstand what that means in this context.

src/API.jl
```diff
@@ -20,14 +20,21 @@ include("generate.jl")
 
 function check_package_name(x::AbstractString)
     if !(occursin(Pkg.REPLMode.name_re, x))
-         pkgerror("$x is not a valid packagename")
+        islikelyurl(x) = any(occursin.(['\\','/'], x))
+        message = "$x is not a valid packagename."
+        if islikelyurl(x)
```
StefanKarpinski: It seems like it would be just as clear to replace `islikelyurl(x)` with its definition from two lines above, no? If you feel it needs more explanation, you can put a comment along the lines of `# not a valid package name, possible path or URL`.

00vareladavid:
>  I'd be willing to go for the hint as you suggest

Awesome, I do think our error messages have a lot of room for improvement.

> How do you see people interacting with Pkg with unregistered packages

In my mind, `Pkg.add(PackageSpec(;url="url"))` would be the idiomatic way to do it. With a user-side wrapper if that operation is used heavily.

> Above you said PackageSpec() was meant to be internal

The constructor thing is a little messy. `Pkg.API.Package` (the API constructor) gets mapped to `Pkg.PackageSpec`: that is the one end-users will be exposed to. `Pkg.Types.PackageSpec` is internal and does not get exported.

src/API.jl
BioTurboNick: I'm personally a fan of the self-documenting code philosophy, but I'll defer to you both.

StefanKarpinski: Taking that to the logical conclusion would mean that any code that requires a comment should be factored out into a function with the comment as its name...

Project.toml
```diff
@@ -3,7 +3,6 @@ keywords = ["package", "management"]
 license = "MIT"
 name = "Pkg"
 version = "1.1.0"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
00vareladavid: Also, the UUID should not be removed. You have to remove it to load package when developing, but you should not commit this change.

src/API.jl
BioTurboNick: You're not wrong. It's a balance of course, but I find aiming for that makes it easier to understand later what I was doing (coming from C#). Reserves comments for explaining why choices were made, don't have to keep comments and code in sync as much, makes refactoring a bit easier. IMO.

BioTurboNick:
That should pretty much do it, unless my minifunction is vetoed :-)

src/API.jl
```diff
@@ -18,16 +18,23 @@ preview_info() = printstyled("───── Preview mode ─────\n"; c
...
-    return PackageSpec(x)
+    return PackageSpec(; name = x)
 end
 
 add_or_develop(pkg::Union{AbstractString, PackageSpec}; kwargs...) = add_or_develop([pkg]; kwargs...)
-add_or_develop(pkgs::Vector{<:AbstractString}; kwargs...) =
-    add_or_develop([check_package_name(pkg) for pkg in pkgs]; kwargs...)
+add_or_develop(pkgs::Vector{<:AbstractString}; mode::Symbol, kwargs...) =
+    add_or_develop([check_package_name(pkg, mode) for pkg in pkgs]; kwargs...)
+add_or_develop(; url = nothing, path = nothing, kwargs...) =
```
00vareladavid: A few concerns regarding this method:
- The source location is initially specified only at `(::PackageSpec).repo.url`, not in `path`. If you look at `API.Package`, you will see that both `url` and `path` get mapped to the same field internally. This is a bit confusing because users expect `path` to mean local filesystem path, but `path` has a different meaning internally.
- It does not make sense to specify both a `url` and a `path` for a single package. `API.Package` guards against this, but this function would circumvent that input checking.

src/Pkg.jl
```diff
@@ -86,10 +86,11 @@ a package, also inside that package.
 
 # Examples
 ```julia
-Pkg.add("Example") # Add a package from registry
+Pkg.add("Example") # Add the latest version of a package from registry
```
```suggestion
Pkg.add("Example") # Add package from registry
```

src/API.jl
BioTurboNick: Oops, that was a leftover, since we agreed to just do PackageSpec(). I had checked that passing `nothing` as default to both url and path to PackageSpec allowed a user to specify just one and it would fail if both were there, but could have missed something. Anyway, I just removed this function version.

fredrikekre:
Ok, thanks GitHub.

PR #1006, opened 2019-01-22
Fixed issue with `show(PackageSpec)` where missing name/uuid would try to print `nothing`

BioTurboNick:


Project.toml
```diff
@@ -3,7 +3,6 @@ keywords = ["package", "management"]
 license = "MIT"
 name = "Pkg"
 version = "1.1.0"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
StefanKarpinski: Can you restore the UUID entry?

BioTurboNick: Done

src/Types.jl
```diff
@@ -157,14 +157,13 @@ PackageSpec(n::AbstractString, u::UUID, v::VersionTypes) = PackageSpec(;name=n,
...
 
 function Base.show(io::IO, pkg::PackageSpec)
     vstr = repr(pkg.version)
-    f = ["name" => pkg.name, "uuid" => has_uuid(pkg) ? pkg.uuid : "", "v" => (vstr == "VersionSpec(\"*\")" ? "" : vstr)]
-    if pkg.repo !== nothing
-        if pkg.repo.url !== nothing
-            push!(f, "url/path" => string("\"", pkg.repo.url, "\""))
-        end
+    f = ["name" => has_name(pkg) ? pkg.name  : "", "uuid" => has_uuid(pkg) ? pkg.uuid : "", "v" => (vstr == "VersionSpec(\"*\")" ? "" : vstr)]
+    if has_repo(pkg)
```
fredrikekre: Why did you change the logic here? Does not seem to be necessary.

00vareladavid: I think @fredrikekre is refering to `has_repo`.

`"name" => has_name(pkg) ? pkg.name  : ""` is sufficient to fix the bug.

00vareladavid: Yup, for example when running `add Example#master`

PR #1009, opened 2019-01-23
Always print the activate environment

### LINKED ISSUE #983: Print a status message on `]activate`
It would be nice if `]activate` in the REPL printed a status message confirming the path to the activated project. I was helping somebody get set up and it turned out that they misunderstood my `]activate .` instruction as a `]activate` instruction and typed in the latter, and the two have quite different effects.

iamed2:
Fixes #983

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -487,7 +487,12 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
 end
 
 
-activate() = (Base.ACTIVE_PROJECT[] = nothing)
+function activate()
+    Base.ACTIVE_PROJECT[] = nothing
+    p = Base.active_project()
+    p === nothing || @info("activating environment at $(pathrepr(p)).")
+    return nothing
+end
 function _activate_dep(dep_name::AbstractString)
     Base.active_project() === nothing && return
     env = nothing
@@ -517,7 +522,6 @@ function activate(path::AbstractString; shared::Bool=false)
             fullpath = _activate_dep(path)
             if fullpath === nothing
                 fullpath = abspath(path)
-                @info("activating new environment at $(Base.contractuser(fullpath)).")
             end
         end
     else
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -487,7 +487,10 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
 end
 
 
-activate() = (Base.ACTIVE_PROJECT[] = nothing)
+function activate()
+    @info("setting active environment to follow working directory")
```
fredrikekre: This is not correct. If anything it should print the value of `Base.active_project()` after `Base.ACTIVE_PROJECT[]` is set to `nothing`.

iamed2: Huh, I don't know where I got that idea from. I clearly remember reading something that says that, but of course nothing like that exists...

Sorry! You're absolutely correct.

src/API.jl
```diff
@@ -487,7 +487,10 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
 end
 
 
-activate() = (Base.ACTIVE_PROJECT[] = nothing)
+function activate()
+    Base.ACTIVE_PROJECT[] = nothing
+    @info("activated home project environment at $(Base.active_project()).")
```
```suggestion
    @info("activated environment at $(pathrepr(Base.active_project())).")
```

fredrikekre: Hmm, `pathrepr` calls `contractuser`: https://github.com/JuliaLang/Pkg.jl/blob/32396f241ac5e957dd078d22e22cee2107f7bdc8/src/Types.jl#L1338 so we should probably use that everywhere instead of `contractuser`. Can be a separate PR though.

src/API.jl
```diff
@@ -487,7 +487,10 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
 end
 
 
-activate() = (Base.ACTIVE_PROJECT[] = nothing)
+function activate()
+    Base.ACTIVE_PROJECT[] = nothing
+    @info("activated environment at $(pathrepr(Base.active_project())).")
```
StefanKarpinski:
```suggestion
    @info("activated home environment at $(pathrepr(Base.active_project())).")
```

fredrikekre:
Pushed an update which now always prints the activated env, and also makes it more robust in the case where there is no active env.

PR #1010, opened 2019-01-23
Fix parsing numbers in TOML

iamed2:
Fixes #968 

Before:
```
julia> Pkg.TOML.parse("a = 1.0\n")
ERROR: MethodError: no method matching get(::String, ::String)
Closest candidates are:
  get(::Any) at /Users/ericdavies/.julia/dev/Pkg/ext/TOML/src/parser.jl:63
Stacktrace:
 [1] numdatetime(::Pkg.TOML.Parser{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Int64) at /Users/ericdavies/.julia/dev/Pkg/ext/TOML/src/parser.jl:391
 [2] value(::Pkg.TOML.Parser{Base.GenericIOBuffer{Array{UInt8,1}}}) at /Users/ericdavies/.julia/dev/Pkg/ext/TOML/src/parser.jl:724
 [3] keyvalues(::Pkg.TOML.Parser{Base.GenericIOBuffer{Array{UInt8,1}}}, ::Pkg.TOML.Table) at /Users/ericdavies/.julia/dev/Pkg/ext/TOML/src/parser.jl:750
 [4] parse(::Pkg.TOML.Parser{Base.GenericIOBuffer{Array{UInt8,1}}}) at /Users/ericdavies/.julia/dev/Pkg/ext/TOML/src/parser.jl:889
 [5] parse(::Base.GenericIOBuffer{Array{UInt8,1}}) at /Users/ericdavies/.julia/dev/Pkg/ext/TOML/src/TOML.jl:32
 [6] parse(::String) at /Users/ericdavies/.julia/dev/Pkg/ext/TOML/src/TOML.jl:40
 [7] top-level scope at none:0
```
After:
```
julia> Pkg.TOML.parse("a = 1.0\n")
Dict{String,Any} with 1 entry:
  "a" => 1.0
```

Looks like this was just a mistake when the code was written to drop 2-arg `get`.

PR #1012, opened 2019-01-24
Test TOML

iamed2:
---
I added a fake uuid for the TOML package, do you think something specific should go there?

Includes #1010 

I modified number printing to fix the printing test. @StefanKarpinski does this impact you? I noticed you made the change that printed quoted values as a fallback.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -44,6 +44,20 @@ jobs:
         - julia --project -e 'using Pkg; Pkg.activate("docs"); Pkg.instantiate(); Pkg.develop(PackageSpec(path = pwd()))'
         - julia --project=docs --color=yes docs/make.jl
       after_success: skip
+    - name: TOML (Julia 1.0)
+      stage: test
+      os: linux
+      julia: 1.0
+      script:
+        - julia --project=ext/TOML -e 'using Pkg; Pkg.test()'
+      after_success: skip
+    - name: TOML (Julia nightly)
+      stage: test
+      os: linux
+      julia: nightly
+      script:
+        - julia --project=ext/TOML -e 'using Pkg; Pkg.test()'
+      after_success: skip
 
 after_success:
   - julia --project -e 'import Pkg; Pkg.add("Coverage"); using Coverage; Codecov.submit(Codecov.process_folder())'
--- a/ext/TOML/Project.toml
+++ b/ext/TOML/Project.toml
@@ -1,0 +1,12 @@
+name = "TOML"
+uuid = "037cace4-c66a-5006-a6b7-c26ba1b2f83e"
+desc = "The TOML parser for Julia's package manager"
+
+[deps]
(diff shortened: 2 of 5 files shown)


StefanKarpinski:
bors try

iamed2:
```
ERROR: failed to clone from https://github.com/JuliaRegistries/General.git, error: GitError(Code:ERROR, Class:OS, failed to send request: The server name or address could not be resolved
```
seems unrelated

.travis.yml
```diff
@@ -44,6 +44,13 @@ jobs:
         - julia --project -e 'using Pkg; Pkg.activate("docs"); Pkg.instantiate(); Pkg.develop(PackageSpec(path = pwd()))'
         - julia --project=docs --color=yes docs/make.jl
       after_success: skip
+  include:
+    - name: TOML
+      stage: test
+      os: linux
+      script:
+        - julia -e 'using Pkg; Pkg.develop(PackageSpec(path="ext/TOML")); Pkg.instantiate(); Pkg.test("TOML")'
```
fredrikekre: `Pkg.instantiate` is not needed here because (i) everything is stdlibs and (ii) `Pkg.test` does its own resolve anyway. `Pkg.develop` is also not needed, we can just run in the TOML project.
```suggestion
        - julia --project=ext/TOML -e 'using Pkg; Pkg.test()'
```

ext/TOML/Project.toml
```diff
@@ -0,0 +1,12 @@
+name = "TOML"
+desc = "The TOML parser for Julia's package manager"
+uuid = "7ab3959e-fc62-457a-88b7-ddb939818b44"  # this is fake
```
fredrikekre: Why not just use the "real" one?

iamed2: Done

fredrikekre:
bors try

StefanKarpinski:
We can also just use the GitHub squash and merge button...

fredrikekre:
Yea. Btw, my suggested way to test on multiple julia versions seems to not work...

iamed2:
Squashed and (I believe) fixed the multiple Julia version testing

fredrikekre:
It's a bit awkward that one must repeat the script part etc for each stage, but seems to be no other way...

bors r+

.travis.yml
```diff
@@ -44,6 +44,22 @@ jobs:
...
       after_success: skip
+  include:
+    - name: TOML (Julia 1.0)
+      stage: test
+      os: linux
+      julia: 1.0
+      script:
+        - julia --project=ext/TOML -e 'using Pkg; Pkg.test()'
+      after_success: skip
+  include:
```
fredrikekre: Looks like this section overwrites the one above. Trying to understamd the Travis docs, but I think it might work if we just remove this line... The `-name:` part on the line below creates a new build I think.

iamed2: Should be fixed now

PR #1016, opened 2019-01-25
Don't throw for missing Versions.toml in the registry

fredrikekre:
Don't throw for missing Versions.toml in the registry, instead consider that the same as no versions, i.e. an empty existing file.


KristofferC:
Oh, so that is what `fakeit` means :D

PR #1018, opened 2019-01-27
Provide more intuitive help interface

### LINKED ISSUE #988: Pkg command `registry-status` doesn't work
This is on julia 1.1.0-rc1.0 on Windows.

The package manager tells me that there is a command ``registry-status`` when I look at the help with ``]help``. When I try to execute that command, I get an error though:
````
(v1.1) pkg> registry-status
ERROR: could not determine command
````

00vareladavid:
This makes help syntax consistent with the command syntax for compound commands e.g. `pkg> registry add` and `pkg> help registry add`.

Also provides filtered and sorted help for "partial" commands e.g. `pkg> help registry`

Resolves #988

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -246,7 +246,7 @@ end
 
 is_opt(word::AbstractString) = first(word) == '-'
 
-function core_parse(words)
+function core_parse(words; only_cmd=false)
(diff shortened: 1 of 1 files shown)


00vareladavid:
bors try

PR #1019, opened 2019-01-28
canonicalize paths to use /

### LINKED ISSUE #1007: Slashes (path seperators) conversion for develop packages between platforms in `Manifest.toml`
I made a `Manifest.toml` and `Package.toml` on Windows for a series of packages for a particular project, and I went to use them on a Unix environment (MacOS) and the manifest failed to work properly.
On Windows I did this:
```
(myproject) pkg> dev ./somepath/MyPackage.jl
```

And then on a Mac, I did this and got an error:
```
(myproject) pkg> resolve
 Resolving package versions...
ERROR: path /path/to/myproject\somepath\MyPackage.jl for package MyPackage no longer exists. Remove the package or `develop` it at a new path
```

Opening the `Manifest.toml` and replacing all the instances of `\\` with `/` fixed it, but it feels like this should be handled by `Pkg` for this case of making the `toml` files in windows and then opening it in a Unix location.

fredrikekre:
canonicalize paths to use / instead of \ when writing Manifest, fix #1007.

### DIFF
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -49,6 +49,14 @@ function safe_version(version::String)
     return version
 end
 
+# turn /-paths into \-paths on Windows
+function safe_path(path::String)
+    if Sys.iswindows() && !isabspath(path)
+        path = joinpath(split(path, "/")...)
+    end
+    return path
+end
+
 read_deps(::Nothing) = Dict{String, UUID}()
 read_deps(deps) = pkgerror("Expected `deps` field to be either a list or a table")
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Question: should we really backport this? It's a functionality change, after all. Might be better to just let the manifest files get canonicalized when someone upgrades to 1.2.

fredrikekre:
bors try

fredrikekre:
bors try

fredrikekre:
bors try

PR #1038, opened 2019-02-04
Continue precompiling even if some packages fail

timholy:
If you copy your environment from one machine to another, and the target machine is, e.g., lacking necessary libs and therefore fails to precompile, there is little reason to stop trying to compile the remaining packages.

I'm not quite sure how to go about testing this; in particular, if I create `BrokenPkg` how do I ensure that `precompile` will try it before moving on to `WorkingPkg`?

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -450,7 +450,10 @@ function precompile(ctx::Context)
         end
         if stale
             printpkgstyle(ctx, :Precompiling, pkg.name)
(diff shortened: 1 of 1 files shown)


ararslan:
Could you have a print or something at the top level that will be triggered during precompilation, redirect stdout to a file, and look for the message?

KristofferC:
I think we can do without a test for now

bors r+

PR #1043, opened 2019-02-05
improve error for nonexisting uuid in manifest

KristofferC:


00vareladavid:
Ah, I think I see the confusion.

One of the main strengths of #989 is that we first compute the subset of the dependency graph that we care about. From this point on, we don't use the manifest file at all. This eliminates synchronization issues between the loaded dependency graph (i.e. `pkgs`) and the manifest file. These synchronization issues were probably the root cause of the "getindex" errors we were seeing.

PR #1045, opened 2019-02-08
Clarify that `dev` does not go through `git`

00vareladavid:


docs/src/getting-started.md
```diff
@@ -126,7 +126,7 @@ We can see `tutorial` now contains `Example` as a dependency.
 
 Say we are working on `Example` and feel it needs new functionality.
 How can we modify the source code?
-We can use `develop` to set up a git clone of the `Example` package.
+We can use `develop` to set up a copy of the `Example` package.
```
StefanKarpinski: But it is a git clone, isn't it? It's just not one created with the `git` command line tool. It's a git clone created with `libgit2`.

00vareladavid: Yes, I understand. I feel the detail that the package will be a git clone is not relevant to the fact that `Pkg` loads the source directly from the filesystem when `develop`ing. I'm fine with going into detail in later sections, but I wanted to keep this section more focused.

StefanKarpinski: Ok, but I guess as a reader it would help me to understand that this copy is a clone where I can safely edit the code with version control, etc. and that it's "my copy" as opposed to the ones that Pkg squirrels away somewhere that I'm not supposed to touch them.

00vareladavid: My intention for this chapter was to help absolute beginners get started. For that purpose, I would like to remove as much detail as possible. The fact that `develop` creates a git clone is not used in the exercise in this section.

Later, this same section states:
> Instead of loading a registered version of Example, Julia will load the source code contained in tutorial/dev/Example.

I think that is sufficient for a beginner to understand the exercise.

I agree that `develop` creating a git clone is important information. I would argue that information belongs in more detailed chapters of the manual.

00vareladavid: If you agree, I can reintroduce that information elsewhere.

00vareladavid: Of course, its questionable how much a two word change will actually end up helping people :laughing:

00vareladavid:
Perhaps not a good idea if it ends up being more confusing

docs/src/getting-started.md
StefanKarpinski: Whatever you see fit. I'm satisfied that my point has been heard 😄

PR #1033, opened 2019-02-02
Basic testing for sandbox

00vareladavid:
- [x] fall back to parent dependency graph
- [x] test `test` sandboxing
- [x] test `build` sandboxing (very basic)

This will also test that the `compat` section of the sandbox project file is honored.

fix #1030, fix #895

PR #1048, opened 2019-02-08
silent activate when testing, fixes #1029.

### LINKED ISSUE #1029: Implementation detail about test env leaking to user
```
(v1) pkg> test Literate
   Testing Literate
[ Info: activating environment at `/tmp/tmpAP0rYi/Project.toml`.          <--- Suppress this.
 Resolving package versions...
```

fredrikekre:
---

#1033 became kinda big and does a lot of things. I think it is better to break it up a bit.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1041,7 +1041,7 @@ function pin(ctx::Context, pkgs::Vector{PackageSpec})
     foreach(pkg -> update_package_pin!(pkg, manifest_info(ctx.env, pkg.uuid)), pkgs)
     load_other_deps!(ctx, pkgs)
     check_registered(ctx, pkgs)
-    
+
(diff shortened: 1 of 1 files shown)


00vareladavid:
Don't we have to set `LOAD_PATH` as well?

fredrikekre:
No, I thought so, but `Base.active_project` (which `EnvCache` uses) looks in `Base.ACTIVE_PROJECT` first even if that does not exist in the load path. This is a bit weird, since you can have an `active_project` but not load any packages. This might not be enough to instantiate, but lets try:

bors try

00vareladavid:
So it does need `LOAD_PATH`?

fredrikekre:
> So it does need `LOAD_PATH`?

I was not able to find a case that needed it, but I feel like its a good idea to `instantate` with the same load path as the tests are run with later.

PR #1049, opened 2019-02-08
Display status for the correct environment when testing.

fredrikekre:
We were printing the status for the context where you call `pkg> test` instead of the environment where the testing is happening.


fredrikekre:
Actually, why are we not printing status if `Types.is_project_uuid(ctx.env, pkg.uuid)`?

00vareladavid:
I'm not sure, I just kept whatever was there before. Honestly I didn't think too much about the display side of the sandbox.

PR #1051, opened 2019-02-08
Introduce env navigation syntax

00vareladavid:
Changes:
- introduce `activate` sandbox support e.g. `activate +test`
- introduce `activate` "backjump" support e.g. `activate -`
- introduce `activate` shared env "@" syntax e.g. `activate @foo`

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -498,6 +498,10 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}; mode=PKGMODE_PROJECT)
 end
 
 
+function _activate_info()
+    p = Base.active_project()
+    p === nothing || @info("activating$(ispath(p) ? "" : " new") environment at $(pathrepr(p)).")
+end
 function activate()
     Base.ACTIVE_PROJECT[] = nothing
     p = Base.active_project()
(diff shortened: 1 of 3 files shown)


00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
Feedback on syntax additions? Otherwise I feel this is good to go

fredrikekre:
Could we perhaps discuss this a bit more? I am not a fan :stuck_out_tongue: 
bors r-

00vareladavid:
Ok, that's understandable

oxinabox:
I wanted to pop up the stack of enviroments.

I am defining 5 subenviroments, with a single common shared enviroment.
And just for purposes of setting them all up 
i would be doing 
```
activate common
activate Special1
add Special1Pkg
activate - 
activate Special2
add Special2Pkg
activate - 
...
```

But since there is no way to just pop the tiop enviroment
I am currently just going to quit between each.

StefanKarpinski:
I have encountered something similar. When I activate the `test` "sub environment" for a project, I want to keep the main environment for the project in the load path as well. I'm not sure what the best mechanism for it is, but currently I've had to just push it into the load path manually, which is awkward.

00vareladavid:
More evidence that people tend to think in terms of the active environment is: #1215. Instead of having to `cd` then `dev`, it would be nicer if users could just e.g. `pkg> dev +relative/path`.

00vareladavid:
I have come around to agreeing with @fredrikekre on this. If anything, a more appropriate place for this is in an extension package

StefanKarpinski:
Could we please have the `activate @foo` syntax part though?

KristofferC:
That's a shorthand for `activate --shared foo`?

StefanKarpinski:
Yes, that was the idea. It would also be a more general syntax: `@foo` anywhere in a command would refer to the named environment `foo`.

fredrikekre:
https://github.com/JuliaLang/Pkg.jl/pull/1488

StefanKarpinski:
Great. I thought that was around somewhere.

PR #1053, opened 2019-02-11
Capture output

00vareladavid:
Aiming for #122 with this one.

Basically just add an `io` field to `Context`, thread the `Context` object through a more functions, and make the display stuff use the `io` field by default. I also switched some functions from taking `EnvCache` to `Context`. Most were necessary because they eventually printed, but some I just changed for consistency's sake.

This should also let us test `Display` more easily, which we are not really doing right now.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -309,6 +309,7 @@ end
 # ENV variables to set some of these defaults?
 Base.@kwdef mutable struct Context
     env::EnvCache = EnvCache()
+    io::IO = stdout
     preview::Bool = false
     use_libgit2_for_all_downloads::Bool = false
     use_only_tarballs_for_downloads::Bool = false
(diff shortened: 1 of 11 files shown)


fredrikekre:
It would be good to decide if we are gonna use `stdout` or the logging system also, right now it is very random what information is sent where.

KristofferC:
FWIW, I think we should just ditch the logging completely.

StefanKarpinski:
I think it would be good to silently log but print info separately.

tkoolen:
Just wanted to bump this PR. I often need Pkg operations to be completely silent, which is currently hard to do given the lack of an easy way to suppress output in general, see https://github.com/JuliaLang/julia/issues/22879 and the fact that Suppressor.jl exists and is not trivial.

FWIW, I like the logging approach precisely because it's easy to control verbosity: no need to learn a new API, no way to forget to pass a `Context`, and it just seems like it's exactly what the logging system is meant for; a good non-`Logging`-based solution is bound to re-implement a lot of `Logging`'s features. But mostly I just care about there being *a* solution.

fredrikekre:
```
$ jlpkg add Example > /dev/null
```
:)

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

src/Types.jl
```diff
@@ -309,6 +309,7 @@ end
 # ENV variables to set some of these defaults?
 Base.@kwdef mutable struct Context
     env::EnvCache = EnvCache()
+    io::IO = stdout
```
KristofferC: I think we should change this to `stderr`.

00vareladavid: Please explain. Its not weird to output to `stderr` by default? What is the benefit over `stdout`?

KristofferC: I don't think it is that weird. Many unix applications work by piping their stdout to each other. If an application starts spewing loginfo to stdout then that breaks. The loggers in Base already use stderr so I don't think it is weird for us to also log to that.

fredrikekre: But e.g. for `Pkg.status` it would make sense with `stdout`, no?

KristofferC: Tbh, I'm not sure.

00vareladavid:
Anyone mind if I merge? My main concern here is getting the `io` field to all the print statements. We can split the output into user feedback vs actual logging at a later time.

KristofferC:
Go for it

src/Types.jl
GunnarFarneback: What would make sense for `Pkg.status` to do would be to print nothing and return some kind of object with the status information. Then the REPL could have it printed the usual way with the `show` method, the Pkg REPL mode could print it any way it finds useful, and a program could look into the object and draw whatever conclusions it finds meaningful.

PR #1057, opened 2019-02-12
Absolutify relative paths in tmp manifest

fredrikekre:
Absolutify relative paths in tmp manifest after it has been copied to tmp directory.


00vareladavid:
This is already included in #1033: https://github.com/JuliaLang/Pkg.jl/pull/1033/files#diff-920ed338dcbd9d62489c6d12dc2fbb18R1145.

PR #1059, opened 2019-02-14
Document toml-files

fredrikekre:


docs/src/toml-files.md
```diff
@@ -0,0 +1,224 @@
...
+Example:
+```toml
+version = "1.2.5"
+```
+Julia uses [Semantic Versioning](https://semver.org/) (SemVer) and the
+`version` field should follow SemVer. The basic rules are:
+* Before 1.0.0, anything goes, but when you make breaking changes
+  the minor version should be incremented.
+* After 1.0.0 only make breaking changes when incrementing the major version.
+* After 1.0.0 no new public API should be added without incrementing the
```
KristofferC: Perhaps we should link to the section about versions specifiers in Pkg here

docs/src/toml-files.md
```diff
@@ -0,0 +1,224 @@
...
+### `Manifest.toml` entries
+
+Each dependency has its own section in the manifest file, and its content varies depending
+on how the dependency was added, see the examples below.
+Every dependency section includes a combination of the following entries:
+
+* `uuid`: the [UUID](https://en.wikipedia.org/wiki/Universally_unique_identifier)
+  for the dependency, for example `uuid = "7876af07-990d-54b4-ab0e-23690620f79a"`.
+* `deps`: a vector listing the dependencies of the dependency, for example
+  `deps = ["Example", "JSON"]`.
```
KristofferC: This is AFAIU only true if there is one dependency called `Example`. If there are multiple dependencies called `Example` a `name => uuid` Dict is instead printed.

fredrikekre: Yea, I can add a section about that maybe.

docs/src/toml-files.md
```diff
@@ -0,0 +1,224 @@
...
+* `version`: a version number, for example `version = "1.2.6"`.
+* `path`: a file path to the source code, for example `path = /home/user/Example`.
+* `repo-url`: a URL to the repository where the source code was found,
+  for example `repo-url = "https://github.com/JuliaLang/Example.jl.git"`.
+* `repo-rev`: a git revision, for example a branch `repo-rev = "master"`
+  or a commit `repo-rev = "66607a62a83cb07ab18c0b35c038fcd62987c9b1"`.
+* `git-tree-sha1`: a content hash of the source tree, for example
+  `git-tree-sha1 = "ca3820cc4e66f473467d912c4b2b3ae5dc968444"`.
+
+Note, however, that `uuid`, `deps` and `version` are present for all dependency sections.
```
00vareladavid: Only `uuid` is required

00vareladavid:
I'm against documenting `Manifest.toml` at a low level. If users want to manipulate it, they should use the API. I've seen a few bugs because users manipulate the fields without fully understanding the implications.

StefanKarpinski:
I think it's ok to document the format while saying that it's autogenerated and should not be edited.

00vareladavid:
Yeah I guess its not a big deal. We also have the nice warning at the top of `Manifest.toml` now i.e. "# This file is machine-generated - editing it directly is not advised "

docs/src/toml-files.md
00vareladavid: To elaborate: stdlibs don't really have a `version` and nodes without dependencies of course wont have a `deps` field :+1:

StefanKarpinski:
Sorry for the late review: this looks great.

Issue #378, opened 2018-06-13
Source of packages should be recorded in the manifest


KristofferC:
Otherwise, we need the same registries that the manifest was created from and that is not recorded.

greimel:
Having this would be really helpful for people that don't want to clutter `General` but also don't want to destroy reproducibility when sharing code. It's really easy to forget that a package from my private registry has been used and then my students or collaborators (not necessarily aware of package registries) get frustrated and cannot use my code.

(In order to avoid that I keep copy-pasting functionality across my teaching notebooks.)

PR #1064, opened 2019-02-17
[POC] Add scoping, e.g. @MyRegistry/MyPackage

EricForgy:
Hello,

This can use some improvements, but I was happy to get it to work 😊 

As an example, I created a new registry at

https://github.com/JuliaFinance/JuliaFinance

This registry contains a package Currencies.jl that is also in the General registry.

Now if add my new registry

```julia
pkg> registry add https://github.com/JuliaFinance/JuliaFinance.git
```

I have two registries and when I try to 

```julia
julia> Pkg.add("Currencies")
```

`Pkg` cleverly notifies me that my package is in two registries and asks me to choose one (very cool btw 😃).

With this PR, I can specify upfront which registry I want to use via

```julia
julia> Pkg.add("@JuliaFinance/Currencies")
```

This is supposed to be reminiscent of scoping on npm.

What do you think?

Let me know if you'd like me to continue work on this or just let it go 😊 

Closes https://github.com/JuliaComputing/Registrator.jl/issues/23

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -61,12 +61,32 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true, kwa
     return
 end
 
-add(pkg::Union{AbstractString, PackageSpec}; kwargs...) = add([pkg]; kwargs...)
+function add(pkg::AbstractString; kwargs...)
+    if pkg[1] == '@'
+        r,p = split(pkg[2:end],'/')
+        add(Context(),[PackageSpec(String(p))],[RegistrySpec(String(r))]; kwargs...)
+    else
+        add([pkg]; kwargs...)
+    end
+    return
+end
+add(pkg::PackageSpec; kwargs...) = add([pkg]; kwargs...)
 add(pkgs::Vector{<:AbstractString}; kwargs...) =
     add([check_package_name(pkg, :add) for pkg in pkgs]; kwargs...)
 add(pkgs::Vector{PackageSpec}; kwargs...)      = add(Context(), pkgs; kwargs...)
 function add(ctx::Context, pkgs::Vector{PackageSpec}; kwargs...)
+    Context!(ctx,kwargs...)
+    println("add(ctx::Context, pkgs::Vector{PackageSpec}; kwargs...)")
+    clone_default_registries()
+    add(ctx,pkgs,collect_registries(); kwargs...)
+    return
+end
+function add(
+    ctx::Context, 
+    pkgs::Vector{PackageSpec}, 
+    regs::Vector{RegistrySpec}; kwargs...)
     pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
+    regs = deepcopy(regs)  # deepcopy for avoid mutating RegistrySpec members
     Context!(ctx; kwargs...)
 
     for pkg in pkgs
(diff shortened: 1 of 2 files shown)


KristofferC:
FYI, syntax like

```
julia> Pkg.add("@JuliaFinance/Currencies")
```

can work in the REPL mode but for the API this should be something like `Pkg.add(PackageSpec(name = "Currencies", registry = "JuliaFinance"))`.

EricForgy:
Sure. Good point 👍

What do you think of the general idea? Should I try to finish it up?

EricForgy:
Would you guys be open to adding a `registries` property to `EnvCache`? This would be a little cleaner and we could pass it into the `Context` as a `kwarg` like the other `kwargs...`.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/issues/378

fredrikekre:
What is the added benefit of
```
pkg> registry add https://github.com/JuliaFinance/JuliaFinance.git

pkg> add @JuliaFinance/Currencies
```
over
```
pkg> registry add https://github.com/JuliaFinance/JuliaFinance.git

pkg> add Currencies
```
?

EricForgy:
In my case, there is an ambiguity because `Currencies` is registered in both `General` and `JuliaFinance`. There is not a nice clever prompt that will ask me to choose which one I wanted, but I'd rather just specify at the command line with no further interaction.

KristofferC:
This is no different to two packages being registered with the same name in a single registry. And then you disambiguate with UUID.

fredrikekre:
> In my case, there is an ambiguity because `Currencies` is registered in both `General` and `JuliaFinance`. There is now a nice clever prompt that will ask me to choose which one I wanted, but I'd rather just specify at the command line with no further interaction.

In this case it is the *same* package though, right? Then you should not even be prompted to choose, they have the same uuid so it does not matter in which registry we get it from.

EricForgy:
True, but disambiguating based on specifying explicitly the registry seems like a better user experience than specifying by UUID.

EricForgy:
> > In my case, there is an ambiguity because `Currencies` is registered in both `General` and `JuliaFinance`. There is now a nice clever prompt that will ask me to choose which one I wanted, but I'd rather just specify at the command line with no further interaction.
> 
> In this case it is the _same_ package though, right? Then you should not even be prompted to choose, they have the same uuid so it does not matter in which registry we get it from.

Right. I was prompted once (otherwise I wouldn't even know there is a prompt 😅, but cannot reproduce it 😅

fredrikekre:
You are only prompted if there are multiple packages with the same name with different uuid.

fredrikekre:
> True, but disambiguating based on specifying explicitly the registry seems like a better user experience than specifying by UUID.

So I guess this would just be a way to narrow the UUID search in `add Package` then. You could still run into name collisions though.

StefanKarpinski:
Alternate proposal for the same syntax: https://github.com/JuliaLang/Pkg.jl/issues/1071. TLDR: the syntax is `@user/package` instead of `@registry/package`. Since the a registry can contain multiple packages with the same name and the same package can appear in multiple registries it doesn't make much sense to use registries to disambiguate packages with the same name. The user/org that maintains a package, on the other hand, is a good already-available, well-known qualifier and it's unlikely that a given user/org maintains two different packages by the same name.

PR #1068, opened 2019-02-18
Update doc strings for pin, free

GregPlowman:
A few changes to help strings for `pin`, `free`.

Correct case for `PackageSpec`
Add examples for `pin`

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -87,7 +87,8 @@ a package, also inside that package.
 # Examples
 ```julia
 Pkg.add("Example") # Add a package from registry
-Pkg.add(PackageSpec(name="Example", version="0.3")) # Specify version
+Pkg.add(PackageSpec(name="Example", version="0.3")) # Specify version; latest release in the 0.3 series
+Pkg.add(PackageSpec(name="Example", version="0.3.1")) # Specify version; exact release
 Pkg.add(PackageSpec(url="https://github.com/JuliaLang/Example.jl", rev="master")) # From url to remote gitrepo
 Pkg.add(PackageSpec(url="/remote/mycompany/juliapackages/OurPackage"))` # From path to local gitrepo
 ```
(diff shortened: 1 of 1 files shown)


src/Pkg.jl
```diff
@@ -87,7 +87,7 @@ a package, also inside that package.
 # Examples
 ```julia
 Pkg.add("Example") # Add a package from registry
-Pkg.add(PackageSpec(name="Example", version="0.3")) # Specify version
+Pkg.add(PackageSpec(name="Example", version="0.3.1")) # Specify version
```
00vareladavid: This change is not really necessary

00vareladavid:
Overall, looks good to me. Particularly useful are the examples for pin :+1:

src/Pkg.jl
GregPlowman: I thought `0.3.1` might be more familiar and helpful to new users, as Julia packages have `x.y.z` version numbering.
Also for someone unfamiliar with versioning and `Pkg` specifics it might be confusing.
For example, I'm unsure whether `version="0.3"` means a specific version because the versioning system only has `x.y` numbering, or whether it means add the _latest_ `0.3` version, say `0.3.7` rather than earlier versions.

00vareladavid: You make a good point. I'm in favor of this.

fredrikekre: Maybe just add a second line and have two example:
```
Pkg.add(PackageSpec(name="Example", version="0.3")) # Specify version; latest release in the 0.3 series
Pkg.add(PackageSpec(name="Example", version="0.3.1")) # Specify version; exact release
```

src/Pkg.jl
```diff
@@ -186,16 +186,22 @@ const installed = API.installed
 
 """
     Pkg.pin(pkg::Union{String, Vector{String}})
-    Pkg.pin(pkgs::Union{Packagespec, Vector{Packagespec}})
+    Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}})
 
-Pin a package to the current version (or the one given in the `packagespec` or a certain
+Pin a package to the current version (or the one given in the `PackageSpec` or a certain
```
```suggestion
Pin a package to the current version (or the one given in the `PackageSpec`) or to a certain
```

GregPlowman: Done

PR #1079, opened 2019-02-20
fix #1077: handle stdlibs in update_package_add and friends

### LINKED ISSUE #1077: Consecutive stdlib add's errors
```
(v1) pkg> add Test
 Resolving package versions...
  [8dfed614] + Test 
[...]

(v1) pkg> add Test
ERROR: MethodError: no method matching iterate(::Pkg.Types.VersionSpec)
```

fredrikekre:
Handle stdlibs explicitly in `update_package_add!`, `update_package_pin!`
and `update_package_test!`, since they are not versioned like other packages,
fixes #1077.

---
bors try

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -900,8 +900,10 @@ function update_package_add(pkg::PackageSpec, entry::PackageEntry, is_dep::Bool)
     if entry.path !== nothing || entry.repo.url !== nothing || pkg.repo.url !== nothing
         return pkg # overwrite everything, nothing to copy over
     end
-    if is_dep && ((isa(pkg.version, VersionNumber) && entry.version == pkg.version) ||
-                 (!isa(pkg.version, VersionNumber) && entry.version ∈ pkg.version))
+    if is_stdlib(pkg.uuid)
+        return pkg # stdlibs are not versioned like other packages
+    elseif is_dep && ((isa(pkg.version, VersionNumber) && entry.version == pkg.version) ||
+                      (!isa(pkg.version, VersionNumber) && entry.version ∈ pkg.version))
         # leave the package as is at the installed version
         return PackageSpec(; uuid=pkg.uuid, name=pkg.name, version=entry.version,
                            tree_hash=entry.tree_hash)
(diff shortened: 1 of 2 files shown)


src/Operations.jl
```diff
@@ -1033,10 +1034,12 @@ function update_package_pin!(pkg::PackageSpec, ::Nothing)
     if pkg.version == VersionSpec() # no version to pin
         pkgerror("Can not `pin` a package which does not exist in the manifest")
     end
+    is_stdlib(pkg.uuid) && pkgerror("cannot `pin` stdlibs.")
     pkg.pinned = true
 end
 
 function update_package_pin!(pkg::PackageSpec, entry::PackageEntry)
+    is_stdlib(pkg.uuid) && pkgerror("cannot `pin` stdlibs.")
```
fredrikekre: Probably does not make sense to `pin` a stdlib. It is an error in 1.0 (`ERROR: KeyError: key "version" not found`) and 1.1 (`ERROR: MethodError: no method matching VersionNumber(::Nothing)`) so this is not a regression, and results in a better error message.

fredrikekre:
bors try

src/Operations.jl
```diff
@@ -900,8 +900,9 @@ function update_package_add(pkg::PackageSpec, entry::PackageEntry, is_dep::Bool)
     if entry.path !== nothing || entry.repo.url !== nothing || pkg.repo.url !== nothing
         return pkg # overwrite everything, nothing to copy over
     end
-    if is_dep && ((isa(pkg.version, VersionNumber) && entry.version == pkg.version) ||
-                 (!isa(pkg.version, VersionNumber) && entry.version ∈ pkg.version))
+    if is_dep && !is_stdlib(pkg.uuid) &&
```
00vareladavid: I would prefer if the stdlib branch was separate. E.g. `is_stdlib(pkg.uuid) && return pkg`

00vareladavid: Otherwise looks good

fredrikekre: Yea, that is cleaner.

Issue #767, opened 2018-09-20
Don't allow generating packages with names that are not legal Julia identifiers


eschnett:
I am teaching a computational physics course using Julia, and taught people how to use `generate` to create new packages. Sure enough, some people created packages with names such as `2018Example`. Of course this doesn't work, because this is not a valid Julia module name, and `using 2018Example` later on fails.

To be beginner friendly, I suggest to disallow package names that are not valid module names.

Nosferican:
Tag as Good-First-Issue?

fredrikekre:
#1083

PR #1083, opened 2019-02-23
Valid Julia identifiers for projects

Nosferican:
Issue #767


src/generate.jl
```diff
@@ -5,6 +5,7 @@ function generate(ctx::Context, path::String; kwargs...)
     Context!(ctx; kwargs...)
     ctx.preview && preview_info()
     dir, pkg = dirname(path), basename(path)
+    Base.isidentifier(pkg) || pkgerror("$pkg is not a legal Julia identifier")
```
fredrikekre:
```suggestion
    Base.isidentifier(pkg) || pkgerror("$(repr(pkg)) is not a valid package name")
```

Nosferican: I think package might be best as usual case, but it might be more proper to say project name in the project/package/application lingo.

PR #1086, opened 2019-02-25
Query interface

00vareladavid:
This provides a low-level interface that handles queries about the dependency graph.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -20,6 +20,26 @@ preview_info() = printstyled("───── Preview mode ─────\n"; c
 
 include("generate.jl")
 
+dependencies() = dependencies(Context())
+function dependencies(ctx::Context)::Dict{UUID, PackageInfo}
+    pkgs = PackageSpec[]
+    Operations.load_all_deps!(ctx, pkgs)
+    find_registered!(ctx, UUID[pkg.uuid for pkg in pkgs])
+    return Dict(pkg.uuid => Operations.package_info(ctx, pkg) for pkg in pkgs)
+end
+
+project() = project(Context())
+function project(ctx::Context)
+    ctx.env.pkg !== nothing || pkgerror("Active environment is not a project.")
+    return ProjectInfo(
+        name         = ctx.env.pkg.name,
+        uuid         = ctx.env.pkg.uuid,
+        version      = ctx.env.pkg.version,
+        dependencies = ctx.env.project.deps,
+        path         = ctx.env.project_file
+    )
+end
+
 function check_package_name(x::AbstractString, mode=nothing)
     if !(occursin(Pkg.REPLMode.name_re, x))
         message = "$x is not a valid packagename."
@@ -252,17 +272,6 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     end
     Operations.test(ctx, pkgs; coverage=coverage, test_fn=test_fn, julia_args=julia_args, test_args=test_args)
     return
-end
-
-installed() = __installed(PKGMODE_PROJECT)
-function __installed(mode::PackageMode=PKGMODE_MANIFEST)
-    diffs = Display.status(Context(), PackageSpec[], mode=mode, use_as_api=true)
-    version_status = Dict{String, Union{VersionNumber,Nothing}}()
-    diffs == nothing && return version_status
-    for entry in diffs
-        version_status[entry.name] = entry.new.ver
-    end
-    return version_status
 end
 
 """
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -49,7 +49,7 @@ is_dep(ctx::Context, pkg::PackageSpec) =
 function load_direct_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Bool=true)
     # load rest of deps normally
     for (name::String, uuid::UUID) in ctx.env.project.deps
-        pkgs[uuid] === nothing || continue # dont duplicate packages
+        pkgs[uuid] === nothing || continue # do not duplicate packages
         entry = manifest_info(ctx, uuid)
         push!(pkgs, entry === nothing ?
               PackageSpec(;uuid=uuid, name=name) :
@@ -66,6 +66,7 @@ end
 
 function load_all_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Bool=true)
     for (uuid, entry) in ctx.env.manifest
+        pkgs[uuid] === nothing || continue # do not duplicate packages
         push!(pkgs, PackageSpec(name=entry.name, uuid=uuid, path=entry.path,
                                 version = version ? something(entry.version, VersionSpec()) : VersionSpec(),
                                 repo=entry.repo, tree_hash=entry.tree_hash))
@@ -1336,6 +1337,27 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
                  join(pkgs_errored, ", "),
                  " errored during testing")
     end
+end
+
+function package_info(ctx::Context, pkg::PackageSpec)::PackageInfo
+    entry = manifest_info(ctx, pkg.uuid)
+    if entry === nothing
+        pkgerror("Can not query `$(pkg.name)` because it does not exist in the manifest.",
+                 " Use `Pkg.resolve()` to populate the manifest.")
+    end
+    package_info(ctx, pkg, entry)
+end
+
+function package_info(ctx::Context, pkg::PackageSpec, entry::PackageEntry)::PackageInfo
+    info = PackageInfo(
+        name         = pkg.name,
+        version      = pkg.version != VersionSpec() ? pkg.version : nothing,
+        ispinned     = pkg.pinned,
+        isdeveloped  = pkg.path !== nothing,
+        source       = project_rel_path(ctx, source_path(pkg)),
+        dependencies = collect(values(entry.deps)),
+    )
+    return info
 end
 
 end # module
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -207,9 +207,6 @@ redirecting to the `build.log` file.
 """
 const build = API.build
 
-# TODO: decide what to do with this
-const installed = API.installed
-
 """
     Pkg.pin(pkg::Union{String, Vector{String}})
     Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}})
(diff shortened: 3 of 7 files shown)


00vareladavid:
Two parts:
- choose a subset of the graph
    - `active`, `direct`, and `all` (could probably use better names)
    - alternatively, can explicitly give a list of UUIDs
- choose properties to query
    - currently supports [`name` `pinned` `rev` `dev` `source` `stdlib` `version` `url`]

KristofferC:
An official interface for asking these type of queries (that we keep stable) would greatly help people working on stuff like the VSCode plugin I think.

fredrikekre:
I had also planned to do something similar to this. I was thinking of just creating an interface to return the TOML-parsed `Dict` of Projetc.toml and then you can do whatever you want with that. Not sure what is better, or if they are orthogonal to each other.

00vareladavid:
Perhaps orthogonal, this is intended for other packages to more easily hook into `Pkg`. It might be too heavy for people who e.g. want to query the version of their own project.

Its possible, but cumbersome as a human interface: `collect(values(Pkg.info(;active=true, version=true)))[1]["version"]`. A user-friendly interface might also support name-based queries, as opposed to UUID-based.

KristofferC:
Putting this as a reference https://doc.rust-lang.org/cargo/reference/external-tools.html#information-about-package-structure.

KristofferC:
One idea which might be a bit more future proof is to make a dedicated `struct` that holds the information instead of a `Dict{String, Any}`. That was we can deprecate things and tab completion will work etc.

davidanthoff:
So in general I would love to see a stable API. In practice, for VS Code, the issue is that we want to support julia 1.0 going forward, given that it is the long term release. So anything that just appears in say in julia 1.2 or julia 1.3 doesn't help us much... In fact, what really would help us would be a package (not in the stdlib) that hides all the differences between different julia 1.x versions for us.

KristofferC:
>In fact, what really would help us would be a package (not in the stdlib) that hides all the differences between different julia 1.x versions for us.

I don't think anyone here really has plan on doing that. Using the "query interface" whenever it gets in a release and "ifdef" for earlier versions would likely be your best option.

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
I feel this is good to merge. Any objections/feedback?

As Kristoffer suggested, the data is returned as a struct instead of a dictionary. I also added documentation and some basic tests. I also remove the old `__installed` API which serves a similar purpose but is based on names, not UUIDs.

Since this is supposed to be a bit more future proof. I went with the conservative approach of requiring users to explicitly specify all the properties that they want to query. The idea is that future properties might be more a lot more expensive to compute (I'm thinking of cases where we have to parse registry info). Feedback on this approach specifically?

I was also conservative with the initial set of properties that users can query. We would definitely add more properties as we understand the kinds of questions that users ask.

src/Pkg.jl
```diff
@@ -264,6 +261,47 @@ const develop = API.develop
...
+Properites:
+
+| `keyword`   | `Type`          | `Description`                                      |
+|:------------|:----------------|:---------------------------------------------------|
+| `name`      | `String`        | The name of the package                            |
+| `version`   | `VersionNumber` | The version of the package                         |
+| `developed` | `Bool`          | Whether a package is directly tracking a directory |
+| `pinned`    | `Bool`          | Whether a package is pinned                        |
+| `source`    | `String`        | The directory containing the package's source code |
+
```
oxinabox: Which of these will be `nothing` for stdlibs?

src/Pkg.jl
```diff
@@ -264,6 +261,47 @@ const develop = API.develop
 #TODO: Will probably be deprecated for something in PkgDev
 const generate = API.generate
 
+"""
+    Pkg.info(; keywords...)::Dict{UUID, PackageInfo}
+    Pkg.info(uuid::UUID; keywords...)::Dict{UUID, PackageInfo}
+    Pkg.info(uuids::Vector{UUID}; keywords...)::Dict{UUID, PackageInfo}
+
```
oxinabox: Document what the uuid is the uuid of.
I am pretty sure it is a package UUID.

Also you probably want to include instructions somewhere on how to retrieve a list of UUIDs for a given package name and a registry / locally.
(I feel like that is `pkgid`?)

src/Types.jl
```diff
@@ -1348,4 +1348,16 @@ function write_env(ctx::Context; display_diff=true)
     write_manifest(env.manifest, env, old_env, ctx; display_diff=display_diff)
 end
 
+###
+### PackageInfo
+###
+
+Base.@kwdef mutable struct PackageInfo
```
oxinabox: I kinda wish this was written in a way that lets this be non-mutable

src/Operations.jl
```diff
@@ -64,6 +64,7 @@ end
 
 function load_all_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Bool=true)
     for (uuid, entry) in ctx.env.manifest
+        pkgs[uuid] === nothing || continue # dont duplicate packages
```
oxinabox:
```suggestion
        pkgs[uuid] === nothing || continue  # don't duplicate packages
```

src/Operations.jl
```diff
@@ -1298,4 +1299,16 @@ function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, test_fn=n
...
+                      name=false, version=false, pinned=false, developed=false, source=false
+                      )::PackageInfo
+    info = PackageInfo()
+    name      && (info.name = pkg.name)
+    version   && (info.version = (pkg.version == VersionSpec() ? nothing : pkg.version))
+    pinned    && (info.pinned = pkg.pinned)
+    source    && (info.source = project_rel_path(ctx, source_path(pkg)))
+    developed && (info.developed = pkg.path !== nothing)
+    return info
+end
```
oxinabox: Actually Package info can be made immutable fairly easily.
I think it would be good to do so, since it would be very confusing if someone mutated it in their code.

```
info = PackageInfo(
    name      = name ? pkg.name : nothing,
    version   = version && pkg.version == VersionSpec() ? pkg.version : nothing,
    pinned    = pinned ? pkg.pinned : nothing,
    source    = source ? project_rel_path(ctx, source_path(pkg)) : nothing,
    developed = developed ? pkg.path !== nothing : nothing,
)
return info
```

src/API.jl
```diff
@@ -18,6 +18,28 @@ preview_info() = printstyled("───── Preview mode ─────\n"; c
...
+info(uuid::UUID; kwargs...) = info([uuid]; kwargs...)
+info(uuids::Vector{UUID}; kwargs...) = info(Context(), [PackageSpec(;uuid=uuid) for uuid in uuids]; kwargs...)
+function info(ctx::Context, pkgs::Vector{PackageSpec};
+              active=false, direct=false, all=false, kwargs...
+              )::Dict{UUID, PackageInfo}
+    if active
+        if ctx.env.pkg === nothing
+            @warn "Active environment is not a project"
+        else
+            pkg = deepcopy(ctx.env.pkg)
```
oxinabox: Why is the `deepcopy` needed?

oxinabox:
Sorry I was so slow in getting around to looking at this.

> Since this is supposed to be a bit more future proof. I went with the conservative approach of requiring users to explicitly specify all the properties that they want to query. The idea is that future properties might be more a lot more expensive to compute (I'm thinking of cases where we have to parse registry info). Feedback on this approach specifically?

I am not enthused about it. I don't hate it but...

What if we used properties & closures, so they could be calculated when and if the field was checked ?
Or rather planned to use properties & closures, for any expensive ones that might turn up later.

timholy:
In case it helps with API design, other questions I *know* we're going to want to ask are:
- "which other registered packages depend on X?"
- "what new incompatibilities are generated if I release version `n` of X"?

Ref https://github.com/JuliaRegistries/Registrator.jl/issues/122. My main concern is that this API seems to focus on "what does X depend on?" whereas I tend to think it's more useful to ask "what depends on X?" Obviously the first is more of a finite set, the latter depends on which registry(ies) you check and there is no guarantee that you know about all the registries that exist in the universe.

00vareladavid:
Thanks @oxinabox for the excellent feedback. I agree lazily loading any expensive properties is a better API. And there is indeed no reason why the properties shouldn't be immutable.

src/Pkg.jl
00vareladavid: All fields should "Just Work" for stdlibs, with the exception of `version` which will always be `nothing`.

00vareladavid:
bors try

00vareladavid:
This has changed a bit from the initial design:
   - I changed the name from `Pkg.info` to `Pkg.dependencies`. I think this makes the purpose more clear. Is "dependencies" too long or should we go with `Pkg.deps`?
   - Simplified the interface by removing all keyword arguments from `Pkg.dependencies`. It will just return the whole set of dependencies (as a `Dict{UUID,PackageInfo}`)
   - added `Pkg.project` function which returns a `ProjectInfo` struct containing info about the active project. @fredrikekre you mentioned a plan similar to this. Do you think this will suffice? I will finish documenting it if it looks good.
   - any bikeshedding on field names?

00vareladavid:
bors try

src/Pkg.jl
```diff
@@ -273,6 +270,30 @@ const develop = API.develop
 #TODO: Will probably be deprecated for something in PkgDev
 const generate = API.generate
 
+"""
+    Pkg.dependencies()::Dict{UUID, PackageInfo}
+
+Query the dependecy graph.
+The result is a `Dict` that maps a package UUID to a `PackageInfo` struct representing that package.
```
oxinabox:
```suggestion
The result is a `Dict` that maps a package UUID to a `PackageInfo` struct representing that the dependency (a package). 
```

oxinabox:
> I changed the name from Pkg.info to Pkg.dependencies. I think this makes the purpose more clear. Is "dependencies" too long or should we go with Pkg.deps?

`dependencies` seems good to me.
I would also be ok with `dependency_packageinfos` to be fully explict.

And also wold go for `projectinfo` over `project` even.

StefanKarpinski:
Since the section is called `deps` it seems like `Pkg.deps` would be more consistent.

oxinabox:
Though it doesn't just list them.it also gets all their `PkgInfo`.

00vareladavid:
bors try

src/Pkg.jl
```diff
@@ -273,6 +270,41 @@ const develop = API.develop
 #TODO: Will probably be deprecated for something in PkgDev
 const generate = API.generate
 
+"""
+    Pkg.dependencies()::Dict{UUID, PackageInfo}
```
KristofferC: Should this include things only in the Project, or also in the Manifest or should that be an option?

src/Pkg.jl
```diff
@@ -273,6 +270,41 @@ const develop = API.develop
...
+| `version`      | The version of the package (this is `Nothing` for stdlibs) |
+| `isdeveloped`  | Whether a package is directly tracking a directory         |
+| `ispinned`     | Whether a package is pinned                                |
+| `source`       | The directory containing the source code for that package  |
+| `dependencies` | The dependencies of that package as a vector of UUIDs      |
+"""
+const dependencies = API.dependencies
+
+"""
+    Pkg.project()::ProjectInfo
```
KristofferC: I wonder if `ProjectInfo(::UUID)` would be a good constructor? Otherwise it is a bit hard to do something with the `.dependencies` field.

PR #1088, opened 2019-02-26
Automatically detect unregistered dependencies

### LINKED ISSUE #1005: Developing local packages can be a bit annoying
Let's say we have created some local unregistered packages (A, B, C, D) in the following dependency graph:

```` 
    A
   / \
  B    C
 /
D
```` 

If we want to `dev A` we need to "manually" resolve the dependency graph by doing:

```
dev path/to/D
dev path/to/B
dev path/to/C
dev path/to/A
```

even if `B, C, A` have Projects / Manifests so they can find all their dependencies. If we just do `dev path/to/A` straight away we will get an error that we don't know about `B` and `C`.

What we could do is have a step where we e.g. search in the `LOAD_PATH` for packages that we currently don't know about. So then we would have `push!(LOAD_PATH, path/to)` and then `dev path/to/A`. We would then find `B` and `C` in the `LOAD_PATH` and then `D` etc recursively.

### PR DESCRIPTION
**Pull request by 00vareladavid**
- [x] support `develop`ed packages
- [x] support packages which are tracking repos

resolves #1005 

---
Let us banish this issue to the past

Example:
```
(B) pkg> status
Project B v0.1.0
    Status `/tmp/B/Project.toml`
  (empty environment)

(B) pkg> dev /tmp/C
 Resolving package versions...
  Updating `/tmp/B/Project.toml`
  [3c22e19c] + C v0.1.0 [`/tmp/C`]
  Updating `/tmp/B/Manifest.toml`
  [3c22e19c] + C v0.1.0 [`/tmp/C`]

(B) pkg> activate A
[ Info: activating environment at `/tmp/A/Project.toml`.

(A) pkg> dev /tmp/B
 Resolving package versions...
  Updating `/tmp/A/Project.toml`
  [391d23c0] + B v0.1.0 [`/tmp/B`]
  Updating `/tmp/A/Manifest.toml`
  [391d23c0] + B v0.1.0 [`/tmp/B`]
  [3c22e19c] + C v0.1.0 [`/tmp/C`]
```

### DIFF
--- a/src/project.jl
+++ b/src/project.jl
@@ -36,7 +36,7 @@ function read_project_deps(raw::Dict{String,Any}, section_name::String)
             uuid = UUID(uuid)
         catch err
             err isa ArgumentError || rethrow()
-            pkgerror("Malfomed value for `$name` in `$(section_name)` section.")
+            pkgerror("Malformed value for `$name` in `$(section_name)` section.")
         end
         deps[name] = uuid
     end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -224,9 +224,9 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
     if haskey(compat, "julia") && !(VERSION in Types.semver_spec(compat["julia"]))
         @warn("julia version requirement for package $(pkg.name) not satisfied")
     end
-    for (deppkg_name, uuid) in project.deps
-        vspec = haskey(compat, deppkg_name) ? Types.semver_spec(compat[deppkg_name]) : VersionSpec()
-        deppkg = PackageSpec(deppkg_name, uuid, vspec)
+    for (name, uuid) in project.deps
+        vspec = haskey(compat, name) ? Types.semver_spec(compat[name]) : VersionSpec()
+        deppkg = PackageSpec(name, uuid, vspec)
         push!(fix_deps_map[pkg.uuid], deppkg)
     end
     if project.version !== nothing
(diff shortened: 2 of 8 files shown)

### CODE REVIEW

**Comment by 00vareladavid** [comment]
The sandbox isolation more heavily reuses existing `Pkg` mechanisms. This means that you can't use unregistered packages as test dependencies (I think there was special code for this before?). So there now exists even more motivation for a solution to be in place.

File: src/project.jl [hunk 1]
```diff
@@ -71,6 +71,19 @@ end
 read_project_compat(raw, project::Project) =
     pkgerror("Expected `compat` section to be a key-value list")
 
+read_project_unreg(::Nothing, section::String) = Dict{String,String}()
+function read_project_unreg(raw::Dict{String,Any}, section_name::String)
+    deps = Dict{String,String}()
+    for (name, path) in raw
+        path isa String ||
+            pkgerror("Malfomed value for `$name` in `$(section_name)` section.")
```
tpapp: `Malfomed` => `Malformed`

00vareladavid: Thanks :slightly_smiling_face:

StefanKarpinski:
Bump. What's the next step for this?

00vareladavid:
The purpose of these recent changes are so that we don't change the format of the Project file. `Pkg` will just check to see if there is a Manifest file available. If so it will walk down the dependency graph and collect any unregistered packages.

I also added some basics tests

00vareladavid:
bors try

KristofferC:
For packages tracking repos, this would only allow a "dependency chain length" of one, right? So if we have

```
A -[url]-> B -[url]-> C
```

where `B` is currently not downloaded, we would fail to find the manifest for `C` and error?

schlichtanders:
Implementing 8 julia packages in sync, this issue is one of the most common issues I encountered.

Developing nicely separated package structure seems not easily supported by now in Julia. 

I checked and tried a lot and have only found complicated solutions:
* adding all dependencies in correct order, i.e. having the need of a custom solution for what this pull request tries to solve
* having a private registry fails short because it is required to fix at least one version to a concrete git commit, however the version is not yet reliable but in development, i.e. everytime something new is developed, the registry is out of date. According to [this discourse question](https://discourse.julialang.org/t/can-versions-toml-point-to-git-branch-instead-of-fixed-git-commit/33442/7) it is intended that a registry always points to fixed versions and hence is not meant for development
* having a custom script which git clones all dependencies into .julia/dev (and adding .julia/dev onto the loadpath)
* [not an option at all for me] build a big and large monolithic julia package which has all packages in one package and makes it impossible to reuse well capsulated subfunctionalitiy
* [not an option any longer because I need to switch from windows to mac and need both to work for some time] you could develop everything locally with one user and one machine so that no one needs to be able to replicate your setting
* [not yet an option] this very pull request which would make it possible for everyone to just add the package (as soon as you have a sound manifest pointing to accessable repositories)

all this is surprisingly tedious and as a new developer for julia at least I myself feel a bit unwelcome by this needed complexity to develop clean code. This is one of these moments where I wished I found julia 2 years later ;-) luckily all my other moments of these kind are already resolved by now (I am intensively developt julia packages since 1.5 years), so I hope this one gets similarly resolved quite soon.

Being pointed to this merge request and realizing it is one year old is a bit hard for me to understand. It suggests that there already is an easy way for developers to developer new interdependent packages, but I looked hard and couldn't find anything but the above. 

Just to make clear, I don't have enough capacity to solve this merge request myself, because, as you can imagine, I have troubles getting my 6 or 7 interdependent julia packages to a state where I can make them public (switching from windows to mac right now brought me this issue with all its force on top of solving my own bugs and missing features)

*Hence I kindly want to suggest raising the priority of this pull request to high. Definitely something higher than "enhacnement".*

00vareladavid:
closing in favor of #1628

Issue #1030, opened 2019-02-01
test resolves from wrong manifest


fredrikekre:
I added `Example` to `Literate/test/Project.toml` (and no `test/Manifest.toml` file):
```
(v1) pkg> st
    Status `~/.julia/environments/v1/Project.toml`
  [7876af07] Example v0.5.0
  [98b081ad] Literate v1.0.4 [`~/dev/Literate`]

(v1) pkg> test Literate
[...]
 Resolving package versions...
  Updating `/tmp/tmpE0z0sa/Project.toml`
  [7876af07] + Example v0.5.1            <-- Should be v0.5.0 from my v1 environment.
  [8dfed614] + Test
```

00vareladavid:
Why does the `v1` env matter here? It should use whatever `Example` is in `Literate/test/Manifest.toml`, no?

fredrikekre:
Right, I should add that I did not have a `test/Manifest.toml` file.

fredrikekre:
It makes more sense to test the package in the installed env. If I have Example@0.5.0 I want to use that, no? This is how things behaved pre-#989.

(This is what I was trying to explain on the last call.)

fredrikekre:
So, IMO, it should behave like this if there is no `test/Manifest.toml`
1. Collect from $current_env/Manifest.toml
2. If still not fully satisfied, do a resolve.

00vareladavid:
I've thought about this some more. I'm worried reusing the parent dependency graph will be too brittle. We currently do not have much ability to gracefully degrade from the best case. In other words: we can try to reuse the dependency graph, and possibly we get lucky. If we don't get lucky, we basically just resolve from scratch. Its basically an "all or nothing" situation. I feel uncomfortable providing a feature that we don't have much control over. How strongly do you feel about supporting this behavior?

A possible alternative would be to match the `compat` section of specific nodes that you care about. I'm more comfortable with this because it would reuse mechanisms that are exposed outside of the `test` operation. This approach would of course be more manual, but (imo) also more transparent behavior.

fredrikekre:
I think this is important. Without this `pkg> test Package` does not mean anything if the package testing can use whatever versions they want.

00vareladavid:
It occurs to me that we are possibly thinking of different testing scenarios. I am focusing on testing a package **as a dependency**. It seems (correct me if I'm wrong) you are focusing on testing a package **as an application**.

[...]

PR #1093, opened 2019-02-28
Make sandbox isolation more flexible

00vareladavid:
After much discussion(#1030), we agreed to make the sandbox mechanism more flexible; specifically when it comes to testing. 

This change simply collects the test environment direct dependencies and tries to preserve them. This is in addition to the dependencies which are specified by `Pkg.test("SomeDep")`

---
I still believe there is merit to a more structured workflow, but that requires more reflection and feedback from users. It will be interesting to see how the relationship between `Pkg` and testing develops. But refining the technical aspects of the sandbox is more important at this point.


00vareladavid:
bors try

PR #1095, opened 2019-02-28
Print contents of build.jl in the documentation

krcools:
Both before and after the build step, the log file contents were displayed


KristofferC:
Oops, thanks!

Issue #1094, opened 2019-02-28
Make Pkg warn on case dependent file names


laborg:
It's completely fine to build a package on Linux that uses case sensitive files, but the same package will error on Mac OS and Windows (E.g. during test).

Having Pkg warn on files that only differ by lower/uppercase would, IMHO, be a nice feature.

With pointers to the source code I might be able to prepare a PR.

StefanKarpinski:
At what point would you want the error about files that differ only by case?

laborg:
At `test` appears natural. Or is there some kind of validation command (comparable to Rs "check")?

In this PR it's only implemented as a warning, but I don't see why it shouldn't be strict.
https://github.com/JuliaLang/Pkg.jl/pull/1099

laborg:
Are there any plans to add general package validation functions into Pkg.jl? See http://r-pkgs.had.co.nz/check.html for inspiration:

- Checking for hidden files and directories. 
- Checking for portable file names.
- Checking for sufficient/correct file permissions.
- Checking installed package size. 
- ...

Or would this be better fit into a new package (e.g. PkgChecks) ?

StefanKarpinski:
Yes, this should all be checked. Can you file an issue on [Registrator.jl](https://github.com/JuliaComputing/Registrator.jl)?

KristofferC:
Don't think there is anything left to do here.

PR #1099, opened 2019-03-01
RFC: Add warning for case dependent file distinction fix: #1094

laborg:
Having filenames which only differ by case leads to problems on case 
insensitive OSes.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1229,6 +1229,17 @@ function update_package_test!(pkg::PackageSpec, entry::PackageEntry)
     pkg.pinned = entry.pinned
 end
 
+function check_case_conflicts(pkg, source_path::String)
+    for (root, dirs, files) in walkdir(source_path)
+        paths = vcat(dirs, files)
+        uniquepaths = unique(lowercase, paths)
+        if length(uniquepaths) != length(paths)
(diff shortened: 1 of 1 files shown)


fredrikekre:
Seems like this check should be done when installing the package.

StefanKarpinski:
Isn’t it too late when installing? On a system where this is a problem you won’t be able to detect this because the problematic setup is impossible to create. On a system where this isn’t a problem it seems a bit weird to warn/error since it’s not a problem there. It seems more like something to check at registration time, no?

StefanKarpinski:
Doing it at test time is a decent time too.

laborg:
In my case it would have been helpful at test time, but having it (at any time) is better than the error message you get at the moment.

00vareladavid:
I would prefer to put this in the registration system, because the check is only run once. I'm resistant to crawling the directory tree on every `test`.

StefanKarpinski:
We probably want some reusable way to sanity check a package before trying to register it which can be reused and enforced during registration.

00vareladavid:
I think `RegistryCI` is a more appropriate location for this code. If so, can we close this and open an issue there?

KristofferC:
Yes, adding this to the registry checks feels like a better idea.

PR #1103, opened 2019-03-04
Build a pdf version of the Pkg.jl manual.

fredrikekre:
---
cc @mortenpi


fredrikekre:
bors try

mortenpi:
Might it be worth guarding the HTML too, so that you could locally just generate the PDF? E.g. `julia docs/make.jl` and `julia docs/make.jl html` generate HTML, `julia docs/make.jl pdf` generates PDF and `julia docs/make.jl html pdf` generates both?

This is a bit off-topic, but I've gotten the feeling that people really appreciate the ability to have PDF manuals. Hence I am wondering if we should retire DocumenterLaTeX and keep the PDF generation as a core feature of Documenter..

fredrikekre:
bors try

StefanKarpinski:
Bump—shall we merge this?

fredrikekre:
Yea, but I don't like the setup so I will change it, rebase and then we can merge.

PR #1023, opened 2019-01-30
Expand `Pkg` guide

00vareladavid:
This simplifies the existing text and adds more material, in order of increasing complexity.

PR #1104, opened 2019-03-05
Fix basedocs.md for building Julia's documentation.

fredrikekre:
Accidentally broken by #1023


fredrikekre:
This is the reason for the failure in https://github.com/JuliaLang/julia/pull/31116.

PR #1114, opened 2019-03-17
doc: coverage=false is the default

fredrikekre:


KristofferC:
Oh, AV nightly is dead...

KristofferC:
Not so tough now are we @bors?

PR #1127, opened 2019-04-04
doc: improve compat examples, fixes #1126

### LINKED ISSUE #1126: Better docs on compatibility version numbers
The [Compatibility section](https://julialang.github.io/Pkg.jl/v1/compatibility/) of the docs give a few examples on version ranges.

At one point, it gives an example as:

```
julia = "1.0"
```

Later, it gives examples as `>= 1.2.3` or `^1`, but it doest not say anything about the `"` character.

Is any of this valid?
```
julia = "^1.0"
julia = ^1.0
julia >= ^"1.0"
```

I would suggest to clarify that in the docs.

felipenoris:


fredrikekre:
Thanks!
bors r+

StefanKarpinski:
Thank you for the improvement! I wouldn't have thought to write it like this but it's much clearer.

PR #1128, opened 2019-04-04
Add `--strict` option to `add` and `dev`

### LINKED ISSUE #1107: add Package --strict
... which doesn't update anything and strictly only adds `Package`.

This will solve the problem, that it can currently take ~ 100s to add a single, 100 LOC package with no dependencies, when many packages are installed and update wasn't called in some time!

00vareladavid:
Adds API and REPL support for a `--strict` option, which simply preserves the whole, existing dependency graph.

Resolves #1107


fredrikekre:
bors try

fredrikekre:
Missing docs?

PR #1133, opened 2019-04-09
script to generate a project file

StefanKarpinski:


bin/gen_project.jl
```diff
@@ -0,0 +1,142 @@
...
+    !force && isfile(project_file) &&
+        @error "$arg already has a project file"
+
+    require_file = joinpath(dir, "REQUIRE")
+    isfile(require_file) ||
+        @error "$arg does not appear to be a package (no REQUIRE file)"
+
+    project = Dict(
+        "name" => name,
+        "uuid" => uuid(name),
```
timholy: If the package is in General, could one add a `"version"` field and set it to the most recently-registered version?

timholy: Is there a technical reason to avoid it, or something else?

Anyway, if your comment was meant to indicate that you intended to address this, it didn't seem to make it into the final version.

StefanKarpinski: No, I figured I'd do it in a separate PR. No reason to block merging this since it's an additional feature. This script is going to improve over time as people use it and provide feedback, which is another reason not to do all the project files at once. Better to get feedback incrementally as people register new versions.

PR #1134, opened 2019-04-09
Include test dependencies in the Project file

timholy:
Note this PR is against #1133. This also looks for a `test/REQUIRE` file and adds the things it finds to the "extras" section; if that ends up non-empty then it also creates a `test` `"target"`.

### DIFF
--- a/bin/gen_project.jl
+++ b/bin/gen_project.jl
@@ -114,23 +114,35 @@ for arg in ARGS
         "uuid" => uuid(name),
         "deps" => Dict{String,String}(),
         "compat" => Dict{String,String}(),
+        "extras" => Dict{String,String}(),
     )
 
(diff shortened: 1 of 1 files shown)


fredrikekre:
For Julia 1.0 compatibility you still need `[targets]` so I think we should generate that, since most packages will still want to support 1.0. `test/Project.toml` is also kinda alpha and has not really been tried in production.

StefanKarpinski:
This wasn’t meant to get merged... but ok I guess?

fredrikekre:
Hmm, maybe that happened because the PR was against `sk/gen_project` which got merged?

StefanKarpinski:
Maybe but I changed the base of that before merging it. Maybe that confused bors.

timholy:
To clarify, this didn't get merged. I'll re-issue it against master.

PR #1135, opened 2019-04-09
Allow comment lines in REQUIRE file

jheinen:


jheinen:
The final version should handle REQUIRE files with empty lines (as in [Plots.jl](https://github.com/JuliaPlots/Plots.jl/blob/master/REQUIRE)). This was the reason for my PR.

timholy:
There seems to have been some merge issue. I've re-issued my PR in #1137

StefanKarpinski:
I changed the Base of this PR because it should have been made against my PR (now merged to master) rather than Tim's PR which was orthogonal. Apparently the bors bot we use to merge PRs either ignores the base branch for PRs or it cannot handle changing the base of a PR.

PR #1139, opened 2019-04-10
Update gen_project.jl

jheinen:
Allow comment lines in REQURE file.


StefanKarpinski:
Merged in https://github.com/JuliaLang/Pkg.jl/pull/1140.

PR #1141, opened 2019-04-10
Add a dummy commit for Julia 1.3 for General deployment

ararslan:
Also add 1.0.3 and change 1.1.0 to use the actual commit.


ararslan:
@StefanKarpinski, there are some TOML files in here that need to be updated for this as well, right?

ararslan:
Seems Nanosoldier is broken without this change. See https://github.com/JuliaLang/julia/commit/62d7ec55fd8063c722bc79b72c6d20b34a887c39.

StefanKarpinski:
I cherry-picked this and pushed to master.

ararslan:
Thanks Stefan!

PR #1145, opened 2019-04-13
re-add section about making changes to existing documentation

mzaffalon:


docs/src/making-changes.md
```diff
@@ -14,75 +14,88 @@ shell> git config --global user.email "EMAIL"
...
-;git checkout -b myfixes         # create a branch for your changes
-<edit code>                      # be sure to add a test for your bug
-Pkg.test("Images")               # make sure everything works now
-;git commit -a -m "Fix foo by calling bar"   # write a descriptive message
-using PkgDev
-PkgDev.submit("Images")
+(v1.1) pkg> add Example#master           # check out the master branch
+(v1.1) pkg> develop --local Example      # make a local copy
+# change path to the directory
+shell> cd(?)
```
mzaffalon: The Pkg documentation mentions the variable `JULIA_PKG_DEVDIR` which is not defined if one types `ENV["JULIA_PKG_DEVDIR"]`. Is there an easy way to access it or would it make sense to simply define it?

docs/src/making-changes.md
```diff
@@ -14,75 +14,88 @@ shell> git config --global user.email "EMAIL"
...
-      * From the shell, run `julia ../test/runtests.jl` from within the package's `src/` folder.
-
+    + From Pkg-REPL, run [`test Example`](@ref): this will run your tests in a separate (new) `julia` process.
+    + From Julia, `include("runtests.jl")` from the package's `test/` folder (it's possible the file has a different name, look for one that runs all the tests): this allows you to run the tests repeatedly in the same session without reloading all the package code; for packages that take a while to load, this can be much faster. With this approach, you do have to do some extra work to make [changes in the package code](@ref man-workflow-tips).
+    + From the shell, run `julia ../test/runtests.jl` from within the package's `src/` folder.
+  * If you are making substantial changes to the documentation, and the package documentation makes use of a dependency program to generate the documentation (e.g. Documenter.jl), you should test that your changes generate the correct documentation. For example, for `Documenter,jl`, from within the directory `Example.jl/docs`, you would invoke in shell mode `julia -e 'push!(LOAD_PATH,"../src/")' make.jl`
```
mzaffalon: `julia -e 'push!(LOAD_PATH,"../src/")' make.jl` this command clearly does not work but I have no idea how to say that one has to push to LOAD_PATH otherwise.

docs/src/making-changes.md
```diff
@@ -14,75 +14,88 @@ shell> git config --global user.email "EMAIL"
...
+    * Alternatively, for committing and pushinig changes, one use the graphical interface `git gui` that by default comes with `git`.
   * The package owner may suggest additional improvements. To respond to those suggestions, you can easily update the pull request (this only works for changes that have not already been merged; for merged pull requests, make new changes by starting a new branch):
 
-      * If you've changed branches in the meantime, make sure you go back to the same branch with `git checkout fixbar` (from shell mode) or [`Pkg.checkout("Foo", "fixbar")`](@ref) (from the Julia prompt).
-      * As above, make your changes, run the tests, and commit your changes.
-      * From the shell, type `git push`.  This will add your new commit(s) to the same pull request; you should see them appear automatically on the page holding the discussion of your pull request.
+    + If you've changed branches in the meantime, make sure you go back to the same branch with `git checkout fixbar` (from shell mode) or [`Pkg.checkout("Foo", "fixbar")`](@ref) (from the Julia prompt).
```
mzaffalon: I will remove Pkg.checkout.

CameronBieganek:
Awesome! This looks pretty nice. I was thinking about adding this documentation myself, and then I found your PR.

I'm not sure how Julia folks normally do this, but I think a slightly different approach to making your local development package might be nicer. If you do `pkg> develop Example`, then you have to work on the local copy created in `~/.julia/dev`. I would prefer not to have to work within the hidden `.julia` directory. Or if you do `pkg> develop --local Example`, it will clone Example.jl in a `dev` folder within the current project. I could be wrong, but it seems to me that this option is intended for when you're developing package `A`, which depends on package `B`, but you need to make an update to package `B`. Then your local package `B` will be cloned in `~/A/dev/B/`.

So, the approach I prefer (until somebody more experienced than me tells me there's a better way), is to clone my forked GitHub repo in whichever directory I prefer, say `~/projects/Example.jl`, and then use `git remote add upstream ORIGINAL-REPO-URL` to add the original repo as the upstream remote. Then I can call `pkg> develop ~/projects/Example.jl` to use the changes to my local repo instead of the registered library repo.

While figuring out this workflow yesterday (which I haven't actually used yet for a real use case!), I summarized the workflow that I came up with on discourse, for future reference:

https://discourse.julialang.org/t/develop-a-local-version-of-a-registered-package-and-submit-changes-as-a-pull-request/24868

mzaffalon:
@CameronBieganek You are giving me credit for something I don't deserve. The text is mostly copy-pasted from a pre-existing document that disappeared when `Pkg.jl` was lifted from Julia base. I found that documentation relevant for people like me who are infrequent contributors.

I hope that your interest in this PR may bring it to the attention of the experienced users and be finally merged.

mzaffalon:
@fredrikekre I don't want to be pushy: is this PR still relevant? I haven't done myself development of existing packages in a while and I have lost the oversight of the procedure.

mzaffalon:
@fredrikekre Is this PR still relevant or can be closed?

PR #1154, opened 2019-04-16
Fix bug in dependency graph creation

carlobaldassi:
This fixes https://github.com/JuliaRegistries/General/pull/49. I'm rather surprised that this had never came up before...

### DIFF
--- a/src/GraphType.jl
+++ b/src/GraphType.jl
@@ -287,7 +287,9 @@ mutable struct Graph
             end
             # The remaining dependencies do not have compatibility constraints
             for uuid in values(n2u)
-                get!(req, pdict[uuid]) do; VersionSpec() end
+                p1 = pdict[uuid]
+                p1 == p0 && continue
+                get!(req, p1) do; VersionSpec() end
             end
             # Translate the requirements into bit masks
             # Hot code, measure performance before changing


fredrikekre:
Great, should make sure to get this into Julia 1.0 and 1.1!

KristofferC:
I think it is because all entries were forced to have some compatability constraints in METADATA.jl?

fredrikekre:
Hi @carlobaldassi , it turns out the error was because the package had a dependency on itself (see https://github.com/JuliaRegistries/General/pull/455), so maybe the assertion was correct?

KristofferC:
I don't think that should give an assertion error though. If any, a resolver error.

carlobaldassi:
Yes, any assertion error is really about going into panic mode, and it's meant just to avoid more nasty stuff like accessing memory out of bounds. I believe the fix to still be valid, but catching a self-dependency is something that should happen earlier than the solver (and throw an informative error, of course).

fredrikekre:
Yea, Pkg errors when you try to add the package itself as a dependency, so in this case the author must have manually edited the toml files.

dehann:
When I first created the IncrementalInference project tomls, I was not quite sure how to do so and used one of community scripts to convert a REQUIRE file to tomls (from discourse a few months ago).  The self dependency may have happened there.  Things were working fine with Attobot, but then quickly failed when transitioning to Registrator a couple days ago.  I expected the fix was easy, but actually had quite a bit of trouble understanding what Pkg was upset about -- but starting to get the hang of tomls now though.  Hope at least these few bugs (there actual 4 issues by my count) helps similar trajectories in the future.  Thanks again for the help!

PR #1156, opened 2019-04-16
Check for empty in gen_project.jl before using ARGS[1] to check for f…

ChrisRackauckas:
…orce

### DIFF
--- a/bin/gen_project.jl
+++ b/bin/gen_project.jl
@@ -87,13 +87,14 @@ function semver(intervals)
     return join(spec, ", ")
 end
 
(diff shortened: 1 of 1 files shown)


bin/gen_project.jl
```diff
@@ -87,13 +87,14 @@ function semver(intervals)
     return join(spec, ", ")
 end
 
+isempty(ARGS) && (push!(ARGS, pwd()))
```
fredrikekre: move this back down

StefanKarpinski:
Alternate fix: https://github.com/JuliaLang/Pkg.jl/pull/1157

PR #1162, opened 2019-04-17
Make `gen_project.jl` exit on finding Project.toml

asinghvi17:
The 'defined behaviour' as far as I am aware is:

> You call it with ARGS as directories where packages live. If the first argument is `-f` then it will overwrite existing `Project.toml` files, otherwise it will refuse to do so. It generates a `[compat]` section that approximates the bounds in the REQUIRE file as closely as possible, but unfortunately since the `[compat]` section doesn’t support version ranges this translation can be a bit messy. It is strongly encouraged that you manually review the `[compat]` section and edit by hand as appropriate.

(from Stefan's comment on Discourse).

### DIFF
--- a/bin/gen_project.jl
+++ b/bin/gen_project.jl
@@ -98,7 +98,7 @@ isempty(ARGS) && (push!(ARGS, pwd()))
 for arg in ARGS
     dir = abspath(expanduser(arg))
     isdir(dir) ||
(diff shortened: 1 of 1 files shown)


asinghvi17:
Will do, that makes sense actually.  I was wondering why it didn't exit on error...

Issue #90, opened 2017-12-27
Ensuring that pre-compilation occurs as part of the add and update process


jlperla:
I recently spent a some time with Julia v0.6 and Pkg2 and thinking like a new user to get a sense of usability issue.  Forgetting about the actual speed of packages, my general feeling is that the **perceived** slowness around packages comes down to 3 main things: (1) there is insufficient visual feedback during add/using/update operations; (2) expectations are not properly set that time is required for both the download/installation and the recompilation; (3) user steps are required between installation and pre-compilation.

For problem (2) and (3)  of perceived slowness, the users can become furious that after they wait out an update or add (with insufficient visual feedback that it is working, see #89 ) they try to use the package and suddenly they need to wait again (again without enough visual feedback on what is happening in the background).

This can be easily solved by two things: (a)  pre-compiling at the end of an add or update so all waiting takes place at the same time, and (b) setting expectations in the output that pre-compilation will take some time.   To do this:

- By default, precompile all packages upon add or update.  You could give an option that turns the automatic pre-compilation off (e.g. `Pkg.update(:precompile=false)` , but most starting users would want to leave it on, so the default should be precompile.
- With the `using` intended to automatically add packages now, I suspect that the separation between the activities would already help things out.
- As in #89, much more output during the precompilation process.  The extra output has the added benefit of grounding expectations that this might be a slow process.

KristofferC:
Yes, an option to precompile packages in batch is a nice feature to have. I wonder if it has to be a part of `Pkg.update` or if we could just have a "precompile all packages that need precompiling" as a separate function.

You could then easily run something like `Pkg.update(); Pkg.precompile()` to do the equivalent of what your `Pkg.update(;precompile = true)`. The advantage is of course that you can run `Pkg.precompile()` by itself if you do not want to do any updating.

yurivish:
Would it be possible to precompile in parallel with downloading? The [REPL video](https://www.youtube.com/watch?v=s8iRe_pSVYE) you linked to in @jlperla's other issue shows incremental installation progress – if it were possible to use the download time for later packages for compiling earlier ones, that would potentially result in massive improvements in both actual and perceived load time.

KristofferC:
It might be possible to try to find packages "at the bottom" e.g. with no dependencies, download these first and start precompiling and move our way upwards the dependency chain... It sounds a bit tricky and the Pkg3 download is usually fast enough that I am not sure how much time would be saved in practie.

[...]

PR #1163, opened 2019-04-18
semantic version compression & compat normalization script

StefanKarpinski:
This PR includes two things:

1. A script that normalizes all the `Compat.toml` files in `General` by loading then and then reapplying the version compression on both the keys and values en masse, thereby "normalizing" anything that's deviated for whatever reason from the standard way to express this.

2. A new version of `compress_versions` that is careful to produce a `VersionSpec` that respects gaps between versions that are considered "potentially breaking" according to semver. For example, if `A` claims compatibility with `B` versions `0.7.0`, `0.7.1`, `0.8.0`, `0.8.1` and `1.0.0` and those are the only versions that exist, then this will get compressed to `0.7-0.8, 1` whereas previously it would get compressed to `0.7-1` since there were no intervening incompatible versions.

We should merge this into Pkg and then deploy Registrator with this change after running the normalization script to update all the package `Compat.toml` files at once.

cc: @nkottary, @KristofferC, @fredrikekre

### DIFF
--- a/bin/normalize.jl
+++ b/bin/normalize.jl
@@ -1,0 +1,36 @@
+import Pkg: TOML, Compress
+import Pkg.Types: VersionSpec, compress_versions
+
+const registry_path = joinpath(DEPOT_PATH[1], "registries", "General")
+const registry_file = joinpath(registry_path, "Registry.toml")
+const packages = TOML.parsefile(registry_file)["packages"]
+
+const julia_uuid = "1222c4b2-2114-5bfd-aeef-88e4692bbb3e"
+const version_map = Dict{String,Vector{VersionNumber}}()
+
+for (uuid, info) in packages
+	path = joinpath(registry_path, info["path"])
+	versions_file = joinpath(path, "Versions.toml")
+	versions = Compress.load(versions_file)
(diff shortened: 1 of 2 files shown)


nkottary:
I downloaded julia 1.1 and patched this to `share/julia/stdlib/v1.1/Pkg`. Is there something else I need to do for the patch to take effect?

I tested this on Registrator with [#90](https://github.com/JuliaComputing/Registrator.jl/pull/90/files) for IPython and got [#131](https://github.com/JuliaRegistries/General/pull/131/files)

Without [#90](https://github.com/JuliaComputing/Registrator.jl/pull/90/files) I get [#46](https://github.com/nkottary/General/pull/46/files)

Both don't look like what we want.

StefanKarpinski:
This was applied to all of General here: https://github.com/JuliaRegistries/General/pull/252. We will probably ultimately want to drop the last commit and enforce strict semver when package authors make compatibility claims, but until we have improved the resolver and automation of package bound bumping, we're going to liberally consider `0.x` versions to be compatible for the purposes of inferring compat ranges at least. This should alleviate much of the "capping pain" that people have been having.

PR #1173, opened 2019-04-24
generate: use uuid4() instead of uuid1()

rfourquet:
---
@StefanKarpinski suggested on slack that `uuid4` is the recommended function for generating packages' UUID, so let's follow this recommendation in `Pkg.generate()`.


StefanKarpinski:
Try again, I guess? What's up with bors today?

PR #1178, opened 2019-04-28
Change the docs to use PrintFileTree instead of `tree` command!

NHDaly:
This changes printing the directory from using the `tree` command to using the `PrintFileTree` package.

Mostly just because it's fun to use Julia stuff where we can, but also maybe because people might be less comfortable installing commandline tools than installing julia packages. :)

Feel free to just close this, though, if you think it's silly. :)

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -11,12 +11,12 @@ To generate files for a new package, use `pkg> generate`.
 (v1.0) pkg> generate HelloWorld
 ```
 
-This creates a new project `HelloWorld` with the following files (visualized with the external [`tree` command](https://linux.die.net/man/1/tree)):
+This creates a new project `HelloWorld` with the following files (visualized with the [PrintFileTree.jl package](https://github.com/NHDaly/PrintFileTree.jl)):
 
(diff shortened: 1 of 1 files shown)


NHDaly:
Friendly ping! :) (I'm going through my old pull requests now, and came across this one. No biggie either way on this)

KristofferC:
I think this complicates things. We are talking about the very basics of a package and bringing up some other package at that point seems like it can be confusing. 

Also, the current version can actually be run by non Windows users while the other one cannot.

NHDaly:
Yeah the complexity is a fair point, and I think convincing enough to close the PR. :) Thanks!

> Also, the current version can actually be run by non Windows users while the other one cannot.

What do you mean by the other one? The PrintFileTree package should work on any OS, and has testing for mac, linux and windows: https://github.com/NHDaly/PrintFileTree.jl#printfiletree.
But oh, cool. I didn't know `tree` would work on windows, that's neat! :)

00vareladavid:
Carlsson has spoken! :stuck_out_tongue_closed_eyes:

Issue #948, opened 2018-12-04
Allow commas


crstnbr:
Is it only me that finds it annoying that it's

```julia
using Dates, BenchmarkTools, LinearAlgebra
```

but in `pkg` REPL mode it's

```julia
add Dates BenchmarkTools LinearAlgebra
```

It would be really convenient to also allow

```julia
add Dates, BenchmarkTools, LinearAlgebra
```

Not only would it be somewhat consistent but also nice for copy-paste. Same for `rm`.

StefanKarpinski:
Yes, this came up once before, we decided against, but I've thought better of it since copy-and-paste would be very handy this way. The only argument against I can think of is that I also want to auto-prompt the user to install packages that are `using`ed and `import`ed, which would make this far less necessary. But why not both?

fredrikekre:
> we decided against

I don't think we came to a conclusion, if you are talking about #589

StefanKarpinski:
Right, thought we'd come to one. I'd say let's just allow commas. Where's the harm? 😀

crstnbr:
A couple of students ran into this once again in a class I'm teaching. Unfortunately I won't be able to make a PR myself any time soon. Maybe this should be a "good first issue"?

Moelf:
Is this as simple as a `replace(args, ","=>" ")` somewhere in https://github.com/JuliaLang/Pkg.jl/blob/63fbfae17786e8cd825e6f0a58192cf434a66230/src/REPL/argument_parsers.jl ?

fredrikekre:
Maybe, why don’t you try?

PR #1187, opened 2019-05-09
allow comma in add, also tests

Moelf:
Fix https://github.com/JuliaLang/Pkg.jl/issues/948

(this is much harder than I thought, and maybe I'm still doing something wrong but all tests passed)

### DIFF
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -170,6 +170,7 @@ Base.@kwdef mutable struct Statement
 end
 
 function lex(cmd::String)::Vector{QString}
+    replace_comma = (nothing!=match(r"^(add|rm|remove)+\s", cmd))
     in_doublequote = false
     in_singlequote = false
     qstrings = QString[]
@@ -207,6 +208,14 @@ function lex(cmd::String)::Vector{QString}
             else # special delimiter
                 push_token!(false)
                 push!(qstrings, QString(";", false))
+            end
+        elseif c == ','
+            if in_doublequote || in_singlequote || !replace_comma# raw char
+                # don't replace ',' in quotes
+                push!(token_in_progress, c)
+            else
+                push_token!(false)
+                push!(qstrings, QString("", false))
             end
         else
             push!(token_in_progress, c)
--- a/src/REPLMode.jl
+++ b/src/REPLMode.jl
@@ -210,7 +210,7 @@ function lex(cmd::String)::Vector{QString}
                 push!(qstrings, QString(";", false))
             end
         elseif c == ','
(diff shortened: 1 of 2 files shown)


fredrikekre:
bors try

fredrikekre:
While this seems to work I think we need to be a bit more careful and only replace `','` with `' '` when we expect a list of package arguments. For example, I think that
```
pkg> status, Example
```
should still error, and everything inside a quoted string should not be touched.

Moelf:
I agree. Then it's a matter of where to patch this, we can add a `if startswith(` at the place, or we can play with `do_add!()`?

Moelf:
I added a section to `lex` to handle commas only when it's one of `add|rm|remove` && `,` is outside of quotes, tests seem passing. @fredrikekre take a look whenever, thx!

Moelf:
bors try

fredrikekre:
bors try

src/REPLMode.jl
```diff
@@ -208,6 +209,14 @@ function lex(cmd::String)::Vector{QString}
                 push_token!(false)
                 push!(qstrings, QString(";", false))
             end
+        elseif c == ','
+            if in_doublequote || in_singlequote || !replace_comma# raw char
```
StefanKarpinski:
```suggestion
            if in_doublequote || in_singlequote || !replace_comma # raw char
```
I don't really follow what this comment means...

Moelf: I'm not sure but it appears several time in the lexer
https://github.com/JuliaLang/Pkg.jl/blob/6d8f0727c8809e4b06da29acda8b55e1cfa4dcb1/src/REPLMode.jl#L199

StefanKarpinski: Ok, I guess carry on then. @00vareladavid, can you explain what the comment means?

00vareladavid: Yeah it means "not a delimiter" i.e. a character that will eventually be used as part of some identifier. Perhaps that should be more clear

00vareladavid:
@Moelf have you checked to see how well this plays with autocompletions? I took a quick shot look at adding commas a while back and I remember autocompletions being a problem. IIRC I took a different approach, so it might not be a problem at all with your implementation

Moelf:
Seems, working:
```
(Pkg) pkg> add Exam <Tab> ---> add Example
(Pkg) pkg> add Example, HTT <Tab> ---> add Example, HTTP
```

`remove` has the same behavior.

Moelf:
Can this be merged?

PR #1188, opened 2019-05-11
add Pkg.precompile

sivapvarma:
some discussion at [julia discourse](https://discourse.julialang.org/t/precompile-only-available-in-repl-mode/24115)

### DIFF
--- a/bin/gen_project.jl
+++ b/bin/gen_project.jl
@@ -139,9 +139,11 @@ for arg in ARGS
                 project["compat"][dep] = semver(req.versions.intervals)
             end
         end
+    end
 
+    for (srcdir, section) in (("src" => "deps"), ("test", "extras"))
         for stdlib in STDLIBS
-            if uses(dirname(file), stdlib)
+            if uses(joinpath(dir, srcdir), stdlib)
                 project[section][stdlib] = uuid(stdlib)
             end
         end


src/Pkg.jl
```diff
@@ -97,6 +97,18 @@ See also [`PackageSpec`](@ref).
...
 const add = API.add
 
+"""
+    Pkg.precompile()
+
+Precompile all the dependencies of the project by running `import` on all of them in a new process.
+
+# Examples
+```julia
+Pkg.precompile() # precompiles all the packages in this project.
```
fredrikekre: Don't think this is needed since it is included in just above:
```suggestion
Pkg.precompile()
```

fredrikekre:
Can you also add `Pkg.precompile` to the list here: https://raw.githubusercontent.com/JuliaLang/Pkg.jl/master/docs/src/api.md for example just after `Pkg.status`?

sivapvarma:
added `Pkg.precompile` in docs as well.

src/Pkg.jl
```diff
@@ -97,6 +97,18 @@ See also [`PackageSpec`](@ref).
 """
 const add = API.add
 
+"""
+    Pkg.precompile()
+
+Precompile all the dependencies of the project by running `import` on all of them in a new process.
```
KristofferC:
```suggestion
Precompile all the dependencies of the project.
```

It does not run `import` on them in a new process anymore.

sivapvarma: Done.

fredrikekre:
Derp. Why do GitHub close a PR when beeing force pushed to?

PR #1191, opened 2019-05-15
`up` should prune manifest

### LINKED ISSUE #1185: Pkg no longer removes unneeded packages from the Manifest file
https://discourse.julialang.org/t/travis-failure-when-testing-required-code-on-julia-v1-0/23986/6

00vareladavid:
Fixes #1185

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1049,6 +1049,7 @@ function up(ctx::Context, pkgs::Vector{PackageSpec}, level::UpgradeLevel)
     end
     check_registered(ctx, pkgs)
     resolve_versions!(ctx, pkgs)
+    prune_manifest(ctx.env)
     update_manifest!(ctx, pkgs)
     new_apply = download_source(ctx, pkgs)
     write_env(ctx) # write env before building
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -747,6 +747,21 @@ end
     end end
 end
 
+@testset "up should prune manifest" begin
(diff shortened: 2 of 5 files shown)


00vareladavid:
bors try

00vareladavid:
Still needs a test

src/API.jl
```diff
@@ -170,7 +170,7 @@ function up(ctx::Context, pkgs::Vector{PackageSpec};
 end
 
 resolve(ctx::Context=Context()) =
-    up(ctx, level=UPLEVEL_FIXED, mode=PKGMODE_MANIFEST, update_registry=false)
```
KristofferC: Why change this?

00vareladavid: Passing `PKGMODE_MANIFEST` to `up` will cause `Pkg` to fix all dependencies at their current version, not just the direct dependencies.

00vareladavid:
bors try

00vareladavid:
Bors failure is a network error, should be good to merge

src/API.jl
fredrikekre: But isn't that what resolve should do?

00vareladavid: Hmm, perhaps I misunderstand the intention behind `resolve`. In any case, that is really a separate issue (we should maybe have a test which checks the difference). I will change it change it back for the sake of merging this PR sooner.

00vareladavid:
bors try

PR #1194, opened 2019-05-18
Add verbose option to Pkg.instantiate

tkf:
This PR adds `verbose` option to `Pkg.instantiate`.  This is useful, for example,if you want to set up multiple projects in installation step of CI.  Using `Pkg.build(verbose=true)` would invoke unnecessary redundant builds while `Pkg.instantiate()` prints only very minimum log.  It can be fixed by supporting `verbose` in `Pkg.instantiate`.

### DIFF
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -936,4 +936,11 @@ end
     end end end
 end
 
+@testset "smoke test `instantiate --verbose`" begin
+    temp_pkg_dir() do project_path
+        Pkg.activate(project_path)
+        pkg"instantiate --verbose"
+    end
+end
+
 end # module
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -939,6 +939,7 @@ end
 @testset "smoke test `instantiate --verbose`" begin
     temp_pkg_dir() do project_path
         Pkg.activate(project_path)
+        Pkg.instantiate(verbose=true)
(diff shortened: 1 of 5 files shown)


fredrikekre:
Seems fine, can you also add the flag for the `pkg> instantiate` REPL command?

tkf:
Thanks. I added the `--version` flag.

00vareladavid:
bors try

test/repl.jl
```diff
@@ -936,4 +936,11 @@ end
     end end end
 end
 
+@testset "smoke test `instantiate --verbose`" begin
+    temp_pkg_dir() do project_path
+        Pkg.activate(project_path)
+        pkg"instantiate --verbose"
```
fredrikekre:
```suggestion
        pkg"instantiate --verbose"
        pkg"instantiate -v"
```

tkf:
I added `Pkg.instantiate(verbose=true)` to the smoke test.

00vareladavid:
Does anyone mind if I merge? Seems useful, simple, and doesn't really interfere with anything else

StefanKarpinski:
No, go for it.

bors r+

fredrikekre:
Squash?

StefanKarpinski:
So annoying that bors can't do that for you.

PR #1205, opened 2019-05-24
Require --diff for diff output in status

fredrikekre:
---
I think the current behaviour causes too much confusion, and you often just want to see the absolute status anyway. See e.g. https://discourse.julialang.org/t/pkg-status-and-project-toml-out-of-sync/24473, https://stackoverflow.com/questions/56050364/how-to-prevent-pkg-jl-to-hold-state-beyond-project-toml-and-manifest-toml etc.

Edit: https://github.com/JuliaLang/Pkg.jl/pull/1205/files?w=1 for easier review.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -490,12 +490,12 @@ end
 
 @deprecate status(mode::PackageMode) status(mode=mode)
 
-status(; mode=PKGMODE_PROJECT) = status(PackageSpec[]; mode=mode)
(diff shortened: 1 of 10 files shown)


StefanKarpinski:
I’m on board as long as I get a `-d` short option :)

fredrikekre:
Okay, will comment out that then.

fredrikekre:
bors try

KristofferC:
Merge?

Issue #1106, opened 2019-03-05
typeassert, expected VersionNumber, got Pkg.Types.VersionSpec


fredrikekre:
Not sure what happened here, but I'll put it here and see if I can reproduce it. In any case this could be a better error.
```
$ julia-master --project -e 'using Pkg; Pkg.activate("docs"); Pkg.instantiate(); Pkg.develop(PackageSpec(path = pwd()))'
[ Info: activating environment at `~/dev/Pkg/docs/Project.toml`.
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating registry at `~/.julia/registries/Staging`
  Updating git-repo `https://github.com/fredrikekre/Staging.git`
ERROR: TypeError: in typeassert, expected VersionNumber, got Pkg.Types.VersionSpec
Stacktrace:
 [1] load_urls(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /home/fredrik/dev/Pkg/src/Operations.jl:438
 [2] #download_source#66 at /home/fredrik/dev/Pkg/src/Operations.jl:562 [inlined]
 [3] download_source at /home/fredrik/dev/Pkg/src/Operations.jl:561 [inlined]
 [4] #instantiate#81(::Nothing, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::typeof(Pkg.API.instantiate), ::Pkg.Types.Context) at /home/fredrik/dev/Pkg/src/API.jl:479
 [5] instantiate at /home/fredrik/dev/Pkg/src/API.jl:461 [inlined]
 [6] #instantiate#80 at /home/fredrik/dev/Pkg/src/API.jl:458 [inlined]
 [7] instantiate() at /home/fredrik/dev/Pkg/src/API.jl:458
 [8] top-level scope at none:1
```

tkf:
I can reproduce this issue using Project.toml and Manifest.toml in https://gist.github.com/tkf/1fac12181ba095cb1ec08241c1659b62 with Julia 1.2.0-DEV.614 (719c0620e4) but not with Julia 1.1.  Procedure:

```console
$ git clone https://gist.github.com/tkf/1fac12181ba095cb1ec08241c1659b62 bug
$ cd bug
$ julia --project=.
...
(bug) pkg> instantiate
  Updating registry at `~/.julia/registries/General`
[...]

PR #1209, opened 2019-05-27
Check for consistent toml files before `instantiate`ing

00vareladavid:
So #1106 occurred because of an inconsistency between the toml files. This just explicitly checks for this case.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -467,6 +467,13 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
         pkgerror("manifest at $(ctx.env.manifest_file) does not exist")
     end
     Operations.prune_manifest(ctx.env)
+    for (name, uuid) in ctx.env.project.deps
+        get(ctx.env.manifest, uuid, nothing) === nothing || continue
+        pkgerror("`$name` is a direct dependency, but does not appear in the manifest.",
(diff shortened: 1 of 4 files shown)


00vareladavid:
bors try

00vareladavid:
bors try

KristofferC:
And just doing a `resolve` instead would just put the dep in the Manifest?

00vareladavid:
Yeah, the user's fix would depend on what they want: either remove the extra direct dependency from the project file or use `resolve` to correctly populate the manifest.

fredrikekre:
Maybe point the user to those solutions in the error message then?

src/API.jl
```diff
@@ -467,6 +467,12 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
         pkgerror("manifest at $(ctx.env.manifest_file) does not exist")
     end
     Operations.prune_manifest(ctx.env)
+    for (name, uuid) in ctx.env.project.deps
+        get(ctx.env.manifest, uuid, nothing) === nothing || continue
+        pkgerror("`$name` is a direct dependency, but does not appear in the manifest.",
+                 " If you intend `$name` to be a direct dependency, run `Pkg.resolve()` to populate the manifest.",
+                 " Otherwise, remove `$name` with `Pkg.rm(\"$name\")`.")
```
```suggestion
                 " Otherwise, remove `$name` with `Pkg.rm(\"$name\")` and instantiate again.")
```

PR #1215, opened 2019-06-02
Highlight a bit more the posibility of relative paths

### LINKED ISSUE #1214: Support relative paths for `dev` in Manifest
It would be very advantageous to tell Manifest.toml to use a relative path instead of a global for some cases, specifically for cases where the path to the package is a direct sub-path of the path of the `Manifest` file.

For example, I have a strcture like:
```
Folder
- src/PackageA/src...
- Project.toml
- Manifest.toml
```
I `dev` packageA but manifest saves the absolute path instead of relative to `Folder`. When I send my project folder to a colleague he cannot use `PackageA` anymore.

related with #867

Datseris:
Closes #1214

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -123,7 +123,7 @@ from that local repo are pulled when packages are updated.
 Note that changes to files in the local package repository will not immediately be reflected when loading that package.
 The changes would have to be committed and the packages updated in order to pull in the changes.
 
-In addition, the local path given could also be relative with respect to the `Manifest.toml` file.
+In addition, it is possible to add packages relatively to the `Manifest.toml` file, see [Developing Packages](@ref) for an example. 
 
 ### Developing packages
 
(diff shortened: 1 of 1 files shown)


docs/src/managing-packages.md
```diff
@@ -123,6 +123,8 @@ from that local repo are pulled when packages are updated.
 Note that changes to files in the local package repository will not immediately be reflected when loading that package.
 The changes would have to be committed and the packages updated in order to pull in the changes.
 
+In addition, the local path given could also be relative with respect to the `Manifest.toml` file.
```
KristofferC: It's relative to pwd I think?

docs/src/managing-packages.md
```diff
@@ -180,6 +182,17 @@ Note that if you add a dependency to a package that tracks a local path, the Man
 out of sync with the actual dependency graph. This means that the package will not be able to load that dependency since it is not recorded
 in the Manifest. To synchronize the Manifest, use the REPL command `resolve`.
 
+Similarly with `add`, `dev` can be given a path relative to the `Manifest.toml` file. Notice that to use add a relative path
```
KristofferC: Same as above, I don't think this is true.

Datseris:
Ok @KristofferC then I don't understand whether the existing state really solves #1214 .

What I want: 

To be able to *use* a package, which has a path relative to the `Manifest.toml` file, **not** relative to the current working directory. Since it is possible to activate a specific project while the working directory has no relation whatsoever to the `Manifest` of the activated project, I must be certain that the package loaded by the Manifest file is really defined with respect to the Manifest file.

What happens now:

You may add a package to your manifest, relative to the *current* working directory, which may have no relation to manifest whatsoever.

So the question is:

Does this achieve what I want? If it does then it is only a matter of documenting it properly by changing the wording of my PR.

fredrikekre:
The path that is stored is relative the project file.

Datseris:
Thanks, this clarifies everything. It is a bit unintuitive that when you *add* a package, the path you add is relative to `pwd` but when you load one the path is relative to the Project.toml. But still, I will try to be transparent about this in the PR.

Datseris:
I've corrected my PR. Please let me know if it is correct now :)

Datseris:
Hello @fredrikekre , will this be merged ?

fredrikekre:
Yes

Datseris:
Perfect, thanks! Once it is merged I'll link it to the documentation of DrWatson, because we already had a user with a related question (https://github.com/JuliaDynamics/DrWatson.jl/issues/62)

Datseris:
bump @fredrikekre

00vareladavid:
This looks good to merge

PR #1220, opened 2019-06-06
Allow path starting with ~/ to be used as a repo-url

tkf:
This PR lets you specify a directory under `$HOME` using `~/` for `repo-url`. It would be useful when sending `Manifest.toml` containing local paths to different users and when such repository does not have a stable public URL. Git CLI also lets you use a path starting with `~/` as `remote.$name.url` so I suppose this is not so crazy idea?

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -83,6 +83,9 @@ function setprotocol!(;
 end
 
 function normalize_url(url::AbstractString)
+    if !Sys.iswindows() && startswith(url, "~/")
+        url = expanduser(url)
+    end
     m = match(GIT_REGEX, url)
     m === nothing && return url
 
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -791,6 +791,23 @@ end
     end end
 end
 
+if !Sys.iswindows()
+    @testset "resolve ~/ in repo-url" begin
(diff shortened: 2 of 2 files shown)


StefanKarpinski:
@KristofferC: you added the repo-url stuff IIRC. How do you feel about this?

src/GitTools.jl
```diff
@@ -83,6 +83,9 @@ function setprotocol!(;
 end
 
 function normalize_url(url::AbstractString)
+    if !Sys.iswindows() && startswith(url, "~/")
```
fredrikekre:
```suggestion
    url = expanduser(url)
```

src/GitTools.jl
```diff
@@ -83,6 +83,9 @@ function setprotocol!(;
...
+        url = expanduser(url)
```
fredrikekre:
```suggestion
```

src/GitTools.jl
```diff
@@ -83,6 +83,9 @@ function setprotocol!(;
...
+    end
```
fredrikekre:
```suggestion
```

fredrikekre:
bors try

IanButterworth:
I think we have all this now with `[sources]` right?

PR #1058, opened 2019-02-13
fix "locking" dependencies when building / testing current project

KristofferC:

PR #1222, opened 2019-06-08
Don't put recursive dependencies in the test/build project, fixes #1144.

### LINKED ISSUE #1144: Test deps not required on nightly
e.g. https://travis-ci.org/JuliaGraphs/LightGraphs.jl/builds/519328333

fredrikekre:


fredrikekre:
The test case added in #1058 actually demonstrates that it was broken since `JSON` should not be in the project there at all :)

bors try

fredrikekre:
bors try

KristofferC:
Ouch. Thanks for cleaning up my mess here.

PR #1226, opened 2019-06-13
[RFC] Allow Pkg.test to pass command-line arguments to test/runtests.jl

### LINKED ISSUE #518: Pkg test with arguments for testing packages
Pkg REPL in Julia-0.7.0 allows only to pass the Package name as argument. But there is no way to give an argument (or a list of arguments) to the test script from Pkg REPL.

From bash, we can do (from the corresponding package directory)
```
$ julia ./test/runtests.jl <args>
```
Even though it is possible to give it through julia REPL, like this
```
ARGS = [ <args> ]; include( joinpath(Pkg.dir(<PackageName>),"test","runtests.jl") )
```
but produces the following warning:
```
 Warning: Pkg.dir is only kept for legacy CI script reasons
└ @ Pkg.API /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v0.7/Pkg/src/API.jl:433
```

### LINKED ISSUE #981: testsets argument to Pkg.test()
@rofinn commented on [Tue Mar 08 2016](https://github.com/JuliaLang/julia/issues/15404)

With the new testsets approach it would be nice if `Pkg.test` took an optional list of testset names to run.

Ex) `Pkg.test(Foo, ["unit", "integration", "lint"]; coverage=true)`

---

@tkelman commented on [Tue Mar 08 2016](https://github.com/JuliaLang/julia/issues/15404#issuecomment-193985292)

`Pkg.test` right now makes no assumptions about the contents of `runtests.jl`. This might be better as a separate test-runner function that would take a file name? Maybe `Pkg.test` could take a list of `ARGS` though, then the contents of `runtests.jl` could handle them however you like.

---

@davidanthoff commented on [Thu Aug 18 2016](https://github.com/JuliaLang/julia/issues/15404#issuecomment-240777010)

Just want to second this issue, any ability to specify the testset to run would be fantastic.

---

@kshyatt commented on [Thu Aug 18 2016](https://github.com/JuliaLang/julia/issues/15404#issuecomment-240864507)

After https://github.com/JuliaLang/julia/pull/17165 is good to go we can add this in a new PR, I think. To that effect: if people can find more ways to break that PR/review it for bad things, it would be great!

---

@musm commented on [Thu Nov 17 2016](https://github.com/JuliaLang/julia/issues/15404#issuecomment-261316450)

A lot of people have shown interest for such a feature, e.g.: https://discourse.julialang.org/t/test-specific-test-sets-in-julia-v0-5/378/3 and https://discourse.julialang.org/t/standard-ways-to-run-common-commands/154 

---

@kkmann commented on [Mon Feb 06 2017](https://github.com/JuliaLang/julia/issues/15404#issuecomment-277670365)

Me too! This would be such a time-saver during development instead of just switching includes on and off by commenting ;)

---

@PearGod commented on [Wed Jan 02 2019](https://github.com/JuliaLang/julia/issues/15404#issuecomment-450966742)

Pinging this as it would be very valuable and we're coming up on two years with no activity on it.

### LINKED ISSUE #1241: Test specific groups of tests
Can we have something like `pkg> test MyPkg tests_names`, where `tests_names` refers to, say, the name of a `testset`? Some packages have very long tests which can take a long time to complete. When you are making modifications to specific part of it, you only want to test the affected areas repeatedly and hopefully quickly, and then maybe do a full test at the end.

DilumAluthge:
Fixes #518 
Fixes #981 
Fixes #1241   

---

In a source build of Julia, you can run specific test sets as such:

```
./julia test/runtests.jl bitarray math
```

In a binary install, you can accomplish the same thing with:

```
Base.runtests(["bitarray", "math"])
```

`Base.runtests(tests)` accomplishes this (see [here](https://github.com/JuliaLang/julia/blob/b7774095f62e47bb2c8467b8a328075ab5a3fb89/base/util.jl#L813)) by passing the contents of the first positional argument `test` as the command line arguments to `test/runtests.jl`. 

However, as far as I can tell, there is currently no functionality built into `Pkg` that allows you to pass any arguments to a package's `test/runtests.jl` file when running the package tests.

This pull request adds two features:
- Pass flags to the `julia` process that runs a package's tests. The flags are specified in the `julia_flags` keyword argument to `Pkg.test`.
- Pass positional command-line arguments to `test/runtests.jl`. The arguments are specified in the `test_args` keyword argument to `Pkg.test` and are available inside `test/runtests.jl` as `ARGS[1]`, `ARGS[2]`, etc.

Example usage:

```julia
# equivalent to running `julia test/runtests.jl`
Pkg.test("HelloWorld")
```

```julia
# equivalent to running `julia test/runtests.jl foo bar baz`
Pkg.test("HelloWorld"; test_args = ["foo", "bar", "baz"])
```

```julia
# equivalent to running `julia --math-mode=ieee test/runtests.jl`
Pkg.test("HelloWorld"; julia_flags = ["--math-mode=ieee"])
```

```julia
# equivalent to running `julia --math-mode=ieee test/runtests.jl foo bar baz`
Pkg.test("HelloWorld"; test_args = ["foo", "bar", "baz"], julia_flags = ["--math-mode=ieee"])
```

---

This pull request only adds the functionality for passing the command line arguments to `test/runtests.jl`. It is up to the author of a package to add the logic for processing the command line arguments inside their `test/runtests.jl` file. However, this should be relatively straightforward. For example, if this is the current `test/runtests.jl` file:

```julia
using HelloWorld
using Test

@testset "HelloWorld.jl" begin
    @testset "foo" begin
        @test HelloWorld.foo(1) == 2
    end
    @testset "bar" begin
        @test HelloWorld.bar(2) == 6
    end
    @testset "baz" begin
        @test HelloWorld.baz(3) == 12
    end
end
```

Then you could modify it to look like this:

```julia
using HelloWorld
using Test

@testset "HelloWorld.jl" begin
    if isempty(ARGS) || "all" in ARGS
        all_tests = true
    else
        all_tests = false
    end
    if all_tests || "foo" in ARGS
        @testset "foo" begin
            @test HelloWorld.foo(1) == 2
        end
    end
    if all_tests || "bar" in ARGS
        @testset "bar" begin
            @test HelloWorld.bar(2) == 6
        end
    end
    if all_tests || "baz" in ARGS
        @testset "baz" begin
            @test HelloWorld.baz(3) == 12
        end
    end
end
```

And now if you run `Pkg.test("HelloWorld"; test_args = ["foo", "baz"])`, then only the `foo` and `baz` testsets will be run.


KristofferC:
I like it. We'll have to discuss a bit on the exact API for the command line args but this feels pretty ok.

bors try

KristofferC:
bors try

KristofferC:
We probably also want to be able to give julia startup command line options. For example `--track-allocations`.

src/Operations.jl
```diff
@@ -1228,7 +1229,10 @@ end
 
 testdir(source_path::String) = joinpath(source_path, "test")
 testfile(source_path::String) = joinpath(testdir(source_path), "runtests.jl")
-function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, test_fn=nothing)
+function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, test_fn=nothing, args=String[])
+    if isa(args, AbstractString)
```
fredrikekre: I know this is how `Base.runtests` does this, but its ugly and I think we should restrict to `Vector{String}`.

DilumAluthge: Makes sense to me. But could we do `AbstractVector{<:AbstractString}` instead of `Vector{String}`?

DilumAluthge:
> We probably also want to be able to give julia startup command line options. For example --track-allocations. So perhaps there should be julia_args, test_args or something?

I like `julia_args`

src/Operations.jl
fredrikekre: Sure.

DilumAluthge: Done!

DilumAluthge: (Will push shortly)

DilumAluthge:
I renamed `args` to `test_args`, and I added a new keyword argument `julia_flags`.

fredrikekre:
Is it not better with `julia_args` and `test_args`?
We should also think of a Pkg REPL syntax for this. I am thinking only `test_args` needs to be supported, since `julia_args` should not be so commonly used. Maybe something like
```
pkg> test Example -- arg1 arg2
```
or something, i.e. using `--` as a separator. This could be implemented in a separate PR of course.

DilumAluthge:
I renamed `julia_flags` to `julia_args`!

DilumAluthge:
> We should also think of a Pkg REPL syntax for this. I am thinking only test_args needs to be supported, since julia_args should not be so commonly used.

Could we support both? Maybe something like:
```julia
pkg> test Example --julia-args --track-allocation=all --test-args arg1 arg2
```

But yeah I definitely agree that we should implement this in a separate pull request.

StefanKarpinski:
That looks quite a bit worse to me. At that point, shouldn't you just be calling the test script manually?

fredrikekre:
I agree, we have not even had a feature request for customizing julia flags

DilumAluthge:
> That looks quite a bit worse to me. At that point, shouldn't you just be calling the test script manually?

Fair enough. So if we add REPL support, it will only be for test args.

src/Operations.jl
```diff
@@ -1130,10 +1130,14 @@ function free(ctx::Context, pkgs::Vector{PackageSpec})
     end
 end
 
-function gen_test_code(testfile::String; coverage=false)
+function gen_test_code(testfile::String;
+        coverage=false,
+        julia_args::AbstractVector{<:AbstractString}=String[],
```
fredrikekre: Maybe this should be `AbstractVector{Cmd}`? And you specify it as
```
Pkg.test("Example"; julia_args = [`--startup-file=no`, `-O3`])
```
?

src/Operations.jl
```diff
@@ -1130,10 +1130,14 @@ function free(ctx::Context, pkgs::Vector{PackageSpec})
...
+        test_args::AbstractVector{<:AbstractString}=String[])
     code = """
         $(Base.load_path_setup_code(false))
         cd($(repr(dirname(testfile))))
+        append!(empty!(ARGS), $(repr(test_args)))
```
fredrikekre: Could `repr` here fail for non-`Vector{String}`? Maybe we should just collect the arguments to `Vector{String}` at the entry point.

src/Operations.jl
```diff
@@ -1145,7 +1149,7 @@ function gen_test_code(testfile::String; coverage=false)
         --inline=$(Bool(Base.JLOptions().can_inline) ? "yes" : "no")
         --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
         --track-allocation=$(("none", "user", "all")[Base.JLOptions().malloc_log + 1])
-        --eval $code
+        $(julia_args) --eval $code
```
fredrikekre:
```suggestion
        $(julia_args) 
        --eval $code
```

src/API.jl
```diff
@@ -229,7 +229,9 @@ test(pkg::Union{AbstractString, PackageSpec}; kwargs...) = test([pkg]; kwargs...
 test(pkgs::Vector{<:AbstractString}; kwargs...)          = test([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
 test(pkgs::Vector{PackageSpec}; kwargs...)               = test(Context(), pkgs; kwargs...)
 function test(ctx::Context, pkgs::Vector{PackageSpec};
-              coverage=false, test_fn=nothing, kwargs...)
+              coverage=false, test_fn=nothing,
+              julia_args::AbstractVector{<:AbstractString}=String[],
+              test_args::AbstractVector{<:AbstractString}=String[], kwargs...)
```
fredrikekre: I suggest to do something like
```
julia_args = String[a for a in julia_args]
test_args = String[a for a in test_args]
```
here at the entry point and use `Vector{String}` only for the internal methods. Or
```
julia_args = Cmd[a for a in julia_args]
```
if you go with the `Cmd` approach.

src/Operations.jl
fredrikekre:
```suggestion
        julia_args::Vector{String}=String[],
```
or
```suggestion
        julia_args::Vector{Cmd}=Cmd[],
```

src/Operations.jl
```diff
@@ -1228,7 +1232,10 @@ end
 
 testdir(source_path::String) = joinpath(source_path, "test")
 testfile(source_path::String) = joinpath(testdir(source_path), "runtests.jl")
-function test(ctx::Context, pkgs::Vector{PackageSpec}; coverage=false, test_fn=nothing)
+function test(ctx::Context, pkgs::Vector{PackageSpec};
+        coverage=false, test_fn=nothing,
+        julia_args::AbstractVector{<:AbstractString}=String[],
```
fredrikekre:
```suggestion
        julia_args::Vector{String}=String[],
```
or
```suggestion
        julia_args::Vector{Cmd}=Cmd[],
```

src/Operations.jl
```diff
@@ -1228,7 +1232,10 @@ end
...
+        test_args::AbstractVector{<:AbstractString}=String[])
```
fredrikekre:
```suggestion
        test_args::Vector{String}=String[])
```

src/Pkg.jl
```diff
@@ -139,13 +139,16 @@ const update = API.up
 
 
 """
-    Pkg.test(; coverage::Bool=false)
-    Pkg.test(pkg::Union{String, Vector{String}; coverage::Bool=false)
-    Pkg.test(pkgs::Union{PackageSpec, Vector{PackageSpec}}; coverage::Bool=false)
+    Pkg.test(; coverage::Bool=false, julia_args::AbstractVector{<:AbstractString}=String[], test_args::AbstractVector{<:AbstractString}=String[])
```
fredrikekre: This looks a bit ugly, maybe rewrite it as 
```
    Pkg.test(; kwargs...)
```
etc and explain allowed kwargs in a bulleted list below

test/sandbox.jl
```diff
@@ -17,6 +17,27 @@ test_test(fn; kwargs...)       = Pkg.test(;test_fn=fn, kwargs...)
...
+            @test true
+        end
+    end end
+    temp_pkg_dir() do project_path; mktempdir() do tmp
+        copy_test_package(tmp, "RuntestsArgs")
+        Pkg.activate(joinpath(tmp, "RuntestsArgs"))
+        test_test(; test_args=["a", "b", "c"], julia_args=["--handle-signals=yes", "--warn-overwrite=no", "--math-mode=ieee"]) do
+            @test true
+        end
+    end end
```
fredrikekre: Tests here should test the actual API; e.g. with `Pkg.test()`

test/test_packages/RuntestsArgs/Manifest.toml
```diff
@@ -0,0 +1,2 @@
+# This file is machine-generated - editing it directly is not advised
```
fredrikekre: rm

test/test_packages/RuntestsArgs/test/Manifest.toml
```diff
@@ -0,0 +1,33 @@
+# This file is machine-generated - editing it directly is not advised
```
fredrikekre: rm this file too?

test/test_packages/RuntestsArgs/test/runtests.jl
```diff
@@ -0,0 +1,6 @@
+using RuntestsArgs
+using Test
+
+@testset "RuntestsArgs.jl" begin
+    @test all(ARGS .== RuntestsArgs.expected_arguments())
+end
```
fredrikekre: I would also add tests for julia_args, e.g. by inspecting `Base.JLOptions()` for some values that you set from `Pkg.test(; julia_args = ...)`

src/Operations.jl
DilumAluthge: I like it.

src/Operations.jl
DilumAluthge: I've implemented your `julia_args = String[a for a in julia_args]`  and `test_args = String[a for a in test_args]` ideas from below, which should solve this.

test/test_packages/RuntestsArgs/Manifest.toml
DilumAluthge: Done.

test/test_packages/RuntestsArgs/test/Manifest.toml
DilumAluthge: Done.

test/test_packages/RuntestsArgs/test/runtests.jl
DilumAluthge: Done.

test/sandbox.jl
DilumAluthge: Done.

src/API.jl
DilumAluthge: Done.

src/Operations.jl
DilumAluthge: Done.

src/Operations.jl
DilumAluthge: Done.

src/Pkg.jl
DilumAluthge: Done.

DilumAluthge:
I've made the requested changes!

Also, in the interest of "eating our own dog food", I've updated the `Pkg` test suite to support running specific testsets. I've [added an example](https://github.com/DilumAluthge/Pkg.jl/blob/db30f565e5c5737287f2aa17b7544fb0d794123f/.travis.yml#L36) to `.travis.yml` to demonstrate.

cmcaine:
> That looks quite a bit worse to me. At that point, shouldn't you just be calling the test script manually?

If you want to run a test script manually with julia I think you have to Pkg.add() all of the deps and extras for the package before it will work, so it could be a moderate pain to do.

Kind of off-topic:

Could solve that in another PR by exposing (if they are not already exposed) some functions for getting the right LOAD_PATH from a Project.toml (like activate, but including extras, and possibly not actually changing LOAD_PATH for you, just returning what you should use).

An extra use case for this is running ancillary test scripts (e.g. `test/compare_examples.jl` in Gadfly (it diffs binary outputs that have been saved to disk from master and the feature branch for regression tests)).

fredrikekre:
> Kind of off-topic:
> Could solve that in another PR by exposing (if they are not already exposed) some functions for getting the right LOAD_PATH from a Project.toml (like activate, but including extras, and possibly not actually changing LOAD_PATH for you, just returning what you should use).

This is exposed(ish) in Julia 1.2; you can there run tests as
```
julia --project=test test/runtests.jl
```

cmcaine:
As mentioned on slack, I think that would require copying the Project.toml to the test directory and making some manual changes first.

DilumAluthge:
I think what @fredrikekre is referring to is that in Julia 1.2 and higher, you specify the test dependencies in `test/Project.toml`, not in the main `Project.toml` file.

fredrikekre:
Looks pretty good now, but

> Also, in the interest of "eating our own dog food", I've updated the Pkg test suite to support running specific testsets. I've added an example to .travis.yml to demonstrate.

can we undo that? I don't think that is necessary and just results in a huge diff.

DilumAluthge:
> > Also, in the interest of "eating our own dog food", I've updated the Pkg test suite to support running specific testsets. I've added an example to .travis.yml to demonstrate.
> 
> can we undo that? I don't think that is necessary and just results in a huge diff.

Alright I think I've reverted all of those changes!

fredrikekre:
bors try

fredrikekre:
I simplified the tests, and fixed running the old code-path.

bors try

rfourquet:
This is really neat, thanks! Also, my +1 to Pkg REPL support for `test_args`.

Issue #843, opened 2018-10-18
semver syntax: support hyphen ranges


StefanKarpinski:
In https://github.com/JuliaLang/Pkg.jl/issues/751 it came up that it would be nice if the semver version spec parsing supported `0.7-1` hyphen range support (as the call it in npm). Quoth @KristofferC:

> Adding things to
> 
> https://github.com/JuliaLang/Pkg.jl/blob/dfb341ae03e34ae81a5f8c2e324b7b14019bdbbe/src/versions.jl#L340-L345
>
> should be fairly simple.

PR #1232, opened 2019-06-17
Allow ranges in spec

goretkin:
Attempt at addressing https://github.com/JuliaLang/Pkg.jl/issues/843

### DIFF
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -85,5 +85,5 @@ Ranges can also be used to specify version ranges:
 
 ```toml
 [compat]
-PkgA = "0.1 - 0.2.3" # [0.1.0,  0.2.3)
+PkgA = "0.1 - 0.2.3"   # [0.1.0,  0.2.3]
 ```
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -78,3 +78,12 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
 PkgC = "= 1.2.3"  # [1.2.3, 1.2.3]
 PkgD = "< 1.2.3"  # [0.0.0, 1.2.2]
 ```
+
+### Range specifiers
+
+Ranges can also be used to specify version ranges:
+
+```toml
+[compat]
+PkgA = "0.1 - 0.2.3" # [0.1.0,  0.2.3)
+```
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -299,11 +299,19 @@ function semver_spec(s::String)
     for ver in split(s, ',')
         range = nothing
         found_match = false
-        for (ver_reg, f) in ver_regs
-            if occursin(ver_reg, ver)
-                range = f(match(ver_reg, ver))
-                found_match = true
-                break
+        try
+            range = VersionRange(ver)
+            found_match = true
+        catch err
+            (err isa ArgumentError) || rethrow()
+        end
+        if !found_match
+            for (ver_reg, f) in ver_regs
+                if occursin(ver_reg, ver)
+                    range = f(match(ver_reg, ver))
+                    found_match = true
+                    break
+                end
             end
         end
         found_match || error("invalid version specifier: $s")
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -109,11 +109,13 @@ VersionRange(b::VersionBound=VersionBound()) = VersionRange(b, b)
 VersionRange(t::Integer...)                  = VersionRange(VersionBound(t...))
 VersionRange(v::VersionNumber)               = VersionRange(VersionBound(v))
 
-function VersionRange(s::AbstractString)
+function VersionRange(s::AbstractString; force_range=false)
     m = match(r"^\s*v?((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*)(?:\s*-\s*v?((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*))?\s*$", s)
     m == nothing && throw(ArgumentError("invalid version range: $(repr(s))"))
     lower = VersionBound(m.captures[1])
-    upper = m.captures[2] != nothing ? VersionBound(m.captures[2]) : lower
+    has_upper = m.captures[2] != nothing
+    force_range && !has_upper && throw(ArgumentError("degenerate version range: $(repr(s))"))
+    upper = has_upper ? VersionBound(m.captures[2]) : lower
     return VersionRange(lower, upper)
 end
 
@@ -300,7 +302,9 @@ function semver_spec(s::String)
         range = nothing
         found_match = false
         try
-            range = VersionRange(ver)
+            # "1.2.3" parses as a VersionRange with one version, unless force_range=true.
+            # It should be instead parsed as a VersionSpec("1.2.3-1") above.
+            range = VersionRange(ver; force_range=true)
             found_match = true
         catch err
             (err isa ArgumentError) || rethrow()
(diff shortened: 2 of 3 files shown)


docs/src/compatibility.md
```diff
@@ -78,3 +78,12 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
...
 PkgD = "< 1.2.3"  # [0.0.0, 1.2.2]
 ```
+
+### Range specifiers
+
+Ranges can also be used to specify version ranges:
+
+```toml
+[compat]
+PkgA = "0.1 - 0.2.3" # [0.1.0,  0.2.3)
```
StefanKarpinski: This is not what a range should mean: the second end-point should be inclusive, and any unspecified trailing numbers should be considered to be wildcards. I.e.
* `0.1-0.2.3` means `[0.1, 0.2.3]`
* `0.1-0.2` means `[0.1, 0.3)`
* `0.1-0` means `[0.1, 1.0)`

This seems to be what you've implemented, but it would be good to have some tests and, of course, the comment here should be fixed.

goretkin: Thanks for taking a look. I didn't notice the ) vs the ]. I just copied in pasted from the doc above.

Did you see the tests already in the commit? I am not really familiar with semver, so if you could recommend any additional tests, that would be helpful.

goretkin:
Hi @fredrikekre, would you mind clarifying whether you are thumbs-downing the feature, or the implementation?

fredrikekre:
Is it really so annoying to list all breaking versions you are compatible with?

StefanKarpinski:
@fredrikekre, why the thumbs down? If I have to guess it's that you don't want to support range syntax in compat sections. The current situation where we have non-overlapping syntaxes for version specs in the registry and in the compat section is a bit of a mess and this fixes at least half of that. We should have a single syntax for specifying version sets that works everywhere. We can encourage/discourage using certain parts of the syntax in different places, but having different syntaxes in different places seems awful to me.

goretkin:
@fredrikekre I don't know if it's annoying. So that I can get a better idea of what you mean, can you tell me how I would specify: `0.5.3-0.10` currently?
(Actual example, EnhancedGJK breaks with StaticArrays 0.11 
https://github.com/JuliaRobotics/EnhancedGJK.jl/blob/f6a6553806a2efc56b0ac5bfd942708492877f06/Project.toml#L18)

fredrikekre:
> can you tell me how I would specify: 0.5.3-0.10 currently?

```toml
[compat]
StaticArrays = "0.5.3, 0.6, 0.7, 0.8, 0.9, 0.10"
```
In this case it becomes a bit long, but it should be **really** unusual to support more than one, maybe two breaking versions of a dependency, so the list will normally don't be that long.

goretkin:
I suppose I don't really understand why there are two syntaxes, either. Why isn't there exactly one way to specify a set of versions? Is it because semver on its own can't specify arbitrary sets of versions?
(I'm looking at https://devhints.io/semver and I don't see any comma-separated anything.)

So then, if that's the case, then there's some other syntax required for specifying arbitrary sets of versions. `Pkg` looks like it uses "," as a union operator of sets to be, uhm, "semver-complete".
If that's the case, then I don't understand why we would impose a limitation on the atoms that you're taking the union of.

StefanKarpinski:
> Why isn't there exactly one way to specify a set of versions?

The "semver syntax"—which isn't actually specified in the semver standard at all—has many different redundant syntaxes, like `~1.2` and `^1.2` and `1.2` so if we are going to support that at all, then we would inherently already have redundant syntaxes for specifying version sets. That's part of why I originally _only_ implemented version range syntax: it seemed like the most intuitive, least redundant way to specify any set of versions (I think). On the other hand, the `~1.2`, `^1.2`, `1.2` set of syntaxes is less general: you cannot, without ranges, specify a range like `0.7-1.3` at all since there are an infinite number of 0.x versions in that range. It's arguable that this is a bad thing, of course, but the syntax is strictly more general.

However, people wanted the normal semver syntax that other languages use for writing compatibility constraints, so that was implemented—without supporting the range syntax, which is a less common syntax in other languages. So that's how we have ended up here: the registry format _only_ supports version range syntax, while the compat section _only_ supports tilde, carrot and equality specifications of version sets. We cannot break things, so that precludes removing support either syntax in order to unify things, which leaves supporting all of these syntaxes in all places as the only way to unify.

fredrikekre:
> I suppose I don't really understand why there are two syntaxes, either.

There are just one user-facing syntax. The range syntax used in the Registry is an implementation detail.

StefanKarpinski:
It’s a very public “implementation detail”.

goretkin:
I thought I had tested my last commit, but it was broken. This should be correctly tested now.

That said, I think I'm still missing some concepts. I would be happy to add some documentation in a separate PR if I understood.

`Pkg.Types.VersionSpec` and `Pkg.Types.VersionRange` are both sets (conceptually) of `Base.VersionNumber`s. A `VersionRange` must be a contiguous interval of consecutive versions.  And a `VersionSpec` is a ___ ? 

And then there's `semver_spec` which seems to be a different kind of `spec`:
```julia
julia> Pkg.Types.semver_spec("1.2.3")
VersionSpec("1.2.3-1")

julia> Pkg.Types.VersionSpec("1.2.3")
VersionSpec("1.2.3")
```

StefanKarpinski:
I don't understand the new `force_range` thing. Can you explain what that's about?

goretkin:
> I don't understand the new `force_range` thing. Can you explain what that's about?

`VersionRange` is happy to parse "v1.2.3" as `lower=upper=Pkg.Types.VersionBound((0x00000001, 0x00000002, 0x00000003), 3)` unless the `force_range` flag is set, in which case it only parses things that look like "v1.2.3-v4.5.6"

I don't know what `Pkg.Types.semver_spec("v1.2.3")  == VersionSpec("1.2.3-1")` means, but that was the existing behavior. I'm not sure if the following transcript helps:

```julia
julia> Pkg.Types.semver_spec("v1.2.3") # maintain existing behavior of semver_spec
VersionSpec("1.2.3-1")

julia> Pkg.Types.VersionRange("v1.2.3")
VersionRange("1.2.3")

julia> Pkg.Types.VersionSpec(Pkg.Types.VersionRange("v1.2.3")) # this would be the wrong thing to return from semver_spec("v1.2.3")
VersionSpec("1.2.3")

julia> Pkg.Types.VersionRange("v1.2.3"; force_range=true)
ERROR: ArgumentError: degenerate version range: "v1.2.3"
```

StefanKarpinski:
Ah, ok. This is related to the fact that by default `semver_spec` interprets `1.2.3` as meaning `^1.2.3`. That is truly unfortunate in retrospect because `1.2.3` by itself means something quite different in registry files. I always had a feeling we shouldn't have chosen `^` as a default even though it's usually what you want. I'm not sure how we can best converge these formats now. @KristofferC, any ideas? Can we deprecate `1.2.3` in compat sections for a while and then make `=` the default later?

tkf:
This would be a great feature to have.  What is blocking from it to be merged?

> Can we deprecate `1.2.3` in compat sections for a while and then make `=` the default later?

Don't you need to still support it during Julia 1.x?  If that's the case, isn't the decision to deprecate it or not orthogonal to this PR?

KristofferC:
> Ah, ok. This is related to the fact that by default semver_spec interprets 1.2.3 as meaning ^1.2.3. That is truly unfortunate in retrospect because 1.2.3 by itself means something quite different in registry files

It was very much intentional that the simplest way to write semver compatible with `1.2.3` should be to just write `1.2.3`. The registry format (which is compressed and not supposed to be written by humans) is orthogonal to this.

KristofferC:
> And then there's `semver_spec` which seems to be a different kind of spec:

`semver_spec` is what takes the string in the `[compat]` entry and transforms it to a `VersionRange` as documented in  https://julialang.github.io/Pkg.jl/v1/compatibility/.

KristofferC:
Regarding this PR, I don't think it should change the `VersionRange` constructor at all. Instead, another entry in

https://github.com/JuliaLang/Pkg.jl/blob/dfb341ae03e34ae81a5f8c2e324b7b14019bdbbe/src/versions.jl#L340-L345

should be added that matches the hyphen syntax and creates the correct `VersionRange`.

KristofferC:
Obsoleted by https://github.com/JuliaLang/Pkg.jl/pull/1410

PR #1237, opened 2019-06-24
Add tilde examples using a major version of 0

omus:
I tend to forget the difference between the tilde and caret specifier behaviour when the major version is 0. I believe this is the behaviour but let me know if my understanding is incorrect.


omus:
I'd like to keep the examples the same between caret and tilde. Should I also delete/change the caret examples?

fredrikekre:
Ah, okay, didn't notice the examples there.

PR #1248, opened 2019-07-09
Fix typo in version limits

jannefiluren:


jannefiluren:
I misunderstood the docs... they were correct

PR #1256, opened 2019-07-18
import Base.get

lewisl:
Enables parser to correctly parse floating point.


KristofferC:
Would be good with a corresponding test if possible.

lewisl:
Instead of taking this pull request, go to Pkg.  That is a more thorough update.

PR #1261, opened 2019-07-24
Generate dummy output during `Pkg.test` to keep Travis from failing

DilumAluthge:
## Summary

This pull request adds the `output_seconds` keyword argument to `Pkg.test`. The default value of `output_seconds` is `0.`

If `output_seconds` is greater than `0`, then a line of output will be printed every `output_seconds` seconds until the test suite finishes.

If `output_seconds` is less than or equal to `0`, then you get the current behavior.

## Example usage
```julia
julia> Pkg.test("Foo") # preserves the current behavior
julia> Pkg.test("Foo"; output_seconds = -1) # preserves the current behavior
julia> Pkg.test("Foo"; output_seconds = 0) # preserves the current behavior
julia> Pkg.test("Foo"; output_seconds = 1) # prints a line of output every 1 second
julia> Pkg.test("Foo"; output_seconds = 300) # prints a line of output every 300 seconds
```

## Motivation

If `Pkg.test` runs for more than 10 minutes without generating any output, Travis will fail the job. The functionality in this pull request allows you to generate output to satisfy Travis while you wait for your jobs to finish.

One way to get around this is the `travis_wait` function. However, `travis_wait` is not defined in subshells, so if you use a custom `script` file, then `travis_wait` is not defined.

Additionally, if you use `travis_wait`, then you'll need to find a different solution for `AppVeyor`, `CircleCI`, etc. In contrast, if we can add support directly to `Pkg.test`, then you can have the same solution for all CI platforms.

## Questions

Is this the best way to implement this feature? I don't love relying on `kill` to end the dummy output generating process. Is there a better way to accomplish the same outcome?

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -229,7 +229,7 @@ test(pkg::Union{AbstractString, PackageSpec}; kwargs...) = test([pkg]; kwargs...
 test(pkgs::Vector{<:AbstractString}; kwargs...)          = test([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
 test(pkgs::Vector{PackageSpec}; kwargs...)               = test(Context(), pkgs; kwargs...)
 function test(ctx::Context, pkgs::Vector{PackageSpec};
-              coverage=false, test_fn=nothing, kwargs...)
+              coverage=false, test_fn=nothing, output_seconds::Integer = 0, kwargs...)
     pkgs = deepcopy(pkgs) # deepcopy for avoid mutating PackageSpec members
     Context!(ctx; kwargs...)
     ctx.preview && preview_info()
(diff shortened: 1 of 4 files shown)


fredrikekre:
I think this is rare enough that the package maintainer itself can just add the printing. The way it is implemented here you need to modify the default CI script which is not very nice.

It is also pretty trivial to add this to your runtests.jl, e.g. something like https://github.com/JuliaLang/docs.julialang.org/blob/741f9627cb4dfc1c03036e708d4ffa4a536012f5/pdf/make.jl#L41-L52

DilumAluthge:
That code you linked is exactly what I was looking for!
```julia
function makedocs(julia_exec)
    @sync begin
        builder = @async begin
            withenv("TRAVIS_REPO_SLUG" => nothing, # workaround Documenter bugs and julia#26314
                    "BUILDROOT" => nothing) do
                run(`make -C $(JULIA_SOURCE)/doc pdf texplatform=docker JULIA_EXECUTABLE=$(julia_exec)`)
            end
        end
        @async begin
            while !istaskdone(builder)
                sleep(60)
                @info "building pdf ..."
            end
        end
    end
end
```

I agree, it would be very easy for a package maintainer to add this to their `runtests.jl` file. We don't need to add it to Pkg.jl.

PR #1265, opened 2019-07-26
Offline mode

fredrikekre:


KristofferC:
Can we merge this?

fredrikekre:
I should add some tests, and also not sure if anything from the artifacts PR needs to be updated

KristofferC:
Yeah, true.

cossio:
This would be nice to have. Is there something holding this PR back?

fredrikekre:
I'll revive it for 1.5.

oschulz:
> I'll revive it for 1.5.

Lot's of thanks in advance, this will be very useful for us here!

fredrikekre:
Rebased. This does not deal with cached clones, but I think it is good enough for now.

test/new.jl
```diff
@@ -2375,4 +2375,23 @@ end
         Pkg.develop(Pkg.PackageSpec(path="B"))
     end
 end
+
+@testset "Offline mode" begin
+    isolate(loaded_depot=false) do
+        # cache this version
+        Pkg.add(PackageSpec(uuid=exuuid, version=v"0.5.1"))
+        @test Pkg.dependencies()[exuuid].version == v"0.5.1"
+        Pkg.offline()
```
KristofferC, line 2384: Need to turn off offline mode after this test?

KristofferC: Also need to use `Pkg.PackageSpec` or `import it`.

timholy:
This PR is a gift waiting to be unwrapped when we can all return to public transit. Thanks @fredrikekre! (And also @KristofferC for reviewing.)

oschulz:
I'm also hoping to use this a lot with read-only software containers.

PR #1273, opened 2019-07-30
use triple equals for testing against nothing

KristofferC:
This is more efficient in Julia (allows for union splitting).


KristofferC:
bors try

Issue #110, opened 2018-01-30
package does not install when another package needs to downgrade


mvhulten:
Pkg3 does nothing when adding a package that depends on an earlier than installed version of another package:
```
pkg> add CSV
INFO: Resolving package versions
INFO: Updating "~/.julia/environments/v0.6/Project.toml"
 [336ed68f] + CSV v0.2.1
INFO: Updating "~/.julia/environments/v0.6/Manifest.toml"
 [336ed68f] + CSV v0.2.1
 [324d7699] + CategoricalArrays v0.3.3
 [944b1d66] + CodecZlib v0.4.2
 [a93c6f00] + DataFrames v0.11.5
 [9a8bc11e] + DataStreams v0.3.4
 [73a701b4] + NamedTuples v4.0.0
 [3bb67fe8] + TranscodingStreams v0.4.1
 [ea10d353] + WeakRefStrings v0.4.1

pkg> add Benchmark
INFO: Resolving package versions
ERROR: Unsatisfiable requirements detected for package DataFrames [a93c6f00]:
 DataFrames [a93c6f00] log:
 ├─possible versions are: [0.1.0, 0.2.0-0.2.5, 0.3.0-0.3.16, 0.4.0-0.4.3, 0.5.0-0.5.12, 0.6.0-0.6.11, 0.7.0-0.7.8, 0.8.0-0.8.5, 0.9.0-0.9.1, 0.10.0-0.10.1, 0.11.0-0.11.5] or uninstalled
 ├─restricted by compatibility requirements with CSV [336ed68f] to versions: 0.11.0-0.11.5
 │ └─CSV [336ed68f] log:
 │   ├─possible versions are: [0.0.1-0.0.9, 0.0.11-0.0.14, 0.1.0-0.1.2, 0.1.4-0.1.5, 0.2.0-0.2.1] or uninstalled
 │   └─restricted to versions 0.2.1 by an explicit requirement, leaving only versions 0.2.1
 └─restricted by compatibility requirements with Benchmark [de6e09b3] to versions: [0.1.0, 0.2.0-0.2.5, 0.3.0-0.3.16, 0.4.0-0.4.3, 0.5.0-0.5.12, 0.6.0-0.6.11, 0.7.0-0.7.8, 0.8.0-0.8.5, 0.9.0-0.9.1, 0.10.0-0.10.1] — no versions left
   └─Benchmark [de6e09b3] log:
[...]

PR #1274, opened 2019-07-31
stop lower/upper bounding packages in Project in add / develop

KristofferC:
Currently, packages in the Project file gets semver restricted (actually, reading the code they might get restricted to exact versions) when a user adds or devs a new package. This introduces some "hysteresis" in the resolver in the sense that what you already happened to have installed will affect the version bounds that gets entered into the resolver.

This has caused quite a lot of confusion when e.g. `add A; add B` gives different results than `add B; add A` or `add A B`. I think it is better to try to have the inputs to the resolver be as stateless as possible by default. We still have the `strict` option if one wants to use that (and of course `[compat]`).

People might complain that a bunch of unrelated stuff gets updated but that is already true for recursive deps and again, there is `--strict`.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/110

TODO: Add test.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -969,7 +969,7 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}, new_git=UUID[]; strict::Bo
     end
     foreach(pkg -> ctx.env.project.deps[pkg.name] = pkg.uuid, pkgs) # update set of deps
     # load dep graph
-    strict ? load_all_deps!(ctx, pkgs) : load_direct_deps!(ctx, pkgs)
+    strict ? load_all_deps!(ctx, pkgs) : load_direct_deps!(ctx, pkgs; version=false)
     check_registered(ctx, pkgs)
     resolve_versions!(ctx, pkgs)
     update_manifest!(ctx, pkgs)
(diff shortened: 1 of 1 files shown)


fredrikekre:
Just to clarify; this allows packages in `Project.toml` to be changed?

00vareladavid:
Yeah, the direct dependencies currently get restricted to the exact version. This seemed to be the preferred behavior when I asked about it.

Perhaps a more flexible solution would be to roll all the desired behaviors into a single keyword e.g. `add --preserve=no Example`. Where `preserve` determines the subset of the dependency graph that is preserved. I have no strong opinion about what the default should be.

I'm willing to make the changes if people think it is a good idea.

KristofferC:
> Yeah, the direct dependencies currently get restricted to the exact version. This seemed to be the preferred behavior when I asked about it.

Previously, I think they were allowed to change as long as it was in a semver compatible fashion.

I'm not sure if people remember but I kind of like my thought of the "tier-based" approach to resolving when doing `add`. Basically, try to hold everything fixed and add the package. If the resolver fails, loosed up the recursive deps, and if that fails, loosen up the project deps. The resolver is fast enough that I don't think it will be a problem in practice.

It is not ideal since we could potentially get away with only loosening a few of the dependencies but it should never change more than we currently do (except in cases where the resolver now errors).

00vareladavid:
I think #1330 closes this

Issue #1234, opened 2019-06-20
Ad-Hoc Artifacts (Data Containers)


staticfloat:
While thinking through https://github.com/JuliaLang/Pkg.jl/issues/841 and https://github.com/JuliaLang/Pkg.jl/issues/796, I thought that we might have the potential to create a powerful data lifecycle primitive; At first I was calling them "data containers", but really they're "ad-hoc artifacts", so I'm just going to use the word "Artifacts" and generalize a bit as compared to the Artifacts as described in https://github.com/JuliaLang/Pkg.jl/issues/841#issuecomment-501415511. Half inspired by docker volumes, the basic idea is that you may want to process some data, then "save it" as something that can be used by other packages on your system.  The scope of this is purposefully very low-level, so that other packages can build on top of it in order to create more high-level behaviors.

The kinds of Artifacts I'm talking about here are similar to those in https://github.com/JuliaLang/Pkg.jl/issues/841#issuecomment-501415511, but would be generated on-machine, on-the-fly, and initially empty.  Design goals: these should be immutable, simple, content-addressable, have well-defined lifecycle, and be composable with https://github.com/JuliaLang/Pkg.jl/issues/841#issuecomment-501415511 Artifacts.

# API

`Pkg` would offer a simple interface to these kinds of Artifacts:
* `create(f::Function)`: Create a new Artifact.  Meant to be used in a `do-block` form, this calls a user callback where code can be run to fill the newly created directory with data, after which the artifact is "frozen", a tree hash is calculated and returned.  That hash is now the primary way to access the artifact from here on out.

* `installed(hash::SHA1)`: Returns `true` if the given Artifact hash exists and is installed.
* `installed(name::String`): same as above, but returns `false` if the given mapping doesn't exist, as well as if the mapping points to a hash that does not exist.

* `get(hash::SHA1)`: Get the path to an installed Artifact by its hash (not necessarily bound in `Artifacts.toml`).  If not installed, but bound in `Artifacts.toml`, installs the artifact.

* `get(name::String)`: Get the path to an installed Artifact by its name (must be bound in `Artifacts.toml`).  Subs out to `get(hash)` once the hash is known, so has identical behavior.

[...]

PR #1277, opened 2019-08-01
Add Artifacts to Pkg

### LINKED ISSUE #841: Pkg + BinaryProvider
> The ponderous forms of Pkg and BinaryProvider slowly intermesh; Hulking behemoths merging their forms like waves from two separate oceans breaking upon the same shore.  The silhouette of one blends seamlessly into the shadow of the other, a möbius strip of darkness and light, beginning and ending within itself.

Let's talk about the possible merging of BinaryProvider and Pkg, to integrate the binary installation story to unheard-of levels.  Whereas:

* Binary installation for us is now as simple as unpacking a tarball
* Pkg knows how to unpack tarballs

I suggest that we do away with the weird indirection we currently have with packages using `build.jl` files to download tarballs, and instead integrate these downloads into Pkg completely.  This implies that we:

* Create a new concept within Pkg, that of a **Binary Artifact**.  The main difference between a Binary Artifact and a Package is that Packages are platform-independent, Binary Artifacts are necessarily not so.  We would need to load over the same kind of platform-matching code as is in BP right now, e.g. dynamically choosing the most specific matching tarball based on the currently running Julia.  (See [`choose_download()` within BP](https://github.com/JuliaPackaging/BinaryProvider.jl/blob/cb403e6e19c5ec130b47cf3d85d8e9530adabd14/src/PlatformNames.jl#L718-L744) for more).

* Modify BinaryBuilder output to generate Binary Artifacts that are then directly imported into the General Registry.  The Binary Artifacts contain within them a small amount of Julia code; things like setting environment variables, mappings from `LibraryProduct` to actual `.so` file, functions to run an `ExecutableProduct`, etc...  This is all auto-generated by BinaryBuilder.

* Change client packages to simply declare a dependency upon these Binary Artifacts when they require a library.  E.g. `FLAC.jl` would declare a dependency upon `FLAC_jll`, which itself declares a dependency upon `Ogg_jll`, and so on and so forth.

* Eliminate the `Pkg.build()` step for these packages, as the build will be completed by the end of the download step.  (We can actually just bake the `deps.jl` file into the Binary Artifact, as we are using relative paths anyway)

Please discuss.

staticfloat:
This adds the artifacts subsystem to Pkg, [read this WIP blog post](https://github.com/JuliaLang/www.julialang.org/pull/417/files?short_path=514f74c#diff-514f74c34d50677638b76f65d910ad17) for more details.  Closes #841 and #1234.

This PR still needs:

- [x] A `pkg> gc` hook that looks at the list of projects that we know about, examines which artifacts are bound, and marks all that are unbound.  Unbound artifacts that have been continuously unbound for a certain time period (e.g. one month, or something like that) will be automatically reaped.
- [x] Greater test coverage (even without seeing the codecov report, I am certain of this), especially as related to the installation of platform-specific binaries.
- [x] `Overrides.toml` support for global overrides of artifact locations


staticfloat:
bors try

src/API.jl
```diff
@@ -496,7 +498,8 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
         mv(tmp_source, sourcepath; force=true)
     end
     new_apply = Operations.download_source(ctx, pkgs)
-    Operations.build_versions(ctx, union(new_apply, new_git))
+    z = union([pkg.uuid for pkg in new_apply], new_git)
```
KristofferC:
```suggestion
```

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+# Levels of API:
+#
+# - Base level:
+#   - create_artifact(): given no information, simply creates an artifact through a user-provided
+#     callback and returns the hash of the resultant file.
+#
+# - Hash level:
+#   - artifact_exists(hash): returns true if the hash exists on disk, false otherwise
+#   - artifact_path(hash):   returns the path to a hash on disk, throws if it does not exist
+#   - remove_artifact(hash): deletes `path(hash)` if it exists, does nothing otherwise.
```
KristofferC: Is this a user facing API? Why is that needed and how can it be safe if e.g. other packages uses the same artifact?

KristofferC:
```suggestion
#   - remove_artifact(hash): deletes `artifact_path(hash)` if it exists, does nothing otherwise.
```

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+
+# Levels of API:
+#
+# - Base level:
+#   - create_artifact(): given no information, simply creates an artifact through a user-provided
+#     callback and returns the hash of the resultant file.
+#
+# - Hash level:
+#   - artifact_exists(hash): returns true if the hash exists on disk, false otherwise
+#   - artifact_path(hash):   returns the path to a hash on disk, throws if it does not exist
```
KristofferC: Just an idea, but could we return `nothing` if the file does not exist and then get rid of the `artifact_exist` function completely?

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+
+
+const ARTIFACT_DIR_OVERRIDE = Ref{Union{String,Nothing}}(nothing)
+"""
+    with_artifacts_directory(f::Function, artifacts_dir::String)
+
+Helper function to allow temporarily changing the artifact installation directory.
+"""
+function with_artifacts_directory(f::Function, artifacts_dir::String)
+    global ARTIFACT_DIR_OVERRIDE
```
KristofferC:
```suggestion
```

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+    try
+        ARTIFACT_DIR_OVERRIDE[] = artifacts_dir
+        f()
+    finally
+        ARTIFACT_DIR_OVERRIDE[] = nothing
+    end
+end
+
+function artifacts_dir(args...)
+    global ARTIFACT_DIR_OVERRIDE
```
KristofferC:
```suggestion
```

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+
+Archive an artifact into a tarball stored at `tarball_path`, returns the SHA256
+of the resultant tarball.  Throws an error if the artifact does not exist.
+"""
+function archive_artifact(hash::SHA1, tarball_path::String)
+    if !artifact_exists(hash)
+        error("Unable to archive artifact $(bytes2hex(hash.bytes)): does not exist!")
+    end
+
+	# Package it up
```
KristofferC: Tab -> space

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+    if !haskey(artifact_dict, name)
+        return nothing
+    end
+    meta = artifact_dict[name]
+
+    # If it's an array, find the entry that best matches our current platform
+    if isa(meta, Array)
+        dl_dict = Dict(unpack_platform(x, name, artifact_toml) => x for x in meta)
+        meta = select_platform(dl_dict, platform)
+
```
KristofferC:
```suggestion

```

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+    meta = artifact_dict[name]
+
+    # If it's an array, find the entry that best matches our current platform
+    if isa(meta, Array)
+        dl_dict = Dict(unpack_platform(x, name, artifact_toml) => x for x in meta)
+        meta = select_platform(dl_dict, platform)
+    
+    # If it's NOT a dict, complain
+    elseif !isa(meta, Dict)
+        @warn("Invalid Artifact.toml at $(artifact_toml): artifact '$name' malformed, must be array or dict!")
```
KristofferC: Should these `@warn` be `@error`?

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+        TOML.print(io, artifact_dict, sorted=true)
+    end
+    return
+end
+
+"""
+    download_artifact(tree_hash::SHA1, tarball_url::String, tarball_hash::String;
+                      verbose::Bool = false)
+
+Download/install an artifact into the artifact store.  Returns `true` on success.
```
KristofferC: And on failure?

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+export create_artifact, artifact_exists, artifact_path, remove_artifact, verify_artifact,
+       artifact_meta, artifact_hash, bind_artifact, unbind_artifact, download_artifact,
+       find_artifact_toml, ensure_artifact_installed, @artifact_str, archive_artifact
+
+## Philosophy of Artifacts:
+#
+# - At the lowest level, artifacts are content-addressed chunks of data.  Similarly to
+#   how packages are identified by UUIDs but have human-readable mappings through the
+#   registry, artifacts are identified by their SHA1 git tree hashes, but have human-
+#   readable mappings through `Artifact.toml` files.
```
KristofferC: For packages we also allow `JuliaProject.toml`. For consistency, it might be worth thinking about allowing `JuliaArtifact.toml` but not sure if it is worth the churn.

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+        TOML.print(io, artifact_dict, sorted=true)
+    end
+    return
+end
+
+
+"""
+    unbind_artifact(name::String, artifact_toml::String; platform = nothing)
+
+Unbind the given `name` from an `Artifact.toml` file.  Silently fails if no such binding
```
KristofferC: What does "silently fails" mean? Is that the same as "do nothing"?

src/Artifacts.jl
```diff
@@ -0,0 +1,589 @@
...
+
+import Base: get, SHA1
+import ..depots1
+import ..GitTools: tree_hash, set_readonly
+using ..BinaryPlatforms
+import ..TOML
+import ..Types: parse_toml
+import ..PlatformEngines: download_verify_unpack, probe_platform_engines!, package
+
+export create_artifact, artifact_exists, artifact_path, remove_artifact, verify_artifact,
```
KristofferC: Are all the functions in here public (i.e. we want to show them in the docs).

src/GitTools.jl
```diff
@@ -164,4 +167,96 @@ function fetch(repo::LibGit2.GitRepo, remoteurl=nothing; header=nothing, kwargs.
...
+    # Finish it off and return the digest!
+    return SHA.digest!(ctx)
+end
+
+"""
+    tree_hash(root::AbstractString)
+
+Calculate the git tree hash of a given path.
+"""
+function tree_hash(root::AbstractString, HashType = SHA.SHA1_CTX)
```
KristofferC: Is there a need to think about `.gitignore` here`?

src/GitTools.jl
```diff
@@ -164,4 +167,96 @@ function fetch(repo::LibGit2.GitRepo, remoteurl=nothing; header=nothing, kwargs.
...
+    ctx = HashType()
+    SHA.update!(ctx, Vector{UInt8}("tree $(content_size)\0"))
+    for (name, hash, mode) in entries
+        SHA.update!(ctx, Vector{UInt8}("$(mode) $(name)\0"))
+        SHA.update!(ctx, hash)
+    end
+    return SHA.digest!(ctx)
+end
+
+function set_readonly(path)
```
KristofferC: This doesn't really feel specific to Git. We use it for the standard julia packages for example.

src/Operations.jl
```diff
@@ -975,13 +1002,18 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}, new_git=UUID[]; strict::Bo
     update_manifest!(ctx, pkgs)
     # TODO is it still necessary to prune? I don't think so..
     new_apply = download_source(ctx, pkgs)
+
+    # After downloading resolutionary packages, search for Artifact.toml files
+    # and ensure they are all downloaded and unpacked as well:
+    download_artifacts(ctx, pkgs; platform=platform)
```
KristofferC: Shuold this happen after writing the env (like `build`) ?

src/GitTools.jl
staticfloat: Yes, we do if we want this to actually be a "git" tree hash.

src/GitTools.jl
staticfloat: Fair enough; this was just the most "utilsy" file I could find to move it into.

src/Operations.jl
staticfloat: Next on my TODO is to modify `write_env()` to include links to the Artifacts.toml files so that we can `gc` things appropriately, so no, I think it should stay where it is.

src/Artifacts.jl
staticfloat: We could; I kind of like being able to easily calculate where an artifact _would_ be though, and I don't like the type instability making things like `joinpath(artifact_path(hash), "bin", "foo")` dangerous.  I'd rather get a "file does not exist" error than a MethodError there.

src/Artifacts.jl
staticfloat: You're right; we shouldn't encourage users to call this; we should leave it up to `gc` in general.

src/Artifacts.jl
StefanKarpinski: You could do `ispath(artifiact_path(hash))`, no? Does the implementation do any more than that?

staticfloat: The point being that if `artifact_path(hash)` returns `nothing` for something that is not installed, then even `ispath(artifact_path(hash))` will throw a `MethodError` (since `ispath()` uses `joinpath()` internally).  If it instead always returns a string, even if the thing is not installed, you get a much more easily-understood `ENOENT`.

src/Artifacts.jl
staticfloat: Since `artifact_meta()` already has a nice failure condition (return `nothing`) I figured it was better to allow programs to continue, assuming that the artifact does not exist if something is horribly wrong with their `Artifact.toml` instead of just error'ing out.  I'm not married to that design decision though.

src/Artifacts.jl
staticfloat: It throws on failure, deep in the `download_verify_unpack()` chain.

src/Artifacts.jl
staticfloat: I'm going to rename to `Artifacts.toml` anyway, so why not.

src/Artifacts.jl
staticfloat: Yes.

src/Artifacts.jl
staticfloat: I see no reason not to.

src/Artifacts.jl
staticfloat: Done

src/GitTools.jl
staticfloat: I think what we should do is have `tree_hash()` take in a `honor_gitignore::Bool = true` but then for all artifact-related usage, we set `honor_gitignore` to `false`.  This way, we can still use this to verify packages on-disk in the future (so as to match what `git` itself thinks the git tree hash is) without having weird git-specific behavior within artifacts.

src/Artifacts.jl
StefanKarpinski: That was my point: if `artifact_path` returns the path of an existent artifact or where the artifact would get unpacked for a non-existent artifact, then `ispath(artifact_path(hash))` does what `artifact_exists` does and seems almost as easy to write.

staticfloat: Yes, it's a [very simple function](https://github.com/JuliaLang/Pkg.jl/blob/4cb4d58e18022a9eaca124b0e039efc1283adfbd/src/Artifacts.jl#L144-L150):

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+       artifact_meta, artifact_hash, bind_artifact, unbind_artifact, download_artifact,
+       find_artifacts_toml, ensure_artifact_installed, @artifact_str, archive_artifact
+
+## Philosophy of Artifacts:
```
oxinabox: should this all move to devdocs?

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+#   parse-time one of the list of hashes will be chosen as the best match for the
+#   current environment (Julia version, OS, processor architecture, etc...), and note
+#   that it is acceptable for there to be multiple possible choices (e.g. an AVX2-
+#   optimized build and a generic x86_64 build that will both run on a modern processor),
+#   however one single hash will be stably chosen from the multiple possibilities.
+#
+# - Artifacts can be created on-machine through the `create()` functional API, and can
+#   also be downloaded as tarballs from external webservers by embedding information into
+#   the `Artifacts.toml` file.  This can be done programmatically through the
+#   `bind_artifact()` function.
```
oxinabox: It may be worth mentioning here that Artifacts is not a general purpose download mechanimism,
and it is  as a matter of design philosophy assumed that any artifact hosted will have been created expressly for use with Pkg's artifacts mechanism. (And that it is only a happpy coincidence when it can work on other things)

staticfloat: Artifacts should work with any tarball (e.g. `.tar.gz` file).  So it won't work with `.tar.xz` files or `.zip` files, but it works fine with anything else.

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+
+    # Temporary directory where we'll do our creation business
+    temp_dir = mktempdir(artifacts_dir(); prefix="create_")
+
+    try
+        # allow the user to do their work inside the temporary directory
+        f(temp_dir)
+
+        # Calculate the tree hash for this temporary directory
+        artifact_hash = SHA1(tree_hash(temp_dir))
```
oxinabox:
```suggestion
        artifact_hash::SHA1 = (tree_hash(temp_dir))
```

`tree_hash` already reaturns a `SHA1`

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+identifying tree hash of this artifact.
+"""
+function create_artifact(f::Function)
+    # Ensure the `artifacts` directory exists.
+    mkpath(artifacts_dir())
+
+    # Temporary directory where we'll do our creation business
+    temp_dir = mktempdir(artifacts_dir(); prefix="create_")
+
+    try
```
oxinabox: This try catch can be replace with `mktempdir(...) do`

src/GitTools.jl
```diff
@@ -164,4 +167,96 @@ function fetch(repo::LibGit2.GitRepo, remoteurl=nothing; header=nothing, kwargs.
...
+    # Return the hash of these entries
+    content_size = sum(((n, h, m),) -> ndigits(UInt32(m); base=8) + 1 + sizeof(n) + 1 + 20, entries)
+
+    ctx = HashType()
+    SHA.update!(ctx, Vector{UInt8}("tree $(content_size)\0"))
+    for (name, hash, mode) in entries
+        SHA.update!(ctx, Vector{UInt8}("$(mode) $(name)\0"))
+        SHA.update!(ctx, hash)
+    end
+    return SHA.digest!(ctx)
```
oxinabox: We should make `tree_hash` already reaturns a `SHA1`

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+    end
+
+    return hash.bytes == tree_hash(artifact_path(hash))
+end
+
+"""
+    archive_artifact(hash::SHA1, tarball_path::String)
+
+Archive an artifact into a tarball stored at `tarball_path`, returns the SHA256
+of the resultant tarball.  Throws an error if the artifact does not exist.
```
oxinabox:
```suggestion
of the resultant tarball, as a hexidecimal string.
Throws an error if the artifact does not exist.
```

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+"""
+function archive_artifact(hash::SHA1, tarball_path::String)
```
oxinabox:
```suggestion
function archive_artifact(hash::SHA1, destination_tarball_path::String)
```
or `save_path`?

Also this feels like it should be the first argument to me.

and also that the function should be `create_tarball(path, artifact::SHA1)` maybe?

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+
+    # Next, architecture, libc, libgfortran version and cxxabi (if given)
+    arch = nosym(get(entry, "arch", nothing))
+    libc = nosym(get(entry, "libc", nothing))
+    libgfortran_version = nover(get(entry, "libgfortran_version", nothing))
+    libstdcxx_version = nover(get(entry, "libstdcxx_version", nothing))
+    cxxstring_abi = nosym(get(entry, "cxxstring_abi", nothing))
+
+    # Construct the actual Platform object
+    return P(arch;
```
oxinabox:
```suggestion
    return Platform(arch;
```

Please no single letter types.

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+    nover(x) = VersionNumber(x)
+
+    # First, extract OS; we need to build a mapping here
+    os_map = Dict(
+        "windows" => Windows,
+        "macos" => MacOS,
+        "freebsd" => FreeBSD,
+        "linux" => Linux,
+    )
+    P = get(os_map, lowercase(entry["os"]), UnknownPlatform)
```
oxinabox:
```suggestion
    platform = get(os_map, lowercase(entry["os"]), UnknownPlatform)
```

src/Artifacts.jl
oxinabox: Can we move the 
```
 os_map = Dict(
        "windows" => Windows,
        "macos" => MacOS,
        "freebsd" => FreeBSD,
        "linux" => Linux,
    )
    P = get(os_map, lowercase(entry["os"]), UnknownPlatform)
```
Stuff to here?

So it is clearer where `P` comes from?

src/Artifacts.jl
StefanKarpinski: I don't get why it would work with `.tar.gz` but not `.tar.xz` or `.zip`? If we have the extraction programs available, why wouldn't it work with those?

src/Artifacts.jl
oxinabox: I wonder if we should define:
`parse(::Type{Type{Platform}}, str)`

src/Artifacts.jl
staticfloat: I just mean right now it doesn't work with those.  We can of course include the ability for it to use those extraction programs.

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+end
+
+function pack_platform!(meta::Dict, p::Platform)
+    os_map = Dict(
+        Windows => "windows",
+        MacOS => "macos",
+        FreeBSD => "freebsd",
+        Linux => "linux",
+    )
+    meta["os"] = os_map[typeof(p)]
```
oxinabox: Maybe break the above into very short functions?
`name(p::Windows)="windows"` etc

src/Artifacts.jl
staticfloat: https://github.com/JuliaLang/Pkg.jl/blob/ee09076490a3a2e2445044cbb11eb4033b0ac1aa/src/BinaryPlatforms.jl#L472-L580

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+    end
+    if libgfortran_version(p) != nothing
+        meta["libgfortran_version"] = string(libgfortran_version(p))
+    end
+    if libstdcxx_version(p) != nothing
+        meta["libstdcxx_version"] = string(libstdcxx_version(p))
+    end
+    if cxxstring_abi(p) != nothing
+        meta["cxxstring_abi"] = string(cxxstring_abi(p))
+    end
```
oxinabox: could do:
```
set!(f::Function) = set(f, f(p))
set!(f::Function, ::Nothing) = nothing
set!(f, val) = meta[string(f)] = string(val)

set!(libc)
set!(libgfortran_version)
set!(libstdcxx_version)
set!(libstdcxx_version)
```

But that is probably

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+
+Get metadata about a given artifact (identified by name) stored within the given
+`Artifacts.toml` file.  If the artifact is platform-specific, use `platform` to choose the
+most appropriate mapping.  If none is found, return `nothing`.
+"""
+function artifact_meta(name::String, artifacts_toml::String;
+                       platform::Platform = platform_key_abi())
+    if !isfile(artifacts_toml)
+        return nothing
+    end
```
oxinabox: why not throw an error in these cases?

src/Artifacts.jl
```diff
@@ -0,0 +1,594 @@
...
+Writes a mapping of `name` -> `hash` within the given `Artifacts.toml` file.  If
+`platform` is not `nothing`, this artifact is marked as platform-specific, and will be
+a multi-mapping.  It is valid to bind multiple artifacts with the same name, but
+different `platform`s and `hash`'es within the same `artifacts_toml`.  If `force` is set
+to `true`, this will overwrite a pre-existant mapping, otherwise an error is raised.
+
+`download_info` is an optional tuple that contains a vector of URLs and a hash.  These
+URLs will be listed as possible locations where this artifact can be obtained.  The
+"""
+function bind_artifact(name::String, hash::SHA1, artifacts_toml::String;
```
oxinabox: I wonder if this should be 
`write_artifact_binding` 
so something, to make it obvious that it writes to disk

src/Artifacts.jl
staticfloat: It is now less simple, because we need to search multiple depots.

src/API.jl
```diff
@@ -261,175 +262,352 @@ function __installed(mode::PackageMode=PKGMODE_MANIFEST)
...
+            paths = process_func(index_file)
+            if paths != nothing
+                # Print the path of this beautiful, extant file to the user
+                println("        $(Types.pathrepr(index_file))")
+                append!(marked_paths, paths)
+            end
+        end
+
+        # Return the list of absolutified, unique, marked paths
+        return Set(unique(marked_paths))
```
oxinabox: Don't need `unique` if putting it in a `Set`

src/GitTools.jl
staticfloat: I'm punting on this; eventually I'd like to use `tree_hash()` to also verify julia packages on-disk, but that would require proper `.gitignore` handling, which is a little more work than I want to put into this PR.  I spent an afternoon trying to make a simple parser, but `.gitignore` files are... flexible.  ;)

src/Artifacts.jl
staticfloat: Unfortunately, it can't, because I need to `mv()` the directory if it's a new artifact, and if I `mv()` it, then `mktempdir()` fails when it tries to delete the temporary directory.

src/GitTools.jl
staticfloat: It only returns a `SHA1` if the `HashType` is `SHA.SHA1_CTX`.  In the future, we are reserving the ability for this to be applied to more secure hash algorithms.

src/Artifacts.jl
staticfloat: Addressed comment above.

src/Artifacts.jl
staticfloat: I like `archive_artifact()` better.  Since most of these artifact functions have the hash as the first argument, I think it's better to keep the consistency, despite the feeling that we are "modifying" the destination path.

src/Artifacts.jl
staticfloat: Because it's completely normal to check for the existence of an artifact that may or may not exist; that shouldn't be an error, just a negative result.  An error should be reserved for some kind of exceptional circumstance, and I don't consider this an exceptional circumstance; there is no guarantee that a certain hash is bound within an `Artifacts.toml` file, because these files can be quite dynamic.

src/Artifacts.jl
staticfloat: I hope that needing to pass in a filename helps get that across.  I am willing to go as far as re-arranging this to be:

```
bind_artifact!(artifacts_toml, name, hash)
```

If you think that is clearer.

src/API.jl
staticfloat: Good catch; that was left over from some refactoring.

staticfloat:
bors try

src/Artifacts.jl
oxinabox: Sold

staticfloat:
bors try-

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
Okay the reason the tests are failing is that, when I attempt to install a new package and then test that package, the Julia that gets started up to run the tests doesn't have the appropriate `Pkg` project.  Hmmm.

staticfloat:
bors try

staticfloat:
bors try-

staticfloat:
bors try

staticfloat:
bors try-

staticfloat:
bors try

staticfloat:
bors try-

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try-

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try-

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try-

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try

staticfloat:
And with the latest bugs crushed and a new docs page, I think this is officially done from my side.  Absent any final comments, let's merge this in, then bump the version in base Julia to let it settle in with the 1.3 RC's once they start rolling out.

staticfloat:
bors try

staticfloat:
bors try

src/Artifacts.jl
staticfloat: done

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
+# [**8.** Artifacts](@id Artifacts)
+
+`Pkg` can reason about containers of binary data that are not Julia packages.  These chunks of data can contain platform-specific binaries, datasets, text, or any other kind of data that would be convenient to place within an immutable, lifecycled datastore.
```
StefanKarpinski:
```suggestion
`Pkg` can install and manage containers of data that are not Julia packages.  These containers can contain platform-specific binaries, datasets, text, or any other kind of data that would be convenient to place within an immutable, life-cycled datastore.
```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+These containers, (called "Artifacts") can be created locally, hosted anywhere, and automatically downloaded and unpacked upon installation of your Julia package.
+They are used to provide the binary dependencies for packages built with [`BinaryBuilder.jl`](https://github.com/JuliaPackaging/BinaryBuilder.jl).
```
StefanKarpinski:
```suggestion
This mechanism is also used to provide the binary dependencies for packages built with [`BinaryBuilder.jl`](https://github.com/JuliaPackaging/BinaryBuilder.jl).
```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+
+## `Artifacts.toml` files
+
+`Pkg` provides an API for working with artifacts, as well as a TOML file format for recording artifact usage in your packages, and to make automatic downloading of artifacts at package install time.
```
StefanKarpinski:
```suggestion
`Pkg` provides an API for working with artifacts, as well as a TOML file format for recording artifact usage in your packages, and to automate downloading of artifacts at package install time.
```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+Artifacts are referred to by content-hash, or optionally by a name that is bound to a hash through an `Artifacts.toml` file.
```
StefanKarpinski:
```suggestion
Artifacts can always be referred to by content hash, but are typically accessed by a name that is bound to a content hash in an `Artifacts.toml` file that lives in a project's source tree.
```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+The `processed_output` artifact contains no `download` stanza, and so cannot be installed.
+An artifact such as this would be the result of code that was previously run, generating a new artifact and binding the resultant hash to a name within this project.
+
+## Using Artifacts
+
+Artifacts can be manipulated using convenient APIs exposed from the `Pkg.Artifacts` namespace.
+As a motivating example, let us imagine that we are writing a package that needs to load the [Iris machine learning dataset](https://archive.ics.uci.edu/ml/datasets/iris).
+While we could just download the dataset during a build step into the package directory, and many packages currently do precisely this, that has some significant drawbacks:
+
+* First, it modifies the package directory, making package installation stateful, which we'd like to avoid.  In the future, we would like to reach the point where packages can be installed completely read-only, instead of being able to modify themselves after installation.
```
StefanKarpinski:
```suggestion
* First, it modifies the package directory, making package installation stateful, which we want to avoid.
  In the future, we would like to reach the point where packages can be installed completely read-only, instead of being able to modify themselves after installation.

```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+* Second, the downloaded data is not shared across different versions of our package. If we have three different versions of the package installed for use by various projects, then we need three different copies of the data, even if it hasn't changed between those versions. Moreover, each time we upgrade or downgrade the package, unless we do something clever (and probably brittle), we will have to download the data again.
```
StefanKarpinski:
```suggestion
* Second, the downloaded data is not shared across different versions of our package.
  If we have three different versions of the package installed for use by various projects, then we need three different copies of the data, even if it hasn't changed between those versions.
  Moreover, each time we upgrade or downgrade the package, unless we do something clever (and probably brittle), we have to download the data again.
```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+
+* **Name-aware** functions deal with bound names within an `Artifacts.toml` file, and as such, typically require both a path to an `Artifacts.toml` file as well as the artifact name.  Name-aware functions include: `artifact_meta()`, `artifact_hash()`, `bind_artifact!()`, `unbind_artifact!()`, `download_artifact()` and `ensure_artifact_installed()`.
+
+* **Utility** functions deal with miscellaneous aspects of artifact life, such as `create_artifact()`, `ensure_all_artifacts_installed()`, and even the `@artifact_str` string macro.
+
+For a full listing of docstrings and methods, see the [Artifacts Reference](@ref) section.
+
+## Overriding artifact locations
+
+It is occasionally convenient to be able to override the location of an artifact.
```
StefanKarpinski:
```suggestion
It is occasionally necessary to be able to override the location and content of an artifact.
```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+A common usecase is a restrictive computing environment where certain versions of a binary dependency must be used.
```
StefanKarpinski:
```suggestion
A common use case is a computing environment where certain versions of a binary dependency must be used, regardless of what version of this dependency a package was published with.
```

docs/src/artifacts.md
```diff
@@ -0,0 +1,172 @@
...
+While a typical Julia configuration would download, unpack and link against a generic library, a system administrator may wish to disable this and instead use a library already installed on the local machine.
+To enable this, `Pkg` supports a per-depot `Overrides.toml` file placed within the `artifacts` depot directory (e.g. `~/.julia/artifacts/Overrides.toml` for the default user depot) that can override the location of an artifact either by content-hash or by package UUID and bound artifact name.
+Additionally, the destination location can be either an absolute path, or a new artifact content-hash, to support sysadmins in creating new, custom artifacts, then overriding other packages to use that new artifact.
```
StefanKarpinski:
```suggestion
Additionally, the destination location can be either an absolute path, or a replacement artifact content hash.
This allows sysadmins to creating their own artifacts which they can then use by overriding other packages to use the new artifact.
```

docs/src/managing-packages.md
```diff
@@ -257,23 +257,32 @@ shell> cat ~/.julia/packages/MbedTLS/h1Vu/deps/build.log
...
-become old and not used from any existing project.
-Pkg keeps a log of all projects used so it can go through the log and see exactly which projects still exist
-and what packages those projects used. The rest can be deleted.
-This is done with the `gc` command:
+As packages are updated and projects are deleted, installed package versions and artifacts that were
+once used will inevitably become old and not used from any existing project.
+`Pkg` keeps a log of all projects used so it can go through the log and see exactly which projects still exist
+and what packages/artifacts those projects used.
+If a package or artifact is not marked as used by any project, it is added to a list of orphaned packages.
+Packages and artifacts that are in the orphan list for 30 consecutive days are deleted from the system.
```
StefanKarpinski:
```suggestion
Packages and artifacts that are in the orphan list for 30 days without being used again are deleted from the system on the next garbage collection.
```

docs/src/managing-packages.md
```diff
@@ -257,23 +257,32 @@ shell> cat ~/.julia/packages/MbedTLS/h1Vu/deps/build.log
...
+This timing is configurable via the `collect_delay` keyword argument to `Pkg.gc()` (With a value of `0` uninstalling everything immediately).
```
StefanKarpinski:
```suggestion
This timing is configurable via the `collect_delay` keyword argument to `Pkg.gc()`.
A value of `0` will cause anything currently not in use immediately, skipping the orphans list entirely;
If you are short on disk space and want to clean out as many unused packages and artifacts as possible, you may want to try this, but if you need these versions again, you will have to download them again.
```

src/Artifacts.jl
KristofferC: `@error` is not the same as `error`. It is just whether to print in red or yellow and call it "Warning" vs "Error".

image

StefanKarpinski:
bors try

StefanKarpinski:
bors try

src/GitTools.jl
staticfloat: Right, I was thinking in terms of verifying that what is on-disk has not changed since you installed it.

src/Artifacts.jl
staticfloat: Oh!  I see what you mean.  Yes, I do think that's better.  Red is better here.

StefanKarpinski:
bors try

StefanKarpinski:
Jesus Christ, bors, just run the f*@&ing tests.

fredrikekre:
Probably was running. But running on 4 Julia versions on AppVeyor takes 2 hours.

appveyor.yml
```diff
@@ -1,6 +1,8 @@
 environment:
   matrix:
-  - julia_version: 1
+  - julia_version: 1.1
+  - julia_version: 1.2
+  - julia_version: 1.3
```
fredrikekre: Should be enough to run on 1.3 and nightly, we will not ship this version of Pkg with older Julia anyway.

docs/src/artifacts.md
```diff
@@ -0,0 +1,177 @@
...
+These containers, (called "Artifacts") can be created locally, hosted anywhere, and automatically downloaded and unpacked upon installation of your Julia package.
+This mechanism is also used to provide the binary dependencies for packages built with [`BinaryBuilder.jl`](https://github.com/JuliaPackaging/BinaryBuilder.jl).
+
+## `Artifacts.toml` files
+
+`Pkg` provides an API for working with artifacts, as well as a TOML file format for recording artifact usage in your packages, and to automate downloading of artifacts at package install time.
+Artifacts can always be referred to by content hash, but are typically accessed by a name that is bound to a content hash in an `Artifacts.toml` file that lives in a project's source tree.
+An example `Artifacts.toml` file is shown here as an example:
+
+```TOML
```
fredrikekre: We use
```suggestion
```toml
```
in other places but not sure whats correct actually.

docs/src/artifacts.md
```diff
@@ -0,0 +1,177 @@
...
+# Example Artifacts.toml file
```
fredrikekre:
```suggestion
```

src/API.jl
```diff
@@ -259,120 +262,373 @@ function __installed(mode::PackageMode=PKGMODE_MANIFEST)
...
-function gc(ctx::Context=Context(); kwargs...)
+"""
+    gc(ctx::Context=Context(); kwargs...)
+
+Garbage-collect package and artifact installations by sweeping over all known
+`Manifest.toml` and `Artifacts.toml` files, noting those that have been deleted, and then
+finding artifacts and packages that are thereafter not used by any other projects.  This
+method will only remove package versions and artifacts that have been continually un-used
+for a period of `collect_delay`; which defaults to thirty days.
+"""
```
fredrikekre: Should this docstring be attached to the binding in `Pkg.jl`?

src/Artifacts.jl
```diff
@@ -0,0 +1,831 @@
...
+    c_simple = "/path/to/c_simple_dir"
+    libfoo = "fb886e813a4aed4147d5979fcdf27457d20aa35d""
+
+This file defines four overrides; two which override specific artifacts identified
+through their content hashes, two which override artifacts based on their bound names
+within a particular package's UUID.  In both cases, there are two different targets of
+the override: overriding to an on-disk location through an absolutet path, and
+overriding to another artifact by its content-hash.
+"""
+global ARTIFACT_OVERRIDES = Ref{Union{Dict,Nothing}}(nothing)
```
fredrikekre:
```suggestion
const ARTIFACT_OVERRIDES = Ref{Union{Dict,Nothing}}(nothing)
```
?

docs/src/artifacts.md
staticfloat: I think they both work

docs/src/artifacts.md
staticfloat: I would rather have that there, so that users that are scanning through quickly can see that (a) comments in TOML file works, and (b) they know the filename of this file.

staticfloat:
I'm sorry `bors`, but you have failed me for the last time.  Here we go, squashing and merging in one fell swoop.  Next up, PR to base Julia.

PR #1279, opened 2019-08-02
Use Cmd for {julia,test}_args keyword arguments

### LINKED ISSUE #1278: Why not julia_args::Cmd and test_args::Cmd?
Thanks to @DilumAluthge, as of #1226 we have

https://github.com/JuliaLang/Pkg.jl/blob/f75bed67ad869e408455fcedfe0ec7362059a7fa/src/API.jl#L231-L234

But why not use `julia_args::Cmd` and `test_args::Cmd`?  Isn't `Cmd` the best API for handling command line arguments?  That is to say, isn't

```julia
Pkg.test(julia_args=`--compiled-modules=no --compile=min`, test_args=`foo bar`)
```

better than

```julia
Pkg.test(julia_args=[`--compiled-modules=no`, `--compile=min`], test_args=["foo", "bar"])
```

?

Previous discussion: https://github.com/JuliaLang/Pkg.jl/pull/1226#discussion_r297584441

tkf:
Before:

```julia
Pkg.test(julia_args=[`--compiled-modules=no`, `--compile=min`], test_args=["foo", "bar"])
```

After:

```julia
Pkg.test(julia_args=`--compiled-modules=no --compile=min`, test_args=`foo bar`)
```

close #1278

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -230,10 +230,8 @@ test(pkgs::Vector{<:AbstractString}; kwargs...)          = test([PackageSpec(pkg
 test(pkgs::Vector{PackageSpec}; kwargs...)               = test(Context(), pkgs; kwargs...)
 function test(ctx::Context, pkgs::Vector{PackageSpec};
               coverage=false, test_fn=nothing,
-              julia_args::AbstractVector{<:Base.AbstractCmd}=Cmd[],
-              test_args::AbstractVector{<:AbstractString}=String[], kwargs...)
-    julia_args = convert(Vector{Cmd}, julia_args)
-    test_args = convert(Vector{String}, test_args)
+              julia_args::Cmd=``,
+              test_args::Cmd=``, kwargs...)
     pkgs = deepcopy(pkgs) # deepcopy for avoid mutating PackageSpec members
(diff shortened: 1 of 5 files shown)


fredrikekre:
Makes sense for `julia_args`, but isn't it easier to reason about `test_args` when it corresponded 1-to-1 with what ended up in `ARGS`?

KristofferC:
Hm, good point, I should probably have waited a bit before merging. Although, I think we want to be consistent and use the same thing for both `test´ and `julia` args.

StefanKarpinski:
A `Cmd` object and a `Vector{String}` are pretty interchangeable. Perhaps accept either and convert to a common form?

tkf:
I think supporting both `Cmd` and `AbstractVector{<:AbstractString}` is a good approach. Although I usually don't think a redundant synonym optimized for writing in API is great, this is a good exception since `Pkg.test` is an interactive interface.  So, supporting a "short-hand" ``Pkg.test(test_args=`foo bar`)`` for `test=["foo", "bar"]` makes sense.  It also helps `test_args` to be consistent with `julia_args`.

PR #1293, opened 2019-08-15
fix tab completions for packages not defining any julia compat

### LINKED ISSUE #1289: Missing tab completion
Tab completing `MLJ<tab>` in `pkg>` mode does not list `MLJModels`, even though it’s a registered package.

KristofferC:
Fixes #1289.


KristofferC:
bors try

PR #1294, opened 2019-08-15
PlatformEngine: respect Base.DOWNLOAD_HOOKS

StefanKarpinski:
Requires https://github.com/JuliaLang/julia/pull/32917


KristofferC:
Huh? Presumably this need to check that `Base.download_url` is actually defined. This now broke CI here.

PR #1297, opened 2019-08-16
Only run CI for LTS, 1.3 and nightly on master

fredrikekre:
Only run CI for 1.3 and nightly on master, this version of Pkg will not ship with earlier Julia anyway.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,8 +1,7 @@
 language: julia
 
 julia:
-  - 1.1
-  - 1.2
+  # - 1.0
   - 1.3
   - nightly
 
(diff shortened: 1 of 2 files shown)


KristofferC:
-1.

People can easily build their own Julias with newer Pkgs. My guess is that for example that quite a lot of people might want to use the Artifact stuff but are stuck on 1.0.x.

fredrikekre:
Fair enough, but we can't run on 4 different Julia versions on AV since they are not built in paralell, it takes 1.5 hours or something.

KristofferC:
Hm, okay. How about running on nightly, latest release and LTS version?

fredrikekre:
Done.

KristofferC:
Let's remove 1.0 for now and open an issue about it?

KristofferC:
Interesting that the tests take 9 min on Linux but 21 min on Mac.

PR #1298, opened 2019-08-16
Fix Base.download_url for earlier Julia.

fredrikekre:


src/PlatformEngines.jl
```diff
@@ -435,7 +435,7 @@ function probe_platform_engines!(;verbose::Bool = false)
         if probe_cmd(`$test`; verbose=verbose)
             # Set our download command generator
             gen_download_cmd = (url, out_path) ->
-                dl_func(Base.download_url(url), out_path)
+                dl_func(VERSION < v"1.3.0-alpha.130" ? url : Base.download_url(url), out_path)
```
staticfloat: IMO its better to be explicit (`isdefined(Base, download_url)`) then to use `VERSION` comparisons like this.

PR #1306, opened 2019-08-18
Artifacts: fix porous platform support

staticfloat:
This branch contains a set of improvements to Artifacts/PlatformEngines:

* Fix "porous" platform support; there is one method (`ensure_all_artifacts_installed()`) that will break if a multi-platform artifact doesn't have support for the current platform.  Oops.

* Fix artifact verification; when downloading, I was not verifying that the artifact's git tree sha actually was what the `Artifact.toml` advertised; there's an issue here where one package could list the incorrect tree hash, then put something completely different in that location (because the download happened directly, rather than to a temporary directory, verified, then moved into place).  This plugs that hole, so there is now one fewer ways we can get the wrong thing in the right place.

* Fix broken symlink support; As part of implementing the above verification, I discovered that broken symlinks can break things.  Fixed to be resilient to them in all places it matters, especially across windows.

* Improve testing of error messages; some of the error messaging methods were a little ad-hoc; this makes things a little more uniform, and tests them more thoroughly.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -670,14 +670,34 @@ function download_artifact(tree_hash::SHA1, tarball_url::String, tarball_hash::S
         return true
     end
 
+    # Ensure that we're ready to download things
     probe_platform_engines!()
 
-    return download_verify_unpack(
-        tarball_url,
-        tarball_hash,
-        artifact_path(tree_hash),
-        verbose=verbose,
-    )
+    # We download by using `create_artifact()`.  We do this because the download may
+    # be corrupted or even malicious; we don't want to clobber someone else's artifact
+    # by trusting the tree hash that has been given to us; we will instead download it
+    # to a temporary directory, calculate the true tree hash, then move it to the proper
+    # location only after knowing what it is, and if something goes wrong in the process,
+    # everything should be cleaned up.  Luckily, that is precisely what our
+    # `create_artifact()` wrapper does, so we use that here.
+    calc_hash = try
+        create_artifact() do dir
+            download_verify_unpack(tarball_url, tarball_hash, dir, ignore_existence=true, verbose=verbose)
+        end
+    catch e
+        # If something went wrong during download, return false
+        return false
+    end
+
+    # Did we get what we expected?  If not, freak out.
+    if calc_hash.bytes != tree_hash.bytes
+        msg  = "Tree Hash Mismatch!\n"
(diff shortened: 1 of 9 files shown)


staticfloat:
bors try

staticfloat:
bors try

staticfloat:
bors try-

staticfloat:
bors try

KristofferC:
Build failure seems to be because the nightly moved to 1.4. Feel free to open an issue about it and ignore it for now.

fredrikekre:
It was some compat errors in the test registry, fixed now.

bors try

fredrikekre:
Seems to be a legit error in artifacts tests now. Also, seems like verbose mode of 7z is used?

staticfloat:
Yep, these are real errors.  This PR is making tree hashing much stricter, so it's finding some small permissions errors on Windows, currently working up a better way to deal with this with Jameson.  As a small example of the kinds of joy that this kind of investigation brings, check out this beautiful example of code polymorphism within Cygwin:

```
$ ls -la $(cygpath -w $(pwd))/foo
-rw-r--r-- 1 Administrator None 4 Aug 19 14:40 'C:\cygwin\tmp\jl_1RMSx4/foo'

$ ls -la $(pwd)/foo
-rwxr-xr-x 1 Administrator None 4 Aug 19 14:40 /tmp/jl_1RMSx4/foo
```

That's right; the same file, accessed through two different pathnames, has different permissions.  (This has to do with the `ls` binary, not with the underlying filesystem, but underscores the difficulty you can have when trying to debug what's going on within a Windows machine)

staticfloat:
bors try

staticfloat:
Alright, the workaround I have instituted for windows is to continue to blindly trust the `Artifacts.toml` file when it says what git hash is contained within the tarball you just downloaded.  We still verify the tarball, but we don't verify the extracted tree hash.  We will be able to transparently stop doing this once Julia gets the capability to read executable ACLs natively; until then, the only downside to this is that if you use `create_artifacts()` on windows, you may get a different hash than if you did it on a POSIX system.  That's just fine for local use, but distributing those artifacts will fail.  This is probably not a problem though, since all of the tools that we are building right now for artifact distribution are based on POSIX systems, so we may be able to fix Julia before anyone tries to build tools that use this particular functionality.

staticfloat:
Great; looks like we're ready to merge.

fredrikekre:
Can maybe backport 90d83cc39acab69d12e162fb3d0638f8c66227b7

Issue #806, opened 2018-10-09
Cannot add packages due to `ERROR: UndefVarError: repo not defined`


Paalon:
From [discourse](https://discourse.julialang.org/t/undefvarerror-repo-not-defined-when-add-pkg-on-v1-0/15860).
I cannot add package due to following error:
```julia
↪ julia-1.0.1
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.0.1 (2018-09-29)
 _/ |\__'_|_|_|\__'_|  |  Official https://julialang.org/ release
|__/                   |

(v1.0) pkg> add Distributions
  Updating registry at `~/.julia/registries/General`
ERROR: UndefVarError: repo not defined
Stacktrace:
 [1] update_registry(::Pkg.Types.Context) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:123
 [2] #handle_repos_add!#32(::Bool, ::Nothing, ::Function, ::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Pkg/src/Types.jl:607
 [3] #handle_repos_add! at ./none:0 [inlined]
 [4] #add_or_develop#13(::Symbol, ::Bool, ::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:54
 [5] (::getfield(Pkg.API, Symbol("#kw##add_or_develop")))(::NamedTuple{(:mode,),Tuple{Symbol}}, ::typeof(Pkg.API.add_or_develop), ::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at ./none:0
[...]

PR #1314, opened 2019-08-22
fix scope of a variable

KristofferC:
fixes https://github.com/JuliaLang/Pkg.jl/issues/806

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -1136,9 +1136,8 @@ function update_registries(ctx::Context, regs::Vector{RegistrySpec} = collect_re
             # Using LibGit2.with here crashes julia when running the
             # tests for PkgDev wiht "Unreachable reached".
             # This seems to work around it.
-            local repo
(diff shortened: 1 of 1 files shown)


src/Types.jl
```diff
@@ -1136,9 +1136,8 @@ function update_registries(ctx::Context, regs::Vector{RegistrySpec} = collect_re
             # Using LibGit2.with here crashes julia when running the
             # tests for PkgDev wiht "Unreachable reached".
             # This seems to work around it.
-            local repo
+            repo = LibGit2.GitRepo(reg.path)
             try
-                repo = LibGit2.GitRepo(reg.path)
```
StefanKarpinski: AFICT this doesn't change the scope of `repo` just where it's assigned. I think the idea here was that the `GitRepo` operation could fail and we want to catch that.

KristofferC: How about this then?

StefanKarpinski: yep, that's a better fix 👍

PR #1317, opened 2019-08-23
Introduce `temp` option for `activate`

### LINKED ISSUE #1201: Feature request: easy "throw away" environments
Would it be nice if we had a `julia --project=@tmp` command (syntax TBD) for creating a "throw away" environment whcih exists only for the duration of the repl session?

E.g. operates in a `tmp` directory and automatically deletes `/Project.toml` and `/ Manifest.toml` afterwards; so `julia --project=@tmp` again gives a new "throw away" environment.

Use case:
- Not editing any specifc package, so don't want `julia --project=@.`,
- Want to test something out, e.g. try reproduce a bug report
- Load `julia` and inevitably end up adding packages, e.g. the bugged package, 
- Your "base" `v1.1` environment ends up having tons of packages <-- something we might wish to avoid.

00vareladavid:
Resolve #1201 

I find myself needing this often as well.

Example:
```
(Pkg) pkg> activate --temp
Activating new environment at `/tmp/tmpVtb1OR/Project.toml`
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -837,7 +837,9 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=
 end
 
 
-function activate()
+function activate(;temp=false,shared=false)
+    shared && pkgerror("Must give a name for a shared environment")
+    temp && return activate(mktempdir())
     Base.ACTIVE_PROJECT[] = nothing
     p = Base.active_project()
     p === nothing || printpkgstyle(Context(), :Activating, "environment at $(pathrepr(p))")
@@ -861,7 +863,8 @@ function _activate_dep(dep_name::AbstractString)
         end
     end
 end
-function activate(path::AbstractString; shared::Bool=false)
+function activate(path::AbstractString; shared::Bool=false, temp::Bool=false)
+    temp && pkgerror("Can not give `path` argument when creating a temporary environment")
     if !shared
         # `pkg> activate path`/`Pkg.activate(path)` does the following
         # 1. if path exists, activate that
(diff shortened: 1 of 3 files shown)


fredrikekre:
Why not just use the already working shorter version:
```
pkg> activate /tmp
```
?

nickrobinson251:
> Why not just use the already working shorter version:
> 
> ```
> pkg> activate /tmp
> ```
> 
> ?

This doesn't have the same behaviour, because `/tmp` won't be removed between repl sessions
```julia
$ julia --project=@.

(SomeEnv) pkg> activate /tmp
Activating new environment at `/tmp/Project.toml`

(tmp) pkg> add AxisArrays
...

julia> exit()

$ julia --project=@.

(SomeEnv) pkg> activate /tmp
Activating environment at `/tmp/Project.toml`

(tmp) pkg> st
    Status `/private/tmp/Project.toml`
  [39de3d68] AxisArrays v0.3.2
```

fredrikekre:
```
pkg> activate /tmp/$(random headbang on keyboard)
```
then? :)

KristofferC:
> Why not just use the already working shorter version:

Windows don't have that. It is nice to have a consistent API.

nickrobinson251:
Let's add `--temp` to the docstring you get from `(Env) pkg> ?activate`? :)

00vareladavid:
> `pkg> activate /tmp/$(random headbang on keyboard)`

FWIW This has been my solution up until now, I even make sure to play proper Swedish metal first!

00vareladavid:
Enough people want this and it is a small enough feature that I've revived this PR

PR #1322, opened 2019-08-28
WIP Add support for zip file download artifacts

davidanthoff:
This seems to work on Windows.

Less clear to me what to do about other platforms. I assume `tar` can't be used to unzip files on Linux and Mac? So we would have to use `unzip`? Or something entirely different?

I'll also need to figure out tests...

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -197,8 +197,12 @@ function probe_platform_engines!(;verbose::Bool = false)
     # the correct 7z given the path to the executable:
     unpack_7z = (exe7z) -> begin
         return (tarball_path, out_path, excludelist = nothing) ->
-        pipeline(`$exe7z x $(tarball_path) -y -so`,
+        if endswith(tarball_path, ".zip")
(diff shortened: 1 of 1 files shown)


CiaranOMara:
Unpacking .zip seems to already work on macOS, but not Linux (debian).

This is the artifact entry that I'm working with.
```
[Trimmomatic]
git-tree-sha1 = "d03af062acaa4098c8d7109d95ad7e064c1a76ce"
lazy = true

    [[Trimmomatic.download]]
    sha256 = "2f97e3a237378d55c221abfc38e4b11ea232c8a41d511b8b4871f00c0476abca"
    url = "http://www.usadellab.org/cms/uploads/supplementary/Trimmomatic/Trimmomatic-0.39.zip"
```

Attempting to automatically load the above artifact on Linux resulted in the following error.
```
gzip: stdin has more than one entry--rest ignored
tar: Child returned status 2
tar: Error is not recoverable: exiting now
ERROR: LoadError: InitError: Unable to automatically install 'Trimmomatic' from '/home/username/.julia/packages/Trimmomatic/Okr9I/Artifacts.toml'
```

PR #1325, opened 2019-08-29
More informative error messages when parsing manifest files.

### LINKED ISSUE #1318: "Dependency name does not match UUID" error could be better
```
(FHE) pkg> st
ERROR: Dependency name does not match UUID
```
I think the problem may be that I renamed the package, but this error message wasn't particularly helpful and coupe be improved (ideally with instruction on how to get out of this mess ;) ).

00vareladavid:
This enhances error messages from the manifest parsing logic.

---
Fix #1318 
The error from #1318 will now read:
```
ERROR: `DocStringExtensions=ffbed154-4ef7-542d-bbb7-c09d3a79fcae` depends on `Pkg=748391e3-3ea3-41aa-8037-d96b5d03e7c6`, but entry with UUID `748391e3-3ea3-41aa-8037-d96b5d03e7c6` has name `Pkg2`.
```

### DIFF
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -4,19 +4,19 @@
 function read_field(name::String, default, info, map)
     x = get(info, name, default)
     x == default && return default
-    x isa String || pkgerror("Expected field `$name` to be a String")
+    x isa String || pkgerror("Expected field `$name` to be a String.")
     return map(x)
 end
 
 read_pinned(::Nothing) = false
 read_pinned(pinned::Bool) = pinned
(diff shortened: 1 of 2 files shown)


KristofferC:
Looks like an improvement. If we don't want to write the UUIDs so verbosely we could write it something like `DocstringExtensions [ffbed154]` which is a bit similar to how we do it in Pkg.status(). But doesn't matter that much since this error should be quite rare, so LGTM.

00vareladavid:
Yeah that error should only happen when someone manually edits their manifest

bors r+

PR #1327, opened 2019-08-30
Tweak API docs for `dependencies` and `project`

00vareladavid:
---

So I noticed just after merging that the formatting was a little weird :sweat_smile:.

@fredrikekre would you happen to know why `Pkg.project` doesn't appear in the API docs? I admit that I'm not clear on how the docs are generated, I just tried following the same format as the other API functions.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -271,43 +271,44 @@ const develop = API.develop
 const generate = API.generate
 
 """
-!!! note
-    This feature is experimental.
-
     Pkg.dependencies()::Dict{UUID, PackageInfo}
+
+!!! compat "Julia 1.4"
+    This feature requires Julia 1.4, and is considered experimental.
 
 Query the dependecy graph.
 The result is a `Dict` that maps a package UUID to a `PackageInfo` struct representing the dependency (a package).
 
-PackageInfo fields:
+# `PackageInfo` fields
 
-| `field`        | `Description`                                              |
-|:---------------|:-----------------------------------------------------------|
-| `name`         | The name of the package                                    |
-| `version`      | The version of the package (this is `Nothing` for stdlibs) |
-| `isdeveloped`  | Whether a package is directly tracking a directory         |
-| `ispinned`     | Whether a package is pinned                                |
-| `source`       | The directory containing the source code for that package  |
-| `dependencies` | The dependencies of that package as a vector of UUIDs      |
+| Field        | Description                                                |
+|:-------------|:-----------------------------------------------------------|
+| name         | The name of the package                                    |
+| version      | The version of the package (this is `Nothing` for stdlibs) |
+| isdeveloped  | Whether a package is directly tracking a directory         |
+| ispinned     | Whether a package is pinned                                |
+| source       | The directory containing the source code for that package  |
+| dependencies | The dependencies of that package as a vector of UUIDs      |
 """
 const dependencies = API.dependencies
 
 """
-!!! note
-    This feature is experimental.
-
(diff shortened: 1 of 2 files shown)


fredrikekre:
> @fredrikekre would you happen to know why Pkg.project doesn't appear in the API docs? I admit that I'm not clear on how the docs are generated, I just tried following the same format as the other API functions.

You didn't include it, but you did include `Pkg.dependencies` https://github.com/JuliaLang/Pkg.jl/commit/996c6b9b69ef0c058e0105427983622b7cc8cb1d#diff-a71ec1546f7cbced1c1eecfc736d1a42

src/Pkg.jl
```diff
@@ -271,43 +271,44 @@ const develop = API.develop
 const generate = API.generate
 
 """
-!!! note
-    This feature is experimental.
-
     Pkg.dependencies()::Dict{UUID, PackageInfo}
 
+!!! compat "Experimental"
```
```suggestion
!!! compat "Julia 1.4"
```

src/Pkg.jl
```diff
@@ -271,43 +271,44 @@ const develop = API.develop
...
+    This feature is experimental.
```
```suggestion
    This feature requires Julia 1.4, and is considered experimental.
```

src/Pkg.jl
```diff
@@ -271,43 +271,44 @@ const develop = API.develop
...
+
 Query the dependecy graph.
 The result is a `Dict` that maps a package UUID to a `PackageInfo` struct representing the dependency (a package).
 
-PackageInfo fields:
+# `PackageInfo` Fields
```
```suggestion
# `PackageInfo` fields
```

src/Pkg.jl
```diff
@@ -271,43 +271,44 @@ const develop = API.develop
...
 
-| `field`        | `Description`                                              |
-|:---------------|:-----------------------------------------------------------|
-| `name`         | The name of the package                                    |
```
fredrikekre: Why not keep these?

src/Pkg.jl
```diff
@@ -271,43 +271,44 @@ const develop = API.develop
...
 """
 const dependencies = API.dependencies
 
 """
-!!! note
-    This feature is experimental.
-
     Pkg.project()::ProjectInfo
 
+!!! compat "Experimental"
```
```suggestion
!!! compat "Julia 1.4"
```

src/Pkg.jl
```diff
@@ -271,43 +271,44 @@ const develop = API.develop
...
+    This feature is experimental.
```
```suggestion
    This feature requires Julia 1.4, and is considered experimental.
```

src/Pkg.jl
```diff
@@ -271,43 +271,44 @@ const develop = API.develop
...
 
-ProjectInfo fields:
-| `field`        | `Description`                                                                               |
-|:---------------|:--------------------------------------------------------------------------------------------|
-| `name`         | The project's name                                                                          |
-| `uuid`         | The project's UUID                                                                          |
-| `version`      | The project's version                                                                       |
-| `dependencies` | The project's direct dependencies as a `Dict` which maps dependency name to dependency UUID |
-| `path`         | The location of the project file which defines the active project                           |
+# `ProjectInfo` Fields
```
```suggestion
# `ProjectInfo` fields
```

00vareladavid:
Thanks! Can't believe I missed that :laughing:.

00vareladavid:
Also, should we reorder the API functions in the docs to be in alphabetical order? If so, I can do that here.

src/Pkg.jl
00vareladavid: It just seems like unnecessary formatting i.e. the "field" title already implies that the names are code.

PR #1330, opened 2019-08-31
Introduce support for tiered resolver

### LINKED ISSUE #110: package does not install when another package needs to downgrade

00vareladavid:
---
Fix #110
Pretty much what Kristoffer recommends in #1274.

Example:
```
afdwer) pkg> status
    Status `/tmp/afdwer/Project.toml`
  [295af30f] Revise v1.1.0

(afdwer) pkg> add MagneticReadHead@0.2.2
 Resolving package versions...
Resolve_Tier All dependency versions.
Resolve_Tier Direct dependency versions.
Resolve_Tier Direct dependency semver constraints.
ERROR: Unsatisfiable requirements detected for package MagneticReadHead [63dd013e]:
<stacktrace here>

(afdwer) pkg> add --preserve=none MagneticReadHead@0.2.2
 Resolving package versions...
 Installed MagneticReadHead ─ v0.2.2
 Installed Cassette ───────── v0.2.6
  Updating `/tmp/afdwer/Project.toml`
  [63dd013e] + MagneticReadHead v0.2.2
  [295af30f] ↑ Revise v1.1.0 ⇒ v2.1.9
```

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -46,8 +46,24 @@ end
 is_dep(ctx::Context, pkg::PackageSpec) =
     any(uuid -> uuid == pkg.uuid, [uuid for (name, uuid) in ctx.env.project.deps])
 
-function load_direct_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Bool=true)
-    # load rest of deps normally
+#TODO rename
+function load_version(version, fixed, preserve::PreserveLevel)
+    if version === nothing
+        return VersionSpec() # stdlibs dont have a version
+    elseif fixed
+        return version # dont change state if a package is fixed
+    elseif preserve == PRESERVE_ALL || preserve == PRESERVE_DIRECT
+        return something(version, VersionSpec())
+    elseif preserve == PRESERVE_SEMVER && version != VersionSpec()
+        return Types.semver_spec("$(version.major).$(version.minor).$(version.patch)")
+    elseif preserve == PRESERVE_NONE
+        return VersionSpec()
+    end
+end
+
+function load_direct_deps(ctx::Context, pkgs::Vector{PackageSpec};
+                          preserve::PreserveLevel=PRESERVE_DIRECT)
+    pkgs = copy(pkgs)
     for (name::String, uuid::UUID) in ctx.env.project.deps
         pkgs[uuid] === nothing || continue # do not duplicate packages
         entry = manifest_info(ctx, uuid)
@@ -58,26 +74,35 @@ function load_direct_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Boo
                 name      = name,
                 path      = entry.path,
                 repo      = entry.repo,
-                tree_hash = entry.tree_hash,
                 pinned    = entry.pinned,
-                version   = version ? something(entry.version, VersionSpec()) : VersionSpec()))
+                tree_hash = entry.tree_hash, # TODO should tree_hash be changed too?
+                version   = load_version(entry.version, isfixed(entry), preserve),
+              ))
     end
+    return pkgs
 end
 
-function load_all_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Bool=true)
+function load_all_deps(ctx::Context, pkgs::Vector{PackageSpec}=PackageSpec[];
+                        preserve::PreserveLevel=PRESERVE_ALL)
+    pkgs = copy(pkgs)
     for (uuid, entry) in ctx.env.manifest
         pkgs[uuid] === nothing || continue # do not duplicate packages
-        push!(pkgs, PackageSpec(name=entry.name, uuid=uuid, path=entry.path,
-                                version = version ? something(entry.version, VersionSpec()) : VersionSpec(),
-                                repo=entry.repo, tree_hash=entry.tree_hash))
+        push!(pkgs, PackageSpec(
+            uuid      = uuid,
+            name      = entry.name,
+            path      = entry.path,
+            pinned    = entry.pinned,
+            repo      = entry.repo,
+            tree_hash = entry.tree_hash, # TODO should tree_hash be changed too?
+            version   = load_version(entry.version, isfixed(entry), preserve),
+        ))
     end
-    load_direct_deps!(ctx, pkgs; version=version)
+    return load_direct_deps(ctx, pkgs; preserve=preserve)
 end
 
 function is_instantiated(ctx::Context)::Bool
     # Load everything
-    pkgs = PackageSpec[]
-    Operations.load_all_deps!(ctx, pkgs)
+    pkgs = load_all_deps(ctx)
     # Make sure all paths exist
     for pkg in pkgs
         sourcepath = Operations.source_path(pkg)
@@ -217,7 +242,8 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
     return
 end
 
-is_fixed(pkg::PackageSpec) = pkg.path !== nothing || pkg.repo.url !== nothing
+istracking(pkg) = pkg.path !== nothing || pkg.repo.url !== nothing
+isfixed(pkg) = istracking(pkg) || pkg.pinned
 
 function collect_fixed!(ctx::Context, pkgs::Vector{PackageSpec}, names::Dict{UUID, String})
     fix_deps_map = Dict{UUID,Vector{PackageSpec}}()
@@ -249,7 +275,6 @@ end
 # adds any other packages which may be in the dependency graph
 # all versioned packges should have a `tree_hash`
 function resolve_versions!(ctx::Context, pkgs::Vector{PackageSpec})
-    printpkgstyle(ctx, :Resolving, "package versions...")
     # compatibility
     proj_compat = Types.project_compatibility(ctx, "julia")
     v = intersect(VERSION, proj_compat)
@@ -263,7 +288,7 @@ function resolve_versions!(ctx::Context, pkgs::Vector{PackageSpec})
 
     # construct data structures for resolver and call it
     # this also sets pkg.version for fixed packages
-    fixed = collect_fixed!(ctx, filter(is_fixed, pkgs), names)
+    fixed = collect_fixed!(ctx, filter(istracking, pkgs), names)
 
     # non fixed packages are `add`ed by version: their version is either restricted or free
     # fixed packages are `dev`ed or `add`ed by repo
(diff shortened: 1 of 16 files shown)


00vareladavid:
bors try

KristofferC:
Cool! I'm in the progress of moving (a lot of time at IKEA and then building) but will look at this as soon as I can.

src/Operations.jl
```diff
@@ -46,8 +46,21 @@ end
...
-function load_direct_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Bool=true)
-    # load rest of deps normally
+#TODO rename
+function load_version(version, fixed, preserve::PreserveLevel)
+    version === nothing && return VersionSpec() # for stdlibs
+    fixed && return version
+    preserve == PRESERVE_ALL && return something(version, VersionSpec())
+    preserve == PRESERVE_DIRECT && return something(version, VersionSpec())
+    if preserve == PRESERVE_SEMVER && version != VersionSpec()
+        return Types.semver_spec("$(version.major)")
```
KristofferC: This doesn't seem completely correct? E.g. if the version is `0.4.2`, calling  `semver_spec("0")` wouldn't be right?

src/Operations.jl
```diff
@@ -961,21 +983,57 @@ function assert_can_add(ctx::Context, pkgs::Vector{PackageSpec})
     end
 end
 
+function tiered_resolve(ctx::Context, pkgs::Vector{PackageSpec})
+    try # do not modify existing subgraph
+        printpkgstyle(ctx, :Resolve_Tier, "All dependency versions.")
```
KristofferC: Presumably this is debug output and won't shown to the user later? At least I don't think we should show it.

src/Operations.jl
```diff
@@ -961,21 +983,57 @@ function assert_can_add(ctx::Context, pkgs::Vector{PackageSpec})
...
+        err isa ResolverError || rethrow()
+    end
+    try # do not modify existing direct deps
+        printpkgstyle(ctx, :Resolve_Tier, "Direct dependency versions.")
+        return targeted_resolve(ctx, pkgs, PRESERVE_DIRECT)
+    catch err
+        err isa ResolverError || rethrow()
+    end
+    printpkgstyle(ctx, :Resolve_Tier, "Direct dependency semver constraints.")
+    return targeted_resolve(ctx, pkgs, PRESERVE_SEMVER)
```
KristofferC: Shouldn't there be a call to `targeted_resolve` with `PRESERVE_NONE` here?

KristofferC:
Some comments but look nice overall. This is pretty hard to test thoroughly. I guess we could use a fixed commit of the registry and just "hardcode" a few explicit test (e.g. adding some package doesn't cause others to upgrade etc).

src/Operations.jl
00vareladavid: Agreed. Its better without the user messages

src/Operations.jl
00vareladavid: I actually wasn't sure about this. Looking at https://semver.org/, the convention seems to be that minor version changes are considered breaking changes? I'll fix it

src/Operations.jl
00vareladavid: The idea is that regular operation shouldn't silently break semver. It would be weird if I did `add Foo` and now my code won't run. From manual testing, it seems `PRESERVE_SEMVER` is actually surprisingly flexible. If the user decides they don't care, they can always just do `add --preserve=none Foo`

KristofferC: I mean we have the `[compat]` section there for a reason though. And we did exactly have the `PRESERVE_SEMVER` strategy before and it leads to a lot of resolver errors in situations where a tries the following:

Add A that installs e.g. version 0.8
Add B that has compat `0.7` with A.

This gives a resolver error because A is not allowed to downgrade from 0.8. If the user had instead installed these in the opposite order, everything would be fine. Adding both packages at the same time would also work ok.

src/Operations.jl
KristofferC: Isn't it enough to call `Types.semver_spec("$(version.major).($version.minor).($version.patch)")`?

src/Operations.jl
00vareladavid: Fair point about `compat`. I'll change this as well

KristofferC: Here is an example in the wild: https://github.com/JuliaStats/StatsKit.jl/issues/20.

KristofferC:
Some small docs and if possible some kind of test and this should be good to go?

00vareladavid:
Yeah I'll try to do some light testing, but a lot of tests would benefit from a small test registry

00vareladavid:
bors try

KristofferC:
> Yeah I'll try to do some light testing, but in the future I think a lot of tests would benefit from a small test registry

Agreed

src/Operations.jl
```diff
@@ -46,8 +46,21 @@ end
 is_dep(ctx::Context, pkg::PackageSpec) =
     any(uuid -> uuid == pkg.uuid, [uuid for (name, uuid) in ctx.env.project.deps])
 
-function load_direct_deps!(ctx::Context, pkgs::Vector{PackageSpec}; version::Bool=true)
-    # load rest of deps normally
+#TODO rename
+function load_version(version, fixed, preserve::PreserveLevel)
+    version === nothing && return VersionSpec() # for stdlibs
+    fixed && return version
+    preserve == PRESERVE_ALL && return something(version, VersionSpec())
```
fredrikekre: Maybe clearer as an `if-ifelse-`block

src/Pkg.jl
```diff
@@ -80,18 +82,34 @@ Used as an argument to  [`PackageSpec`](@ref) or as an argument to [`Pkg.update`
...
+The `preserve` keyword argument allows you to key into a specific tier in the resolve algorithm.
+The following table describes the argument values for `preserve` (in order of strictness):
+
+| Value             | Description                                                                         |
+|:------------------|:------------------------------------------------------------------------------------|
+| `PRESERVE_ALL`    | Preserve the state of all existing dependencies (including recursive dependencies)  |
+| `PRESERVE_DIRECT` | Preserve the state of all existing direct dependencies                              |
+| `PRESERVE_SEMVER` | Preserve semver-compatible versions of direct dependencies                          |
+| `PRESERVE_NONE`   | Do not attempt to preserve any version information                                  |
+| `PRESERVE_TIERED` | Use the tier which will preserve the most version information (this is the default) |
```
fredrikekre: Worth clarifying here that we try them in order? Or maybe write that in the text block above, I think we could explain the default there too.

src/Operations.jl
00vareladavid: yea, looks pretty ugly like how it is now :laughing:

src/Pkg.jl
```diff
@@ -80,18 +82,34 @@ Used as an argument to  [`PackageSpec`](@ref) or as an argument to [`Pkg.update`
...
-    Pkg.add(pkg::Union{PackageSpec, Vector{PackageSpec}})
+    Pkg.add(pkg::Union{String, Vector{String}}; preserve=PRESERVE_TIERED)
+    Pkg.add(pkg::Union{PackageSpec, Vector{PackageSpec}}; preserve=PRESERVE_TIERED)
 
 Add a package to the current project. This package will be available by using the
 `import` and `using` keywords in the Julia REPL, and if the current project is
 a package, also inside that package.
 
+## Resolution Tiers
+`Pkg` resolves the set of packages in your environment using a tiered algorithm.
```
```suggestion
!!! compat "Julia 1.4"
    Resolution tiers, and the `preserve` keyword argument requires at least Julia 1.4.

`Pkg` resolves the set of packages in your environment using a tiered algorithm.
```

src/REPL/command_declarations.jl
```diff
@@ -106,9 +106,23 @@ specified by `#branch` or `#commit`.
 If a local path is used as an argument to `add`, the path needs to be a git repository.
 The project will then track that git repository just like it would track a remote repository online.
 
+`Pkg` resolves the set of packages in your environment using a tiered approach.
+The `--preserve` command line option allows you to key into a specific tier in the resolve algorithm.
+The following table describes the command line arguments to `--preserve` (in order of strictness).
+
```
fredrikekre:
```suggestion

!!! compat "Julia 1.4"
    Resolution tiers, and the `--preserve` argument requires at least Julia 1.4.
```

KristofferC:
Yay!

PR #1333, opened 2019-08-31
More printing updates

fredrikekre:
Print more stuff to `ctx.io`, which now defaults to `stderr` except when the user has explicitly asked for the output (e.g. `pkg> status`, `pkg> registry status` and `pkg> help`).

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -16,7 +16,11 @@ using ..BinaryPlatforms
 using ..Artifacts: artifact_paths
 
 
-preview_info() = printstyled("───── Preview mode ─────\n"; color=Base.info_color(), bold=true)
+function preview_info(ctx::Context)
+    if ctx.preview
+        printstyled(ctx.io, "───── Preview mode ─────\n"; color=Base.info_color(), bold=true)
+    end
+end
 
 include("generate.jl")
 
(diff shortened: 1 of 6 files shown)


00vareladavid:
Wouldn't it be better to instead move all the stuff which should be going to `stderr` but is using `ctx.io` to use the logging mechanism?

fredrikekre:
IMO no, there is no way for us to control the logging output, see e.g. https://github.com/JuliaLang/julia/pull/28229

00vareladavid:
Didn't know about that. In that case, this change does seem necessary :+1:

src/API.jl
```diff
@@ -772,6 +775,7 @@ status(pkgs::Vector{<:AbstractString}; kwargs...) =
 status(pkgs::Vector{PackageSpec}; kwargs...) = status(Context(), pkgs; kwargs...)
 function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=PKGMODE_PROJECT,
                 kwargs...)
```
00vareladavid:
```suggestion
                io::IO=stdout, kwargs...)
        Context!(ctx; io=io, kwargs...)
```

So the user can still override if necessary

fredrikekre: Should be equivalent to what I have now. If a user has `io` as a kwarg it will be overwritten at L779.

fredrikekre:
bors try

fredrikekre:
bors ping

fredrikekre:
bors try

fredrikekre:
bors try

PR #1334, opened 2019-08-31
Fix test isolation

00vareladavid:
This was causing problems for me when testing locally. The tests would switch to using the built-in version of `Pkg` instead of the development one.

I don't understand exactly why this happens and why it does not cause problems on CI.

### DIFF
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -792,20 +792,20 @@ end
 end
 
 @testset "create manifest file similar to project file" begin
-    cd_tempdir() do dir
+    temp_pkg_dir() do project_path; mktempdir() do dir
         touch(joinpath(dir, "Project.toml"))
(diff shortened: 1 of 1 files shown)


fredrikekre:
Another one (?):
```
ERROR: LoadError: LoadError: LoadError: UndefVarError: dependencies not defined
Stacktrace:
 [1] (::getfield(Main.PkgTests.OperationsTest.REPLTests, Symbol("##23#32")))() at /home/fredrik/dev/Pkg/test/repl.jl:69
```

00vareladavid:
:slightly_frowning_face: yes I think so. Did you get that when running this branch?

fredrikekre:
No, master.

00vareladavid:
will be closed by #1384

PR #1344, opened 2019-09-02
Give a proper error for registered_name(...) = nothing

oxinabox:


src/Operations.jl
```diff
@@ -425,6 +425,7 @@ function deps_graph(ctx::Context, uuid_to_name::Dict{UUID,String}, reqs::Require
     for uuid in uuids
         uuid == uuid_julia && continue
         if !haskey(uuid_to_name, uuid)
+            uuid === nothing && @error "Something is very wrong." uuid pkg_name=registered_name(ctx, uuid)
```
oxinabox: Maybe better is:

```suggestion
            uuid === nothing && error("Something is very wrong as attempting to work work with: uuid=nothing,  pkg_name=$(registered_name(ctx, uuid))")
```

fredrikekre:
It's not `uuid` that is nothing, but the return value of `registered_name(ctx, uuid)`. I suggest something like:
```diff
diff --git a/src/Operations.jl b/src/Operations.jl
index ab0960b..8740041 100644
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -425,7 +425,9 @@ function deps_graph(ctx::Context, uuid_to_name::Dict{UUID,String}, reqs::Require
     for uuid in uuids
         uuid == uuid_julia && continue
         if !haskey(uuid_to_name, uuid)
-            uuid_to_name[uuid] = registered_name(ctx, uuid)
+            name = registered_name(ctx, uuid)
+            name === nothing && pkgerror("cannot find name corresponding to UUID $(uuid)")
+            uuid_to_name[uuid] = name
             entry = manifest_info(ctx, uuid)
             entry ≡ nothing && continue
             uuid_to_name[uuid] = entry.name
```

oxinabox:
ah yeah, my bad.

src/Operations.jl
```diff
@@ -425,7 +425,9 @@ function deps_graph(ctx::Context, uuid_to_name::Dict{UUID,String}, reqs::Require
     for uuid in uuids
         uuid == uuid_julia && continue
         if !haskey(uuid_to_name, uuid)
-            uuid_to_name[uuid] = registered_name(ctx, uuid)
+            name = registered_name(ctx, uuid)
+            name === nothing && pkgerror("cannot find name corresponding to UUID $(uuid)")
```
oxinabox: I feel we should do more to emphisize the seriousness of this error.
Since it is not something someone should be expected to resolve on their own, or by reading docs.

oxinabox: maybe we should have `criticalpkgerror` for this kinda thing?

fredrikekre: I'm not sure why this error would be more serious than any other error.

oxinabox:
```
(12) pkg> test
ERROR: trying to test unnamed project
```

is something the user can fix.
It is a user error.

This on the other hand,  it took you and I over an hour.
And you know a whole lot more about the package system internals than the average user could ever be expected to.

fredrikekre: In this case it was a broken install, but this can happen in normal use also, e.g. if you send me a project + manifest with some private package. I just don't see how the extra `"this is serious"` helps anything, that doesn't give the user any more information. I'll add something like
```
"cannot find name corresponding to UUID $(uuid) in a registry"
```

oxinabox: Sounds good to me

PR #1348, opened 2019-09-03
[RFC] Modularize `Types.jl`

00vareladavid:


KristofferC:
Funny, I had started a branch that removed all submodules :P

fredrikekre:
Yea, the submodules are mostly annoying.

00vareladavid:
@KristofferC Interesting, can you explain the merit of using less modules than we have now?

@fredrikekre I agree that they are a bit annoying

KristofferC:
We chunks like this in almost every file:

https://github.com/JuliaLang/Pkg.jl/blob/f596c0e10d95a8f48b5a52fb254b7522b4c0494f/src/API.jl#L11-L16

It's just kinda a pain.

00vareladavid:
Yes, that's exactly what I am trying to avoid. The interface for `Types` is even worse: https://github.com/JuliaLang/Pkg.jl/blob/f596c0e10d95a8f48b5a52fb254b7522b4c0494f/src/Types.jl#L5-L33

Compare that to the interface for `Manifests`: https://github.com/JuliaLang/Pkg.jl/blob/43d5520c5b888e08c35370e2e029211ae8640161/src/Type/Manifests.jl#L3-L7

KristofferC:
Sure, if the file is smaller then it will need fewer things. But I don't think that change the fact that we are just adding blindly import the things we use in all files. So what is the point of the modules at all then?

00vareladavid:
> Sure, if the file is smaller then it will need fewer things.

Yes, we can agree on this.

> So what is the point of the modules at all then?

The utility of smaller modules for me is that they allow me to more easily reason about a set of functions. It reinforces what functions are supposed to be more tightly coupled (by being in the same module) and which of those functions will be useful to the rest of the code base (this is signaled by `export`). I am not sure I understand your argument against them (your second sentence).

Please note that I am trying to understand your point of view, and I appreciate you taking the time to explain. Of course, I am on board with whatever direction is better for the codebase.

00vareladavid:
To phrase things more productively: ideally, in what way would the other maintainers organize the codebase? Can we at least agree that the current situation is suboptimal?

I am aware that I am being a bit picky, but it really bothers me when I see dead code lying around or a function defined in one module only to be used once in a separate module.

KristofferC:
> Can we at least agree that the current situation is suboptimal?

Yes. And the worst file is indeed the Types.jl file (which right now doesn't really have anything to do with Types).

Here are my personal thoughts on submodules.

In order for submodules to work, they basically need to be "mini packages" that you can properly document. What I mean with that is that you can write them without thinking almost anything about the other code of the package. Some examples of this where it would make sense is:
1) The resolver. The resolver has a quite small API interface and we don't really need any of the internals in there. So we could have a `Resolver` submodule that we make a few call into.
2) The TOML module. Self explanatory, it is just a full fledged package.
3) The "compat string -> semver" functionality. It is also a limited API (`String`-> `VersionBound`) and in other languages there are packages called SemVer that does pretty much it. Right now it is `Pkg.Types.semver_spec` but it could have been nicer with `Pkg.SemVer.spec`.

When submodules don't work very well when you basically just put a type you have and some operations with the type and shove that into a submodule. There is no clear external API so what ends up happening is just that we export (or import) the functionality that is currently used and call that the "public" API. But then when the code evolves and we need more stuff from the submodule, we add those to the export / import list and go on. With time, more and more gets exposed and you end up with these monster export / import list which is what we kind of have now. As an example from this PR:

```
import ..Pkg
import ..RegistryOps, ..GitRepos, ..Operations, ..depots, ..logdir, ..TOML, ..GitTools,
       ..Projects, ..GitOps
using  ..Artifacts: artifact_paths
using  ..VersionTypes: VersionSpec
using  ..PackageSpecs: _VersionTypes
using  ..BinaryPlatforms: Platform, platform_key_abi
using  ..PackageResolve, ..Contexts, ..PackageSpecs, ..PkgErrors, ..Utils, ..Infos
```

We even import a `_VersionTypes`. To support this, it seems the refactoring in this PR increases the lines of code with ~10%

I personally also don't really like very small files. Navigating a code base with a bunch of small files is kinda annoying because you can't scroll or search in the file very well. There are of course limits to this but, for example, I don't think there is anything really wrong with the `Operations.jl` file, even though it is quite large.

So the question then is, what do we do with the current code?  In my opinion, it would almost make sense to try out a "fresh start" with no submodules (except the obvious ones) and then see if there really are contained "sub projects" in the package that are worth factoring out into modules. But then we should almost enforce some kind of internal documentation for the submodule so it is clear upfront what it does and what the API it exposes is.

fredrikekre:
I agree with Kristoffer. The only advantage with the current situation is that most stuff is hidded from `Pkg.<TAB>`.

00vareladavid:
K, back to the drawing board. Thank you both for the feedback

StefanKarpinski:
I wanted to add, not only are the module annoying, they're pretty arbitrary. We're mostly just putting random stuff in random submodules. If there was more of a clear delineation it would both be less annoying and less arbitrary.

00vareladavid:
Yes, I was _trying_ to make the modules less arbitrary in this PR. But I fell into the mindset of "class -> module interface" (a philosophy I do not agree with!!). Good thing you guys talked some sense into me :grin:

PR #1350, opened 2019-09-04
Fix `compress_versions()` confusion with build number VersionNumber's

staticfloat:
This is not a permanent fix, it's to paper over the fact that we have
two, slightly incompatible concepts of a `VersionNumber` within `Pkg`;
the `VersionNumber` and the `VersionBound`.  The latter does not support
`build` or `prerelease` specifiers, which causes much confusion within
`compress_versions()` when it attempts to compare equality of
`VersionNumber` objects with their supposedly-equivalent `VersionBound`
objects.  To fix this, we explcitly strip those pieces of information
before doing any comparisons.

### DIFF
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -262,7 +262,12 @@ subset of the pool of available versions, this function computes a `VersionSpec`
 includes all versions in `subset` and none of the versions in its complement.
 """
 function compress_versions(pool::Vector{VersionNumber}, subset::Vector{VersionNumber})
-    subset = sort(subset) # must copy, we mutate this
+    # Explicitly drop prerelease/build numbers, as those can confuse this.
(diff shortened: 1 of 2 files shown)


staticfloat:
bors try

staticfloat:
bors try

KristofferC:
Will this be a problem for pre 1.3 Pkgs?

staticfloat:
The only time I've seen `compress_versions()` triggered in this way is when running `Registrator`.

KristofferC:
Yeah, true.

fredrikekre:
Registrator has its own implementation of this function, so you need to update there.

staticfloat:
Bors r+

PR #1354, opened 2019-09-06
Offload `PackageSpec` checking from REPL to API layer

### LINKED ISSUE #589: Bad error message for `]add Foo, Bar`
```
(v0.7) pkg> add Images, TestImages, ImageInTerminal
  Cloning git-repo `Images,`
ERROR: failed to clone from Images,, error: GitError(Code:ERROR, Class:Net, unsupported URL protocol)
```

Could also just allow that syntax and it shouldn't drastically complicate the implementation since `,` are (hopefully) not allowed in package names.

### LINKED ISSUE #1257: "Error:Invalid Token" when dev-ing specific branch from registered package
Some of the various combination I've tried  (please excuse the DataFrames typos)
image

00vareladavid:
This checking is necessary in the API layer, so there is no need to duplicate it in the REPL layer. Also tweaked the API error handling to be a bit more clear/verbose.

---
fix #1257 
Example:
```
(foobar) pkg> dev JSON Example#master
ERROR: It is invalid to specify a git revision. git revision `master` was specified for package `Example`.
```

close #589
Example:
```
(v1.0) pkg> dev /tmp/idonotexist
[ Info: Resolving package identifier `/tmp/idonotexist` as a URL.
   Cloning git-repo `/tmp/idonotexist`
ERROR: failed to clone from /tmp/idonotexist, error: GitError(Code:ERROR, Class:Net, unsupported URL protocol)
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -56,6 +56,14 @@ function check_package_name(x::AbstractString, mode=nothing)
     return PackageSpec(x)
 end
 
+function err_rep(pkg::PackageSpec)
+    x = pkg.name !== nothing && pkg.uuid !== nothing ? x = "$name [$(string(pkg.uuid)[1:8])]" :
+        pkg.name !== nothing ? pkg.name :
+        pkg.uuid !== nothing ? string(pkg.uuid)[1:8] :
+        pkg.repo.url
+    return "`$x`"
+end
+
 develop(pkg::Union{AbstractString, PackageSpec}; kwargs...) = develop([pkg]; kwargs...)
 develop(pkgs::Vector{<:AbstractString}; kwargs...) =
     develop([check_package_name(pkg, :develop) for pkg in pkgs]; kwargs...)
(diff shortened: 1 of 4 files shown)


src/API.jl
```diff
@@ -67,12 +75,20 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
...
-        pkg.version == VersionSpec() ||
-            pkgerror("Can not specify version when tracking a repo.")
+        if pkg.name == "julia"
+            pkgerror("`julia` is not a valid package name.")
+        end
+        if pkg.name === nothing && pkg.uuid === nothing && pkg.repo.url === nothing
+            pkgerror("Packages must be specified by name, UUID, URL, or filesystem path.")
+        end
+        if pkg.repo.rev !== nothing
+            pkgerror("It is invalid to specify a git revision.",
```
KristofferC: Not always so it would be good to be clear about why it is invalid here. Also, I personally prefer one sentence error messages, so perhaps something like

```
git revision cannot be specified in `develop` , got `$(pkg.repo.rev)` for package $(err_rep(pkg))
```

?

00vareladavid: Agree completely, I'll change it

PR #1356, opened 2019-09-08
remove support for REQUIRE files

KristofferC:
all packages must now have a Project file to be used in a checked out state


KristofferC:
bors try

KristofferC:
bors try

KristofferC:
bors cancel

KristofferC:
bors try

StefanKarpinski:
What a relief to get rid of all that junk!

test/repl.jl
```diff
@@ -358,37 +337,6 @@ temp_pkg_dir() do project_path; cd(project_path) do
         @test "Example" in c
         c, r = test_complete("rm Exam")
         @test isempty(c)
-        Pkg.REPLMode.pkgstr("develop $(joinpath(@__DIR__, "test_packages", "RequireDependency"))")
```
00vareladavid: Will this be brought back? This test is still useful and not `REQUIRE` specific

00vareladavid: It just needs a test package with a project file like this:
```
name = "Simple"
uuid = "632ccb1b-42db-4992-98c9-5cd245e24532"
version = "0.1.0"

[deps]
```

src/Operations.jl
```diff
@@ -254,7 +229,7 @@ function collect_fixed!(ctx::Context, pkgs::Vector{PackageSpec}, names::Dict{UUI
 
         found_project = collect_project!(ctx, pkg, path, fix_deps_map)
         if !found_project
-            collect_require!(ctx, pkg, path, fix_deps_map)
```
00vareladavid: If we changed this line: https://github.com/JuliaLang/Pkg.jl/blob/b51b4b1db554cc4afc46e435eac6b010c2a2bbce/src/Operations.jl#L200 to something like
```
project_file !== nothing || pkgerror("No project file found at path `$path`.")
```
then we wouldn't have to do error handling outside of the function

KristofferC: Yeah, but I wasn't sure if it ever was ok to return `nothing` from this. Will look where the function is used.

test/repl.jl
KristofferC: Added back.

src/Operations.jl
KristofferC: Updated

Issue #515, opened 2018-07-20
Context kwargs should just be a single kwarg in API functions instead of API just passing kwargs... to it


KristofferC:
Bad error messages otherwise and there is currently no documentation about it so for now it is just an implementaion.

00vareladavid:
I've been thinking about this. What about something like this?

- `Pkg.develop([:preview => true], "Example"; devdir=true)` : for callers who don't care about `Context` and just want to preview
- `Pkg.develop(ctx, "Example"; devdir=true)` : for callers who already have a `Context` object (they should just modify it themselves)
- `Pkg.develop("Example"; devdir=true)` : this one will just create a `Context` object with default values

All context options will go in their own explicit argument. Arguments specific to the API function will be keyword arguments.

KristofferC:
2 years later, `Context` shouldn't exist at all.

PR #1358, opened 2019-09-08
Add a more concise way to call the API functions

KristofferC:
When using the `Pkg.API` calls and wanting to e.g. add something by URL, it is a bit verbose to have to write

```jl
Pkg.add(PackageSpec(url = "..."))
```

Now it is possible to write

```jl
Pkg.add(url = "...")
```

in other words, relevant keyword arguments gets passed directly to the PackageSpec constructor. For multiple packages we can use a Vector of NamedTuples where each named tuple gets passed:

```jl
Pkg.add([(url = "...",), (url = "...",)])
```

Previously, all keyword arguments were passed directly to the `Context!` modifier but that is a real waste because no one uses that.

I need to expand the docs a bit and ideally this should be added for the `Pkg.Registry` module as well but that might be a later project...

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -119,11 +119,11 @@ The following table describes the argument values for `preserve` (in order of st
 ```julia
 Pkg.add("Example") # Add a package from registry
 Pkg.add("Example"; preserve=Pkg.PRESERVE_ALL) # Add the `Example` package and preserve existing dependencies
-Pkg.add(PackageSpec(name="Example", version="0.3")) # Specify version; latest release in the 0.3 series
-Pkg.add(PackageSpec(name="Example", version="0.3.1")) # Specify version; exact release
-Pkg.add(PackageSpec(url="https://github.com/JuliaLang/Example.jl", rev="master")) # From url to remote gitrepo
-Pkg.add(PackageSpec(url="/remote/mycompany/juliapackages/OurPackage")) # From path to local gitrepo
-Pkg.add(PackageSpec(url="https://github.com/Company/MonoRepo", subdir="juliapkgs/Package.jl)")) # With subdir
+Pkg.add(name="Example", version="0.3") # Specify version; latest release in the 0.3 series
+Pkg.add(name="Example", version="0.3.1") # Specify version; exact release
+Pkg.add(url="https://github.com/JuliaLang/Example.jl", rev="master") # From url to remote gitrepo
+Pkg.add(url="/remote/mycompany/juliapackages/OurPackage") # From path to local gitrepo
+Pkg.add(url="https://github.com/Company/MonoRepo", subdir="juliapkgs/Package.jl)") # With subdir
 ```
 
 See also [`PackageSpec`](@ref).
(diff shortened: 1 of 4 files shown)


DilumAluthge:
Awesome!

src/Pkg.jl
```diff
@@ -403,16 +404,26 @@ in the vector.
...
+
+!!! compat "Julia 1.4"
+    In Julia 1.4, many functions that take a `PackageSpec` or a `Vector{PackageSpec}` can be called with a more concise notation.
+    For example, `Pkg.add` can be called either as the explicit or concise versions as:
+
+    | Explicit                                                            | Concise
+    |:--------------------------------------------------------------------|:-----------------------------------------------|
+    | `Pkg.add(PackageSpec(name="Package))`                               | `Pkg.add(name = "Package")`                    |
+    | `Pkg.add(PackageSpec(url="www.myhost.com/MyPkg")))`                 | `Pkg.add(name = "Package")`                    |
+    |` Pkg.add([PacakgeSpec(name="Package"), PackageSpec(path="/MyPkg"])` | `Pkg.add([(name="Package",), (path="MyPkg",)])`|
```
DilumAluthge: Minor typo in the docstring: `PacakgeSpec`

00vareladavid:
I would strongly prefer to keep the API as dumb as possible. I believe we have generally agreed that if users want something more user friendly, they should just use the REPL.

DilumAluthge:
I'd be on board with just using the REPL if pasting in REPL commands worked correctly.

But, for example, if you copy the following code and paste it into a Julia REPL:
```julia
] status
```

You'll get the error `ERROR: syntax: unexpected "]"`.

Given that copy-pasting into the Pkg REPL mode doesn't work, I feel like it'd be good to have a concise syntax for certain API functions so that I can tell a user to copy and paste this:
```julia
Pkg.add(url = "...")
```

And have it "just work".

00vareladavid:
I think copy and pasting is a separate issue. In any case, pkg"add ..." works just fine

KristofferC:
> I would strongly prefer to keep the API as dumb as possible. 

Well, I would say that this is "dumber" than before.

>  I believe we have generally agreed that if users want something more user friendly, they should just use the REPL.

Not the way I see it, we should always try to be user-friendly. What we have agreed on is that it is ok for the REPL mode to be slightly ambiguous while the API should be very unambiguous. Nothing here changes the ambiguity level, it gives an arguably nicer way to write things.

00vareladavid:
> What we have agreed on is that it is ok for the REPL mode to be slightly ambiguous while the API should be very unambiguous.

Yes, that's a good distinction. Apologies, I didn't mean to put words in your mouth.

I'm just a bit wary of the error messages e.g. `Pkg.add(URL="foo")` does not hint that the kwarg is supposed to be `url`(lowercased) but `Pkg.add(Pkg.PackageSpec(URL="foo"))` does. I suppose if people want better error messages we can just tell them to use the more verbose option.

00vareladavid:
Closes #515?

DilumAluthge:
Bump?

DilumAluthge:
Bump. Whatever happened to this?

KristofferC:
Rebased

PR #1357, opened 2019-09-08
WIP: Some refactoring of submodules

KristofferC:
This is a PR to make some incremental progress on how we structure the code in Pkg. The PR is built upon #1356 so I put the merge branch to that one to not make the diff noisy.

Changes:

- Move `compress.jl` into `bin` since it is never used within Pkg itself.
- Put GraphType into the Resolve module since it is basically only used there.
- Put `Resolve.jl` into the folder `Resolve` to have the whole resolve functionality contained there.
- Unsubmodularize some of the files in `Resolve`, they mostly just exported everything anyway.
- Move `REPLMode.jl` into the (renamed from `REPL`) folder `REPLMode.jl` where now all files for `REPLMode` lives.
- Move out `set_readonly` from `GitTools.jl` into a separate `utils.jl` file, it didn't make sense in `GitTools.jl`.

Current file structure:

```
.
├── API.jl
├── Artifacts.jl
├── BinaryPlatforms.jl
├── Display.jl
├── GitTools.jl
├── Operations.jl
├── Pkg.jl
├── PlatformEngines.jl
├── REPLMode
│   ├── REPLMode.jl
│   ├── argument_parsers.jl
│   ├── command_declarations.jl
│   └── completions.jl
├── Registry.jl
├── Resolve
│   ├── Resolve.jl
│   ├── graphtype.jl
│   ├── maxsum.jl
│   └── versionweights.jl
├── Types.jl
├── backwards_compatible_isolation.jl
├── generate.jl
├── manifest.jl
├── project.jl
├── utils.jl
└── versions.jl
```

As I said, this is just an incremental improvement, not the refactor to end all refactorings in Pkg.

TODO: Fix tests :)

PR #1361, opened 2019-09-09
[release-1.3] Backports for 1.3.0

fredrikekre:


fredrikekre:
bors try

src/versions.jl
```diff
@@ -262,7 +262,12 @@ subset of the pool of available versions, this function computes a `VersionSpec`
 includes all versions in `subset` and none of the versions in its complement.
 """
 function compress_versions(pool::Vector{VersionNumber}, subset::Vector{VersionNumber})
-    subset = sort(subset) # must copy, we mutate this
+    # Explicitly drop prerelease/build numbers, as those can confuse this.
+    # TODO: Rewrite all this to use VersionNumbers instead of VersionBounds
```
KristofferC: As a heads up, in #1357 I am moving this stuff into `bin` since it is unused and untested in Pkg.

KristofferC: Eeh, actually it is tested...

KristofferC: Oh  because the relevant PR added tests... Anyway, this seems like a registry specific thing (there is no need to compress anything from Pkgs side so it seems it should live in Registrator.jl (which we will refactor out a RegisterTools.jl from).

fredrikekre:
bors try

PR #1362, opened 2019-09-09
Some refactoring of submodules

KristofferC:
Something weird happened with #1357


KristofferC:
bors retry

KristofferC:
compress is in Registrator: https://github.com/JuliaRegistries/Registrator.jl/blob/e16dee52bbd9b2d5ff87f34546199fe0c32afc7f/src/regedit/register.jl#L154-L214

KristofferC:
But you could of course always load the compress module with `include` and  `pathof(Pkg)`

staticfloat:
the function is, but it still attempts to import `Pkg.Compress`:

```
https://github.com/JuliaRegistries/Registrator.jl/blob/e16dee52bbd9b2d5ff87f34546199fe0c32afc7f/src/regedit/register.jl#L190
```

PR #1363, opened 2019-09-09
Verify that content of unpacked tarball matches git-tree-sha1.

fredrikekre:


KristofferC:
bors try

KristofferC:
Let's just do this?

StefanKarpinski:
May not work on Windows? cc @staticfloat

KristofferC:
It passed CI on windows at least. But yeah, might be some stuff with the executable bit. Perhaps best to save this for 1.4 at least.

staticfloat:
Yes, if someone has put an executable file (as marked by permissions bits) inside their git repo, this won't work until we fix `stat()` on Windows.  Probably best to leave for 1.4.

StefanKarpinski:
How about we do this but just not on Windows? It's better to be safe on most platforms than on none. If it's possible to easily check if any files in the tarball have the executable bit set, then we could do this check unless `Sys.iswindows() && contains_executable_files(tarball)`.

staticfloat:
Hmmm, yes, `7z` does give us the ability to parse out the executable bits, I believe.

StefanKarpinski:
It may even be possible to check directly that a tarball contains the correct tree before unpacking it.

staticfloat:
Ah, you mean we unpack it in-memory and hash the contents, so that we don't pollute the disk with a tarball that has been corrupted?  The benefit there being we don't extract-then-delete on hash failure, the downside being that we would extract everything twice.

StefanKarpinski:
Eh, probably better to just fix stat and chmod on Windows.

fredrikekre:
I will disable on windows for now. Is there an open issue to track the problem?

staticfloat:
Here's an issue: https://github.com/JuliaLang/julia/issues/33212

fredrikekre:
It's always Julia 1.3 on macos that can't connect to internet. Wonder why.

Issue #1091, opened 2019-02-27
implicit roots for manifest without project


StefanKarpinski:
It's becomes somewhat common to send just a manifest file and a source file in order to provide a MWE of a bug. In this case, the recipient of the code and manifest needs to reconstruct a project file in order to run the code in question. To facilitate this situation, I propose that we consider a directory with just a manifest file and no project file to implicitly have a project file with just a `[deps]` section with entries for every package in the manifest whose name is unique. This will require changes to code loading, so technically it's a Base Julia issue as well, but I figured I'd post it here first to see what people think. cc @JeffBezanson who motivated the idea for me.

tpapp:
A less automatic but still rather convenient way would be exposing a function that reconstructs a `Project.toml` from the `Manifest.toml` as part of the API. Given a `Manifest.toml`, one could just call that.

Detecting a `Manifest.toml` without a `Project.toml` while loading code in an interactive session, Julia could prompt offering to call this function.

fredrikekre:
I have been thinking about this too and my idea was to use `instantiate` for this (`instantiate` would be the API to construct the `Project.toml` as tpapp suggests). I think this makes sense and you can then set up an environment with any combination of the files, and the workflow would be the same no matter what combination of files you get; you always activate the project and `instantiate`. `instantiate` would then have the following behavior:
1. `instantiate` a `Project.toml` file creates a `Manifest.toml` file, resulting in an environment with `Project.toml/[deps]` available.
2. `instantiate` a `Project.toml` + `Manifest.toml` pair results in an environment with `Project.toml/[deps]` available, and versions fixed by the given manifest.
3. `instantiate` a `Manifest.toml` results in an environment with all deps in `Manifest.toml` available, and versions fixed.

> with entries for every package in the manifest whose name is unique.

With the `instantiate` proposal above we can (in an interactive environment) query the user for which of the deps to put in the project file.

> This will require changes to code loading

One advantage of this idea is that this should not be needed.

StefanKarpinski:
Yes, that's an even better idea, @fredrikekre 😁

PR #1364, opened 2019-09-09
add support for instantiating lonely manifest

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/1091.

~~Not sure if the presence of a manifest file in the current directory is the best way to activate this feature or if we should enforce a flag to instantiate to do it.~~

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -709,6 +709,19 @@ instantiate(; kwargs...) = instantiate(Context(); kwargs...)
 function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
                      update_registry::Bool=true, verbose::Bool=false, kwargs...)
     Context!(ctx; kwargs...)
+    if !isfile(ctx.env.project_file) && isfile(ctx.env.manifest_file)
+        _manifest = Pkg.Types.read_manifest(ctx.env.manifest_file)
+        deps = Dict()
+        for (uuid, pkg) in _manifest
+            if pkg.name in keys(deps)
+                # TODO, query what package to put in Project when in interactive mode?
+                pkgerror("cannot instantiate a manifest without project file when the manifest has multiple packages with the same name ($(pkg.name))")
+            end
+            deps[pkg.name] = string(uuid)
+        end
+        Types.write_project(Dict("deps" => deps), ctx.env.project_file)
+        return instantiate(Context(); manifest=manifest, update_registry=update_registry, verbose=verbose, kwargs...)
+    end
     if (!isfile(ctx.env.manifest_file) && manifest == nothing) || manifest == false
         up(ctx; update_registry=update_registry)
         return
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -311,12 +311,14 @@ const project = API.project
 """
     Pkg.instantiate(; verbose = false)
 
-If a `Manifest.toml` file exists in the current project, download all
+If a `Manifest.toml` file exists in the active project, download all
 the packages declared in that manifest.
 Otherwise, resolve a set of feasible packages from the `Project.toml` files
 and install them.
 `verbose = true` prints the build output to `stdout`/`stderr` instead of
 redirecting to the `build.log` file.
+If no `Project.toml` exist in the current active project, create one with all the
+dependencies in the manifest and instantiate the resulting project.
 """
 const instantiate = API.instantiate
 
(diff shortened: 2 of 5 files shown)


fredrikekre:
> Not sure if the presence of a manifest file in the current directory is the best way to activate this feature or if we should enforce a flag to instantiate to do it.

My thinking was that we do this if the project is activated and there is no project file.

KristofferC:
Ah, makes sense.

KristofferC:
How about this?

src/API.jl
```diff
@@ -709,6 +709,22 @@ instantiate(; kwargs...) = instantiate(Context(); kwargs...)
...
+        _manifest = Pkg.Types.read_manifest(manifestfile)
+        deps = Dict()
+        for (uuid, pkg) in _manifest
+            if pkg.name in keys(deps)
+                # TODO, query what package to put in Project when in interactive mode?
+                pkgerror("cannot instantiate a manifest without project file when the manifest has multiple packages with the same name ($(pkg.name))")
+            end
+            deps[pkg.name] = string(uuid)
+        end
+        Types.write_project(Dict("deps" => deps), "Project.toml")
```
fredrikekre: The project might not be in the currect directory:
```suggestion
        Types.write_project(Dict("deps" => deps), Types.find_project_file())
```

fredrikekre: Or I guess:
```suggestion
        Types.write_project(Dict("deps" => deps), ctx.env.project_file)
```

KristofferC: Yeah, I accidentally left some of that assumption in, thanks.

src/API.jl
```diff
@@ -709,6 +709,22 @@ instantiate(; kwargs...) = instantiate(Context(); kwargs...)
 function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
                      update_registry::Bool=true, verbose::Bool=false, kwargs...)
     Context!(ctx; kwargs...)
+    if !isfile(ctx.env.project_file) && isfile(ctx.env.manifest_file)
+        if manifest == false
+            pkgerror("cannot instantiate without a project file with `manifest=false` option")
+        end
+        _manifest = Pkg.Types.read_manifest(manifestfile)
```
```suggestion
        _manifest = Pkg.Types.read_manifest(ctx.env.manifest_file)
```

KristofferC:
bor try

KristofferC:
bors try

KristofferC:
bors retry

src/Pkg.jl
```diff
@@ -311,12 +311,14 @@ const project = API.project
...
 
-If a `Manifest.toml` file exists in the current project, download all
+If a `Manifest.toml` file exists in the active project, download all
 the packages declared in that manifest.
 Otherwise, resolve a set of feasible packages from the `Project.toml` files
 and install them.
 `verbose = true` prints the build output to `stdout`/`stderr` instead of
 redirecting to the `build.log` file.
+If no `Project.toml` exist in the current active project, create one with all the
+dependencies in the manifest.
```
00vareladavid:
```suggestion
dependencies in the manifest, then install all the packages declared in the manifest.
```
Perhaps this is more clear that you proceed with a normal `instantiate` after the project file is created.

src/API.jl
```diff
@@ -709,6 +709,22 @@ instantiate(; kwargs...) = instantiate(Context(); kwargs...)
...
+        deps = Dict()
+        for (uuid, pkg) in _manifest
+            if pkg.name in keys(deps)
+                # TODO, query what package to put in Project when in interactive mode?
+                pkgerror("cannot instantiate a manifest without project file when the manifest has multiple packages with the same name ($(pkg.name))")
+            end
+            deps[pkg.name] = string(uuid)
+        end
+        Types.write_project(Dict("deps" => deps), ctx.env.project_file)
+        return instantiate(Context(); manifest=manifest, update_registry=update_registry, verbose=verbose)
```
```suggestion
        return instantiate(Context(); manifest=manifest, update_registry=update_registry, verbose=verbose, kwargs...)
```

KristofferC: The kwargs are already applied to the context though?

00vareladavid: But this line creates a new context as well, it doesn't seem that e.g. `io=stdout` would go through

KristofferC: Oh yeah...

Issue #996, opened 2019-01-15
preview up doesn't work properly


jebej:
I am assuming that the point of `preview` is to see what the effect of an operation would be, and would therefore expect that `preview up` would tell me which packages would get updated. This does not see to be the case on 1.0.3:

```
(v1.0) pkg> preview up                                                           
───── Preview mode ─────                                                         
[ Info: Skipping updating registry in preview mode                               
 Resolving package versions...                                                   
  Updating `C:\Users\Jeremy\.julia\environments\v1.0\Project.toml`               
 [no changes]                                                                    
  Updating `C:\Users\Jeremy\.julia\environments\v1.0\Manifest.toml`              
 [no changes]                                                                    
  Building skipping building in preview mode                                     
───── Preview mode ─────                                                         
                                                                                 
(v1.0) pkg> up                                                                   
  Updating registry at `C:\Users\Jeremy\.julia\registries\General`               
  Updating git-repo `https://github.com/JuliaRegistries/General.git`             
 Resolving package versions...                                                   
 Installed PyPlot ────────── v2.7.0                                              
 Installed SoftGlobalScope ─ v1.0.9                                              
 Installed Tokenize ──────── v0.5.3                                              
 Installed IJulia ────────── v1.15.2                                             
 Installed StaticArrays ──── v0.10.2                                             
 Installed StatsBase ─────── v0.27.0                                             
 Installed Atom ──────────── v0.7.12                                             
 Installed OrdinaryDiffEq ── v4.19.0                                             
 Installed CodeTools ─────── v0.6.4                                              
 Installed DataStructures ── v0.15.0                                             
 Installed WebSockets ────── v1.2.0                                              
 Installed Cassette ──────── v0.2.1                                              
  Updating `C:\Users\Jeremy\.julia\environments\v1.0\Project.toml`               
  [c52e3926] ↑ Atom v0.7.11 ⇒ v0.7.12                                            
  [7073ff75] ↑ IJulia v1.14.1 ⇒ v1.15.2                                          
  [1dea7af3] ↑ OrdinaryDiffEq v4.18.3 ⇒ v4.19.0                                  
  [d330b81b] ↑ PyPlot v2.6.3 ⇒ v2.7.0                                            
  Updating `C:\Users\Jeremy\.julia\environments\v1.0\Manifest.toml`              
  [c52e3926] ↑ Atom v0.7.11 ⇒ v0.7.12                                            
[...]

PR #1367, opened 2019-09-10
remove preview command

KristofferC:
`preview` had good intentions. By allowing to check what happened when executing a command without it taking effect, it was intended for the user to avoid pkg commands that mess up their project. However, in practice, no one think they will do a mistake so no one uses preview. It is also quite invasive in the code base. You constantly need to think if you are running things in preview mode and in that case try to emulate the normal usage but at the same time don't do any persistent mutations.

What is needed is instead `undo` + `redo`. This just removes `preview` and I will implement `undo` + `redo` to replace it (unless someone beats me to it)

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -123,7 +123,7 @@ from that local repo are pulled when packages are updated.
 Note that changes to files in the local package repository will not immediately be reflected when loading that package.
 The changes would have to be committed and the packages updated in order to pull in the changes.
 
-In addition, it is possible to add packages relatively to the `Manifest.toml` file, see [Developing Packages](@ref) for an example. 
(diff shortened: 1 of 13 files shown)


KristofferC:
bors try

00vareladavid:
I think this will close #996

essenciary:
For what is worth, I religiously used preview to see the versions of updated packages and whether or not potential updates introduced breaking changes. 

As for undo/redo wouldn't that be just versioning Manifest.toml (potentially using Git)?

KristofferC:
With the undo redo it would just be

```
pkg > up
...

# didn't like it
pkg> undo
```

It would be "versioning" and in memory and not as painful as having to commit all the time.

PR #1369, opened 2019-09-10
Update link to gen_project.jl script.

fredrikekre:


KristofferC:
Did I remove this... maybe should be put back :P

fredrikekre:
Nah, we can link to an old commit.

Issue #1305, opened 2019-08-16
More structured/faster tests


00vareladavid:
This seems to be annoying everyone.

- First, how do people feel about modifying existing tests? I tend to just pile on the tests. I'm afraid of breaking any existing tests :grimacing: 
- Is there really that much of an expected difference between Linux and Mac/Windows? Or are we hitting some bad case regarding the latter? Was there a point where was not such a huge difference?
- test "Isolation". Perhaps they way we are `temp_pkg_dir` is wasteful? I definitely feel like I abuse it at points. Is some part of creating temp directories expensive on Windows?
- redundancy in what is being tested
- separation between network facing and internal operations. More structured tests would mean we can be more deliberate in choosing when we hit the network.

I'm just spitballing here. Any advice/concerns? Would it be a good idea to restructure tests? Any ideas about what is causing Windows to be so slow? I'm willing to put in some elbow grease to push this forward

KristofferC:
It would be good if we could get an overview of what is actually taking time.

00vareladavid:
K, I guess a Windows virtualbox + something like https://github.com/vchuravy/PProf.jl is a good start?

KristofferC:
Using Julias built-in profiler will not work so well, it gives too "fine-grained" results. My guess is that trying to update the registry over and over takes quite a lot of time.

We could turn that off when testing because we are getting a fresh clone in the start.
We could also try using the Windows bots on Travis, maybe they are faster than AV.

00vareladavid:
Ok, I finally got around to poking around Windows. And holy crap! :grimacing: updating the registry takes an obscene amount of time.

I noticed the 'Antimalware Service' spike at the same time. Could that be the cause of the slowdown?

StefanKarpinski:
We've seen evidence elsewhere of AV and other security systems not liking git. Maybe we should default to just downloading a complete compressed copy of the registry? Compressed it's about 1.2MB which is not that large these days.

KristofferC:
I'm also fine to just redo the tests from scratch, going feature by feature and making sure that coverage looks good. It is a bit scary but if we identify the purpose of every test and structure things a bit better, it's probably beneficial in the long run.

KristofferC:
The first step would probably be to structure up the existing tests a bit.

00vareladavid:
> We could also set things up so the tests could be run in parallel. Assuming we get more than one core per worker on AV, perhaps that could speed things up.

[...]

PR #1372, opened 2019-09-10
Rework repo handling

### LINKED ISSUE #709: more informative error message for add'ing and dev'ing non-existent path
```julia
julia> VERSION
v"1.0.1-pre.0"

(v1.0) pkg> dev /tmp/nonexistent-path
   Cloning git-repo `/tmp/nonexistent-path`
ERROR: failed to clone from /tmp/nonexistent-path, error: GitError(Code:ERROR, Class:Net, unsupported URL protocol)
```

Same error message appears in #677, #589, but those are about new features, this issue is just about making the error message *per se* more informative.

### LINKED ISSUE #1343: Double update of repo when adding by URL
```
(Debugger) pkg> add https://github.com/JuliaDocs/Highlights.jl
   Cloning git-repo `https://github.com/JuliaDocs/Highlights.jl`
  Updating git-repo `https://github.com/JuliaDocs/Highlights.jl`
  Updating git-repo `https://github.com/JuliaDocs/Highlights.jl`
```

Happens on 1.2+. Seems unnecessary?

00vareladavid:
Goals:
- Treat relative paths consistently for `dev` and `add`. `dev` does this correctly, but the behavior for `add` is confusing (see #1215).
- Refactor `handle_repos_*` in hopes of making the internal structure more clear.
- Prevent an unnecessary cache update on `add` (fix #1343).
- Prevent an unnecessary update of cached clone on instantiate. We should only fetch if the tree does not already exist in the clone.

---
close #709

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -3,7 +3,7 @@
 ## Adding packages
 
 There are two ways of adding packages, either using the `add` command or the `dev` command.
-The most frequently used one is `add` and its usage is described first.
+The most frequently used is `add` and its usage is described first.
 
 ### Adding registered packages
 
@@ -97,7 +97,7 @@ To go back to tracking the registry version of `Example`, the command `free` is 
 
 ### Adding unregistered packages
 
-If a package is not in a registry, it can still be added by instead of the package name giving the URL to the repository to `add`:
+If a package is not in a registry, it can be added by specifying a URL to the repository:
 
 ```
 (v1.0) pkg> add https://github.com/fredrikekre/ImportMacros.jl
@@ -118,9 +118,10 @@ For unregistered packages we could have given a branch name (or commit SHA1) to 
 ### Adding a local package
 
 Instead of giving a URL of a git repo to `add` we could instead have given a local path to a git repo.
-This works similarly to adding a URL. The local repository will be tracked (at some branch) and updates
+This works similar to adding a URL. The local repository will be tracked (at some branch) and updates
 from that local repo are pulled when packages are updated.
-Note that changes to files in the local package repository will not immediately be reflected when loading that package.
+Note tracking a package through `add` is distinct from `develop`:
+changes to files in the local package repository will not immediately be reflected when loading that package.
 The changes would have to be committed and the packages updated in order to pull in the changes.
 
 In addition, it is possible to add packages relatively to the `Manifest.toml` file, see [Developing Packages](@ref) for an example.
@@ -147,7 +148,7 @@ Let's try to `dev` a registered package:
 The `dev` command fetches a full clone of the package to `~/.julia/dev/` (the path can be changed by setting the environment variable `JULIA_PKG_DEVDIR`).
 When importing `Example` julia will now import it from `~/.julia/dev/Example` and whatever local changes have been made to the files in that path are consequently
 reflected in the code loaded. When we used `add` we said that we tracked the package repository, we here say that we track the path itself.
-Note that the package manager will never touch any of the files at a tracked path. It is therefore up to you to pull updates, change branches etc.
+Note the package manager will never touch any of the files at a tracked path. It is therefore up to you to pull updates, change branches etc.
 If we try to `dev` a package at some branch that already exists at `~/.julia/dev/` the package manager we will simply use the existing path.
 For example:
 
@@ -157,8 +158,8 @@ For example:
 [ Info: Path `/Users/kristoffer/.julia/dev/Example` exists and looks like the correct package, using existing path instead of cloning
 ```
 
-Note the info message saying that it is using the existing path. As a general rule, the package manager will
-never touch files that are tracking a path.
+Note the info message saying that it is using the existing path.
+When tracking a path, the package manager will never modify the files at that path.
 
 If `dev` is used on a local path, that path to that package is recorded and used when loading that package.
 The path will be recorded relative to the project file, unless it is given as an absolute path.
@@ -182,15 +183,12 @@ Note that if you add a dependency to a package that tracks a local path, the Man
 out of sync with the actual dependency graph. This means that the package will not be able to load that dependency since it is not recorded
 in the Manifest. To synchronize the Manifest, use the REPL command `resolve`.
 
-Similarly with `add`, `dev` can track packages with a path relative to the `Manifest.toml` file. This is done by giving a relative path to `add` or `dev`. Notice that due to the way Julia handles relative paths, when you *give* the path, it is always assumed to be with respect to the current working directory, `pwd()`. However, when the package is loaded its path is indeed resolved with respect to the `Manifest.toml` file. To be on the safe side, you can always change path to the `Manifest.toml` file before adding the relative path. For example:
-```
-julia> cd("path/to/manifest"); # This is the project's folder
-
-(v1.0) pkg> activate .
-
-(SomeProject) pkg> dev src/ExamplePackage # similarly for add
-```
-
+In addition to absolute paths, `add` and `dev` can accept relative paths to packages.
+In this case, the relative path from the active project to the package is stored.
+This approach is useful when the relative location of tracked dependencies is more important than their absolute location.
+For example, the tracked dependencies can be stored inside of the active project directory.
+The whole directory can be moved and `Pkg` will still be able to find the dependencies
+because their path relative to the active project is preserved even though their absolute path has changed.
 
 ## Removing packages
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -37,8 +37,8 @@ function project(ctx::Context)
 end
 
 function check_package_name(x::AbstractString, mode=nothing)
-    if !(occursin(Pkg.REPLMode.name_re, x))
-        message = "$x is not a valid packagename."
+    if !Base.isidentifier(x)
+        message = "`$x` is not a valid package name."
         if mode !== nothing && any(occursin.(['\\','/'], x)) # maybe a url or a path
             message *= "\nThe argument appears to be a URL or path, perhaps you meant " *
                 "`Pkg.$mode(PackageSpec(url=\"...\"))` or `Pkg.$mode(PackageSpec(path=\"...\"))`."
@@ -46,14 +46,6 @@ function check_package_name(x::AbstractString, mode=nothing)
         pkgerror(message)
     end
     return PackageSpec(x)
-end
-
-function err_rep(pkg::PackageSpec)
-    x = pkg.name !== nothing && pkg.uuid !== nothing ? x = "$name [$(string(pkg.uuid)[1:8])]" :
-        pkg.name !== nothing ? pkg.name :
-        pkg.uuid !== nothing ? string(pkg.uuid)[1:8] :
-        pkg.repo.url
-    return "`$x`"
 end
 
 develop(pkg::Union{AbstractString, PackageSpec}; kwargs...) = develop([pkg]; kwargs...)
@@ -70,7 +62,7 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
         if pkg.name == "julia"
             pkgerror("`julia` is not a valid package name.")
         end
-        if pkg.name === nothing && pkg.uuid === nothing && pkg.repo.url === nothing
+        if pkg.name === nothing && pkg.uuid === nothing && pkg.repo.source === nothing
             pkgerror("When calling `develop`, packages must be specified by name, UUID, URL, or filesystem path.")
         end
         if pkg.repo.rev !== nothing
@@ -106,21 +98,23 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PR
         if pkg.name == "julia"
             pkgerror("`julia` is not a valid package name.")
         end
-        if pkg.name === nothing && pkg.uuid === nothing && pkg.repo.url === nothing
+        if pkg.name === nothing && pkg.uuid === nothing && pkg.repo.source === nothing
             pkgerror("When calling `add`, packages must be specified by name, UUID, URL, or filesystem path.")
         end
-        if (pkg.repo.url !== nothing || pkg.repo.rev !== nothing)
-            pkg.version == VersionSpec() ||
-                pkgerror("Can not specify version when tracking a repo.")
+        if pkg.repo.source !== nothing || pkg.repo.rev !== nothing
+            if pkg.version != VersionSpec()
+                pkgerror("It is invalid to specify a version when tracking a git repository:",
+                         " `$(pkg.version)` given for package $(err_rep(pkg)).")
+            end
         end
     end
 
-    Types.update_registries(ctx)
-
-    repo_pkgs = [pkg for pkg in pkgs if (pkg.repo.url !== nothing || pkg.repo.rev !== nothing)]
+    repo_pkgs = [pkg for pkg in pkgs if (pkg.repo.source !== nothing || pkg.repo.rev !== nothing)]
     new_git = handle_repos_add!(ctx, repo_pkgs)
-    # repo + unpinned -> name, uuid, repo.rev, repo.url, tree_hash
+    # repo + unpinned -> name, uuid, repo.rev, repo.source, tree_hash
     # repo + pinned -> name, uuid, tree_hash
+
+    Types.update_registries(ctx)
 
     project_deps_resolve!(ctx, pkgs)
     registry_resolve!(ctx, pkgs)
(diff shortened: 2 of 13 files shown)


src/manifest.jl
```diff
@@ -133,13 +137,14 @@ function Manifest(raw::Dict)::Manifest
...
-            entry.version  = read_field("version",       nothing, info, safe_version)
-            entry.path     = read_field("path",          nothing, info, safe_path)
-            entry.repo.url = read_field("repo-url",      nothing, info, identity)
-            entry.repo.rev = read_field("repo-rev",      nothing, info, identity)
-            entry.tree_hash = read_field("git-tree-sha1", nothing, info, safe_SHA1)
+            entry.pinned      = read_pinned(get(info, "pinned", nothing))
+            uuid              = read_field("uuid",          nothing, info, safe_uuid)
+            entry.version     = read_field("version",       nothing, info, safe_version)
+            entry.path        = read_field("path",          nothing, info, safe_path)
+            entry.repo.remote = read_remote(get(info, "repo-remote", nothing))
```
KristofferC: What is this field?

KristofferC: Ah its just a new name for `repo.url`.

00vareladavid: `repo.source` was `repo.url`. And `repo.remote` means that the source is not on the local filesystem

00vareladavid: I'll rework this so it does not introduce manifest changes

test/artifacts.jl
```diff
@@ -338,7 +338,7 @@ end
         copy_test_package(project_path, "ArtifactInstallation")
         Pkg.activate(joinpath(project_path))
         add_this_pkg()
-        Pkg.add(Pkg.Types.PackageSpec(
+        Pkg.develop(Pkg.PackageSpec(
```
00vareladavid: @staticfloat It seems that these tests were set up incorrectly. Due to error handling `Pkg.Types.PackageSpec` is not the same as `Pkg.PackageSpec` (one is for the API one is for internal use). API functions assume the API version is used. `path` maps to `repo.url` because previously `Pkg.develop` allowed the use of git revisions. `Pkg.add` requires the package to be version controlled.

I haven't taken the time to understand the tests but I get:
```
Override.toml: Error During Test at /home/david/pkg/Pkg.jl/test/artifacts.jl:458
  LoadError: Cannot automatically install 'arty'; no download section in '/tmp/jl_pAilNr/ArtifactOverrideLoading/Artifacts.toml'
```

The artifact file seems to confirm this error message.

00vareladavid:
What does the team think of this direction for #1305?

The biggest change is to use a `loaded_depot` points the tests to an _isolated_ test depot which comes with the test packages already installed. It also comes with a loaded registry. This way we don't have to redownload a bunch of stuff. Most of the time, we are not directly testing this aspect of `Pkg` anyway. I figure we can introduce some separate tests to check the correctness of depot changes separately. If anything goes horribly wrong, we can always turn off loading a depot in `isolate` anyway.

Some preliminary performance numbers (for just the new tests!):
- clean depot: `80.826406 seconds (57.00 M allocations: 3.242 GiB, 4.91% gc time)`
- loaded depot: `30.814060 seconds (52.64 M allocations: 2.903 GiB, 11.80% gc time)`

This is on Linux. I suspect we will see a greater performance increase on Windows.

There are more changes, but right now I am mostly hoping for feedback on the testing changes.

fredrikekre:
I think we should do something like this, but I don't think we should put a fully loaded depot in the repository.

00vareladavid:
I was thinking the same. I just pushed it up like this as a demo (also I got excited about :grin:). We can just download it from somewhere else at the start of the tests (and possible move some tests packages over there too).

fredrikekre:
How long time does it take to just set this up once at the beginning? We are at most using the General registry, and install Example and JSON and maybe some other package, so just pre-downloading those should help a lot.

src/API.jl
```diff
@@ -23,22 +23,23 @@ function dependencies(ctx::Context)::Dict{UUID, PackageInfo}
...
-function project(ctx::Context)
-    ctx.env.pkg !== nothing || pkgerror("Active environment is not a project.")
-    return ProjectInfo(
-        name         = ctx.env.pkg.name,
-        uuid         = ctx.env.pkg.uuid,
-        version      = ctx.env.pkg.version,
+dependencies(fn::Function, uuid::UUID) = fn(dependencies()[uuid])
+
+environment() = environment(Context())
+function environment(ctx::Context)::EnvironmentInfo
```
KristofferC: I think this should still be called "project"? AFAIU, environment includes the stack of projects while this only describes a specific project.

src/API.jl
```diff
@@ -23,22 +23,23 @@ function dependencies(ctx::Context)::Dict{UUID, PackageInfo}
...
+    return EnvironmentInfo(
+        name         = ctx.env.pkg === nothing ? nothing : ctx.env.pkg.name,
+        uuid         = ctx.env.pkg === nothing ? nothing : ctx.env.pkg.uuid,
+        version      = ctx.env.pkg === nothing ? nothing : ctx.env.pkg.version,
+        isproject    = ctx.env.pkg !== nothing,
```
KristofferC: This seems like it should be called `ispackage`.

src/API.jl
00vareladavid: Ah yes, I was using different terminology in my head :grimacing:. I'll change it back

00vareladavid:
> How long time does it take to just set this up once at the beginning?

I have not checked this yet.

> We are at most using the General registry, and install Example and JSON and maybe some other package, so just pre-downloading those should help a lot.

We have to install multiple different versions, and also create clones. I suppose we can cycle through the necessary versions and branches. I think downloading a tarball with everything in it should be way faster in any case (especially on Windows).

00vareladavid:
I'll try checking tomorrow though, so we can have some more concrete numbers.

00vareladavid:
Maybe we can pull a two-for-one and use the process of setting up the depot as a test for depot changes? It might be a bit "clever" for testing purposes, but we could end up saving some time if we do it correctly.

KristofferC:
It is a bit hard to review when the repo handling fixes are coupled with test overhaul. Perhaps it is best to focus on just the repo handling fixes here and then the test overhaul can be a separate PR?

00vareladavid:
> It is a bit hard to review when the repo handling fixes are coupled with test overhaul.

Ok, that's fair. I'll need a bit more time

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

src/Types.jl
```diff
@@ -165,6 +170,14 @@ PackageSpec(name::AbstractString, uuid::UUID) = PackageSpec(;name=name, uuid=uui
 PackageSpec(name::AbstractString, version::VersionTypes) = PackageSpec(;name=name, version=version)
 PackageSpec(n::AbstractString, u::UUID, v::VersionTypes) = PackageSpec(;name=n, uuid=u, version=v)
 
+function err_rep(pkg::PackageSpec)
```
KristofferC: But this in `utils.jl` perhaps?

src/API.jl
```diff
@@ -718,12 +726,12 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
...
         return instantiate(Context(); manifest=manifest, update_registry=update_registry, verbose=verbose, kwargs...)
     end
-    if (!isfile(ctx.env.manifest_file) && manifest == nothing) || manifest == false
+    if (!isfile(ctx.env.manifest_file) && manifest === nothing) || manifest == false
         up(ctx; update_registry=update_registry)
         return
     end
     if !isfile(ctx.env.manifest_file) && manifest == true
-        pkgerror("manifest at $(ctx.env.manifest_file) does not exist")
+        pkgerror("Manifest at `$(ctx.env.manifest_file)` does not exist.")
```
KristofferC: Not a big deal but I typically like to start with non-capital letters and base does that a lot too, e.g.:

```
julia> [][1]
ERROR: BoundsError: attempt to access ..
```

src/GitTools.jl
```diff
@@ -103,8 +103,15 @@ function normalize_url(url::AbstractString)
     end
 end
 
-ensure_clone(ctx, target_path, url; kwargs...) =
-    ispath(target_path) ? LibGit2.GitRepo(target_path) : GitTools.clone(ctx, url, target_path; kwargs...)
+function ensure_clone(ctx, target_path, url; kwargs...)
+    if ispath(target_path)
+        @debug "Using existing repo"
```
KristofferC: This seems like it needs a bit more context if it is going to be useful in general. Perhaps it is just for intermediate debugging.

src/REPLMode/argument_parsers.jl
```diff
@@ -71,11 +71,11 @@ end
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
 function parse_package_identifier(word::AbstractString; add_or_develop=false)::PackageSpec
-    if add_or_develop && casesensitive_isdir(expanduser(word))
+    if add_or_develop && isdir(expanduser(word))
```
KristofferC: What made this not require `casesensitive_isdir` anymore? Was it never needed?

src/Types.jl
```diff
@@ -140,14 +142,17 @@ end
 
 const VersionTypes = Union{VersionNumber,VersionSpec,UpgradeLevel}
 
-# The url field can also be a local path, rename?
 Base.@kwdef mutable struct GitRepo
-    url::Union{Nothing,String} = nothing
+    source::Union{Nothing,String} = nothing
```
KristofferC: I wonder if `remote` instead of `source` is a better name. It is the remote which can both be a url and a local path.

test/artifacts.jl
```diff
@@ -326,6 +326,7 @@ end
     end
 end
 
+#=
```
KristofferC: TODO?

test/pkg.jl
```diff
@@ -785,6 +785,7 @@ end
     end end
 end
 
+#=
```
KristofferC: TODO?

test/runtests.jl
```diff
@@ -3,6 +3,8 @@
 module PkgTests
 
 include("utils.jl")
+include("repl.jl")
+#=
```
KristofferC: TODO

src/Types.jl
00vareladavid: Yeah, `remote` is a better name. `Pkg.dependencies()` already uses `source` to mean something else so it might get confusing.

src/GitTools.jl
00vareladavid: Yeah I'll just get rid of this for now

src/Types.jl
KristofferC: I thought about this a bit more and perhaps it is not such a good name after all. If you have a repo on your local disc, the path to that repo is not the remote of the repo. There can also be multiple remotes of a repo etc... Hmm.

00vareladavid:
bors try

00vareladavid:
bors try

staticfloat:
Note that, pursuant to our discussion in the `#pkg-dev` channel, I have just disabled bors.  You will need to rebase on top of https://github.com/JuliaLang/Pkg.jl/pull/1400

00vareladavid:
Ok, thanks!

test/utils.jl
```diff
@@ -160,20 +160,11 @@ function copy_test_package(tmpdir::String, name::String; use_pkg=true)
...
-        path=test_pkg_dir,
-    )
-    Pkg.add(spec)
-end
 
 function add_this_pkg()
     pkg_dir = dirname(@__DIR__)
     pkg_uuid = Pkg.TOML.parsefile(joinpath(pkg_dir, "Project.toml"))["uuid"]
-    spec = Pkg.Types.PackageSpec(
+    spec = Pkg.PackageSpec(
```
00vareladavid: @staticfloat some of these tests were set up incorrectly. I made a comment earlier but perhaps it got drowned out by everything else (the earlier comment is here: https://github.com/JuliaLang/Pkg.jl/pull/1372#discussion_r323056817).

src/API.jl
00vareladavid: Sure, I mostly just care about the consistency. I'm planning a separate PR to make formatting for the error messages we have consistent, I can change it then

src/REPLMode/argument_parsers.jl
```diff
@@ -71,22 +71,20 @@ end
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
 function parse_package_identifier(word::AbstractString; add_or_develop=false)::PackageSpec
-    if add_or_develop && casesensitive_isdir(expanduser(word))
+    if add_or_develop && Types.isdir_windows_workaround(expanduser(word))
```
00vareladavid: @KristofferC It seems that `casesensitive_isdir` is not valid for `.` and `..`. Things were working before because we were just accepting it in the "guess it is a url" part. Given that we have to check the casing of directories in the API layer anyway, I just moved it out REPL layer.

test/utils.jl
```diff
@@ -160,20 +160,11 @@ function copy_test_package(tmpdir::String, name::String; use_pkg=true)
         end
     end
 end
-function add_test_package(name::String, uuid::UUID)
```
00vareladavid: Nothing was using this, so I just removed it here

test/utils.jl
staticfloat: Yes, it did get drowned out.  Do you still see the overriding test error with this change?  That error message means that Julia did not load the `Overrides.toml` file that it should have, which probably means that the depot path wasn't set correctly, or something like that, since it keys off of depot paths in order to determine which override file it should be paying attention to.

00vareladavid: I am not able to reproduce it anymore. :man_shrugging: I guess I'll just keep an eye out for it

00vareladavid:
Does anyone object to a merge? I updated the list of changes at the top. I have also added more complete test coverage at #1384

docs/src/managing-packages.md
```diff
@@ -118,9 +118,10 @@ For unregistered packages we could have given a branch name (or commit SHA1) to
 ### Adding a local package
 
 Instead of giving a URL of a git repo to `add` we could instead have given a local path to a git repo.
-This works similarly to adding a URL. The local repository will be tracked (at some branch) and updates
+This works similar to adding a URL. The local repository will be tracked (at some branch) and updates
 from that local repo are pulled when packages are updated.
-Note that changes to files in the local package repository will not immediately be reflected when loading that package.
+Note tracking a package through `add` is distinct from `develop`:
```
KristofferC:
```suggestion
Note that tracking a package through `add` is distinct from `develop`:
```

KristofferC: Oh, you removed the `that` again later. Imo it should either have the "that" or be something like "Note:"

src/API.jl
```diff
@@ -742,16 +749,33 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
...
+        # determine canonical form of repo source
+        if isurl(pkg.repo.source)
+            repo_source = pkg.repo.source
+        else
+            repo_source = normpath(joinpath(dirname(ctx.env.project_file), pkg.repo.source))
+        end
+        clonepath = Types.clone_path(repo_source)
+        # We try to avoid updating the clone if the target already exists
+        if tree_hash_exists(clonepath, string(pkg.tree_hash))
+            tmp_source = Types.repo_checkout(ctx, clonepath, string(pkg.tree_hash))
```
KristofferC: This line and the one in the `else` below are identical so couldn't this whole thing be changed to

```jl
if !tree_hash_exists(clonepath, string(pkg.tree_hash))
    if !isurl(repo_source) && !isdir(repo_source)
        pkgerror("Expected $(err_rep(pkg)) to exist at `$(repo_source)`, but no such directory.")
    end
    Types.clone_path!(ctx, repo_source) # TODO a more direct way of updating a clone
end
tmp_source = Types.repo_checkout(ctx, clonepath, string(pkg.tree_hash))
```

KristofferC:
LGTM, just some minor comments. Could be fixed some other time.

test/utils.jl
00vareladavid: You can see a stack trace here: https://travis-ci.org/JuliaLang/Pkg.jl/jobs/588723217.
All I did was change `add` to `develop` and used the API facing `PackageSpec`, since that seemed like the most straightforward mapping.

I'm reasonably sure I also tried `add` (tracking by repo) and I hit the same error. Naively, this makes it seem unlikely that a bad `DEPOT_PATH` is the issue.

PR #1381, opened 2019-09-12
small refactor of write_env functions

KristofferC:
I didn't like how we passed in `env.project, env, old_env, ctx` to one function (all of this can be gotten from `ctx` and the first argument `project` was immediately overwritten etc) and also options like `display_diff = true` seems to indicate that that part should just be in its own function. Therefore, factor out the project diff printing into a new function into `Display` and call that function.

Also, slight refactoring to the `write_manifest` and `write_project` to make them look more similar to each other.


KristofferC:
bors retry

PR #1315, opened 2019-08-22
try disabling updating registry on update when running CI

KristofferC:
Experiment to see how this imfact CI time.

PR #1384, opened 2019-09-14
Refactor tests

### LINKED ISSUE #1305: More structured/faster tests

00vareladavid:
close #1305


00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
:man_facepalming: ran the wrong tests

bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
Some rudimentary numbers:
```
                                  CI Time (sec)
                    ┌                                        ┐ 
   Windows (before) ┤■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 378     
   Windows  (after) ┤■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 361       
   Windows  (depot) ┤■■■■■■■■■■■■■ 153                         
       OSX (before) ┤■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 403   
       OSX  (after) ┤■■■■■■■■■■■■■■■■■■■■■■■■■■■■ 317          
       OSX  (depot) ┤■■■■■■■■■■■■■■■■■ 197                     
     Linux (before) ┤■■■■■■■■■■■■■■■■■■ 210                    
     Linux  (after) ┤■■■■■■■■■■■■■■■ 176                       
     Linux  (depot) ┤■■■■■■■■■■■■■ 147                         
                    └                                        ┘ 
```

KristofferC:
> Is that something which is enabled on the CI machines?

That is a good question. We should see if we could disable that, if that is the case.

00vareladavid:
bors try

00vareladavid:
:slightly_frowning_face: does anyone know what is going on with bors?

00vareladavid:
I think I may have found the problem: we are calling `abspath` in a hot loop. Could that be it??

00vareladavid:
bors try

KristofferC:
`abspath` should be really really cheap...

00vareladavid:
bors try

00vareladavid:
> abspath should be really really cheap...

Yea I guess I was hoping that it touched the filesystem in some way :grimacing:, but it does not seem that way from a cursory look.

By the way, I did some testing on Windows, and (at least for `Pkg.add`) there doesn't seem to be much difference (compared to Linux) in where it spends its time.

In any case, Windows and OSX performance seems pretty close for these tests. I'll just keep transferring tests and see if they begin to diverge again.

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
bors try

00vareladavid:
@KristofferC if you have the time, could you take a quick glance at `test/new.jl`? Do you think this is an overall good direction for the tests?

test/utils.jl
```diff
@@ -1,6 +1,65 @@
...
 
+const LOADED_DEPOT = joinpath(@__DIR__, "loaded_depot")
+const FORCE_CLEAN_UNITS = false
+
+function test_depot_setup(; loaded_depot=false)
+    if FORCE_CLEAN_UNITS
+        return mktempdir()
+    end
+
+    Pkg.UPDATED_REGISTRY_THIS_SESSION[] = loaded_depot
```
KristofferC: This still doesn't prevent the registry from updating on `pkg> update`, which we do quite often. Maybe we should add https://github.com/JuliaLang/Pkg.jl/pull/1315 after all.

test/utils.jl
```diff
@@ -1,6 +1,65 @@
...
+        if loaded_depot
+            cp(LOADED_DEPOT, test_depot; force=true)
+        else
+            cp(joinpath(LOADED_DEPOT, "registries"), joinpath(test_depot, "registries"))#TODO should remove in final version?
+        end
+    end
+    return test_depot
+end
+
+function isolate(fn::Function; kwargs...)
```
KristofferC: We should still probably have the`registry` URL hack in here that we have with `temp_pkg_dir` for tests where we do need a new `DEPOT` but don't want to close the registry from the internet anyway.

KristofferC: Oh, I see the copy for registries now. Maybe that is faster.

test/utils.jl
```diff
@@ -1,6 +1,65 @@
...
+        Base.HOME_PROJECT[] = nothing
+        Base.ACTIVE_PROJECT[] = nothing
+        Pkg.UPDATED_REGISTRY_THIS_SESSION[] = false
+        withenv("JULIA_PROJECT" => nothing,
+                "JULIA_LOAD_PATH" => nothing,
+                "JULIA_PKG_DEVDIR" => nothing) do
+            test_depot = nothing
+            try
+                test_depot = test_depot_setup(; kwargs...)
+                push!(LOAD_PATH, "@", "@v#.#", "@stdlib")
```
KristofferC: I wonder if we can remove `"@stdlib"` from this.

KristofferC:
Yeah, identifying the tests that need a fresh depot to test for correcteness and only put those in a new depot is good. I think those tests are things like:

- `instantiate`
- `"libgit2 downloads"`
- `"tarball downloads"`
- some artifact tests that check stuff like `@test !artifact_exists(arty_hash)` and `!isfile(usage_path)`.

KristofferC:
Another thing I have thought about, in the REPL tests we also do a bunch of Pkg correctness tests. It might be worth to have the REPL tests only test that it passes the correct thing to Pkg.API but never actually execute any Pkg commands in the REPL tests  (except maybe a few basic ones).

KristofferC:
> Another thing I have thought about, in the REPL tests we also do a bunch of Pkg correctness tests. It might be worth to have the REPL tests only test that it passes the correct thing to Pkg.API but never actually execute any Pkg commands in the REPL tests (except maybe a few basic ones).

But maybe this is a bit tricky since some of the functions in the REPLMode end up unconditionally calling into API. But we should at least try to minimize correctness checking of Pkg through the REPL tests.

00vareladavid:
bors try

00vareladavid:
> REPL tests only test that it passes the correct thing to Pkg.API but never actually execute any Pkg commands

done :grin:

00vareladavid:
bors try

test/utils.jl
00vareladavid: I think we will probably need the URL hack in some tests so that we can test that registries are being cached correctly. From what I can see, we are not doing that yet

00vareladavid:
This is very close to a mergable state, if anyone would like to take a look.

There is an intermittent failure on Windows, but I wish to avoid going down the Windows rabbit hole yet again for this PR. I'm leaning to just disabling it on windows systems.

src/API.jl
```diff
@@ -58,27 +58,29 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
...
+                     " `$(pkg.version)` specified for package $(err_rep(pkg))")
         end
     end
 
     new_git = handle_repos_develop!(ctx, pkgs, shared)
 
-    any(pkg -> Types.collides_with_project(ctx, pkg), pkgs) &&
-        pkgerror("Cannot `develop` package with the same name or uuid as the project")
+    for pkg in pkgs
+        Types.collides_with_project(ctx, pkg) || continue
```
KristofferC: I would rather just write this as either

```jl
if Types.collides....
    pkgerror()
end
```
I think it is worth the extra line :)

src/Operations.jl
```diff
@@ -932,11 +932,12 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
...
 
-update_package_add(pkg::PackageSpec, ::Nothing, is_dep::Bool) = pkg
-function update_package_add(pkg::PackageSpec, entry::PackageEntry, is_dep::Bool)
+update_package_add(ctx::Context, pkg::PackageSpec, ::Nothing, is_dep::Bool) = pkg
+function update_package_add(ctx::Context, pkg::PackageSpec, entry::PackageEntry, is_dep::Bool)
     if entry.pinned
-        pkg.version == VersionSpec() ||
-            @warn "`$(pkg.name)` is pinned at `v$(entry.version)`. Maintaining pinned version."
+        if pkg.version == VersionSpec()
+            println(ctx.io, "`$(pkg.name)` is pinned at `v$(entry.version)`: maintaining pinned version")
```
KristofferC: We should probably make our own `pkgwarning` or something because I like that you get the color and `WARNING:` with `@warn`. Now it will just be standard text.

src/Types.jl
```diff
@@ -1354,20 +1360,34 @@ end
 Base.@kwdef struct PackageInfo
     name::String
     version::Union{Nothing,VersionNumber}
+    tree_hash::Union{Nothing,String}
     ispinned::Bool
     isdeveloped::Bool
+    is_tracking_registry::Bool
```
KristofferC: What does this one do?

src/Types.jl
```diff
@@ -1354,20 +1360,34 @@ end
...
+    git_revision::Union{Nothing,String}
+    git_source::Union{Nothing,String}
+    git_cache::Union{Nothing,String}
```
KristofferC: What is this one storing?

test/new.jl
```diff
@@ -0,0 +1,1817 @@
...
+        @test isempty(Pkg.dependencies())
+        # A simple `add` should set up some things for us:
+        Pkg.add(Pkg.PackageSpec(;name="Example", version="0.5.3"))
+        # - `General` should be initiated by default.
+        regs = Pkg.Registry.status(;as_api=true)
+        @test length(regs) == 1
+        reg = regs[1]
+        @test reg.name == "General"
+        @test reg.uuid == general_uuid
+        # - The package should be installed correctly.
```
KristofferC: This only checks if the project + manifest files were correctly updated. Don't we want to use the `isinstalled` here to see that the files are actually downloaded?

KristofferC: Oh, that's the `isdir(pkg.source)`?

test/new.jl
```diff
@@ -0,0 +1,1817 @@
...
+            @test isdir(pkg.source) # TODO check for full git clone, have to implement saving original URL first
+        end
+        # Check that the original installation was undisturbed.
+        regs = Pkg.Registry.status(;as_api=true)
+        @test length(regs) == 1
+        reg = regs[1]
+        @test reg.name == "General"
+        @test reg.uuid == general_uuid
+        @test mtime(source053) == source053_time
+        # Now we clean up so that `isolate` can reuse the loaded depot properly
```
KristofferC: I don't really understand the "loaded depot" terminology. The point is that the depot does not get reset, right? There is no need to "load" it initially with what we are going to use. Whatever ends up there ends up there.

src/API.jl
00vareladavid: :laughing: fair enough

src/Types.jl
00vareladavid: it means the package is not tracking a path or tracking a repo

00vareladavid: perhaps there is a better name?

src/Types.jl
00vareladavid: the directory in ".julia/clones" where we cache the git repo

src/Operations.jl
KristofferC: Doesnt have to be done here though.

src/Types.jl
KristofferC: Do we really want to expose that to users? Isn't that just some function of the `git_source`?

src/Types.jl
00vareladavid: No because we don't set a `git_source` when tracking by path. But it is redundant because it is equivalent to `!pkg.isdeveloped && git_source === nothing`

src/Types.jl
00vareladavid: Yes, it doesn't seem like something end users should be concerned with. I'll get rid of it

00vareladavid:
Thanks!

Sure, I'm just using the `test/new.jl` to keep track of things for now. I don't have a strong opinion on how to organize the tests once this process is done.

KristofferC:
Sorry for causing a merge conflict here. I forgot this touched more code than just the tests.

00vareladavid:
No worries :slightly_smiling_face:

PR #1387, opened 2019-09-15
fix not returning to project in artifact test

KristofferC:
This causes us to have the wrong Project file when exiting the Artifact tests. This means that in latter test modules it is possible that we load the sysimg baked Pkg instead of the one we are testing.


KristofferC:
Oh yeah... I'll revert this.

PR #1398, opened 2019-09-18
store the initial manifest and project in the envcache

KristofferC:
Instead of having to reparse Project + Manifest again just before we write out the new ones. I think this makes more sense (avoids reparsing the manifest + project and is not sensitive for writing the diff before the files update). I wanted this for the undo/redo stuff but factored out this change here.


KristofferC:
bors try

PR #1399, opened 2019-09-18
Implement undo and redo functionality

KristofferC:
Should be rebased after #1398 goes through.

Screenshot 2019-09-18 at 17 46 22

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -788,6 +788,7 @@ function activate()
     Base.ACTIVE_PROJECT[] = nothing
     p = Base.active_project()
     p === nothing || printpkgstyle(Context(), :Activating, "environment at $(pathrepr(p))")
+    add_snapshot_to_undo()
     return nothing
 end
 function _activate_dep(dep_name::AbstractString)
@@ -844,6 +845,7 @@ function activate(path::AbstractString; shared::Bool=false)
         n = ispath(p) ? "" : "new "
         printpkgstyle(Context(), :Activating, "$(n)environment at $(pathrepr(p))")
     end
+    add_snapshot_to_undo()
     return nothing
 end
 function activate(f::Function, new_project::AbstractString)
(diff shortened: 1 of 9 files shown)


fredrikekre:
`redo` seems not necessary?

KristofferC:
Seems maybe useful. Like you go back and then go "oh yeah, that makes sense" and then want to go forward.

KristofferC:
bors try

src/API.jl
```diff
@@ -780,6 +780,7 @@ function activate()
     Base.ACTIVE_PROJECT[] = nothing
     p = Base.active_project()
     p === nothing || printpkgstyle(Context(), :Activating, "environment at $(pathrepr(p))")
+    add_snapshot_to_undo()
```
fredrikekre: This could perhaps be done in `print_env` instead.

src/API.jl
```diff
@@ -848,6 +850,55 @@ function activate(f::Function, new_project::AbstractString)
...
 end
 
+########
+# Undo #
+########
+
+struct UndoSnapshot
+    date::DateTime
+    project::Types.Project
+    manifest::Types.Manifest
```
StefanKarpinski: Why not just store blobs with the entire content of the files here? That way if someone had some formatting or comments and are unhappy that Pkg has messed them up / delete them, they can get them back at least.

src/API.jl
KristofferC: It is done there as well for the modified state but you want a "base state" as well. I know what you think and I have tried it but it became messy.

src/API.jl
KristofferC: I don't think people have formatting or comments in their Project file because that is removed the first thing that happens on every Pkg operations.

We don't deal with the blob anywhere else in Pkg so it quite a few special cases to handle it.
Right now I can just set the `.project` and `.manifest` fields to the ones at the state after the undo/redo and everything falls out (diff printing, writing the project etc).

StefanKarpinski: Ok, fair enough.

PR #1400, opened 2019-09-18
Enable Travis and Appveyor

staticfloat:


.travis.yml
```diff
@@ -15,20 +15,11 @@ notifications:
 before_script:
   - export PATH=$HOME/.local/bin:$PATH
 
-branches:
-  only:
-    - master
-    - staging
-    - trying
-    - /^v[0-9]+\.[0-9]+\.[0-9]+$/ # version tags
```
fredrikekre: keep these?

appveyor.yml
```diff
@@ -13,12 +13,6 @@ platform:
 #   allow_failures:
 #   - julia_version: latest
 
-branches:
-  only:
-    - staging
-    - trying
-    - /^v[0-9]+\.[0-9]+\.[0-9]+$/ # version tags
```
fredrikekre: keep?

.travis.yml
staticfloat: If we keep these, that means there will be no CI run on pull requests.

appveyor.yml
staticfloat: If we keep this, that means there will be no CI run on pull requests.

.travis.yml
fredrikekre: No.

appveyor.yml
KristofferC: I think those are only for branch pushes,not PRs.

.travis.yml
KristofferC: I think these are only for branch pushes, not PRs.

appveyor.yml
staticfloat: So according to https://github.com/JuliaLang/Pkg.jl/pull/1401 it looks like it's half-true; we don't get the build of the tip of the PR branch (Which I think is valuable, it helps to debug bad interactions and general flakiness) so perhaps what we should do is to add `/^refs/pull/.*/` or something, so that we get both, but we don't get pushes of branches that don't have a PR open, to keep the number of unnecessary CI runs down?

staticfloat:
Merging this as well, to get CI working again.

Issue #1304, opened 2019-08-16
stop using bors


StefanKarpinski:
I think that using bors on this repo was an interesting experiment, but I feel that it is a failed one. I find interacting with bors incredibly annoying and confusing and I think the benefits are negligible and far outweighed by the downsides. For example, @staticfloat and I both ended up giving up on trying to convince bors to run CI on various PRs and just yolo merged instead. The only things that bors is actually good for are:

1. Merging a PR whenever it passes tests. There are various GitHub integrations that can do this without being as annoying to use as bors.

2. Making sure that the build doesn't break when someone merges a PR and someone else wants to merge another PR that passes tests independently, doesn't have any merge conflicts, yet somehow manages to break the build. This has happened maybe two, three times in the entire history of Julia development on https://github.com/JuliaLang/julia. I doubt that Pkg is any more prone to this problem than Julia itself is. Maybe Rust code has this problem more and actually needs bors, but Julia code doesn't seem to—if two changes both pass tests and don't conflict, the chances of their merge breaking tests is vanishing.

So I propose that we stop using bors and use normal CI like Julia itself does.

KristofferC:
I find 1. really convenient so would be good if we could find a replacement for that.

I think Bors itself has worked pretty well (with the exception of no squashing functionality), it is just that:

1. Our Windows CI runs serially for many Julia versions and the tests are slow so CI takes hours to run.
2. We hit the network a lot and any connection error means the CI will fail.

So it feels a little bit that perhaps Bors is taking a bit of the blame here for the fact that our CI will be annoying no matter what system we use.

For 1, https://github.com/JuliaLang/Pkg.jl/pull/1297 should help. For 2, not sure what to do.

StefanKarpinski:
Even just getting bors to rerun CI is a problem though, whereas without it, I can at least just restart individual CI runs. Bors also majorly increases the chances of accidentally merging without squashing.

KristofferC:
Yeah, the flakiness combined with that it is hard to restart individual CI runs (+ that CI takes a long time) is a good argument that causes the bors workflow to be kind of annoying.

KristofferC:
Maybe something like https://github.com/apps/mergewhenready for 1.?

StefanKarpinski:
Yeah, I've been looking at a few options. That's one, here are some others:

* https://github.com/bobvanderlinden/probot-auto-merge
* https://mergify.io/

[...]

PR #1402, opened 2019-09-18
Make tests less flaky by wrapping them in a bounded retry loop

DilumAluthge:
Per @notriddle’s suggestion [here](https://github.com/JuliaLang/Pkg.jl/issues/1304#issuecomment-522758509), this pull request wraps the tests on Travis in a bounded retry loop.

Whether or not we keep using Bors, I think that this will make life a little easier by reducing the number of times that someone needs to manually restart a failed CI build.

I have no idea how to write a shell script in Windows, but we should probably do this for the AppVeyor tests as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -33,7 +33,14 @@ stages:
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
   - julia --project -e 'using UUIDs; write("Project.toml", replace(read("Project.toml", String), r"uuid = .*?\n" =>"uuid = \"$(uuid4())\"\n"));'
+  - try_count=0
+  - while [ $try_count != 100 ]; do
   - julia --project --check-bounds=yes -e 'import Pkg; Pkg.build(); Pkg.test(; coverage=true)'
+  - if [ $? = 0 ]; then exit 0; fi
+  - try_count=`expr $try_count + 1`
+  - sleep 1
+  - done
+  - exit 1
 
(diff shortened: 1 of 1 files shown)


staticfloat:
I'm 👎 on this, because it means that if we write a failed test, it will attempt to run the test suite 100 times, which will definitely just always eat up the whole time budget for each run.  IMO the solution to flaky tests is to fix the tests, not to run them multiple times.

staticfloat:
Heh, I suppose 2x isn't so bad.  ;)

DilumAluthge:
Ahh fair enough. I’ve changed the `100` to `2`.

I definitely agree that the best approach is to fix flaky tests. But I think the network failures (“could not resolve URL github.com”, etc) are going to be tricky to fix.

KristofferC:
If we should do this, we imo need to do it on a testset level. Not just rerun things from scratch.

00vareladavid:
> But I think the network failures (“could not resolve URL github.com”, etc) are going to be tricky to fix.

#1384 will allow us to hit the network less

DilumAluthge:
Sounds like we have a good plan!

The plan is to see https://github.com/JuliaLang/Pkg.jl/pull/1384 through. If we still end up having nondeterministic test failures after https://github.com/JuliaLang/Pkg.jl/pull/1384 is merged, we can revisit the idea of retrying tests, but at the `@testset` level.

Issue #1376, opened 2019-09-11
7z is running verbose mode on Windows


fredrikekre:

00vareladavid:
Is this all the feedback that begins with `tar:`?

PR #1404, opened 2019-09-19
Silence 7z output

musm:
Silence 7z output


musm:
Fix https://github.com/JuliaLang/Pkg.jl/issues/1376

note 7zip is annoying in that 7zip does not have silent mode.

musm:
cc @staticfloat, same change as in BinaryProvider

fredrikekre:
Looks like this is included in #1405

musm:
Ok, I did open this PR first.

PR #1405, opened 2019-09-20
7z improvements

staticfloat:
This silences verbose `7z` output, and prefers `7z` on all platforms.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -488,7 +488,9 @@ function install_archive(
         url_success || continue
         dirs = readdir(dir)
         # 7z on Win might create this spurious file
+        @show dirs
         filter!(x -> x != "pax_global_header", dirs)
+        @show dirs
         @assert length(dirs) == 1
         # Assert that the tarball unpacked to the tree sha we wanted
         # TODO: Enable on Windows when tree_hash handles
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -352,7 +352,7 @@ function probe_platform_engines!(;verbose::Bool = false)
     rm(tmpfile, force = true)
 
     # For windows, we need to tweak a few things, as the tools available differ
-    @static if Sys.iswindows()
+    if Sys.iswindows()
         # For download engines, we will most likely want to use powershell.
         # Let's generate a functor to return the necessary powershell magics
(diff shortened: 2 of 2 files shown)


staticfloat:
It would be ideal to have Julia 1.3-rc3 available on CI to test this; `7z` bundling hasn't landed in 1.3-rc2 yet, but the current CI run will still be valuable to figure out if fallbacks are still working properly, etc...

KristofferC:
Nightly has the 7z bundling though (and we apparently fail test on nightly now).

musm:
@staticfloat  did you see https://github.com/JuliaLang/Pkg.jl/pull/1404 ? 

Don't think it's a good idea to pipe stderr to devnull btw

staticfloat:
Yeah, I didn't see yours before I did this.

> Don't think it's a good idea to pipe stderr to devnull btw

Isn't that what [you just did to `BinaryProvider.jl`](https://github.com/JuliaPackaging/BinaryProvider.jl/commit/d57da3f92ad843ea8529a666ca4ea2151e9e0c64) though?

musm:
Yes it's exactly the same

musm:
It seems that this repo just copied the file.

staticfloat:
No, what I mean is that you piped `stderr` to `devnull` in BinaryProvider, so why would we do that there but not here?

musm:
Hmm it appears that I did that there as well. It probably doesn't matter since julia would throw an error anyways.

KristofferC:
`dirs = ["JuliaIO-JSON.jl-723a515", "JuliaIO-JSON.jl-b34d7ce"]`

Wat...

staticfloat:
Fredrik and I have been debugging this; turns out `@testset` does `Random.seed!()` so we re-use temporary directories a _lot_.  That, combined with paths through that function that don't cleanup properly means that we end up in bad state a lot.

KristofferC:
Oh... I knew using `randstring` would come back to bite me...

StefanKarpinski:
I knew that setting the seed automatically in testsets would come back to haunt us 😐

KristofferC:
f

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/1412

PR #1408, opened 2019-09-20
`rm` should always preserve `julia` compat

### LINKED ISSUE #1407: failure to preserve compat entries for Julia itself
When removing a package from a project, any packages that are no longer in `[deps]` or `[extras]` get deleted from `[compat]`, which is all well and good, but this seems to be applied to any `julia` entry that might be in there as well, which we should keep since there is always an implicit dependency on `julia`. To reproduce:

1. Manually put a `julia = "1.3"` entry in the `[compat]` section.
2. Add any package, say `] add JSON`.
3. Remove that package: `] rm JSON`.
4. Now the `julia = "1.3"` entry has been deleted.

00vareladavid:
---
fix #1407
caused by #1324

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -917,8 +917,10 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
         println(ctx.io, "No changes")
         return
     end
+    # only declare `compat` for direct dependencies
+    # `julia` is always an implicit direct dependency
     filter!(ctx.env.project.compat) do (name, _)
-        name in keys(ctx.env.project.deps)
+        name == "julia" || name in keys(ctx.env.project.deps)
     end
     deps_names = append!(collect(keys(ctx.env.project.deps)),
                          collect(keys(ctx.env.project.extras)))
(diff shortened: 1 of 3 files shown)


00vareladavid:
This makes me wonder, is declaring `compat` on packages which are not direct dependencies useful?

DilumAluthge:
> This makes me wonder, is declaring compat on packages which are not direct dependencies useful?

Yeah. E.g. I want to specify compat for test dependencies

StefanKarpinski:
Agree, one should probably be allowed to declare compat with anything in `[extras]` as well.

KristofferC:
You are allowed to do that and it will be active when the target is run (i.e. when testing).

StefanKarpinski:
That's not what the logic in this PR looks like it's doing—it looks like it's deleting compat for anythign that's not in `[deps]` but maybe I'm misunderstanding the logic...

KristofferC:
Oh, yeah, if that happens, then it is a bug.

fredrikekre:
Registrator doesn't allow it I think.

StefanKarpinski:
If so then that should be fixed too. Why does registrator have an opinion about it...

PR #1409, opened 2019-09-20
Better cleanup of temporary files in install_archive.

fredrikekre:


src/Operations.jl
```diff
@@ -504,11 +508,11 @@ function install_archive(
...
         !isdir(version_path) && mkpath(version_path)
         mv(unpacked, version_path; force=true)
-        Base.rm(path; force = true)
-        Base.rm(dir; force = true)
-        return true
+        break # successful install
     end
-    return false
+    # Clean up and exit
+    foreach(x -> Base.rm(x; force=true), tmp_objects)
```
staticfloat: Do we not need `recursive=true` here as well?

PR #1349, opened 2019-09-03
Document the interpretation of compat ranges in registries

timholy:
This is a start on #1190. @StefanKarpinski or others, feel free to modify at will.

PR #1410, opened 2019-09-22
Allow hyphenated ranges in compatability specs

### LINKED ISSUE #843: semver syntax: support hyphen ranges

DilumAluthge:
Fixes #843 
Closes #1232 

This pull request implements @KristofferC's suggestion in [this PR comment](https://github.com/JuliaLang/Pkg.jl/pull/1232#issuecomment-532582871).

cc: @KristofferC @StefanKarpinski @00vareladavid @fredrikekre @goretkin @tkf


docs/src/compatibility.md
```diff
@@ -104,3 +104,12 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
...
 PkgD = "< 1.2.3"  # [0.0.0, 1.2.2]
 ```
+
+### Hyphen specifiers
+
+Hyphen syntax can also be used to specify version ranges:
+
+```toml
+[compat]
+PkgA = "0.1 - 0.2.3"   # [0.1.0,  0.2.3]
```
KristofferC: This needs more docs about what happens when the upper bound does not specify a patch or minor version, e.g. `0.1 - 0.2`. Is that equivalent to `[0.1.0 - 0.2.0]`?

DilumAluthge: Good catch. Yes, I should add more docs. And frankly, should also add tests to cover those specific cases.

From Stefan's comment [here](https://github.com/JuliaLang/Pkg.jl/pull/1232#discussion_r294876288), we have:
 
>     * `0.1-0.2.3` means `[0.1, 0.2.3]`
> 
>     * `0.1-0.2` means `[0.1, 0.3)`
> 
>     * `0.1-0` means `[0.1, 1.0)`

I'll add some docs and tests.

DilumAluthge:
> This feels like a "feature" that we might want to think about backporting to 1.x

I wouldn't oppose that, but I am curious why you think we should backport this.

KristofferC:
Because otherwise packages that want to support this have to drop the LTS version of julia. Maybe it isn't a big deal.

test/pkg.jl
```diff
@@ -88,6 +88,57 @@ import Pkg.Types: semver_spec, VersionSpec
     @test   v"1.2.3" in semver_spec(">=1.2.3")
     @test !(v"1.2.2" in semver_spec(">=1.2.3"))
 
+    @test semver_spec("0.1.0 - 0.2.2") == VersionSpec("0.1.0 - 0.2.2")
```
KristofferC: There are 3x3 branches in the code so the number of tests here feel a bit excessive.

DilumAluthge: Will keep the tests per triage

docs/src/compatibility.md
```diff
@@ -104,3 +104,20 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
 PkgC = "= 1.2.3"  # [1.2.3, 1.2.3]
 PkgD = "< 1.2.3"  # [0.0.0, 1.2.2]
 ```
+
+### Hyphen specifiers
+
```
fredrikekre: Did we decide to backport this? Maybe we can add something like this for now and update later when we backport?

```suggestion

!!! compat "Julia 1.4"
    Hyphen specifiers requires at least Julia 1.4, so it is recomended to also add
    ```toml
    [compat]
    julia = "1.4"
    ```
    to the project file when using them.
```

docs/src/compatibility.md
```diff
@@ -104,3 +104,20 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
...
+Hyphen syntax can also be used to specify version ranges:
+
+```toml
+[compat]
+PkgA = "0.1 - 0.2.3"   # [0.1, 0.2.3]
```
fredrikekre: Given https://discourse.julialang.org/t/how-can-we-encourage-julia-package-developers-to-release-version-1-0-0/29124, maybe one way is to use post 1.0 numbers in examples like this? :smile: At least in addition, to the current example.

docs/src/compatibility.md
```diff
@@ -104,3 +104,20 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
...
+```
+
+Any unspecified trailing numbers in the second end-point will be considered to be wildcards:
+
+```toml
+[compat]
+PkgA = "0.1 - 0.2"     # [0.1, 0.3)
```
fredrikekre: I assume wildcards hold for the first number in the range too? Maybe clarify like:
```suggestion
PkgA = "0.1 - 0.2"     # [0.1.0, 0.3.0)
```

docs/src/compatibility.md
```diff
@@ -104,3 +104,20 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
...
+PkgA = "0.1 - 0"       # [0.1, 1.0)
```
fredrikekre:
```suggestion
PkgA = "0.1 - 0"       # [0.1.0, 1.0.0)
```

docs/src/compatibility.md
```diff
@@ -104,3 +104,20 @@ PkgB = "≥ 1.2.3"  # [1.2.3,  ∞)
...
+```
```
fredrikekre: Maybe add a post 1.0 example here too.

docs/src/compatibility.md
KristofferC: > I assume wildcards hold for the first number in the range too? 

I think they just pad with zeros for the first number

fredrikekre: Yea, thats what I meant, effectively a wildcard though

KristofferC: Yeah, sure

timholy:
I like this. To fully obsolete #1349, this would also need to document `0.7-*`.

StefanKarpinski:
A concern that should be considered is if we want to support pre-release versions then `1.2.3-4.5.6` is both a valid version range and a pre-release of `1.2.3` with pre-release tuple `(4, 5, 6)`. See https://github.com/npm/node-semver#prerelease-tags.

StefanKarpinski:
We could just disambiguate that as a version range, which seems like the more common case, although then there isn’t any way to write the pre-release version, whereas if we interpreted this as the pre-release version the version range can still be written as `1.2.3 - 4.5.6` since that is clearly not a single version number.

KristofferC:
Maybe enforce having a space on the sides of the hypen for now?

docs/src/compatibility.md
tkf: Is package's Project.toml loaded in normal installation operations (i.e., `Pkg.add("Package"); using Package`)?  If not, isn't requiring `julia = "1.4"` overly restricting?  Though I suppose it can be problematic if you want to use `Pkg.test` to run tests (but there are other ways to run tests).

KristofferC: > Is package's Project.toml loaded in normal installation operations (i.e., Pkg.add("Package"); using Package)? 

No, it isn't.

>  If not, isn't requiring julia = "1.4" overly restricting? Though I suppose it can be problematic if you want to use Pkg.test to run tests (but there are other ways to run tests).

Yeah, it's a bit odd situation that you can add the package but you cannot test it or dev it etc.

StefanKarpinski:
Would it be nuts to use a different character than `-` which already means something in the actual standardized semver syntax? We could write `1.2.3:4.5.6` for version ranges. The `:` character has no meaning in single-version syntax. That's a bit of an unfortunate mismatch with the registry range syntax, but 🤷‍♂

DilumAluthge:
> Would it be nuts to use a different character than `-` which already means something in the actual standardized semver syntax? We could write `1.2.3:4.5.6` for version ranges. The `:` character has no meaning in single-version syntax. That's a bit of an unfortunate mismatch with the registry range syntax, but 🤷‍♂

`1.2.3:4.5.6` seems kind of hard to read. So many dots! How about using the word `to`? Then you’d have `1.2.3 to 4.5.6`.

StefanKarpinski:
At that point I'd rather have `1.2.3 - 4.5.6` with required spaces.

DilumAluthge:
`→`?

tkf:
Another option may be `--`. In LaTeX it's compiled to an en-dash which is used to represent a range of values: https://en.wikipedia.org/wiki/Dash#Ranges_of_values

oschulz:
> I wouldn't oppose that, but I am curious why you think we should backport this.

I think with the current push to encourage dependency version limits (upper bounds now required for automatic merging on General), this feature would be very very helpful, and I'd be very glad if it were backported to v1.0!

I just ran into a situation where this package resolution failed, because I upgraded "ElasticArrays" to v1.0 and updated another package to require it - but another dependency that I wanted required "ElasticArrays" v0.2. I can work around that by just setting "ElasticArray <= 2.0" on my other package - but what if I need a lower bound as well?

I have a distinct feeling that without the ability to specify deps version ranges, will run into (extremely frustrating) situations in which package resolution fails more and more often, as package authors put tighter bounds on dependency versions.

StefanKarpinski:
> this feature would be very very helpful, and I'd be very glad if it were backported to v1.0!

As a general rule, features aren't backported. Pkg is a bit special though, and I've been toying around with the idea of making releases of 1.0 with an updated version of Pkg.

oschulz:
@fredrikekre reminded me that it's already possible to specify a comma-separated list of versions -somehow I missed that or forgot about it, sorry! So since there's a decent workaround, I'll have take back my earlier statement about this needing a backport to 1.0.

DilumAluthge:
I completely forgot about this pull request. What was the latest decision? Enforce a space on either side of the hyphen? I.e. `1.2.3 - 4.5.6` is allowed, but `1.2.3-4.5.6` is not allowed?

docs/src/compatibility.md
DilumAluthge: Done

DilumAluthge:
Alright, I have updated the PR with all of the review comments.

I am enforcing the rule that there must be whitespace on both sides of the hyphen. That way, there is no confusion/ambiguity with prerelease notation.

I think this is ready to merge. cc: @StefanKarpinski @KristofferC @fredrikekre

DilumAluthge:
Bump @StefanKarpinski @KristofferC @fredrikekre. 

Where did you land on this PR? Do you want to merge it? Or do we want to close it and push users towards using caret and tilde notation instead?

DilumAluthge:
Bump. Any update on merging this?

00vareladavid:
FWIW node seems to use the same "space padded hyphen" notation for ranges: https://docs.npmjs.com/misc/semver

Also, the tests seem pretty comprehensive.

PR #1418, opened 2019-09-24
Update README links.

fredrikekre:
Point README links for AV to JuliaLang, remove links for BORS.


fredrikekre:
AV now uses the JuliaLang account.

PR #1419, opened 2019-09-24
WIP: refactor some of the repo handling code

KristofferC:
I was working on adding subdir support (multiple packages in the same repo) but even with https://github.com/JuliaLang/Pkg.jl/pull/1372 I found the repo handling code a bit tricky to follow.

This is an attempt to go through the code, add some comments, and arguably clean up some things when going through the code again.
It ended up with about -100 lines which I think is a good indication that it is arguably a bit simpler now.

Some things I did in no particular order:

- Push the `GitRepo` uses "up the stack" so we don't open and close it in all small helper functions.
- Move uses of `CredentialsCache` inside `GitTools.fetch` and `GitTools.clone` instead of creating them all over the place
- Rename some things to be a bit more precise / consistent with other uses, for example:
    - `resolve_projectfile!` instead of `parse_package!`, we are setting `uuid` and `name`, that is `resolve`.
    - `dev_repo_cache_path` / `add_repo_cache_path` instead of `clone_path`
- Sometimes inlined small helper functions that were used in one place when their name was confusing (for the uninitiated) and it was hard to context swap to the other function instead of reading it inline.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -12,7 +12,7 @@ using REPL.TerminalMenus
 
 using ..TOML
 import ..Pkg, ..UPDATED_REGISTRY_THIS_SESSION
-import Pkg: GitTools, depots, depots1, logdir, set_readonly
+import Pkg: GitTools, depots, depots1, logdir, set_readonly, safe_realpath
 import ..BinaryPlatforms: Platform
 
 import Base: SHA1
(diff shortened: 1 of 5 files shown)


src/Types.jl
```diff
@@ -451,293 +452,184 @@ function relative_project_path(ctx::Context, path::String)
...
     if !isurl(pkg.repo.source)
-        given_abspath = isabspath(pkg.repo.source)
-        pkg.repo.source = try
-            realpath(pkg.repo.source)
-        catch
+        if isdir(pkg.repo.source)
+            if !isdir(joinpath(pkg.repo.source, ".git"))
+                pkgerror("Did not find a git repository at `$(pkg.repo.source)`")
+            end
+            pkg.repo.source = safe_realpath(pkg.repo.source)
```
00vareladavid: It seems this does not preserve the relative v absolute path to a local repo. (e.g. as opposed to https://github.com/JuliaLang/Pkg.jl/blob/b29b799c667e6bcf5bb954eec9f05b0a273ddfbe/src/Types.jl#L715-L716)

KristofferC: Yep, fixed.

PR #1420, opened 2019-09-24
Clarify backward-compatible feature additions require only minor bump in pre-1.0 packages

### LINKED ISSUE #1417: Answer to be documented: Is it OK to register a package as 0.x.(y+1) when it is backward-compatible to 0.x.y but has feature additions?
When writing #1416, I realized that Pre-1.0 behavior section doesn't indicate that it is invalid to register a pre-1.0 post-0.1 package without a minor version bump when features are added if it is backward-compatible.  SemVer only says that:

> How should I deal with revisions in the 0.y.z initial development phase?
>
> The simplest thing to do is start your initial development release at 0.1.0 and then increment the minor version for each subsequent release.
>
> --- https://semver.org/#how-should-i-deal-with-revisions-in-the-0yz-initial-development-phase

But, I've been thinking that I should bump the minor version whenever I introduce new features in pre-1.0 packages.  Am I missinterpreting or misremembering something?

If the SemVer rule "feature additions requires minor version bump" only applies to post-1.0 packages in Julia, how about mentioning it explicitly in at the end of Pre-1.0 behavior section?

> In particular, it is valid to register a package with version 0.2.4 when it has feature additions compared to 0.2.3 as long as it remains backward compatible with 0.2.0.

I think it has some benefits:

* It makes it more reasonable to write `= "0.x"` in `[compat]` as minor version updates become less frequent.

    * As a corollary, it helps advocating or even banning https://github.com/JuliaRegistries/General/issues/3673 registering packages with no upper bounds in `[compat]` (although this point may be solved soon by the range specification syntax #1410).

* It helps packages to _not_ going 1.0 before solidifying the API.  If pre-1.0 packages cannot express backward-compatibility, the only option to do so is to use non-zero major versions.  However, a package with ever-changing major versions splits Julia ecosystem in incompatible sub-parts.

* It helps package authors practicing backward-compatibility issues in the wild before going 1.0.

tkf:
... as discussed in #1417.

close #1417

### DIFF
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -56,8 +56,8 @@ Example = "0.2.1"
 
 results in a versionbound on Example as `[0.2.1, 0.3.0)`.
 
-In particular, the depended package (`Example` in above example) may set `version = "0.2.4"` when it has feature additions
-compared to 0.2.3 as long as it remains backward compatible with 0.2.0.  See also [The `version` field](@ref).
+In particular, a package may set `version = "0.2.4"` when it has feature additions compared to 0.2.3 as long as it
+remains backward compatible with 0.2.0.  See also [The `version` field](@ref).
 
 ### Caret specifiers
 
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -56,8 +56,8 @@ Example = "0.2.1"
 
 results in a versionbound on Example as `[0.2.1, 0.3.0)`.
 
-In particular, it is valid to register a package with version 0.2.4 when it has feature additions compared to 0.2.3 as
-long as it remains backward compatible with 0.2.0.
+In particular, the depended package (`Example` in above example) may set `version = "0.2.4"` when it has feature additions
+compared to 0.2.3 as long as it remains backward compatible with 0.2.0.  See also [The `version` field](@ref).
 
 ### Caret specifiers
 
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -56,6 +56,9 @@ Example = "0.2.1"
 
 results in a versionbound on Example as `[0.2.1, 0.3.0)`.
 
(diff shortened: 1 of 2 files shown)


docs/src/compatibility.md
```diff
@@ -56,6 +56,9 @@ Example = "0.2.1"
 
 results in a versionbound on Example as `[0.2.1, 0.3.0)`.
 
+In particular, it is valid to register a package with version 0.2.4 when it has feature additions compared to 0.2.3 as
```
fredrikekre: We have not really used the term `register` in this text, since this is talking about what Pkg considers compatible. Maybe rewrite it without that?

tkf: Thanks for the feedback. I reworded it in the new commit.

docs/src/compatibility.md
```diff
@@ -56,6 +56,9 @@ Example = "0.2.1"
 
 results in a versionbound on Example as `[0.2.1, 0.3.0)`.
 
+In particular, the depended package (`Example` in above example) may set `version = "0.2.4"` when it has feature additions
```
KristofferC: Is "depended package" a thing one can say? Maybe , "the package depended on"?

tkf: OK, this was something that was not super clear to me. Maybe it's better to say ``the upstream package (`Example` in above example)``?

KristofferC: or just say `a package`?

tkf: Sounds good to me. I changed it. (I tried to stick with the view from the downstream package since it's about `[compat]`. But I guess that was unnecessary consideration.)

KristofferC:
I just did a small rewording, hopefully that's ok. Thanks for the PR.

tkf:
Thanks for reviewing and fixing up the wording!

PR #1422, opened 2019-09-25
add subdir functionality

KristofferC:
```
julia> Pkg.add(PackageSpec(path="MainRepo", subdir="SubDir.jl"))
  Updating git-repo `/home/kc/Documents/JuliaTests/MainRepo`
 Resolving package versions...
  Updating `~/Pkg.jl/Project.toml`
  [900e8cb0] + SubDir v0.1.0 #master (/home/kc/Documents/JuliaTests/MainRepo[/SubDir.jl])
  Updating `~/Pkg.jl/Manifest.toml`
  [900e8cb0] + SubDir v0.1.0 #master (/home/kc/Documents/JuliaTests/MainRepo[/SubDir.jl])
```

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1558,9 +1558,8 @@ function stat_rep(x::PackageSpec; name=true)
     if x.repo.rev !== nothing
         rev = occursin(r"\b([a-f0-9]{40})\b", x.repo.rev) ? x.repo.rev[1:7] : x.repo.rev
     end
-    repo = Operations.is_tracking_repo(x) ?
-    string("`$(x.repo.source)", (x.repo.subdir == nothing ? "" : "/[$(x.repo.subdir)]"), "#$(x.repo.rev)`") :
-        ""
+    subdir_str = x.repo.subdir == nothing ? "" : "/[$(x.repo.subdir)]"
+    repo = Operations.is_tracking_repo(x) ? "`$(x.repo.source)$(subdir_str)#$(rev)`" : ""
     path = Operations.is_tracking_path(x) ? "$(pathrepr(x.path))" : ""
     pinned = x.pinned ? "⚲" : ""
     return join(filter(!isempty, [name,version,repo,path,pinned]), " ")
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1558,7 +1558,9 @@ function stat_rep(x::PackageSpec; name=true)
     if x.repo.rev !== nothing
         rev = occursin(r"\b([a-f0-9]{40})\b", x.repo.rev) ? x.repo.rev[1:7] : x.repo.rev
     end
-    repo = Operations.is_tracking_repo(x) ? "`$(x.repo.source)#$(rev)`" : ""
+    repo = Operations.is_tracking_repo(x) ?
+    string("`$(x.repo.source)", (x.repo.subdir == nothing ? "" : "/[$(x.repo.subdir)]"), "#$(x.repo.rev)`") :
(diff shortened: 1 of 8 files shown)


fingolfin:
This is something we are rather interested in and would love to have.

However, it seems this PR has stalled for now. I was wondering if there is any interest in / need for help from outsiders? I may have some time in a few weeks to dig into this, but obviously I don't want to step on any toes or waste time in case somebody is actually working on this elsewhere, or perhaps there are reasons for the stall (e.g. perhaps problems where discovered with the approach, or further design discussions are needed, or ...)

Anyway, just wanted to say I'd be interested in chipping in, but of course I understand you guys may have other priorities.

KristofferC:
IIRC, this PR was pretty much done, with the exception of the Pkg REPL implementation of it, and bashing my head against https://github.com/JuliaLang/Pkg.jl/blob/54eae6c1b591070918a65a581b7ab3a0c6d9d90c/src/REPLMode/argument_parsers.jl#L21 made me kind of lose my interest in the whole thing.

I do think we could just punt on supporting this from the REPL though.

Help to get this rebased and a bit more tested would be appreciated because I think it is a good feature to have.

KristofferC:
I rebased this, should be good to go. It doesn't include any REPL syntax but that can be added later.

src/Operations.jl
```diff
@@ -1558,7 +1558,9 @@ function stat_rep(x::PackageSpec; name=true)
     if x.repo.rev !== nothing
         rev = occursin(r"\b([a-f0-9]{40})\b", x.repo.rev) ? x.repo.rev[1:7] : x.repo.rev
     end
-    repo = Operations.is_tracking_repo(x) ? "`$(x.repo.source)#$(rev)`" : ""
+    repo = Operations.is_tracking_repo(x) ?
+    string("`$(x.repo.source)", (x.repo.subdir == nothing ? "" : "/[$(x.repo.subdir)]"), "#$(x.repo.rev)`") :
+        ""
```
fingolfin, line 1563: So the old code use `rev`, now you use `x.repo.rev`, leaving `rev` unused. So either `rev` should go, or perhaps this be done? (I don't claim to fully understand that code, so I am guessing):
```suggestion
    repo = Operations.is_tracking_repo(x) ?
    string("`$(x.repo.source)", (x.repo.subdir == nothing ? "" : "/[$(x.repo.subdir)]"), "#$(rev)`") :
        ""
```
And if I might go further, to my personal subjective taste, code would be more readable if you introduce an auxiliary variable:
```suggestion
    subdir = x.repo.subdir == nothing ? "" : "/[$(x.repo.subdir)]"
    repo = Operations.is_tracking_repo(x) ? string("`$(x.repo.source)$(subdir)#$(rev)`") : ""
```

src/manifest.jl
```diff
@@ -139,6 +138,7 @@ function Manifest(raw::Dict)::Manifest
             entry.path        = read_field("path",          nothing, info, safe_path)
             entry.repo.source = read_field("repo-url",      nothing, info, identity)
             entry.repo.rev    = read_field("repo-rev",      nothing, info, identity)
+            entry.repo.subdir = read_field("repo-subdir",        nothing, info, identity)
```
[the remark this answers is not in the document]
KristofferC: Fixed, thanks.

src/Operations.jl
KristofferC: Thanks, updated with something like this. And not using `rev` was an error on my part, indeed.

PR #1427, opened 2019-09-27
Refactor and test `Pkg.test`

### LINKED ISSUE #1287: Test deps should not be direct deps
E.g. here:
```
[~/sandbox/Foo]$ cat Project.toml 
name = "Foo"
uuid = "167bd52b-5e3d-4bc7-bd89-d284fca55e53"
authors = ["Fredrik Ekre <ekrefredrik@gmail.com>"]
version = "0.1.0"

[extras]
ForwardDiff = "f6369f11-7733-5829-9624-2563aa707210"

[targets]
test = ["ForwardDiff"]
[~/sandbox/Foo]$ cat src/Foo.jl 
module Foo

using ForwardDiff

end # module
[~/sandbox/Foo]$ cat test/runtests.jl 
using Foo
@assert true
```
we end up with this manifest entry in the temporary test env:
```
[[Foo]]
deps = ["ForwardDiff"]
path = "/Users/fredrik/sandbox/Foo"
uuid = "167bd52b-5e3d-4bc7-bd89-d284fca55e53"
version = "0.1.0"
```

### LINKED ISSUE #1423: Testing another package from a package environment is broken
```sh
$ cd $(mktemp -d) &&
  export JULIA_DEPOT_PATH=$(mktemp -d) &&
  git clone https://github.com/JuliaIO/JSON.jl.git JSON &&
  git clone https://github.com/fredrikekre/Literate.jl.git Literate &&
  pkg --project=JSON dev Literate &&
  pkg --project=JSON test Literate
```
fails when `Literate` tries to load `JSON`. The problem is that when setting up the test env for Literate we resolve a release version of `JSON` when we should just use the one from the given project. Probably this happens because when we look in the current manifest, `JSON` is not there, since we are running in the `JSON` project. We should look there too. The reason for the failure later is that the version of `JSON` that we resolved is never installed/downloaded, but we shouldn't have to do this since we should just use the version of `JSON` from the current project.

00vareladavid:
I want "targets" based testing to reuse the same machinery as "test/Project.toml" based testing. This will allow us to get rid of `backwards_compatible_isolation.jl`.

fix #1287 fix #1423


KristofferC:
Regarding #311, it doesn't make much sense to me. You should probably set the environment variable yourself in that case.

src/Operations.jl
```diff
@@ -243,22 +243,46 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
...
-istracking(pkg) = pkg.path !== nothing || pkg.repo.source !== nothing
-isfixed(pkg) = istracking(pkg) || pkg.pinned
+is_tracking_path(pkg) = pkg.path !== nothing
+is_tracking_repo(pkg) = pkg.repo.source !== nothing
+is_tracking_registry(pkg) = !is_tracking_path(pkg) && !is_tracking_repo(pkg)
+isfixed(pkg) = !is_tracking_registry(pkg) || pkg.pinned
+
+function collect_developed!(ctx::Context, pkg::PackageSpec, developed::Vector{PackageSpec})
+    source = project_rel_path(ctx, source_path(pkg))
+    Pkg.activate(source) do
```
KristofferC: Instead of activating here, can't we give the path to `Context`? Something like `Context(env = EnvCache(source))`.

src/Operations.jl
```diff
@@ -243,22 +243,46 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
...
+        pkgs = load_all_deps(Context())
+        for pkg in filter(is_tracking_path, pkgs)
+            # normalize path
+            pkg.path = project_rel_path(Context(), source_path(pkg))
```
KristofferC: Probably shouldn't remade the `Context `here every time?

src/Operations.jl
```diff
@@ -243,22 +243,46 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String, fix_deps
...
+            push!(developed, pkg)
+            collect_developed!(ctx, pkg, developed)
```
KristofferC: Is this trying to help with https://github.com/JuliaLang/Pkg.jl/issues/1005?

src/Operations.jl
```diff
@@ -767,6 +793,16 @@ function dependency_order_uuids(ctx::Context, uuids::Vector{UUID})::Dict{UUID,In
     return order
 end
 
+function gen_build_project(ctx::Context, pkg::PackageSpec, source_path::String)
+    build_project = Types.Project()
+    Pkg.activate(source_path) do
```
KristofferC: Same here with path to `Context`

KristofferC:
Would be great to get rid of that compat file!

src/API.jl
```diff
@@ -22,7 +22,13 @@ function dependencies(ctx::Context)::Dict{UUID, PackageInfo}
     pkgs = Operations.load_all_deps(ctx)
     return Dict(pkg.uuid => Operations.package_info(ctx, pkg) for pkg in pkgs)
 end
-dependencies(fn::Function, uuid::UUID) = fn(dependencies()[uuid])
+function dependencies(fn::Function, uuid::UUID)
+    dep = get(dependencies(), uuid, nothing)
+    if dep === nothing
```
KristofferC: I guess you could write this as

```
dep = get(dependencies, uuid) do
    pkgerror("depenendency with UUID `$uuid` does not exist")
end
```

But I am not sure this is preferable even.
```

00vareladavid:
I feel this is in a mergeable state now, if anyone wants to take another look.

I've tested most of the basic cases I can think of, hopefully we can pin down most of the test behavior with these tests. Perhaps @fredrikekre you can take a quick look at the tests? Since you have been good about finding holes in `Pkg.test`.

I'll be happy to test any other cases that people can think of.

00vareladavid:
Added even more tests. I'm happy with this PR now. I'll merge in a few days if no one vetos.

KristofferC:
Great job! Nice to get rid of that duplication.

00vareladavid:
I appreciate it Kristoffer. And thanks for the review!

KristofferC:
I think this caused packages to fail during tests if they don't have a project file (for example: https://github.com/maleadt/BasePkgEvalReports/blob/adfa530cc2662ca9abd85544d3ed41aa6d0ad95f/pkgeval-daily_2019_12_16/logs/Permutations/1.4.0-DEV-a6bf74a2f9.log#L36-L65).

I think we need to try fix that.

00vareladavid:
Definitely. Taking a look now

Issue #1454, opened 2019-10-17
Possible mistake in [[compat]] docs


dehann:
With the [recent zeptodoctor auto-merging, I was forced back](https://github.com/JuliaRegistries/General/pull/4440#issuecomment-542928302) to the Pkg.jl docs and found this part to be confusing:
>  ```
>  [compat]
>  Example = "1.2, 2"
>  ```
>  will result in [1.2.0, 3.0.0).

Since I get error messages locally that would suggest the actual behavior is in fact (not as advertised):
```
Example = [1.2.0-1.3.0, 2.0.0-3.0.0]
```
but not any of the versions including `1.3.1, 1.4, 1.5, ...`.

EDIT:  Perhaps its specific to the leading zero case -- i.e. `ExampleB = 0.2, 1`

It might be a bug or documentation error, but after some trail and error I have been forced to do this (and is really the only way I could get broad version apertures):
https://github.com/JuliaRobotics/IncrementalInference.jl/pull/414/files#diff-910a7b3ca0075d545dcec45cb7335ca9R47

Part of the problem might be that the following case is not documented either:
```
Example = "^0.1" == "0.1.0-0.2.0"
```

My current understanding is that there is no way to define a compatibility which stretches from 
`0.x through 1.0 or 1.x` 
or is there and I'm just not finding it?

#### Reference

In the docs just above this:
https://julialang.github.io/Pkg.jl/v1/compatibility/index.html#Caret-specifiers-1

- I'm on Julia 1.2

DilumAluthge:
Your issue here is specific to the leading zero case.

DilumAluthge:
What compat range are you trying to use?

DilumAluthge:
I suspect that https://github.com/JuliaLang/Pkg.jl/pull/1410 will meet your needs. But that PR is not ready yet.

DilumAluthge:
If you need to document `0.x` through `1.x`, you’ll need to list out all of the breaking versions, for example `0.5, 0.6, 0.7, 0.8, 1.0, 1.1, 1.2, 1.3`

DilumAluthge:
But ask yourself: do I really need to maintain compatibility with all of those different versions. Can’t I just drop compat for the `0.x` versions and use a nice simple compat entry such as `^1.0.0`?

DilumAluthge:
[...]

PR #1456, opened 2019-10-17
improvement on composite leading zero description in docs

dehann:
see #1454

### DIFF
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -29,7 +29,7 @@ The union of multiple version specifiers can be formed by comma separating indiv
 [compat]
 Example = "1.2, 2"
 ```
-will result in `[1.2.0, 3.0.0)`.
+will result in `[1.2.0, 3.0.0)`.  Note leading zeros are treated differently, e.g. `Example = "0.2, 1"` would only result in `[0.2.0-0.3.0, 1.0.0-2.0.0]`.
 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
The very next section heading is:
```
### Behavior of versions with leading zeros (0.0.x and 0.x.y)
```

docs/src/compatibility.md
```diff
@@ -29,7 +29,7 @@ The union of multiple version specifiers can be formed by comma separating indiv
 [compat]
 Example = "1.2, 2"
 ```
-will result in `[1.2.0, 3.0.0)`.
+will result in `[1.2.0, 3.0.0)`.  Note leading zeros are treated differently, e.g. `Example = "0.2, 1"` would only result in `[0.2.0-0.3.0, 1.0.0-2.0.0]`.
```
StefanKarpinski:
```suggestion
will result in `[1.2.0, 3.0.0)`.  Note leading zeros are treated differently, e.g. `Example = "0.2, 1"` would only result in `[0.2.0-0.3.0, 1.0.0-2.0.0]`. See the next section for more information on versions with leading zeros.
```

dehann:
@DilumAluthge , thanks -- the explicit version range for this syntax is useful; and thanks @StefanKarpinski the transition sentence was directly added.

PR #1477, opened 2019-10-29
Try to revert breaking changes to the Pkg API

rofinn:
Possible cleanup to #936


omus:
I don't think this will be reverted.

PR #1484, opened 2019-11-04
Insert compiler barrier within `@artifact_str`

staticfloat:
This greatly reduces the load time of JLL artifacts, as it severs all
connection between user (JLL) code and Pkg; the compiler no longer needs
to manage the large number of backedges within `Pkg.Artifacts` leading
back to each and every JLL package.


staticfloat:
I'm going to close this in favor of #1485, since it subsumes these changes.

PR #1486, opened 2019-11-04
use `===` for comparison to `nothing` and Symbols

JeffBezanson:
This is helpful to the compiler, since e.g. `DataValue(nothing) == nothing` is true.


fredrikekre:
Do we want this for 1.3.0?

JeffBezanson:
It's not crucial.

PR #1487, opened 2019-11-05
`bin/recompress.jl`: Don't try to parse version numbers for JLL packages

DilumAluthge:
Currently, if you try to run the `bin/recompress.jl` script on the General registry, you will quickly encounter an error similar to the following:
```
ERROR: LoadError: ArgumentError: invalid version range: "4.1.0+1"
```

This happens because JLL packages use version numbers like `4.1.0+1`, which the compress script does not like.

The solution is: do not try to parse version numbers for JLL packages.

This pull request implements that solution.

cc: @KristofferC since I think he's the last person to look at this code.

### DIFF
--- a/bin/recompress.jl
+++ b/bin/recompress.jl
@@ -10,10 +10,13 @@ const packages = TOML.parsefile(registry_file)["packages"]
 const version_map = Dict{String,Vector{VersionNumber}}()
 
 for (uuid, info) in packages
-    path = joinpath(registry_path, info["path"])
-    versions_file = joinpath(path, "Versions.toml")
-    versions = Compress.load(versions_file)
-    version_map[uuid] = sort!(collect(keys(versions)))
+    name = info["name"]
+    if !endswith(name, "_jll")
+        path = joinpath(registry_path, info["path"])
+        versions_file = joinpath(path, "Versions.toml")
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Hmm... that seems like a bit of a hack.

DilumAluthge:
Correct.

DilumAluthge:
I suppose that the better approach would be to figure out how to turn these things into VersionRanges. But how do we do that? What would be the correct VersionRange generated by the entry `4.1.0+1`?

DilumAluthge:
You could say that the upper bound is the

DilumAluthge:
Since you can only specify major, minor, and patch for a VersionBound, I’m not sure how we’d set the lower VersionBound in this example.

StefanKarpinski:
You'd change the definition of VersionBound to support build and patch components :)

DilumAluthge:
Yeah you are right.

Would that be a breaking change though? I guess we just be changing the internal fields of a struct. Presumably we’re allowed to do that in a non-breaking release.

DilumAluthge:
I'm closing this in favor of Stefan's idea:
> You'd change the definition of VersionBound to support build and patch components :)

Not sure if I'll have time anytime soon to make this change, but hopefully someone will.

PR #1496, opened 2019-11-10
Deprecate the `Pkg.generate` function

DilumAluthge:
Now that https://github.com/invenia/PkgTemplates.jl/pull/61 has been merged, PkgTemplates.jl no longer uses `Pkg.generate`.

Therefore, I think that we should officially deprecate `Pkg.generate` and direct users to PkgTemplates.jl. This will allow us to remove `Pkg.generate` in Julia 2.0.

This pull request adds the deprecation warning to `Pkg.generate`.

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -2,6 +2,7 @@
 
 generate(path::String; kwargs...) = generate(Context(), path; kwargs...)
 function generate(ctx::Context, path::String; kwargs...)
+    Base.depwarn("Pkg.generate is deprecated. Please use PkgTemplates.jl instead.", Core.Typeof(generate).name.mt.name)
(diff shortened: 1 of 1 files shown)


aminya:
Agree. PkgTemplate has much more features

DilumAluthge:
Bump. Any opposition to deprecating the `Pkg.generate` function?

rapus95:
@DilumAluthge We'll also lose the REPL mode integration which is VERY convenient. Is there any chance that PkgTemplates.jl can hook into the REPL mode (including pkg"generate" macro)?

DilumAluthge:
I'm closing this as stale. We can revisit in the future.

PR #1498, opened 2019-11-11
Documentation: Fix a typo in the "Tilde specifiers" section

DilumAluthge:
Currently, the "Tilde specifiers" section of the "Compatibility" documentation says the following:
```toml
PkgE = "~0.0.3" # [0.0.3, 0.0.4)
```

But this is not how tilde specifiers are currently implemented in Pkg. The correct documentation is:
```toml
PkgE = "~0.0.3" # [0.0.3, 0.1.0)
```


DilumAluthge:
To illustrate, here is what I get on the latest Julia master:
```julia
julia> import Pkg

julia> v"0.0.3" in Pkg.Types.semver_spec("~0.0.3")
true

julia> v"0.0.4" in Pkg.Types.semver_spec("~0.0.3")
true

julia> v"0.0.5" in Pkg.Types.semver_spec("~0.0.3")
true

julia> v"0.0.6" in Pkg.Types.semver_spec("~0.0.3")
true

julia> v"0.1.0" in Pkg.Types.semver_spec("~0.0.3")
false

julia> v"0.0.3" in Pkg.Types.semver_spec("^0.0.3")
true

julia> v"0.0.4" in Pkg.Types.semver_spec("^0.0.3")
false

julia> v"0.0.5" in Pkg.Types.semver_spec("^0.0.3")
false

julia> v"0.0.6" in Pkg.Types.semver_spec("^0.0.3")
false

julia> v"0.1.0" in Pkg.Types.semver_spec("^0.0.3")
false

julia> v"0.0.3" in Pkg.Types.semver_spec("0.0.3")
true

julia> v"0.0.4" in Pkg.Types.semver_spec("0.0.3")
false

julia> v"0.0.5" in Pkg.Types.semver_spec("0.0.3")
false

julia> v"0.0.6" in Pkg.Types.semver_spec("0.0.3")
false

julia> v"0.1.0" in Pkg.Types.semver_spec("0.0.3")
false
```

KristofferC:
Isn't this a bug and not a doc error?

DilumAluthge:
> A tilde specifier provides more limited upgrade possibilities. When specifying major, minor and patch versions, or when specifying major and minor versions, only the patch version is allowed to change.

DilumAluthge:
If my compat entry is `~0.0.3`, then I have specified major, minor, and patch. So per that description, only the patch should be allowed to change. Which means `0.0.4`, `0.0.5`, `0.0.6`, etc are allowed.

KristofferC:
But

> A tilde specifier provides more limited upgrade possibilities

And `^0.0.3` is [0.0.3 - 0.0.3]` so to be more limited (or at leat equally limited) that is what tilde also has to do?

DilumAluthge:
Ah, I see... yes it probably doesn’t make sense to have tilde be less limited than caret.

DilumAluthge:
I’m closing this PR in favor of fixing the bug, as discussed above.

PR #1501, opened 2019-11-13
Allow packages to easily retrieve their version

omus:
Allows for packages to codify their own deprecations for future revisions. I example of deprecating the return type of a function:
```julia
if @__VERSION__() < v"1.4"
    """
        f() -> String
    """
    function f(; classic_return::Bool=true)
        if classic_return
            Base.depwarn("`f` will return `nothing` in version 1.4")
            return ""
        else
            return nothing
        end
    end
elseif v"1.4" <= @__VERSION__ < v"1.5"
    """
        f() -> Nothing
    """
    function f(; classic_return=nothing)
        if classic_return !== nothing
            Base.depwarn("`f` will no drop support for the `class_return` keyword in version 1.5")
        end
        return nothing
    end
else
    """
        f() -> Nothing
    """
    f() = nothing
end
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -8,7 +8,7 @@ import Random
 using Dates
 import LibGit2
 
-import ..depots, ..depots1, ..logdir, ..devdir
+import ..depots, ..depots1, ..logdir, ..devdir, ..pkgdir
 import ..Operations, ..Display, ..GitTools, ..Pkg, ..UPDATED_REGISTRY_THIS_SESSION
 using ..Types, ..TOML
 using ..Types: VersionTypes
@@ -985,7 +985,7 @@ Get the `VersionNumber` of the package which expands this macro. If executed out
 package `nothing` will be returned.
 """
 macro __VERSION__()
-    pkg_dir = Base.pkgdir(__module__)
+    pkg_dir = pkgdir(__module__)
 
     if pkg_dir !== nothing
         project_data = TOML.parsefile(Types.projectfile_path(pkg_dir))
--- a/src/API.jl
+++ b/src/API.jl
@@ -985,17 +985,14 @@ Get the `VersionNumber` of the package which expands this macro. If executed out
 package `nothing` will be returned.
 """
 macro __VERSION__()
-    ctx = Types.Context()
-    pkg_id = Base.PkgId(__module__)
-    pkg_id.uuid === nothing && return nothing
+    pkg_dir = Base.pkgdir(__module__)
 
-    pkg_info = if ctx.env.project.name == pkg_id.name
-        ctx.env.project
+    if pkg_dir !== nothing
+        project_data = TOML.parsefile(joinpath(pkg_dir, "Project.toml"))
+        return VersionNumber(project_data["version"])
     else
-        Pkg.Types.manifest_info(ctx, pkg_id.uuid)
+        return nothing
     end
-
-    return pkg_info.version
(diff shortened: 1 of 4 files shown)


KristofferC:
I think this should just look at the Project file of the package itself, not of the project/manifest of the active environment which seems like a very roundabout way of getting to the place where the version actually was defined.

So something like (using https://github.com/JuliaLang/julia/pull/33128) and with error checking

```jl
project_data = Pkg.TOML.parsefile(joinpath(pkgdir(@__MODULE__), "Project.toml"))
return VersionNumber(project_data["version"])
```

omus:
> I think this should just look at the Project file of the package itself, not of the project/manifest of the active environment which seems like a very roundabout way of getting to the place where the version actually was defined.

How can we locate the path of the package when in sub-module? I'm under the impression you meant `pathof` when you said `pkgdir`?

omus:
> This seems to encourage having a bunch of dead code around though.

I agree this will leave dead code around for a while. This however is better then forgetting to remove a deprecation when you do a version bump

fredrikekre:
https://github.com/JuliaLang/julia/pull/33128

KristofferC:
This should only be used at top level so the result ends up in the precompile file and so that no runtime parsing of the project file is needed.

src/API.jl
```diff
@@ -977,4 +977,22 @@ Package(name::AbstractString) = PackageSpec(name)
...
+    @__VERSION__ -> Union{VersionNumber, Nothing}
+
+Get the `VersionNumber` of the package which expands this macro. If executed outside of a
+package `nothing` will be returned.
+"""
+macro __VERSION__()
+    pkg_dir = Base.pkgdir(__module__)
+
+    if pkg_dir !== nothing
+        project_data = TOML.parsefile(joinpath(pkg_dir, "Project.toml"))
```
00vareladavid: `joinpath(pkg_dir, "Project.toml"))` should be `Types.projectfile_path(pkg_dir)` because `Project.toml` is not the only name for a project file

omus:
Code should be ready to go

StefanKarpinski:
While this is better, it seems like if we're going to officially support this, we should it right and record the tree hash and version number at the time something was loaded, rather than this, which can become out of sync after loading.

src/API.jl
```diff
@@ -977,4 +977,22 @@ Package(name::AbstractString) = PackageSpec(name)
...
+    @__VERSION__ -> Union{VersionNumber, Nothing}
+
+Get the `VersionNumber` of the package which expands this macro. If executed outside of a
+package `nothing` will be returned.
+"""
+macro __VERSION__()
+    pkg_dir = pkgdir(__module__)
+
+    if pkg_dir !== nothing
+        project_data = TOML.parsefile(Types.projectfile_path(pkg_dir))
```
tkf: Maybe it makes sense to re-compile when the version in Project.toml is changed?

```suggestion
        toml_path = Types.projectfile_path(pkg_dir)
        include_dependency(toml_path)
        project_data = TOML.parsefile(toml_path)
```

IanButterworth:
I believe `Base.pkgversion` (new in 1.9) provides the same functionality https://docs.julialang.org/en/v1.9.0-rc3/base/base/#Base.pkgversion-Tuple{Module}

Please feel free to reopen if not, just a tidyup close

PR #1502, opened 2019-11-13
Improve "No changes" message for rm() & behavior during build

ianshmean:
The message `Info: No changes` is confusing in some cases. For instance,  when you know that `build.jl` has actually changed the package build state. This is designed to make it clearer that it's referring to dependencies

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -903,7 +903,7 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
         uuid ∉ drop
     end
     if length(ctx.env.project.deps) == n
-        println(ctx.io, "No changes to package dependencies")
+        println(ctx.io, "No dependencies have been removed")
         return
     end
     # only declare `compat` for direct dependencies
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -903,7 +903,7 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
         uuid ∉ drop
     end
     if length(ctx.env.project.deps) == n
-        println(ctx.io, "No changes")
+        println(ctx.io, "No changes to package dependencies")
         return
     end
     # only declare `compat` for direct dependencies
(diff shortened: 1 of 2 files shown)


src/Operations.jl
```diff
@@ -903,7 +903,7 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
         uuid ∉ drop
     end
     if length(ctx.env.project.deps) == n
-        println(ctx.io, "No changes")
+        println(ctx.io, "No changes to package dependencies")
```
00vareladavid: What about: `"no direct dependencies have been removed"`?

ianshmean: That does seem more accurate! Updated. Thanks

ianshmean:
This message is a bit more accurate here, but it does make me wonder whether it needs to be printed every time for packages with `deps/build.jl`, given that it's not printed for packages without. For instance, current 1.3 behavior, where Spinnaker has a `build.jl`. This is what is printed whether or not `build.jl` changes anything.

```
(v1.3) pkg> build CSV

(v1.3) pkg> build Spinnaker
  Building Spinnaker → `~/.julia/packages/Spinnaker/pl5GO/deps/build.log`
[ Info: No changes
```

ianshmean:
It turns out that this message was being printed because `Pkg.API.rm` was being called on an empty array at https://github.com/ianshmean/Pkg.jl/blob/03cdb3b29e427bdfdcda09c5e7985d9c8d52f165/src/backwards_compatible_isolation.jl#L516

I've made that only call if `length(not_loadable) > 0`

ianshmean:
@fredrikekre what do you think? Relates to https://github.com/JuliaLang/Pkg.jl/pull/1222

KristofferC:
Why are we printing anything at all here? Was there a discussion somewhere about that?

00vareladavid:
I dont think anyone meant to print something. Someone probably just made a call to `Pkg.rm`, we didn't have the option to disable printing through `ctx` before.

KristofferC:
What's the status of this now when `backwards_compatible_isolation.jl` is removed? Is it still needed?

00vareladavid:
I don't believe this is an issue anymore. We just call `prune_manifest` directly now (instead of `rm`) so spurious printing is not an issue.

KristofferC:
Alright.

PR #1504, opened 2019-11-14
(Updated) Remove OS-specific shell commands from Pkg docs

00vareladavid:
All credit should go to Chris Rackauckas, I just updated #765 for the current doc layout we are using.

close #765


fredrikekre:
Much more ugly but 🤷‍♂

PR #1512, opened 2019-11-26
Don’t remove whitespace inside the string

### LINKED ISSUE #1511: Version bounds with a space not a comma
I made the mistake of writing this
```
julia = "0.7 1.0"
```
over [here](https://github.com/JuliaMath/VML.jl/pull/24/commits/1e2c6200b405d44ee879c034b1d7daa05cd37a2f), which leads to warnings when loading the package on Julia 1.3.

@fredrikekre points out that this is parsed as follows:
```
julia> Pkg.Types.semver_spec("0.7 1.0")
VersionSpec("0.71”)
```
Surely I’m not the only guy who’ll ever make this mistake. Perhaps it should either be an error, or be accepted?

DilumAluthge:
Fixes #1511

### DIFF
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -256,11 +256,8 @@ Base.show(io::IO, s::VersionSpec) = print(io, "VersionSpec(\"", s, "\")")
 ###################
 
 function semver_spec(s::String)
-    s = replace(s, r"\s-\s" => "→") # replace " - " with "→"
-    s = replace(s, " " => "") # remove all spaces
-    s = replace(s, "→" => " - ") # replace "→" with " - "
     ranges = VersionRange[]
-    for ver in split(s, ',')
+    for ver in strip.(split(strip(s), ','))
         range = nothing
         found_match = false
         for (ver_reg, f) in ver_regs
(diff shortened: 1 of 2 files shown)


src/versions.jl
```diff
@@ -223,7 +223,7 @@ function Base.intersect(A::VersionSpec, B::VersionSpec)
         i += 1
     end
     VersionSpec(ranges)
-end
+endi
```
fredrikekre: 🤔

DilumAluthge: GitHub web editor + iPad on-screen keyboard.

DilumAluthge: Actually it’s my new syntax I’m proposing for Julia 2.0. Deprecate `end` in favor of `endi`.

DilumAluthge:
Alright, this is ready to go. @fredrikekre @KristofferC 

As you can see from the new test, "0.7 1.0" et cetera will now throw an error.

DilumAluthge:
Bump. Any opposition to merging this?

KristofferC:
Thanks!

PR #1513, opened 2019-11-28
fix GitTools.tree_hash executableness criterion

StefanKarpinski:
Git only looks at the user's executable bit. This seems to be causing issues for users with strict umasks: https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/527.

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -202,7 +202,7 @@ function gitmode(path::AbstractString)
     elseif isdir(path)
         return mode_dir
     # We cannot use `Sys.isexecutable()` because on Windows, that simply calls `isfile()`
-    elseif filemode(path) & 0o010 == 0o010
+    elseif !iszero(filemode(path) & 0o100)
         return mode_executable
     else
         return mode_normal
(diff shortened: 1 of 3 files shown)


fredrikekre:
Possible to add a test?

StefanKarpinski:
@staticfloat, looks like this fix breaks one of the artifact known hash tests:

https://travis-ci.org/JuliaLang/Pkg.jl/jobs/618610010#L802

Can you check whether the hash should change or not?

StefanKarpinski:
I looked into it myself: using the git command-line client, the existing reference hash is correct and the tree in question doesn't involve any executable files, so this fix has broken something else.

StefanKarpinski:
Mystery solved: it was the other fix that broke this. I now understand this code much better: it was using throwing an error for an empty tree to signal that the tree was empty; I've refactored so that it explicitly skips empty subdirectories during recursion instead. This leads to a double call to `readdir` which is not ideal, but it seems better to support the valid hashing of the empty tree and to not use try/catch for control flow like that.

StefanKarpinski:
Ok, tests pass, finally; now splitting the code into logical commits with the same end state.

StefanKarpinski:
Or did we turn bors off?

staticfloat:
Thanks for this Stefan; I had previously thought that we should mimic `git`'s behavior of not allowing empty directories, but it turns out that I was wrong about that as well; you can pass `--allow-empty` and `git` will give you hashes like this as well (as corroborated by your added tests).

PR #1519, opened 2019-12-02
WIP: Store Julia version in Manifest.toml

### LINKED ISSUE #1458: Additional Manifest entries
From Slack,
`Manifest.toml` should contain all the needed information in order to recreate an environment. However, a key piece of information used in solving the graph is not tracked: the Julia version. We should start including an entry on the Manifest for the Julia version used. When recreating an environment  using `instaniate -m`, it should warn if the Julia version is different from the one used to create that `Manifest.toml`.

DilumAluthge:
Fixes #1458


DilumAluthge:
This breaks more things than it fixes.

KristofferC:
I do think this is probably a good idea (the implementation can however likely be much more simple than in this PR). What I don't think is viable is starting to commit multiple manifest with some `Manifest_1.3` name scheme that depends on the julia verision. Being reproducible on one Julia version is imo good enough.

DilumAluthge:
I wasn’t able to get this working, but maybe @KristofferC or @fredrikekre could take a stab at it?

My idea was to have an entry in the Manifest.toml file for the Julia version. And then if you load a Manifest.toml file, and the Julia entry in the Manifest.toml file is different from the value of Base.VERSION in your current Julia session, then print a warning.

PR #1522, opened 2019-12-03
Fix tree hashing with nested empty directories

staticfloat:
When testing for directories to exclude from hashing, we must exclude
not only empty directories, but also directories that themselves contain
nothing but empty directories; in essence, we suppress adding
directories that have no files contained within their entire subtree.

While fixing this, it seemed prudent to eliminate the `names` argument
to `tree_hash()`, especially as it was not actually used to iterate
over.

This will fix the error reported in https://github.com/JuliaLang/julia/pull/33979#issuecomment-561041847

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -251,16 +251,29 @@ function blob_hash(path::AbstractString, HashType = SHA.SHA1_CTX)
 end
 
 """
+    contains_no_files(root::AbstractString)
+
+Helper function to determine whether a directory contains no files; e.g. it is
+either empty or contains only other directories that contain nothing but other
+directories.  This is used to exclude directories from tree hashing.
+"""
+function contains_no_files(root::AbstractString)
+    for (root, dirs, files) in walkdir(root)
+        if !isempty(files)
+            return false
+        end
+    end
+    return true
+end
+    
+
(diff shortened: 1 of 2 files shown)


staticfloat:
I'm explicitly requesting a review from @StefanKarpinski just to make sure that my changing the signature of `tree_hash()` doesn't effect anything anywhere else; I figured the `names` parameter was intended only for efficiency reasons, but since it wasn't used anyway (and has less of a chance to be used now) I figure it's okay to drop it.

StefanKarpinski:
Not using the names argument was a mistake. I meant to replace the call to readdir with it. Looks good to me. I have to say this makes it seems even wonkier there git doesn’t just allow empty directories.

staticfloat:
Yeah, `git` is very "files" focused; so it does kind of feel like directories are kind of this strange addition that was bolted on, rather than designed in from the beginning.

StefanKarpinski:
It's bizarre for a number of reasons:
1. a feature that people want
2. it would be more straightforward to compute tree hashes if it was allowed
3. there exists workarounds like putting an empty `.gitignore` file in the directory to commit it
4. it would be mostly non-breaking for git to start allowing it

PR #1526, opened 2019-12-04
Try a little harder to figure out Julia's libgfortran restriction

staticfloat:
If you use `MKL.jl` to rebuild a system image that uses MKL instead of
OpenBLAS, you can end up with a Julia that doesn't eagerly load
`libgfortran` into its process.  This means that when you ask `Pkg` what
its platform key is, it doesn't realize that it actually has a
libgfortran dependency imposed upon it by the _other_ packages (such as
`SuiteSparse`) that were compiled against the libgfortran version that
we ship with Julia.

Slowly but surely, we are moving into a world where Julia will not be
shackled to a particular libgfortran version at all, but we're not quite
there yet.  Right now what happens is that Julia looks to see if a
`libgfortran` is already loaded, sees that one isn't, and concludes that
there is no constraint upon `libgfortran`, and happily downloads the
latest version of things like `OpenSpecFun_jll` available.  This doesn't
work for two reasons:

* We assume that the proper `libgfortran` is bundled with Julia (it is
bundled, but since Julia hadn't loaded it yet, it didn't know that)
* Even if we shipped the proper `libgfortran` alongside
`OpenSpecFun_jll` (which we could, it's called
`CompilerSupportLibraries_jll`) we shouldn't load two different versions
of `libgfortran` within the same Julia, which is what would happen if we
were to load `SuiteSparse` after loading `OpenSpecFun_jll`.

Ideally, if we had no dependence on `libgfortran` within Julia at all,
you would `Pkg.add()` packages such as `SuiteSparse` and `OpenSpecFun`
and they would both choose the most recent build available (e.g.
`libgfortran5`), but since `SuiteSparse` is still shipped alongside
Julia, we need to try a little harder to lock ourselves in to the
`libgfortran` version that Julia supports.

To do that, I'm just attempting to `dlopen()` all the different
versions, starting from the oldest versions first.  I do this first
looking only in `Base.LIBDIR`, then finally globally.  This allows for
Julia to pick up a global libgfortran and thereby, in the far-off
future, when we do have a completely libgfortran-free Julia, default to
using a version that is most probably compatible with the host system.

### DIFF
--- a/src/BinaryPlatforms.jl
+++ b/src/BinaryPlatforms.jl
@@ -693,6 +693,30 @@ function detect_libgfortran_version(libgfortran_name::AbstractString, platform::
 end
 
 """
+    probe_dlopen_libgfortran()
+
+Attempts to `dlopen()` libgfortran by guessing its SONAME.
+"""
+function probe_dlopen_libgfortran(;platform::Platform = default_platkey,
+                                   versions::Vector = [3, 4, 5])
+    libgfortran_name(p::MacOS, ver)                = "libgfortran.$(ver).dylib"
+    libgfortran_name(p::Union{Linux,FreeBSD}, ver) = "libgfortran.so.$(ver)"
+    libgfortran_name(p::Windows, ver)              = "libgfortran-$(ver).dll"
+
+    # First, look ONLY in Base.LIBDIR; if that doesn't work, look globally.
+    for prefix in (Base.LIBDIR, "")
+        for version in versions
+            libgfortran_path = joinpath(prefix, libgfortran_name(platform, version))
+            hdl = dlopen(libgfortran_path; throw_error=false)
+            if hdl != nothing
+                return hdl
+            end
+        end
+    end
+    return nothing
+end
+
+"""
     detect_libgfortran_version()
 
 Inspects the current Julia process to determine the libgfortran version this Julia is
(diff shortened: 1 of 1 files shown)


staticfloat:
Marking this as a draft until I have time to write coherent tests.  @giordano this is the kind of thing I think we should do to fix https://github.com/JuliaMath/SpecialFunctions.jl/issues/192

src/BinaryPlatforms.jl
```diff
@@ -692,6 +692,30 @@ function detect_libgfortran_version(libgfortran_name::AbstractString, platform::
...
+Attempts to `dlopen()` libgfortran by guessing its SONAME.
+"""
+function probe_dlopen_libgfortran(;platform::Platform = default_platkey,
+                                   versions::Vector = [3, 4, 5])
+    libgfortran_name(p::MacOS, ver)                = "libgfortran.$(ver).dylib"
+    libgfortran_name(p::Union{Linux,FreeBSD}, ver) = "libgfortran.so.$(ver)"
+    libgfortran_name(p::Windows, ver)              = "libgfortran-$(ver).dll"
+
+    # First, look ONLY in Base.LIBDIR; if that doesn't work, look globally.
+    for prefix in (Base.LIBDIR, "")
```
giordano: Maybe `Base.LIBDIR` and `joinpath(Base.LIBDIR, "julia")`?  Do we need to prepend `Sys.BINDIR` too?

staticfloat:
Upcoming CompilerSupportLibraries_jll integration with Julia will make this obsolete.

PR #1531, opened 2019-12-06
Pkg client: authentication support

StefanKarpinski:
This is WIP currently. First step is adding the ability to pass headers to download agents.


StefanKarpinski:
Decided that since this is standalone and was passing tests, I might as well merge and let it settle.

PR #1536, opened 2019-12-09
When writing Manifest.toml, add a comment with the current Julia version

DilumAluthge:
This does NOT fully solve #1458. However, I think it is a good stopgap solution until someone figures out how to actually solve #1458. (I tried, see #1519, but I couldn’t get it to work.)

Anyway, this pull request is very simple:

**Before this pull request:** the following comment is added to the top of every `Manifest.toml` file generated by Pkg:
```toml
```

**After this pull request:** the following comments are added to the top of every `Manifest.toml` file generated by Pkg:
```toml
# Generated by Julia version 1.2.3
```

I think this pull request is better than the status quo because at least a human will be able to look at the top of the `Manifest.toml` file and easily see which version of Julia generated it. The current situation is that there is no easy way for a human to figure out the version of Julia used to generate a given `Manifest.toml` file.

---

In the updated version of this pull request, `-DEV` versions are shortened, to make life a little easier when working on master.

So here is the correspondence between version number and the message in the manifest:

| `Base.VERSION` | Message in the manifest |
| - | - |
| `1.2.3` | `# Generated by Julia version 1.2.3` |
| `1.2.3-DEV` | `# Generated by Julia version 1.2.3-DEV` |
| `1.2.3-DEV.456` | `# Generated by Julia version 1.2.3-DEV` |


KristofferC:
Going to be a lot of diffs on the Manifest when working of julia master on a package or something. Perhaps worth it, but I know it will be annoying :P

DilumAluthge:
> Going to be a lot of diffs on the Manifest when working of julia master on a package or something. Perhaps worth it, but I know it will be annoying :P

Well if you are working on Julia master, then it’s all the more important to keep track of where you were on master when you generated the manifest. After all, `1.4.0-DEV.557` and `1.4.0-DEV.558` could be very different. A package or dependency that is broken on `1.4.0-DEV.557` might be fixed on `1.4.0-DEV.558`.

DilumAluthge:
Bump. Any chance that we could add the `triage` label to this so that it can be discussed during the next Pkg call?

DilumAluthge:
> Going to be a lot of diffs on the Manifest when working of julia master on a package or something. Perhaps worth it, but I know it will be annoying :P

Alright, so I have changed this PR so that you won't have this issue. If you are working on Julia `master` and your version number is e.g. `1.2.3-DEV.456`, then I won't include the `.456` in the message I print to the manifest, i.e. the message will be `# Generated by Julia version 1.2.3-DEV`.

DilumAluthge:
Bump. @KristofferC @00vareladavid What do you think of this PR now that I've removed the annoying behavior when working on Julia master?

00vareladavid:
Sorry for the late response @DilumAluthge. Can you elaborate on the benefit of this approach? It still seems like a lot of manual work for the user.

Say I get a project with a manifest. Now I would have to check what the version of the manifest is. If it is not the correct version. Then I would conceivably stash the current manifest somewhere and then resolve a new one.

Where you planning on some different workflow?

DilumAluthge:
Here’s the current situation: I send you a Manifest.toml file. You have literally no idea what version of Julia I used to generate that Manifest. You open e.g. Julia 1.0.1, run `Pkg.instantiate` and cross your fingers that everything works. If it doesn’t work, I guess you have to try Julia 1.0.1. And in fact you have to try every version of Julia until you find a version of Julia that works with the given Manifest. In my opinion, this is a terrible situation. If Manifests are supposed to provide reproducible environments, shouldn’t they include all of the information that you need to reproduce the environment?

Here’s the situation with this PR: I send you a Manifest.toml file. You peek at the top of the file with `head` and quickly learn which version of Julia I used to generate the Manifest. You download (or compile) the necessary version of Julia if you don’t already have it. Then you open that version of Julia, and run `Pkg.instantiate`. You know it will work because you are using the same version of Julia that I used to generate the Manifest. Now you have reproduced the exact environment that I wanted to send you.

DilumAluthge:
> Then I would conceivably stash the current manifest somewhere and then resolve a new one.

If I send you both a Project.toml file and a Manifest.toml file, and you delete the Manifest and generate a new one, then you are not reproducing the same environment that I sent you. But isn’t that the whole point of Manifests? Reproducible environments?

DilumAluthge:
> It still seems like a lot of manual work for the user.

Yeah I agree that a more "automatic" or "natural" approach would be better. I tried that with #1519, but I couldn’t get it to work. (My PR broke the undo/redo feature, among other things.)

I think the best approach would be something where when you load a Manifest, if the version of Julia that you are running does not match the version of Julia used to generate the Manifest, then Pkg automatically prints a warning message that tells you which version of Julia was used to generate the Manifest. But I don’t know how to implement that.

The purpose of this PR is to be a stopgap until we get a better solution. 

Right now the Julia ecosystem is full of Manifests, and none of them contain the information on which version of Julia was used to generate them, which makes it very difficult to reproduce those environments. At least this PR adds that information to the Manifest.

DilumAluthge:
I’d love to see a better solution but right now we have nothing.

DilumAluthge:
Bump.

What are the current thoughts on storing Julia versions in `Manifest.toml` files? As far as I can tell, this is the last piece of information that is keeping Manifests from being truly reproducible. Once we have the Julia version in the Manifest file, then a Manifest really does provide a way of reproducing an exact environment. But until we do that, Manifests do not deliver on that feature.

As I see it, we have two options:
1. Store the Julia version in the `Manifest.toml` file
2. Have different `Manifest.toml` files for each version of Julia, with a naming convention that we establish. So you might have `Manifest.1.2.0.toml`, `Manifest.1.3.0.toml`, `Manifest.master.toml`, etc. Then you can have multiple Manifests living side by side in harmony, with a different Manifest for each version of Julia for which you want to provide this reproducibility guarantee.

Personally I think option 2 is the best. But I know @KristofferC hates it. So that's why I've opened this PR, which is the simplest way of implementing option 1.

Is there a third option that I have not considered?

StefanKarpinski:
Let’s try it this way and see how annoying it actually is.

fredrikekre:
Who will read this comment? If you create a project + manifest to distribute, can't you just put Julia as a `[compat]` entry?

KristofferC:
A comment is pretty useless. If anything, it should be something like an entry `julia_version = "..."`.

DilumAluthge:
> If you create a project + manifest to distribute, can't you just put Julia as a `[compat]` entry?

You can distribute a Manifest.toml without a Project.toml file.

> If anything, it should be something like an entry julia_version = "...".

I tried that in https://github.com/JuliaLang/Pkg.jl/pull/1519 but I couldn’t get the tests to pass. It broke undo/redo functionality and I couldn’t fix it. I agree that this would be a better option, if someone is able to figure out how to implement it.

StefanKarpinski:
If this info is going into the manifest as actual TOML data, it should be a `[[julia]]` entry with the appropriate UUID for Julia itself and the version number and commit hash. The main issue is that Pkg needs to know that there's always an implicit top-level dependency on `julia` so it doesn't delete the manifest entry.

KristofferC:
> My understanding is that the Manifest.toml file by itself should be able to completely reproduce the environment.

Should put the OS in there as well then? And the architecture, and the libgfortran dependency so you get the same artifacts? Etc. Reproduciblility is clearly a continuous measure.

> The main issue is that Pkg needs to know that there's always an implicit top-level dependency on julia so it doesn't delete the manifest entry.

I mean, we can just tell it not to delete it?  Or do you mean for older versions of Pkg?

StefanKarpinski:
Right, so the real question is whether the Julia version is something we can count on as a given that can remain implicit or is part of the reproducibility story. As @KristofferC says, we abstract over many things: operating system, libc kind/version, libgfortran, etc. Is the Julia version one of those things that we abstract over or one of those things that we capture for reproducibility. Part of the reason for capturing package and artifact versions exactly is that it's quite common for packages to break compatibility in various ways, intentionally or by accident. Julia maintains a very high standard for not breaking things: I run master all the time with frequent updates and rebuilds and it's generally not a concern, everything keeps working. So is the Julia version a sufficiently good abstraction like the operating system, or is a leaky abstraction like packages?

DilumAluthge:
Take the same `Project.toml` file. Put it on two different computers with the same version of Julia, but different operating system, different architecture, and different version of gfortran. In a fresh depot, instantiate the `Project.toml` file on each computer at the same time. You should get the same exact `Manifest.toml` file on both computers. The resolver does not consider operating system, architecture, or version of gfortran.

Maybe I am mistaken about this. If so, perhaps someone could provide a counter-example?

Take the same `Project.toml` file. Put it on two different computers that have different versions of Julia but are otherwise identical systems. In a fresh depot, instantiate the `Project.toml` file on each computer at the same time. It is very likely that you will get two different `Manifest.toml` files. The resolver will only consider packages that are compatible with the current version of Julia.

Suppose for example that this is the `Project.toml` file:
```toml
[deps]
Foo = "uuid-for-foo"
```

Suppose also that there are two versions of the Foo package registered in the General registry: Foo 1.0 and Foo 1.1.

Suppose that Foo 1.0 has the following `[compat]` entry for Julia:
```toml
[compat]
julia = "1"
```

And suppose that Foo 1.1 uses multithreading and therefore has the following `[compat]` entry for Julia:
```toml
[compat]
julia = "1.3”
```

If you put this `Project.toml` file on two computers, where one is running Julia 1.0 and the other is running Julia 1.3, but the systems are otherwise identical, and you instantiate the same `Project.toml` on each computer in a fresh depot, you will get two different `Manifest.toml` files. The computer running Julia 1.0 will have a manifest with Foo 1.0. The computer running Julia 1.3 will have a manifest with Foo 1.1.

Therefore, as I see it, the Julia version is a leaky abstraction. In this example, it clearly leaks into the manifests.

In contrast, I cannot think of an example in which any of the other factors impacts the creation of the manifests.

DilumAluthge:
> Should put the OS in there as well then? And the architecture, and the libgfortran dependency so you get the same artifacts? Etc. Reproduciblility is clearly a continuous measure.

Sure. A `Manifest.toml` file won’t reproduce the exact same artifacts. But it should be able to reproduce the exact same "environment", right? By "environment", I mean the exact list of packages and their versions. In other words, the dependency graph.

StefanKarpinski:
Reproducibility of version resolution isn't really a goal though. Yes, it depends on Julia version, but it also depends on registry state, which is generally ephemeral. If you add things in a different order you also might get a different set of package versions. (Unlike Pkg1/2 where this was generally considered a huge annoyance because doing any package operation caused a full resolve, which would change all versions potentially, so it was really hard to conservatively change as little as possible to add/remove dependencies.)

DilumAluthge:
Hmmm. If that is the case, then why would you ever share a manifest or check it into source control?

DilumAluthge:
If I understand correctly, the difference between a project file and a manifest file is that the manifest contains the results of version resolution. If reproducing version resolution is not a goal, then what is the purpose of the manifest file?

StefanKarpinski:
> If that is the case, then why would you ever share a manifest or check it into source control?
> If reproducing version resolution is not a goal, then what is the purpose of the manifest file?

You seem to be mixing up two things:
1. reproducibility of versions
2. reproducibility of version resolution

The manifest solves the first one: it tells you exactly what version of packages to use. The second one is not a goal: I don't care how you got your manifest or if I can get the same one somehow, I just care that this was your manifest when I'm running the code. That's what the manifest gives you: a record of what versions of things were used. Whether you can produce the same manifest or not doesn't matter, which is kind of the point of the manifest.

Which brings us back to the abstraction question. Part of the reason we _have_ to abstract over operating system and libc version and hardware is that we have no control over it. It's informative, yes, but it's not like we're going to say "oops, you're running the wrong operating system, please install Linux on this desktop and then try again."

So the question is, what are we going to do about the version of Julia being different? Currently, nothing. We will just try loading the recorded versions of all the packages and hope that it works. We could conceivably do something where manage Julia versions the same way we manage other binary dependencies and switch to whatever version of Julia is recorded in a manifest. But are we really going to do that? Maybe, but we're certainly not there at this point.

DilumAluthge:
> You seem to be mixing up two things:
> 
> 1. reproducibility of versions
> 2. reproducibility of version resolution
> 
> The manifest solves the first one: it tells you exactly what version of packages to use. The second one is not a goal: I don't care how you got your manifest or if I can get the same one somehow, I just care that this was your manifest when I'm running the code. That's what the manifest gives you: a record of what versions of things were used. Whether you can produce the same manifest or not doesn't matter, which is kind of the point of the manifest.

Ah yes that’s exactly the confusion I was having.

DilumAluthge:
> So the question is, what are we going to do about the version of Julia being different? Currently, nothing. We will just try loading the recorded versions of all the packages and hope that it works. We could conceivably do something where we manage Julia versions the same way we manage other binary dependencies and switch to whatever version of Julia is recorded in a manifest. But are we really going to do that? Maybe, but we're certainly not there at this point.

I guess my vision would be a warning: `This Manifest was created by Julia 1.3.0 but you are running Julia 1.0`. And then the user, if they want to, can close Julia and reopen a different version of Julia.

StefanKarpinski:
If we were to emit such a warning, how fussy should it be? If I generated a manifest with Julia 1.0.5 and someone is running Julia 1.0.4, should I warn them? Wouldn't the more accurate criterion for warning be if the project or any of the dependencies in the manifest had a compat bound on the Julia version that the current version doesn't satisfy? What if it happens to work despite the compat bounds not being satisfied?

timholy:
One option would be to mention this only if there's a version incompatibility in the project. For example, recent versions of packages with extensive binary dependencies require at least Julia 1.3; if I share such a project with a colleague who is running Julia 1.2, s/he is going to get an error when trying to run the project. Having this information present in the Manifest file might make it easier to issue an informative message on any resolver error.

DilumAluthge:
How about a user-configurable setting?
- `warn_all`: print a warning whenever the manifest Julia version is not exactly the same as your Julia version 
- `warn_smart`: the default value. Print a warning whenever the manifest Julia version is not exactly the same as your Julia version AND there exists as least one package version in the manifest is not compatible with your Julia version
- `warn_never`: don’t print any warnings

That way, people that don’t want the warnings can disable them.

StefanKarpinski:
Making things behave differently everywhere is rarely helpful. It's better to pick one good solution.

DilumAluthge:
> Making things behave differently everywhere is rarely helpful. It's better to pick one good solution.

In that case, it probably makes the most sense to pick:
Print a warning whenever the manifest Julia version is not exactly the same as your Julia version AND there exists as least one package version in the manifest is not compatible with your Julia version

DilumAluthge:
It's looking like the consensus in this discussion is to close this PR in favor of the following:
1. Store the Julia version in the manifest as actual TOML data
2. Print a warning whenever the manifest Julia version is not exactly the same as your Julia version AND there exists as least one package version in the manifest is not compatible with your Julia version

Does that seem correct?

DilumAluthge:
> It's looking like the consensus in this discussion is to close this PR in favor of the following:
> 
>     1. Store the Julia version in the manifest as actual TOML data
> 
>     2. Print a warning whenever the manifest Julia version is not exactly the same as your Julia version AND there exists as least one package version in the manifest is not compatible with your Julia version
> 
> 
> Does that seem correct?

Alright, I am going to close this PR in favor of the above. Maybe someone can figure out how to implement the first step (store the Julia version in the manifest as actual TOML data). Perhaps you could salvage my failed attempt in https://github.com/JuliaLang/Pkg.jl/pull/1519. Or maybe it would be easier to start from scratch.

PR #1537, opened 2019-12-09
PlatformEngines: use `tar -m` to ignore mtimes

StefanKarpinski:
People are getting complaints from `tar` that their mtimes are in the future: https://discourse.julialang.org/t/artifacts-toml-timestamp-is-in-the-future/32047. We don't actually care about mtimes at all, so we should pass the `-m` flag to tar when extracting to tell it to just create files with natural system mtimes based on their actual time of modification.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -329,7 +329,7 @@ function probe_platform_engines!(;verbose::Bool = false)
             elseif endswith(tarball_path, ".bz2")
                 Jjz = "j"
             end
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
This looks good from my perspective, @staticfloat, if this looks good to you, please merge.

StefanKarpinski:
@fredrikekre, we probably want to make sure this ends up in 1.3.1 so that people stop getting complaints about [future sailors](https://www.youtube.com/watch?v=eAbkh4TMRqg).

PR #1543, opened 2019-12-11
Disable needless specialization in Artifacts code

staticfloat:
After spending some quality time with `--trace-compile=stderr` and `@snoopi`, I have found a fairly minimal set of changes that speeds up loading of [`TestGtkMegaJLL.jl`](https://github.com/giordano/TestGtkMegaJLL.jl) from `2.12s` to `0.69s` on my machine.  It's all essentially nonspecialization hacks.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -540,8 +540,13 @@ const precompile_script = """
     # Create simple artifact, bind it, then use it:
     foo_hash = Pkg.Artifacts.create_artifact(dir -> touch(joinpath(dir, "foo")))
     Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo", foo_hash)
+    # Also create multiple platform-specific ones because that's a codepath we need precompiled
+    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=platform_key_abi())
+    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=Linux(:x86_64), force=true)
+    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=Windows(:x86_64), force=true)
+    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=MacOS(:x86_64), force=true)
     # Because @artifact_str doesn't work at REPL-level, we JIT out a file that we can include()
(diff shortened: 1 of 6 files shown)


staticfloat:
@KristofferC I tried simplifying the artifact code to eliminate the download path, but it didn't actually help anything, so I elected for the smaller changes (with no change in behavior, which that would constitute for 1.3.1)

staticfloat:
The cumulative impact of all the optimizations we've done so far is that `using GTK3_jll` has gone from `12.5s` on 1.3.0 to `6.1s` on this branch.  Not bad!

KristofferC:
What's the status here? Should it be merged for 1.3.1?

staticfloat:
This should eventually be merged into 1.3.X (and 1.4 of course); it's just performance improvements so it's not time-critical, but it's a nice-to-have.  There's nothing critical missing in this PR; although the test isn't useful yet, because I don't know how to make the precompile tester actually fail, so our precompile script can still bitrot.

StefanKarpinski:
Isn’t improving the package load time on 1.3 fairly pressing?

KristofferC:
Yeah, but I know Jameson didn't like us putting more Pkg precompile into the sysimage. But I think we should merge this and put it in 1.3.1 and keep thinking about how it can be improved.

KristofferC:
The way to check the precompilation script is to check that the output after each command is as expected.

src/Pkg.jl
```diff
@@ -537,6 +537,17 @@ const precompile_script = """
     ] add Te\t\t$CTRL_C
     ] st
     $CTRL_C
+    # Create simple artifact, bind it, then use it:
+    foo_hash = Pkg.Artifacts.create_artifact(dir -> touch(joinpath(dir, "foo")))
+    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo", foo_hash)
+    # Also create multiple platform-specific ones because that's a codepath we need precompiled
+    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=platform_key_abi())
```
KristofferC: Did this every work?

```
julia> using Pkg.Artifacts

julia> platform_key_abi
ERROR: UndefVarError: platform_key_abi not defined
```

giordano: It's in Pkg.BinaryPlatforms if I remember correctly

KristofferC: Doesn't seem like it..?

staticfloat: Perhaps it got lost in a rebase somewhere; I distinctly remember adding a `using Pkg.BinaryPlatforms` as well as testing this and having it work.

PR #1547, opened 2019-12-16
Fix `safe_realpath` for shared network drives

ckingdon95:
See discussion here:
https://github.com/JuliaLang/julia/issues/33956

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -14,7 +14,13 @@ end
 
 # try to call realpath on as much as possible
 function safe_realpath(path)
-    ispath(path) && return realpath(path)
+    if ispath(path) 
+        try
+            return realpath(path)
+        catch
+            return path
+        end
+    end
     a, b = splitdir(path)
     return joinpath(safe_realpath(a), b)
 end


StefanKarpinski:
I guess the question is does this fix the issue for you?

ckingdon95:
Yes! I can now activate environments on my network drive. I'm not sure if there are other cases though that I should test out

ckingdon95:
Actually, I spoke too soon. The activation works, but trying to do a `status` or using one of the packages from the environment results in the same error. I will look at where these other calls to `realpath` are coming from.

ckingdon95:
Unfortunately, the line causing these next errors is in julia base, not in Pkg. here: https://github.com/JuliaLang/julia/blob/master/base/loading.jl#L120

The function `dummy_uuid` is being called when I do a `Pkg.status` or when I try to use a package in the environment. It calls `realpath` on the project file, which in my case is on the shared network drive.

StefanKarpinski:
Exciting! You get to make a PR to Base 😀. It would be fine for that function to do something similar, i.e. wrap the call to `realpath` in a try catch and just use `project_path` if `realpath` fails.

StefanKarpinski:
I've approved but, @ckingdon95, I need you to make this no longer a draft before I can merge.

ckingdon95:
@StefanKarpinski should we be concerned about the one failing Travis build on this one?

StefanKarpinski:
It doesn't look related, but I've restarted it just to be sure.

fredrikekre:
I wonder if we should do this Windows only. Maybe doesn't matter.

davidanthoff:
It would actually be interesting to know if one uses a Linux/Mac client to access the same network share... But that probably shouldn't hold up this PR, right?

PR #1549, opened 2019-12-17
provide a fallback when sandboxing a package with no Project.toml

### LINKED ISSUE #1550: Packages fail during tests if they don't have a `Project.toml` file
For example:
- https://github.com/maleadt/BasePkgEvalReports/blob/adfa530cc2662ca9abd85544d3ed41aa6d0ad95f/pkgeval-daily_2019_12_16/logs/Permutations/1.4.0-DEV-a6bf74a2f9.log#L36-L65

Right now, we believe that this was probably caused by https://github.com/JuliaLang/Pkg.jl/pull/1427

cc: @00vareladavid

00vareladavid:
reintroduces some functionality I missed in #1427 

close #1550

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -804,15 +804,6 @@ function dependency_order_uuids(ctx::Context, uuids::Vector{UUID})::Dict{UUID,In
     return order
 end
 
-function gen_build_project(ctx::Context, pkg::PackageSpec, source_path::String)
-    build_project = Types.Project()
-    source_ctx = Context(env = EnvCache(projectfile_path(source_path)))
-    source_project = source_ctx.env.project
-    build_project.deps = source_project.deps
-    build_project.compat = source_project.compat
-    return build_project
-end
-
 function gen_build_code(build_file::String)
     code = """
         $(Base.load_path_setup_code(false))
@@ -848,7 +839,7 @@ function build_versions(ctx::Context, uuids::Vector{UUID}; might_need_to_resolve
             else
                 pkgerror("Could not find either `git-tree-sha1` or `path` for package $name")
             end
-            version = v"0.0"
+            version = something(entry.version, v"0.0")
         end
         ispath(path) || error("Build path for $name does not exist: $path")
         ispath(buildfile(path)) && push!(builds, (uuid, name, path, version))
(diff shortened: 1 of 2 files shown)


DilumAluthge:
Travis failure is a network error.

DilumAluthge:
I wonder if there is a way to run NewPkgEval with this commit of Pkg, just to make sure that it solves the problem before we merge it into Pkg master.

@maleadt

maleadt:
Create a branch on JuliaLang/julia where Pkg is bumped accordingly (or somehow checked out to include this commit) and trigger a `runtests` on that branch its commit.

KristofferC:
> I wonder if there is a way to run NewPkgEval with this commit of Pkg, just to make sure that it solves the problem before we merge it into Pkg master.

There is a test in the PR?

fredrikekre:
Does this also fix  https://travis-ci.org/JuliaPackaging/BinaryBuilder.jl/jobs/626451620#L313 ? Looks like it is failing there because one of the deps does not have a project file rather than the tested package itself.

00vareladavid:
@fredrikekre yes, it should. I will add a test for the `build` case to be extra careful.

 To clarify, the general issue was that the sandbox calls `Pkg.develop` on the target, but `Pkg.develop` errors if it does not see a `Project.toml` file. With the changes in the PR, `develop` (when called through the sandbox) will just use the registry information if it does not see a `Project.toml` file.

KristofferC:
> will just use the registry information if it does not see a Project.toml file.

Why is there need for registry information? Everything is in the current Manifest?

00vareladavid:
Correct, there is no need to go into the registry. Latest commit fixes that

DilumAluthge:
>> I wonder if there is a way to run NewPkgEval with this commit of Pkg, just to make sure that it solves the problem before we merge it into Pkg master.

> There is a test in the PR?

Sure, but it would still be nice to make sure that this solves the problem we are seeing in NewPkgEval. Unit tests don’t always tell the whole story.

<sub>Sent with <a href="http://githawk.com">GitHawk</a></sub>

00vareladavid:
I now realize we can use `Pkg.resolve` instead of `Pkg.develop` and avoid a whole lot of special casing

00vareladavid:
Ok, I'm fairly happy with the patch now.

If we find no `test/Project.toml`, `gen_target_project` will normalize the "targets" format to `test/Project.toml` format. Additionally, if the target has no `Project.toml` file, we spoof one with the `deps` table matching the current manifest entry, and proceed normally.

fredrikekre:
Not needed in 1.3, right?

KristofferC:
No

PR #1555, opened 2019-12-18
make Pkg protocol opt-out instead of opt-in

StefanKarpinski:
Don't merge until after 1.4 branch.


KristofferC:
So the commit before this is the Pkg 1.4 branch point?

StefanKarpinski:
Sure, we’re supposed to be feature frozen for a while anyway. I already branched Julia bit Alex told me I did it wrong.

PR #1557, opened 2019-12-22
Do not ever attempt to set owner when extracting tarballs

staticfloat:
When extracting files as `root`, `tar` will attempt to set ownership, which doesn't make much sense when moving from one machine to another.  (Props to Stefan for calling this out in the `Tar.jl` readme at the exact same time as it was reported in an issue).

Should fix https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/555

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -330,7 +330,7 @@ function probe_platform_engines!(;verbose::Bool = false)
             elseif endswith(tarball_path, ".bz2")
                 Jjz = "j"
             end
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
It’s amazing the hoops you have to jump through to make command line `tar` suitable for the data transfer use case. That invocation is getting pretty gnarly.

EricForgy:
Hi,

Just checking. It looks like this was merged on Dec 22 and v1.3.1 is dated Dec 30. Does this mean this is fixed on v1.3.1?

00vareladavid:
Its not on 1.3.1 but it is in 1.4.0-rc1

StefanKarpinski:
It's marked for backport so it should be in 1.3.2.

PR #1561, opened 2019-12-23
hook up `develop` to `tiered_resolve`

00vareladavid:
it would definitely be good to have this in 1.4 since a lot of people are having version resolution issues now


00vareladavid:
Will merge soon. Honestly it was a bit of an oversight that this was not already included in #1330. It will probably be annoying to have different resolution strategies for `add` and `dev`.

Is it too late to get into 1.4? Not sure how much of a hassle it would be. I think on Pkg side it is enough to find the commit which corresponds with to [this hash](https://github.com/JuliaLang/julia/pull/34186/files), create a 1.4 branch, and cherry pick this commit. I'm not super clear on what to do beyond that.

Issue #1554, opened 2019-12-18
Artifact downloads should be it's own "step" and have more output


KristofferC:
With that I mean that the Pkg should output something like

```
Downloading Artifacts...
Installed MKL.lkgfhdlkghfdl
Installed ...
```

Right now, it just looks like Pkg has hung for largish artifacts
Also, since these downloads can be large, it would be good with a progress bar.

KristofferC:
I'll put a priority on this because the current situation is frankly unacceptable from a UI standpoint (with MKL being installed so easily now).

cc @staticfloat

KristofferC:
No longer really a priority since we have some indication now but it would be good to still improve this to make it fit in with the design of the rest of the Pkg output.

PR #1564, opened 2019-12-25
Force verbose downloads during `Pkg` operations

staticfloat:
This should add a progress bar (and some text about what is being downloaded) for #1554.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -772,7 +772,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
                  " Finally, run `Pkg.instantiate()` again.")
     end
     # Download artifacts for this here package before fast-exiting
-    Operations.download_artifacts([dirname(ctx.env.manifest_file)]; platform=platform, verbose=verbose)
+    Operations.download_artifacts(ctx, [dirname(ctx.env.manifest_file)]; platform=platform, verbose=verbose)
     Operations.is_instantiated(ctx) && return
 
     Types.update_registries(ctx)
@@ -812,7 +812,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     end
 
     # Ensure artifacts are installed for the dependent packages, and finally this overall project
-    Operations.download_artifacts(pkgs; platform=platform, verbose=verbose)
+    Operations.download_artifacts(ctx, pkgs; platform=platform, verbose=verbose)
 
     new_apply = Operations.download_source(ctx, pkgs)
     Operations.build_versions(ctx, union(UUID[pkg.uuid for pkg in new_apply], new_git); verbose=verbose)
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -724,6 +724,7 @@ function download_artifact(
     tarball_url::String,
     tarball_hash::Union{String, Nothing} = nothing;
     verbose::Bool = false,
+    quiet_download::Bool = false,
 )
     if artifact_exists(tree_hash)
         return true
@@ -742,7 +743,8 @@ function download_artifact(
         # hash.  This will be fixed in a future Julia release which will properly interrogate
         # the filesystem ACLs for executable permissions, which git tree hashes care about.
         try
-            download_verify_unpack(tarball_url, tarball_hash, dest_dir, ignore_existence=true, verbose=verbose)
+            download_verify_unpack(tarball_url, tarball_hash, dest_dir, ignore_existence=true,
+                                   verbose=verbose, quiet_download=quiet_download)
         catch e
             # Clean that destination directory out if something went wrong
             rm(dest_dir; force=true, recursive=true)
@@ -840,18 +842,21 @@ Ensures an artifact is installed, downloading it via the download information st
 function ensure_artifact_installed(name::String, artifacts_toml::String;
                                    platform::Platform = platform_key_abi(),
                                    pkg_uuid::Union{Base.UUID,Nothing}=nothing,
-                                   verbose::Bool = false)
+                                   verbose::Bool = false,
+                                   quiet_download::Bool = false)
     meta = artifact_meta(name, artifacts_toml; pkg_uuid=pkg_uuid, platform=platform)
     if meta === nothing
         error("Cannot locate artifact '$(name)' in '$(artifacts_toml)'")
     end
 
-    return ensure_artifact_installed(name, meta, artifacts_toml; platform=platform, verbose=verbose)
+    return ensure_artifact_installed(name, meta, artifacts_toml; platform=platform,
+                                     verbose=verbose, quiet_download=quiet_download)
 end
 
 function ensure_artifact_installed(name::String, meta::Dict, artifacts_toml::String;
                                    platform::Platform = platform_key_abi(),
-                                   verbose::Bool = false)
+                                   verbose::Bool = false,
+                                   quiet_download::Bool = false)
     hash = SHA1(meta["git-tree-sha1"])
 
     if !artifact_exists(hash)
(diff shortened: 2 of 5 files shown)


DilumAluthge:
I'm guessing that it is too late to get this into an RC for 1.3.1, but perhaps we could backport this to 1.3.2 or something?

KristofferC:
This seems a bit excessive maybe?

```
[ Info: No hash cache found
[ Info: Calculated hash 13fc17b97be41763b02cbb80e9d048302cec3bd3d446c2ed6e8210bddcd3ac76 for file /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_5fc40a/download_target.tar.bz2
[ Info: Unpacking /var/folders/nz/vv4_9tw56nv9k3tkvyszvw÷80000gn/T/jl_5fc40a/download_target.tar.bz2 into /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_5fc40a/target...
[ Info: Destination file /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_5fc40a/download_target.tar.bz2 already exists, verifying...
[ Info: Hash cache is consistent, returning true
[ Info: Destination directory /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_5fc40a/target already exists, returning
[ Info: Destination file /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_5fc40a/download_target.tar.bz2 already exists, verifying...
[ Info: Hash cache is consistent, returning true
[ Info: Destination directory /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_5fc40a/target already exists, returning
[ Info: Downloading https://github.com/staticfloat/small_bin/raw/f1a92f5eafbd30a0c6a8efb6947485b0f6d1bec3/socrates.tar.xz to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_d1OMig/download_target.tar.xz...

```

Is there any way to have something similar we have for the git clones, like: 

```
Downloading artifacts:
 MKL: [=====>                           ] 40%
```

staticfloat:
Right now, the progress bar is just being printed by `curl`.  The rest of the messages are useful when there is any possibility of corrupted state on disk, but for artifacts, it's true, it's not likely that this will happen since the tarballs are not cached.  I've threaded the `quiet_download` argument through a few more things now, so we can get just the downloading messages, which looks a lot neater.

For 1.5, I would prefer if we could instead use `HTTP.download()` and report progress through a callback function tailored to `Pkg`; We may want to modify the Artifacts API a bit to give us a list of all the things that need to be downloaded so that we can do parallel downloads or something like that as well.  Perhaps we can modify `HTTP.jl` to have a threadpool of workers, then we just run `ensure_all_artifacts_installed()` in parallel or something?

KristofferC:
> For 1.5, I would prefer if we could instead use HTTP.download() and report progress through a callback function tailored to Pkg; 

It is still a bit unclear to me how HTTP.jl should be added to stdlib (I don't think a full source code import is that good of an idea) but getting a nice progress bar should be pretty easy then, yes.

Would it be possible to use libcurl and something like https://curl.haxx.se/libcurl/c/CURLOPT_PROGRESSFUNCTION.html (which is pretty much how we do it with LibGit2 currently). But adding new binary deps might be annoying.

> Perhaps we can modify HTTP.jl to have a threadpool of workers, then we just run ensure_all_artifacts_installed() in parallel or something?

I don't think that is needed, just create a bunch of tasks with `@async` that does the downloading (similar to how we do it downloading packages)?

staticfloat:
> But adding new binary deps might be annoying.

Yeah, I would prefer to reduce these as much as possible.

>> Perhaps we can modify HTTP.jl to have a threadpool of workers, then we just run ensure_all_artifacts_installed() in parallel or something?

> I don't think that is needed, just create a bunch of tasks with @async that does the downloading (similar to how we do it downloading packages)?

This doesn't matter to this PR, but wouldn't that create a storm of downloads?  E.g. if you have 65 artifacts to download, you really only want ~4 of them downloading at a time, don't you?

KristofferC:
> E.g. if you have 65 artifacts to download, you really only want ~4 of them downloading at a time, don't you?

Yeah but you do something like

```
for i in 1:4
    @async while !isempty(downloads)
        download(popfirst!(downloads))
    end
end
```

KristofferC:
I see this in the logs now

```
[ Info: Downloading https://pkg.julialang.org/artifact/43563e7631a7eafae1f9f8d9d332e3de44ad7239 to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_AxhclR-download.gz...

curl: (22) The requested URL returned error: 404 Not Found

[ Info: Downloading https://github.com/staticfloat/small_bin/raw/master/socrates.tar.gz to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_Q2M9Ic-download.gz...

[ Info: Downloading https://pkg.julialang.org/artifact/43563e7631a7eafae1f9f8d9d332e3de44ad7239 to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_AoYZcP-download.gz...

curl: (22) The requested URL returned error: 404 Not Found

[ Info: Downloading https://github.com/staticfloat/small_bin/raw/master/socrates.tar.gz to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_1IVDcb-download.gz...

[ Info: Downloading https://pkg.julialang.org/artifact/43563e7631a7eafae1f9f8d9d332e3de44ad7239 to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_uO6lH9-download.gz...

curl: (22) The requested URL returned error: 404 Not Found

[ Info: Downloading https://github.com/staticfloat/small_bin/raw/master/socrates.tar.gz to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_YHRTWp-download.gz...

[ Info: Downloading https://pkg.julialang.org/artifact/43563e7631a7eafae1f9f8d9d332e3de44ad7239 to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_0SGRRx-download.gz...

curl: (22) The requested URL returned error: 404 Not Found

[ Info: Downloading https://github.com/staticfloat/small_bin/raw/master/socrates.tar.gz to /var/folders/nz/vv4_9tw56nv9k3tkvyszvwg80000gn/T/jl_cMNNxK-download.gz...

```

wonder what that is about.

Anyway, this should be an improvement over the status quo but could still be improved imo.
No need to show where we download the artifact for example since it is just to a temp dir.

KristofferC:
Is this really how the progress bar is supposed to look? Kinda trippy.

Alt Text

staticfloat:
That means the server didn’t give you a total length, so it doesn’t know
how to present progress.

KristofferC:
Alright, is that a problem with Pkg server?

StefanKarpinski:
Ah, probably, we should send the total content length as a header.

fredrikekre:
`Downloading artifacts...` is printed even though everything is downloaded already:
```
(@v1) pkg> up
Downloading artifacts...
 [no changes]
   Updating `~/.julia/environments/v1/Manifest.toml`
 [no changes]
```

KristofferC:
Hm, yeah. In general, I think it is usually good to have a "collection phase" where you just gather all urls you need to download, and then a "download phase" which does the downloading. That makes it easier to check if something actually need to be downloaded (as well as concurrently do downloads as well).

StefanKarpinski:
We should also figure out in advance the methods by which a thing should be attempted to be downloaded (Pkg protocol, GitHub release, GitHub tarball, artifact downloads, git clone). It would be good to have a phase (as Kristoffer says) where we figure out all the things to download and what methods to try for each one, and then another phase where we just do all the downloads, with a configurable throttle of how many downloads we try concurrently, regardless of method.

PR #1567, opened 2019-12-28
BinaryPlatforms: Add `armv6l` architecture

staticfloat:
It turns out we can actually build Julia for `armv6l` architectures (for use on things like the Raspberry Pi Zero W).  Unfortunately, I had been led to believe that the only 32-bit arm architecture Julia would ever run on would be `armv7l`, so we went for the shortcut of identifying tarballs as `arm-linux-gnueabihf` instead of the full `armv7l-linux-gnueabihf` (since many other projects did this as well).

It turns out, we should have just done the correct, if slightly harder thing, first.  So this PR changes us to emitting `armv7l-linux-gnueabihf` and `armv6l-linux-gnueabihf` by default, but still parsing `arm-linux-gnueabihf` as `armv7l`, so as to maintain backward compatibility (and compatibility with other systems).

### DIFF
--- a/src/BinaryPlatforms.jl
+++ b/src/BinaryPlatforms.jl
@@ -165,7 +165,7 @@ struct Linux <: Platform
                    libc::Union{Nothing,Symbol}=nothing,
                    call_abi::Union{Nothing,Symbol}=nothing,
                    compiler_abi::CompilerABI=CompilerABI())
-        if !in(arch, [:i686, :x86_64, :aarch64, :powerpc64le, :armv7l])
+        if !in(arch, [:i686, :x86_64, :aarch64, :powerpc64le, :armv6l, :armv7l])
             throw(ArgumentError("Unsupported architecture '$arch' for Linux"))
         end
 
@@ -178,8 +178,8 @@ struct Linux <: Platform
             throw(ArgumentError("Unsupported libc '$libc' for Linux"))
         end
 
-        # Auto-map the `call_abi` to be `eabihf` on armv7l
-        if call_abi === nothing && arch === :armv7l
+        # Auto-map the `call_abi` to be `eabihf` on armv6l/armv7l
+        if call_abi === nothing && arch in (:armv6l, :armv7l)
             call_abi = :eabihf
         end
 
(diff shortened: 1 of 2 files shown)


src/BinaryPlatforms.jl
```diff
@@ -477,7 +477,8 @@ function platform_key_abi(machine::AbstractString)
         :x86_64 => "(x86_|amd)64",
         :i686 => "i\\d86",
         :aarch64 => "aarch64",
-        :armv7l => "arm(v7l)?",
+        :armv7l => "arm(v7l)?", # if we just see `arm-linux-gnueabihf`, we assume it's `armv7l`
```
giordano: Maybe add a comment here about the why we need this for backward compatibility?

StefanKarpinski:
Should this be backported to 1.3 or is this too much of a feature?

staticfloat:
Let's not backport for now; it doesn't add much until we get the BB support and Julia builds.  We can backport in the future if armv6l support really takes off.

staticfloat:
Hmmm, actually, we might need to backport it to avoid warnings when Julia 1.3 parses an `Artifacts.toml` and finds platform keys that reference `armv6l`.

src/BinaryPlatforms.jl
```diff
@@ -178,22 +178,22 @@ struct Linux <: Platform
             throw(ArgumentError("Unsupported libc '$libc' for Linux"))
         end
 
-        # Auto-map the `call_abi` to be `eabihf` on armv7l
-        if call_abi === nothing && arch === :armv7l
+        # Auto-map the `call_abi` to be `eabihf` on armv6l/armv7l
+        if call_abi === nothing && arch in (:armv6l, :armv7l)
```
ararslan: You could define a helper function like
```julia
isarm32(arch::Symbol) = arch === :armv6l || arch === :armv7l
```
then replace all of the `arch in (:armv6l, :armv7l)`s with `isarm32(arch)`.

KristofferC:
Should this go into 1.4?

staticfloat:
I'd like to get this into 1.4, yes.  And also backported to 1.3.  I'll mark the appropriate labels.

KristofferC:
There probably won't be a new 1.3 but maybe. Merge?

staticfloat:
> There probably won't be a new 1.3 but maybe. Merge?

We need a 1.3.2 with https://github.com/JuliaLang/Pkg.jl/pull/1581 at least, since without it, git tree hashing breaks in some rare cases.

PR #1573, opened 2019-12-31
Check for "permissions-blindness" before calculating Artifacts git tree hashes

staticfloat:
We need this to be more general than just skipping Windows; if we are on
Linux but using an NFS share or FAT32 filesystems, (for instance) we can
run into the same issues.  So we instead probe for the ability to have a
sane `chmod()`/`isexecutable()` run, and if that fails, we do not bother
to calculate git tree hashes.  This should transparently start to
calculate the proper git tree hashes on Windows once
https://github.com/JuliaLang/julia/issues/33212 is fixed

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -716,6 +716,65 @@ function unbind_artifact!(artifacts_toml::String, name::String;
     return
 end
 
+_permissions_blind_filesystem_cache = Dict{String,Bool}()
+"""
+    permissions_blind_filesystem(path::AbstractString)
+
+Determine whether the given directory is a permissions-blind filesystem,
+such as a FAT32 volume, or an NFS mount with permissions-modifications.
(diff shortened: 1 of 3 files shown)


staticfloat:
Ah, we explicitly run some git tree hash checks, so the `Sys.isexecutable()` change here breaks things.

KristofferC:
What's the status of this?

staticfloat:
This is good to go; I reverted the change to make use of `Sys.isexecutable()` until https://github.com/JuliaLang/julia/pull/35625 is merged; once that's merged we can change the behavior here (although I suppose it should probably be version-gated)

staticfloat:
On further testing, it appears that when we use `7z` as an extraction engine, it doesn't save permissions properly anyway.  So let's go with the other option and just let the disk deal with things as it will.

Issue #1579, opened 2020-01-02
@artifact_str(artifactname) erroring: ERROR: UndefVarError: name not defined


ianshmean:
While `artifact"myartifact"` works, `@artifact_str(artifactname)` (with a variable arg for the name) errors with
```
ERROR: UndefVarError: name not defined
```
With the Pkg update this is now occurring in 1.3.1

cc. @staticfloat

PR #1580, opened 2020-01-02
Properly `esc()` variable in `@artifact_str`

staticfloat:
We need to properly escape a macro input variable.  This should fix https://github.com/JuliaLang/Pkg.jl/issues/1579

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -1029,7 +1029,7 @@ macro artifact_str(name)
         # Use invokelatest() to introduce a compiler barrier, preventing many backedges from being added
         # and slowing down not only compile time, but also `.ji` load time.  This is critical here, as
         # artifact"" is used in other modules, so we don't want to be spreading backedges around everywhere.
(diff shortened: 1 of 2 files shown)


ssfrr:
Note this still has the issue that the error message printed [here](https://github.com/JuliaLang/Pkg.jl/pull/1580/files#diff-a1e7cfbd723206272a351c94f21308deR1014-R1022) will list the artifact name as whatever the variable name is, rather than the actual name that variable is bound to (e.g. if the caller does `foo="myartifact"; @artifact_str(foo)` then the displayed name in the error message will be "foo", not "myartifact"). Doesn't seem like a big deal though.

staticfloat:
I would rather throw an error at compile time than at runtime there, so I don't think it's that big a deal.  It will work correctly when it's a literal string, and the variable name itself could be helpful, (especially since the error is that there's no `.toml` file to begin with) so I think it's okay.

ssfrr:
Yeah, that seems reasonable. Just wanted to bring it up in case

PR #1581, opened 2020-01-02
Include directories that contain only symlinks to other directories

staticfloat:
When hashing, we were accidentally excluding directories that contained
only symlinks that pointed to other directories.  This fixes that, and
adds a regression test

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -258,7 +258,9 @@ direct parent of a file or it contains some other directory that itself is a
 direct parent of a file. This is used to exclude directories from tree hashing.
 """
 function contains_files(path::AbstractString)
-    isdir(lstat(path)) || return true
+    st = lstat(path)
+    ispath(st) || throw(ArgumentError("non-existent path: $(repr(path))"))
+    isdir(st) || return true
     for p in readdir(path)
         contains_files(joinpath(path, p)) && return true
     end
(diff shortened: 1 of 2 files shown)


KristofferC:
Seems like this fail some hash tests.

src/GitTools.jl
```diff
@@ -251,19 +251,18 @@ function blob_hash(path::AbstractString, HashType = SHA.SHA1_CTX)
...
+direct parent of a file or it contains some other directory that itself is a
+direct parent of a file. This is used to exclude directories from tree hashing.
 """
-function contains_no_files(root::AbstractString)
-    for (root, dirs, files) in walkdir(root)
-        if !isempty(files)
-            return false
-        end
+function contains_files(path::AbstractString)
+    isdir(lstat(path)) || return true
```
StefanKarpinski:
```suggestion
    st = lstat(path)
    ispath(st) && !isdir(st) && return true
```
This handles non-existent paths by returning `false`, which seems like the right answer. Perhaps erroring if called on a non-existent path would be better though.

StefanKarpinski:
The new test was broken. I don't know what it was attempting to test, but if was making an absolute symlink to a temp path, which is going to change every time the test runs and therefore not have a stable git hash. I've changed the test to make `bar_link` a relative symlink to `bar` inside of `foo`. Since both `bar` and `bar_link` would have appeared as empty directories previously, this would have failed previously, now it correctly hashes `foo` as non-empty.

PR #1594, opened 2020-01-06
New verspec parser

bicycle1885:
This pull requests overhauls the implementation of the `Pkg.Types.semver_spec` function to resolve some inconsistent behaviours such as:
- `"<1.2.3"` is allowed but `">1.2.3"` is not.
- `"≥1.2.3"` and `">=1.2.3"` are allowed but `"≤1.2.3"` and `"<= 1.2.3"` are not.
- `"=  1.2.3"` and `"<  1.2.3"` are allowed but `"^  1.2.3"` and `"~  1.2.3"` are not.
- `"= vv1.2.3"` is allowed, which seems to be a bug.
- Trailing comma is not allowed (e.g., `"1, 2,"`).

The new implementation uses no regular expressions and thus it is easier to report more user-friendly error messages. The performance is also improved.

If you like the idea, I'm going to refine the code and add tests for new specifiers.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,4 @@
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 keywords = ["package management"]
 license = "MIT"
 desc = "The next-generation Julia package manager."
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -274,11 +274,61 @@ end
 function parse_spec1(spec::String, i::Int)
     # parse specifier
     i = skipws(spec, i)
+    next = parse_specifier(spec, i)
+    isnothing(next) && return
+    specifier, i = next
+
+    # parse version number
+    next = parse_vernum(spec, i)
+    isnothing(next) && error("found no version number")
+    ver, i = next
+    next = iterate(spec, i)
+    if isnothing(next) || next[1] == ','
+        isnothing(next) || (i = next[2])  # comsume comma
+        specifier == '?' && (specifier = '^')  # implicit caret
+    elseif isspace(next[1])
+        i = skipws(spec, next[2])
+        next = iterate(spec, i)
+        if isnothing(next) || next[1] == ','
+            isnothing(next) || (i = next[2])  # consume comma
+            specifier == '?' && (specifier = '^')  # implicit caret
+        elseif next[1] == '-'
+            specifier == '?' || error("invalid hyphen specifier syntax")
+            specifier = '-'
+            next = iterate(spec, next[2])
+            isnothing(next) && error("incomplete hyphen specifier")
+            isspace(next[1]) || error("no space after hyphen specifier")
+            i = next[2]
+        else
+            error("unrecognizable character: $(repr(next[1]))")
+        end
+    elseif next[1] == '-'
+        error("no space before hyphen specifier")
+    else
+        error("unrecognizable character: $(repr(next[1]))")
+    end
+
+    # if not a hyphen specifier, parsing ends.
+    specifier == '-' || return interpret_spec(specifier, ver), i
+
+    # parse version number after hyphen
+    i = skipws(spec, i)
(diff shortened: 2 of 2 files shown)


Project.toml
```diff
@@ -1,5 +1,4 @@
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
00vareladavid: I think you accidentally removed this

bicycle1885: Uh, right. Thank you.

bicycle1885:
A question regarding the hyphen specifier: why does it need space around it? This pull request follows the rule since it is explicitly declared in the docs but I don't know why it is required.

00vareladavid:
Because otherwise it can be confused with a prerelase version e.g. `1.3.4-DEV`. Requiring a space removes that ambiguity

bicycle1885:
I see. IMHO, it is a matter of stylistic recommendation. But it's not important here.

DilumAluthge:
> Because otherwise it can be confused with a prerelase version e.g. 1.3.4-DEV. Requiring a space removes that ambiguity

Yeah exactly.

> IMHO, it is a matter of stylistic recommendation.

Unfortunately it’s not just a style question. The prerelease format is part of the semver spec. If we allow hyphen specification without the spaces around the hyphen, then we’ll be violating semver.

bicycle1885:
We cannot write a version with a prerelease suffix in version specification of the compatibility section in `Project.toml` anyway. So, syntactically space characters are not required.

DilumAluthge:
> We cannot write a version with a prerelease suffix in version specification of the compatibility section in `Project.toml` anyway. So, syntactically space characters are not required.

Well... not yet. But because JLL packages have weird version numbers, we will soon need to allow you to specify build metadata in compat entries. And it would be weird to support build metadata but not prerelease suffixes. So I think we’ll end up supporting both prerelease suffixes and build metadata.

bicycle1885:
Ah, I got it. Thank you for clarification.

KristofferC:
> - "<1.2.3" is allowed but ">1.2.3" is not.
> - "≥1.2.3" and ">=1.2.3" are allowed but "≤1.2.3" and "<= 1.2.3" are not.

These are intentional. Upper bounds are given as non-inclusive while lower bounds are given as inclusive. If we start allowing `-dev` in version specifiers then `<1.2.3` would include `1.2.3-dev` which is probably not intended.

KristofferC:
It is great that error messages can be improved but the code here reads a little bit like low level, optimized parsing code (this is quite unlikely to ever be a performance hotspot).
The focus should imo be on the clarity of the implementation. Having to add functions like `parse_decimal` and `skipws` feels like this is operating on a lower level than needed.

Since the work here has already been done and this code probably won't have to change that much (it is unlikely we will much around too much with the compat specifiers) maybe it's fine to just put it in but it seems a little bit odd to me that we need a function that parses a decimal number in Pkg.

bicycle1885:
So, would it be better to remove `>` and `≤` specifiers? I think having these is good for completeness and I don't think they are harmful.

The implementation may be low level but I think it is straightforward to follow. Some people may think regular expressions are more cryptic than low-level parsing functions. I think it's kind of a matter of taste.

KristofferC:
> So, would it be better to remove > and ≤ specifiers? I think having these is good for completeness and I don't think they are harmful.

What I wrote about regarding `>` didn't make sense (I was writing about `<`...). Anyway, I think it is best to try to have things be as consistent as possible instead of having multiple ways of saying the same thing. So yes, I think it is best to remove them.

StefanKarpinski:
I think that people are going to expect a full set of inequality operators. Yes, it's often best to have only one way to write things, but that ship has already sailed when we added support for "semver syntax" instead of just using explicit ranges for everything: semver syntax has many different ways to write the exact same thing. When one inequality operator is allowed, people automatically assume the others will work. Having some work and others not is just irritating.

PR #1599, opened 2020-01-08
Update precompile script to block cloning of registry

### LINKED ISSUE #1320: The Pkg precompile script does a clone of the default registry
This can take a while making Julia slower than necessary to build. At the same time, we do want to precompile these paths since they are important for the latency of Pkg. Ideally, there would be some static, mini registry, that could be used.

Also, it is not ideal that we will precompile less in places that have no internet access. We should probably have a mini local registry, and a mini local package that we bundle with Pkg and use that in the precompilation script.

fredrikekre:
Update precompile script to properly block cloning of the General registry, fixes #1320.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -527,9 +527,9 @@ const precompile_script = """
     cd(tmp)
     empty!(DEPOT_PATH)
     pushfirst!(DEPOT_PATH, tmp)
-    # Prevent cloning registry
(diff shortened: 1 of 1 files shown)


fredrikekre:
Confirmed that this builds with Julia master.

KristofferC:
> Confirmed that this builds with Julia master.

Should also check that the script works as intended (#1618)

PR #1603, opened 2020-01-08
require handling

00vareladavid:
not sure if I have permissions to push to the original branch


00vareladavid:
@KristofferC feel free to merge either PR. I did it here because I couldn't push to your PR branch. I'm pretty sure that the error you were seeing was because the missing `copy` was causing unintended mutation.

KristofferC:
Great, thanks for fixing it.

PR #1604, opened 2020-01-08
telemetryinfo: function to show what's sent

StefanKarpinski:
cc @oxinabox


StefanKarpinski:
So that users can see what telemetry info we're sending.

PR #1607, opened 2020-01-09
add an API option to ignore compat

KristofferC:
Since this change is so small, I might as well probe what people think about this.

This adds an option that just turns off all compat (except the compat in your project) and gives you the latest versions of everything.

```jl
julia> Pkg.update()
   Updating `~/.julia/environments/v1.3/Project.toml`
 [no changes]
   Updating `~/.julia/environments/v1.3/Manifest.toml`
 [no changes]

julia> Pkg.update(; ignore_compat=true)
   Updating `~/.julia/environments/v1.3/Project.toml`
 [no changes]
   Updating `~/.julia/environments/v1.3/Manifest.toml`
  [68821587] ↑ Arpack_jll v3.5.0+2 ⇒ v3.7.0+0
  [a81c6b42] ↑ Compose v0.7.3 ⇒ v0.8.0
  [31c24e10] ↑ Distributions v0.21.12 ⇒ v0.22.0
  [ae029012] ↑ Requires v0.5.2 ⇒ v1.0.0
```

No docs yet because that will be the majority of the work for this PR and I want to see some opinions first.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -323,7 +323,7 @@ function resolve_versions!(ctx::Context, pkgs::Vector{PackageSpec})
     for pkg in pkgs
         compat = project_compatibility(ctx, pkg.name)
         v = intersect(pkg.version, compat)
-        if isempty(v)
+        if !ctx.ignore_compat && isempty(v)
             throw(Resolve.ResolverError(
                 "empty intersection between $(pkg.name)@$(pkg.version) and project compatibility $(compat)"))
         end
@@ -422,11 +422,13 @@ function deps_graph(ctx::Context, uuid_to_name::Dict{UUID,String}, reqs::Resolve
                             other_uuid in uuids || push!(uuids, other_uuid)
                         end
                     end
-                    for (vr, cd) in compat_data
-                        all_compat_u_vr = get_or_make!(all_compat_u, vr)
-                        for (name,vs) in cd
-                            # check conflicts??
-                            all_compat_u_vr[name] = vs
+                    if !ctx.ignore_compat
+                        for (vr, cd) in compat_data
+                            all_compat_u_vr = get_or_make!(all_compat_u, vr)
+                            for (name,vs) in cd
+                                # check conflicts??
+                                all_compat_u_vr[name] = vs
+                            end
                         end
                     end
                 end
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -335,6 +335,7 @@ Base.@kwdef mutable struct Context
     # the future. It currently stands as an unofficial workaround for issue #795.
     num_concurrent_downloads::Int = haskey(ENV, "JULIA_PKG_CONCURRENCY") ? parse(Int, ENV["JULIA_PKG_CONCURRENCY"]) : 8
     graph_verbose::Bool = false
+    ignore_compat::Bool = false
     currently_running_target::Bool = false
 end
 


StefanKarpinski:
Generally best to avoid double negatives. How about `compat=[true|false]` or `respect_compat=[true|false]`?

KristofferC:
You will only call this with `ignore_compat=true` though which I think is more "direct" than `respect_compat=false` because you are directly saying what to do with the "ignore compat" (the bool is almost unnecessary).

DilumAluthge:
If we want to avoid double negatives, how about something like `force_latest = true` or `override_compat = true`?

StefanKarpinski:
Nah, I guess this is ok.

DilumAluthge:
Bump. Is there any objection to merging this? Seems like it could be very useful.

ExpandingMan:
How hard would it be to (perhaps as an additional option) add a comprehensive set of warnings about what compats are no longer being respected?  I think the ideal situation would not be to permanently just always go around ignoring compats, but to make it extremely simple to resolve specific issues.  Seems this PR would get us halfway there, but the other half is transparency about what's happening.

KristofferC:
Not too hard, just go through all compat in the end and see what is incompatible. However, since I don't think that this should really be used I am personally not very interested in putting much more effort than this into it.

tpapp:
To second @cstjean, can you please elaborate on the use case for this?

KristofferC:
For people that complain about compat bounds.

tpapp:
My concern is that after an upgrade that ignores compat bounds, they would complain about broken software instead.

NightMachinary:
[**@tpapp**](https://github.com/tpapp) commented on [Aug 7, 2020, 5:14 PM GMT+4:30](https://github.com/JuliaLang/Pkg.jl/pull/1607#issuecomment-670498055 "2020-08-07T12:44:21Z - Replied by Github Reply Comments"):
> My concern is that after an upgrade that ignores compat bounds, they would complain about broken software instead.

I have been using Python a lot, and I have yet to see problems out of ignored compat bounds. Yet Julia's strict insistence on compat bounds, coupled with badly defined compats by package authors, is giving me trouble almost 1 in 3 times I install a package. I think a major problem is that when the package authors don't specify a compat bound, e.g., see [TerminalExtensions](https://github.com/Keno/TerminalExtensions.jl/blob/master/Project.toml), Pkg defaults to respecting their current used versions, and thus installing one package with old deps will cause a cascade of downgrades. My own philosophy is to always use the latest versions, and if sth breaks, stop using it or create a specific env for just that package. It's madness to do snowball-downgrading by default. See also my issue on [LanguageServer.jl](https://github.com/julia-vscode/LanguageServer.jl/issues/814), where they state that they use ugly hacks just to circumvent this behavior.

A much better approach to preserving compatibility between packages is to use Rich Hickey's philosophy; Don't change existing API in breaking ways. If you really need a breaking change, add a new API, and preserve the old one. Something as simple as `dosth` and `dosthv2`. You can even stop maintaining the old API, but don't just drop it to hell. This reduces maintenance costs by a lot, at the price of not introducing new goodies to old dogs. Which is fine, unless security is a major concern, which it is not in most usecases.

tpapp:
> Julia's strict insistence on compat bounds, coupled with badly defined compats by package authors, is giving me trouble almost 1 in 3 times

It is hard to address this without specifics.

> when the package authors don't specify a compat bound, e.g., see TerminalExtensions

Note that this particular package *does* have compat bounds (since the modules it uses are standard libraries, compat bounds on `julia` is sufficient). Generally, compat bounds are now [required](https://github.com/JuliaRegistries/RegistryCI.jl#automatic-merging-guidelines) for automatic merging into the General registry.

> much better approach to preserving compatibility between packages is to use Rich Hickey's philosophy; Don't change existing API in breaking ways

This has been discussed extensively on Github and Discourse, so I would rather not open that question here. Briefly, it is not the model Julia subscribes to.

About this PR: my expectation is that instead of complaints about compat bounds, package maintainers would get bug reports about a broken state that would require tracking down problems that happen mostly in *other* packages (since that's where compat would be turned off).

StefanKarpinski:
We DO suggest that packages follow semver like Julia itself does, which includes not making any breaking changes except in major versions. However, the reality of life is that people don't always do what you suggest they do.

DilumAluthge:
What if, when someone uses this feature, they get a warning message saying that it may break things, and it is their responsibility (not a package maintainer's) to solve any problems that arise when using this feature?

NightMachinary:
[**@tpapp**](https://github.com/tpapp) commented on [Aug 7, 2020, 8:31 PM GMT+4:30](https://github.com/JuliaLang/Pkg.jl/pull/1607#issuecomment-670588915 "2020-08-07T16:01:57Z - Replied by Github Reply Comments"):
> Note that this particular package _does_ have compat bounds (since the modules it uses are standard libraries, compat bounds on `julia` is sufficient). 

Yes, I had mistaken the source of the issue. I edited my comment to point the real culprit, but unfortunately, it doesn't get reflected in emails. (Is it a good practice to reply to comments instead of editing them to make sure people get the update in their email clients?)

> Generally, compat bounds are now [required](https://github.com/JuliaRegistries/RegistryCI.jl#automatic-merging-guidelines) for automatic merging into the General registry.

Perhaps this issue will get dissolved after a year or two then. That (latest) problematic package that
broke my setup is actually fine if I use `add ...#master`, but it's stable package version uses bad bounds.

> About this PR: my expectation is that instead of complaints about compat bounds, package maintainers would get bug reports about a broken state that would require tracking down problems that happen mostly in other packages (since that's where compat would be turned off).

I don't quite understand you here. Can you elaborate further? My impression is that ignoring compat bounds will mostly harm unmaintained packages that haven't updated their deps?

A workaround for this problem might be to add a issue reporting helper tool to Pkg, that outputs all relevant info (OS version, Julia version, Python used by PyCall, ...), and include the ignored bounds in this output. When people open issues, this will be a clear hint that they have ignored bounds, and the author can ask them to try to reproduce under strict bounds. These kinds of stuff are standard practice in, e.g., reporting Emacs issues. The author will ask you to reproduce under a clean env (`emacs -Q`), if they suspect that other packages are the source of the problem.

KristofferC:
I mostly did this PR to have something to point to when people complain that they are not getting the latest version. They can use this option and when everything breaks, they will reach enlightenment.

tpapp:
> I don't quite understand you here. Can you elaborate further? My impression is that ignoring compat bounds will mostly harm unmaintained packages that haven't updated their deps?

Not necessarily. A maintained package may not update their `[compat]` section to the very latest versions for a while for valid reasons (eg it takes time to update & test code and no one got around to it yet). Please note that "maintained" != "everything happens instantly".

> when everything breaks, they will reach enlightenment.

I wish it was so, but in practice I think some people will complain vocally. Being able to say "told you so" doesn't really help anyone in this situation.

That said, if this is one of those pedagogical Zen PRs, it is a nice one.

src/Operations.jl
```diff
@@ -422,11 +422,13 @@ function deps_graph(ctx::Context, uuid_to_name::Dict{UUID,String}, reqs::Resolve
                             other_uuid in uuids || push!(uuids, other_uuid)
                         end
                     end
-                    for (vr, cd) in compat_data
-                        all_compat_u_vr = get_or_make!(all_compat_u, vr)
-                        for (name,vs) in cd
-                            # check conflicts??
-                            all_compat_u_vr[name] = vs
+                    if !ctx.ignore_compat
```
DilumAluthge, line 425:
```suggestion
                    if ctx.ignore_compat
                        @warn "You are using the ignore_compat feature. Use of this feature may cause packages to break. It is the responsibility of the user (not the maintainers of any packages) to solve any problems that arise from the use of this feature." maxlog = 1
                    else
```

DilumAluthge:
Needs a rebase?

PR #1611, opened 2020-01-10
Some more compat with 1.3

KristofferC:
From looking at PkgEval results: https://docs.google.com/spreadsheets/d/1ngalhKl9FEplWCWAWOfVKamZj-pkt4157pFGimb5wCw/edit#gid=609759912

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -307,6 +307,7 @@ The result is a `Dict` that maps a package UUID to a `PackageInfo` struct repres
 |:------------------|:-----------------------------------------------------------|
 | name              | The name of the package                                    |
 | version           | The version of the package (this is `Nothing` for stdlibs) |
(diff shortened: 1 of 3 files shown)


src/Pkg.jl
```diff
@@ -307,6 +307,7 @@ The result is a `Dict` that maps a package UUID to a `PackageInfo` struct repres
 |:------------------|:-----------------------------------------------------------|
 | name              | The name of the package                                    |
 | version           | The version of the package (this is `Nothing` for stdlibs) |
+| in_project        | The package is in the project file (a direct dependency)   |
```
00vareladavid: can `in_project` be `is_direct_dep` (or similar) instead? it would be consistent with all the other `is_*` boolean fields. It is also more precise because technically test deps can also be in the project file

KristofferC: yeah, makes sene,

Issue #1618, opened 2020-01-16
First Pkg.status is much slower on 1.4 than 1.3


KristofferC:
Likely due to changes in precompile script?

1.3:

```
time julia -e 'import Pkg; Pkg.status()'
julia -e 'import Pkg; Pkg.status()'  0.42s user 0.15s system 162% cpu 0.348 total
```

1.4 (note not master because that doesn't have updated Pkg):

```
➜  MyPackage git:(master) time ~/julia/julia -e 'import Pkg; Pkg.status()'
~/julia/julia -e 'import Pkg; Pkg.status()'  2.94s user 0.20s system 110% cpu 2.847 total
```

KristofferC:
This is the bad line:

https://github.com/JuliaLang/Pkg.jl/blob/f0fc6c9cc683967acccc9132481403789173a75a/src/Pkg.jl#L564

Breaks the precompile script:

```
julia>
#### inputting statement: ####
"    open(io -> print(io, \"name = \"Registry\""
####
julia>     open(io -> print(io, "name = "Registry"
ERROR: syntax: cannot juxtapose string literal
Stacktrace:
 [1] top-level scope at REPL[19]:0

julia>
#### inputting statement: ####
"uuid = \"37c07fec-e54c-4851-934c-2e3885e4053e\""
####
julia> uuid = "37c07fec-e54c-4851-934c-2e3885e4053e"
"37c07fec-e54c-4851-934c-2e3885e4053e"

julia>
#### inputting statement: ####
"repo = \"https://github.com/JuliaRegistries/Registry.git\""
####
julia> repo = "https://github.com/JuliaRegistries/Registry.git"
"https://github.com/JuliaRegistries/Registry.git"

julia>
#### inputting statement: ####
"[packages]"
####
julia> [packages]
ERROR: UndefVarError: packages not defined
Stacktrace:
 [1] top-level scope at REPL[23]:1
```

PR #1619, opened 2020-01-16
fix precompile script

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/1618.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -561,7 +561,7 @@ const precompile_script = """
     pushfirst!(DEPOT_PATH, tmp)
     # Prevent cloning the General registry by adding a fake one
     mkdir("registries"); mkdir("registries/Registry")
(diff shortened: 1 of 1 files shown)


src/Pkg.jl
```diff
@@ -573,7 +573,8 @@ const precompile_script = """
     foo_hash = Pkg.Artifacts.create_artifact(dir -> touch(joinpath(dir, "foo")))
     Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo", foo_hash)
     # Also create multiple platform-specific ones because that's a codepath we need precompiled
-    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=platform_key_abi())
+    using Pkg.BinaryPlatforms
+    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=Pkg.Artifacts.platform_key_abi())
```
```suggestion
    Pkg.Artifacts.bind_artifact!("./Artifacts.toml", "foo_plat", foo_hash; platform=Pkg.BinaryPlatforms.platform_key_abi())
```

KristofferC: Forgot to push the change where I removed the prefix over just `using BinaryPlatforms`.

Issue #1600, opened 2020-01-08
Pkg unconditionally prints "Downloading artifacts..."


KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/1564#issuecomment-571973313

KristofferC:
Bump @staticfloat

PR #1620, opened 2020-01-16
make artifact download more similar to rest of Pkg "theme"

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/1600, fixes https://github.com/JuliaLang/Pkg.jl/issues/1554

I didn't really like how the artifact download looked for some reasons:

- It printed `Downloading artifacts...` unconditionally.
- It used an `@info` log to print download url and location which showed some temp directory and (for a user) garbage hash name
- It kept the progress bar even after the download was finished.

This is a (arguably) ugly attempt to remedy this. The download now looks like:

Peek 2020-01-16 15-53

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -5,7 +5,7 @@ import ..depots1, ..depots, ..set_readonly
 import ..GitTools
 using ..BinaryPlatforms
 import ..TOML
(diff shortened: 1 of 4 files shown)


staticfloat:
You are clearly better at making beautiful terminal UIs than I am.  👍

PR #1623, opened 2020-01-21
Misc doc changes

fredrikekre:


fredrikekre:
First commit should be backported to the same branches as https://github.com/JuliaLang/Pkg.jl/pull/1567

PR #1626, opened 2020-01-22
make telemetry file writing more likely to be atomic

StefanKarpinski:
Alternative to https://github.com/JuliaLang/Pkg.jl/pull/1625. Write the telemetry file to a temp file and then move it into place.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -741,14 +741,15 @@ function load_telemetry_file(file::AbstractString)
         delete!(info, "ci_variables")
         changed = true
     end
-    if changed
-        mkpath(dirname(file))
-        mktemp() do tmp, io
-            TOML.print(io, info, sorted=true)
-            mv(tmp, file, force=true)
-        end
+    changed || return info
+    # write telemetry file atomically (if on same file system)
+    mkpath(dirname(file))
+    mktemp() do tmp, io
+        TOML.print(io, info, sorted=true)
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
@fredrikekre do you have a situation where this was causing problems so we can see if this fixes it? Or should I just merge this and we can see if the problems persist?

fredrikekre:
Btw, this doesn't solve what https://github.com/JuliaLang/Pkg.jl/pull/1625 tried to solve -- multiple tasks can still try to read the file at the same time, which was the problem.

StefanKarpinski:
How doesn’t it? It writes the file in a temporary location and then moves it into place only once it’s complete. Assuming the mv operation is atomic, it should make it impossible to read a half-written file.

StefanKarpinski:
However, we could apply that fix as well since it’s a much more graceful way to synchronize between different tasks in the same process. The temp file maneuver should fix the interprocess issue.

fredrikekre:
The problem wasn't that a task read a file that was partially written by another, but when there was already a broken file on disk that multiple files read. But yea, good to have a look anyway I think, since otherwise all tasks would generate their own `client_uuid` and all tasks would display the legal notice etc.

StefanKarpinski:
Clearly I didn't understand what your fix was for. Sorry about that. I included a version of your PR in https://github.com/JuliaLang/Pkg.jl/pull/1871 so that should address the problem in any case.

PR #1627, opened 2020-01-23
tweak the code for getting the registry from PkgServer

### LINKED ISSUE #1622: Big try catch in code for getting the registry from Pkg server
The code:

https://github.com/JuliaLang/Pkg.jl/blob/f78c1e1ad6d4ade6d72ab6b283f8c9b65a879c2a/src/Types.jl#L862-L882

has a very big try catch block.  This could easily disquise typos which might go unnoticed on CI since it will fallback to getting the registry from git. 

It would be better if the try catch could only enclose the actual `download` call, and also perhaps check the type of the exception.

cc @StefanKarpinski

KristofferC:
Fixes #1622 and https://github.com/JuliaLang/Pkg.jl/issues/1622#issuecomment-577614170.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -863,22 +863,24 @@ function pkg_server_registry_url(uuid::UUID)
     server = pkg_server()
     server === nothing && return nothing
     probe_platform_engines!()
-    hash = nothing
+    tmp_path = tempname()
+    download_ok = false
     try
-        mktemp() do tmp_path, io
-            download("$server/registries", tmp_path, verbose=false)
-            for line in eachline(io)
-                if (m = match(r"^/registry/([^/]+)/([^/]+)$", line)) !== nothing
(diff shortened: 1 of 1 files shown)


src/Types.jl
```diff
@@ -863,22 +863,24 @@ function pkg_server_registry_url(uuid::UUID)
...
-    hash === nothing ? nothing : "$server/registry/$uuid/$hash"
+    if download_ok
+        for line in eachline(tmp_path)
+            if (m = match(r"^/registry/([^/]+)/([^/]+)$", line)) !== nothing
+                uuid == UUID(m.captures[1]) || continue
+                hash = String(m.captures[2])
+                return "$server/registry/$uuid/$hash"
+            end
+        end
+    end
```
StefanKarpinski:
```suggestion
    end
    rm(tmp_path, force=true)
```

PR #1628, opened 2020-01-23
[WIP] [RFC] more complete support for unregistered dependencies

### LINKED ISSUE #492: Store some location infor for deps in Project.toml
Currently `Project.toml` is not enough to use a package, unless all it's deps can be found and installed from registries. I think it would be nice to include some information in `Project.toml` about where to find the deps. In particular two usecases come to mind:

1. Projects which depend on unregistred packages could put a url to the package
```
name = "Foo"
uuid = "4bdd1c6b-68bb-4dfe-a423-fa0d59eb7384"
version = "0.1.0"

[deps]
Bar = "ce02910b-3083-40c5-958c-8fd1036218f1"

[where]
Bar = "https://www.bar.com"
```
2. Projects with deps in the same repo can put the (relative) path
```
name = "Foo"
uuid = "4bdd1c6b-68bb-4dfe-a423-fa0d59eb7384"
version = "0.1.0"

[deps]
Bar = "ce02910b-3083-40c5-958c-8fd1036218f1"

[where]
Bar = "modules/Bar"
```

This would make `Project.toml` a bit more standalone; I can give that to someone and things will work, but if I want that someone to obtain the exact state I have, then I can also send `Manifest.toml`.

### LINKED ISSUE #1005: Developing local packages can be a bit annoying

00vareladavid:
Still WIP, but opening early for feedback. Basic tests are passing.

TLDR This adds a `source` field in the active project for any packages we are tracking by repo or by path. On `add`/`dev`/`instantiate` we crawl over any unregistered dependencies and download as necessary. After we collect all unregistered dependencies, we continue with `resolve` normally.

close #492 close #1005 supersedes #1088


tkf:
How about looking into all manifest files checked in to the repository?  You can use them if they include the "root" project as a dependency (which then implies that this Manifest file includes the dependencies of the "root" project).

I implemented it in [`Rogue.add`](https://tkf.github.io/Rogue.jl/dev/#Rogue.add-Tuple{AbstractString}) and it is quite useful.

src/Operations.jl
```diff
@@ -1065,6 +1079,45 @@ function _resolve(ctx::Context, pkgs::Vector{PackageSpec}, preserve::PreserveLev
...
+                new_pkg = Packagespec(;name=name, uuid=uuid, repo=entry.repo, tree_hash=entry.tree_hash)
+                Types.handle_repo_add!(ctx, new_pkg) && push!(new_git, new_pkg.uuid)
+            end
+        elseif (source = get(src_project.source, name, nothing)) !== nothing
+            new_pkg = PackageSpec(;repo=GitRepo(;source=source))
+            Types.handle_repo_add!(ctx, new_pkg) && push!(new_git, new_pkg.uuid)
+        end
+        if new_pkg !== nothing
+            push!(unreg, new_pkg)
+            collect_unregistered!(ctx, new_pkg, new_git, unreg)
```
tkf: I wonder why you have to recurse? In principle, the manifest of `pkg` is likely to contain all URLs for unregistered dependencies, isn't it? Otherwise, such manifest cannot be instantiated.

KristofferC: What about when you do an add on a url and that package has dependencies on other unregistered packages?

tkf: In this case, I guess you'd need to parse the manifest file(s) contained in the newly to-be-added package?  But what I was commenting was that, if this new package has a manifest file, it is likely that you don't need to parse the manifest files of _dependencies_ of this newly added package.  Or at least you can try to minimize the I/O by stop recursing if all direct and transitive dependencies are found.  It could be an unnecessary premature optimization but don't need something like this to prevent infinite recursion anyway?

00vareladavid: @tkf Perhaps you missed it, but this PR also introduces a `source` field (as discussed in #492). So there could be no manifest files involved.

tkf: Thanks for the clarification. I was confused since I came to this PR via another PR #1088.

00vareladavid:
Getting very close now.

Here is an example of toggling between tracking a repo and tracking a path:
```
(demo) pkg> add https://github.com/00vareladavid/Unregistered.jl
   Updating git-repo `https://github.com/00vareladavid/Unregistered.jl`
  Resolving package versions...
   Updating `/tmp/demo/Project.toml`
   dcb67f36 + Unregistered v0.2.0 `https://github.com/00vareladavid/Unregistered.jl#master`
   Updating `/tmp/demo/Manifest.toml`
   7876af07 + Example v0.5.3
   dcb67f36 + Unregistered v0.2.0 `https://github.com/00vareladavid/Unregistered.jl#master`

(demo) pkg> dev --local Unregistered
    Cloning git-repo `https://github.com/00vareladavid/Unregistered.jl`
  Resolving package versions...
   Updating `/tmp/demo/Project.toml`
   dcb67f36 ~ Unregistered v0.2.0 `https://github.com/00vareladavid/Unregistered.jl#master` ⇒ v0.2.0 `dev/Unregistered`
   Updating `/tmp/demo/Manifest.toml`
   dcb67f36 ~ Unregistered v0.2.0 `https://github.com/00vareladavid/Unregistered.jl#master` ⇒ v0.2.0 `dev/Unregistered`

(demo) pkg> add Unregistered
   Updating git-repo `https://github.com/00vareladavid/Unregistered.jl`
  Resolving package versions...
   Updating `/tmp/demo/Project.toml`
   dcb67f36 ~ Unregistered v0.2.0 `dev/Unregistered` ⇒ v0.2.0 `https://github.com/00vareladavid/Unregistered.jl#master`
   Updating `/tmp/demo/Manifest.toml`
   dcb67f36 ~ Unregistered v0.2.0 `dev/Unregistered` ⇒ v0.2.0 `https://github.com/00vareladavid/Unregistered.jl#master`
```

And adding a package with multiple unregistered dependencies:
```
(demo) pkg> add https://github.com/00vareladavid/UnregisteredA
   Updating git-repo `https://github.com/00vareladavid/UnregisteredA`
   Updating git-repo `https://github.com/00vareladavid/UnregisteredD`
   Updating git-repo `https://github.com/00vareladavid/UnregisteredB`
   Updating git-repo `https://github.com/00vareladavid/UnregisteredC`
  Resolving package versions...
   Updating `/tmp/demo/Project.toml`
   b686e103 + UnregisteredA v0.1.0 `https://github.com/00vareladavid/UnregisteredA#master`
   Updating `/tmp/demo/Manifest.toml`
   b686e103 + UnregisteredA v0.1.0 `https://github.com/00vareladavid/UnregisteredA#master`
   4a15487d + UnregisteredB v0.1.0 `https://github.com/00vareladavid/UnregisteredB#master`
   576bedce + UnregisteredC v0.1.0 `https://github.com/00vareladavid/UnregisteredC#master`
   295405fa + UnregisteredD v0.1.0 `https://github.com/00vareladavid/UnregisteredD#master`
```

KristofferC:
When you think this is ready, it would be good with a bit of a higher level description of what this implements and how it works etc. Just to aid in reviewing it :)

00vareladavid:
The way that this currently works is that, whenever we track a dependency by path or by repo, we add the source url to the `source` table in the project file. This solves the basic problem of storing the location for unregistered packages.

Whenever Pkg tracks an package by repo/path, we read its project file and see if it declares any dependencies on unregistered packages by checking its `source` table. We add any unregistered packages to the dependency graph and recurse until we have collected all reachable unregistered dependencies. Once we have collected unregistered dependencies, their versions are fixed, and we can continue resolving the rest of the dependency graph as normal.

When collecting unregistered packages, if the declared source of the package is a local directory and that directory is not a git repo, we fall back to tracking the dependency by path. This fallback allows #1005 to integrate seamlessly with the rest of the scheme.

00vareladavid:
I will rebase this soon. Any feedback for this, even if high level?

StefanKarpinski:
My feedback is "doooo it".

00vareladavid:
The goal for this PR is to make the use of unregistered dependencies as seamless as possible. Relying on the presence of manifests has two problems: (1) people have to remember to commit it and (2) we already use the presence of a manifest as a signal to `Pkg.test` to use that dependency graph.

fredrikekre:
Alright. In which cases is Pkg "allowed" to look at the source field? E.g. I don't think it should be allowed when sticking to registered versions for example.

KristofferC:
> Alright. In which cases is Pkg "allowed" to look at the source field?

Whenever tracking a package by path or url maybe?

fredrikekre:
Yea, just wanted to make sure that is how this is implemented. Also; why not store the source for all packages https://github.com/JuliaLang/Pkg.jl/issues/378 ?

KristofferC:
That's in the manifest though. But yeah, that's a good idea as well. Also, I notice that Cargo stores the URL to the registry where the information for the package was retrieved instead of the URL to the git repo. Something to think about perhaps.

casabre:
Any progress here? I would love to see that feature backed in the next release :+1:

TheCedarPrince:
Bump - any more progress or work here? Saw the checks did not finish...

evanphilip:
bors try

kmsherbertvt:
Was this PR supposed to have been superseded by #3783?

**On the one hand:**
Issue #492, which seems to have been requesting the same thing this PR seeks to address, has been closed, and @KristofferC asserted there that #3783 solved it.

**On the other hand:**
A couple people pointed out in #492 that the [`[sources]` documentation](https://pkgdocs.julialang.org/v1/toml-files/#The-[sources]-section) indicates it is only used in active environments, _not_ when tracing dependencies:

> Note that this information is only used when this environment is active, i.e. it is not used if this project is a package that is being used as a dependency.

Moreover, I just tried it and I still get the "no known versions" error for an unregistered dependency I've always gotten, when trying to add an unregistered package. (And the most recent mention of this PR seems to be due to another group encountering the same problem.)

PR #1630, opened 2020-01-24
Authentication Hooks

pfitzseb:
This PR adds a hook into `get_auth_header` which is called when Pkg can't find authentication for the current package server.

#### Public API (unexported):
```    
    register_auth_error_handler(f)

Registers `f` as the topmost handler for failures in package server authentication.
`f` must be a function that takes three input arguments `(url, pkgserver, err)`, where `url` is the
URL currently being downloaded, `pkgserver = Pkg.pkg_server()` the current package server, and
`err` is one of `no-auth-file`, `insecure-connection`, `malformed-file`, `no-access-token`,
`no-refresh-key`, `insecure-refresh-url`,  `malformed-file`, or `no-access-token`.
The handler `f` needs to return a tuple of `Bool`s `(handled, should_retry)`. If `handled` is `false`,
the next handler in the stack will be called, otherwise `get_auth_header` is called again if `should_retry`
is `true`.
`register_auth_error_handler` returns a zero-arg function that can be called to deregister the handler.
```
```
    deregister_auth_error_handler(f)

Removes `f` from the stack of authentication error handlers.
```

#### Example handler:
```
julia> dispose = Pkg.PlatformEngines.register_auth_error_handler((url, svr, err) -> begin
                PkgAuth.authenticate(svr*"/auth")
                return true, true
              end)
```

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -617,7 +617,7 @@ function handle_auth_error(url, err; verbose::Bool = false)
 end
 
 function register_auth_error_handler(f)
-    pushfirst!(AUTH_ERROR_HANDLERS, f)
+    unique!(pushfirst!(AUTH_ERROR_HANDLERS, f))
     return () -> deregister_auth_error_handler(f)
 end
 
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -609,19 +609,43 @@ const AUTH_ERROR_HANDLERS = []
 function handle_auth_error(url, err; verbose::Bool = false)
     handled, should_retry = false, false
     for handler in AUTH_ERROR_HANDLERS
-        handled, should_retry = handler(url, err)
+        handled, should_retry = handler(url, Pkg.pkg_server(), err)
         handled && break
     end
     handled && should_retry && return get_auth_header(url; verbose = verbose)
     return nothing
 end
 
+"""
+    register_auth_error_handler(f)
+
+Registers `f` as the topmost handler for failures in package server authentication.
+
+`f` must be a function that takes three input arguments `(url, pkgserver, err)`, where `url` is the
+URL currently being downloaded, `pkgserver = Pkg.pkg_server()` the current package server, and
+`err` is one of `no-auth-file`, `insecure-connection`, `malformed-file`, `no-access-token`,
+`no-refresh-key`, `insecure-refresh-url`,  `malformed-file`, or `no-access-token`.
+
+The handler `f` needs to return a tuple of `Bool`s `(handled, should_retry)`. If `handled` is `false`,
+the next handler in the stack will be called, otherwise `get_auth_header` is called again if `should_retry`
+is `true`.
+
+`register_auth_error_handler` returns a zero-arg function that can be called to deregister the handler.
+"""
 function register_auth_error_handler(f)
     unique!(pushfirst!(AUTH_ERROR_HANDLERS, f))
     return () -> deregister_auth_error_handler(f)
 end
 
-deregister_auth_error_handler(f) = filter!(handler -> handler !== f, AUTH_ERROR_HANDLERS)
+"""
+    deregister_auth_error_handler(f)
+
+Removes `f` from the stack of authentication error handlers.
+"""
+function deregister_auth_error_handler(f)
+    filter!(handler -> handler !== f, AUTH_ERROR_HANDLERS)
+    return nothing
+end
 
 function get_auth_header(url::AbstractString; verbose::Bool = false)
     server_dir = get_server_dir(url)
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -608,8 +608,9 @@ const AUTH_ERROR_HANDLERS = []
 
 function handle_auth_error(url, err; verbose::Bool = false)
     handled, should_retry = false, false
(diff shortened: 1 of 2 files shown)


KristofferC:
Test would be good

pfitzseb:
Added some basic tests. Anything more elaborate would require a mocked up pkg server, I think.

src/PlatformEngines.jl
```diff
@@ -604,27 +604,70 @@ function get_server_dir(url::AbstractString)
...
+"""
+function deregister_auth_error_handler(f)
+    filter!(handler -> handler !== f, AUTH_ERROR_HANDLERS)
+    return nothing
+end
+
 function get_auth_header(url::AbstractString; verbose::Bool = false)
     server_dir = get_server_dir(url)
-    server_dir === nothing && return
+    server_dir === nothing && return handle_auth_error(url, "no-auth-file"; verbose=verbose)
```
StefanKarpinski: Slight problem here: this should not call `handle_auth_error` when the request doesn't need authentication, i.e. when `pkg_server() == nothing` or when `!startswith(url, server)`. A few ways I can think to handle it:
1. factor the part where `get_server_dir` checks those two things out and handle them separately here;
2. change `get_server_dir` to return `Union{String, Bool}` where `false` means "this request doesn't need authentication" and `true` means "this request needs authentication but doesn't have an auth directory yet";
3. change `handle_auth_error` to check `pkg_server()` and `startswith(url, server)` again an bail out if the request doesn't need to be authenticated.

Unfortunately, they're all a bit messy.

pfitzseb: My thought was that whatever handler is registered also has the responsibility to check whether it's applicable (e.g. by checking the `Pkg.pkg_server()` argument).

StefanKarpinski: That seems like something that should be done once, not by each handler. If you're not talking to the current `pkg_server` then you shouldn't be sending auth headers at all.

pfitzseb: Fair enough. Imho the best solution would be to register a pair of `Union{Regex, String}, Function` and check whether any given handler is applicable with `occursin`.

pfitzseb: Should be resolved with the current design.

src/PlatformEngines.jl
```diff
@@ -604,27 +604,74 @@ function get_server_dir(url::AbstractString)
...
+
+Registers `f` as the topmost handler for failures in package server authentication.
+
+A handler is only invoked if `occursin(urlscheme, url)` is true (where `url` is the URL Pkg
+is currently trying to download.)
+
+`f` must be a function that takes three input arguments `(url, pkgserver, err)`, where `url` is the
+URL currently being downloaded, `pkgserver = Pkg.pkg_server()` the current package server, and
+`err` is one of `no-auth-file`, `insecure-connection`, `malformed-file`, `no-access-token`,
+`no-refresh-key`, `insecure-refresh-url`,  `malformed-file`, or `no-access-token`.
```
StefanKarpinski:
```suggestion
`no-refresh-key` or `insecure-refresh-url`.
```

src/PlatformEngines.jl
```diff
@@ -604,27 +604,74 @@ function get_server_dir(url::AbstractString)
...
+
+The handler `f` needs to return a tuple of `Bool`s `(handled, should_retry)`. If `handled` is `false`,
+the next handler in the stack will be called, otherwise `get_auth_header` is called again if `should_retry`
```
StefanKarpinski:
```suggestion
the next handler in the stack will be called, otherwise handling terminates; `get_auth_header` is called again if `should_retry`
```

src/PlatformEngines.jl
StefanKarpinski: Was resolved by reverting the first `return` not to call any handlers.

StefanKarpinski:
Now failing tests, but the code itself looks good to me.

pfitzseb:
Tests are working once again, but I have no clue why [this](https://github.com/JuliaLang/Pkg.jl/pull/1630/commits/76f5efc1895695520e27c39ce6c2454ac5edd9a5) is necessary.

KristofferC:
Would was the error without it? What made you put it there? Would be good to figure that out. It should also probably be withing a try finally.

pfitzseb:
The error was 
```
Authentication Header Hooks: Error During Test at /Users/travis/build/JuliaLang/Pkg.jl/test/platformengines.jl:242

  Test threw exception

  Expression: PlatformEngines.get_auth_header("https://foo.bar/baz") == nothing

  no depots found in DEPOT_PATH
```

KristofferC:
Should be backported?

StefanKarpinski:
It think it's not necessary. If we backport it then it will be possible to add hooks to vanilla Julia 1.4 (as opposed to JuliaPro) so that it can automatically acquire an auth.toml file via a nice workflow, but it won't work out of the box anyway since users would need to setup the hooks themselves. It seems just as easy to instruct anyone using vanilla Julia 1.4 to download an `auth.toml` file to the right location manually. In Julia 1.5, I'd like to bake the whole process in (I have a plan to make this fairly generic and independent of the server's URL schema), so that vanilla Julia 1.4 can follow a nice workflow to get an auth.toml file from the server in a safe way, but that requires switching Pkg to using HTTP, which is clearly too much to backport.

PR #1635, opened 2020-01-27
relax project file validation

### LINKED ISSUE #1571: relax the project file validation
Currently, if you're working on a project without a UUID, you get this error:
```
project appears to be a package but has no uuid
```
This is particularly annoying when trying to test Pkg itself, which requires commenting out the `uuid` in order to load the dev version of Pkg instead of the built-in one. The logic also seems wrong: a project can have a name without being a package and a non-package is allowed to have no UUID. I would propose just removing this test, but there's probably logic elsewhere that assumes that there is a UUID. Thoughts, @00vareladavid?

00vareladavid:
fix #1571

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -278,9 +278,7 @@ function EnvCache(env::Union{Nothing,String}=nothing)
     # read project file
     project = read_project(project_file)
     # initialize project package
-    if any(x -> x !== nothing, [project.name, project.uuid, project.version])
-        project.name === nothing && pkgerror("project appears to be a package but has no name")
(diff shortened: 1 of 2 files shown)


00vareladavid:
I think this is good to merge. Should it be backported to 1.4?

StefanKarpinski:
It's a functional change, but it's pretty low risk. Thoughts, @KristofferC?

KristofferC:
It just changes errors to non-errors and I have hit this (and gotten annoyed by it) a couple of times already. So I say we put it in.

PR #1636, opened 2020-01-27
change REPL path resolution

00vareladavid:
Currently `pkg> add Example` will resolve to a path if `Example` exists in the current working directory. This can be surprising since `pkg> add Example` usually resolves to a registered package. It is also annoying since we have no way to force a registered package: you have to change your working directory in order to add the registered package.

We should only interpret the argument as a path when the user is more explicit i.e. when they include an explicit path separator in the argument

Example:
```
shell> ls
Example

(ex3) pkg> add Example
  Resolving package versions...
   Updating `/tmp/ex3/Project.toml`
  [7876af07] + Example v0.5.3
   Updating `/tmp/ex3/Manifest.toml`
  [7876af07] + Example v0.5.3

(ex3) pkg> add ./Example
   Updating git-repo `/tmp/ex1/dev/Example`
  Resolving package versions...
   Updating `/tmp/ex3/Project.toml`
  [7876af07] ↑ Example v0.5.3 ⇒ v0.5.4 #master (../ex1/dev/Example)
   Updating `/tmp/ex3/Manifest.toml`
  [7876af07] ↑ Example v0.5.3 ⇒ v0.5.4 #master (../ex1/dev/Example)
```

### DIFF
--- a/src/REPLMode/argument_parsers.jl
+++ b/src/REPLMode/argument_parsers.jl
@@ -73,10 +73,7 @@ end
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
 function parse_package_identifier(word::AbstractString; add_or_develop=false)::PackageSpec
-    if add_or_develop && isdir_windows_workaround(expanduser(word))
-        if !occursin(Base.Filesystem.path_separator_re, word)
-            @info "Resolving package identifier `$word` as a directory at `$(Base.contractuser(abspath(word)))`."
-        end
+    if add_or_develop && (occursin(Base.Filesystem.path_separator_re, word) || word in (".", ".."))
         return PackageSpec(repo=Types.GitRepo(source=expanduser(word)))
     elseif occursin(uuid_re, word)
         return PackageSpec(uuid=UUID(word))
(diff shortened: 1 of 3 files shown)


KristofferC:
I dunno... I use dev on a package directory all the time. Can't we just use `isdir_casesensitive` or whatever it is called to avoid the common `add Test` getting the test directory.

StefanKarpinski:
Maybe prompt the user or error if it's ambiguous? I.e. the argument is a name that could either be a package name or a local path that already exists?

00vareladavid:
@KristofferC although I was prompted by the slack discussion, this is meant to be a more general behavior fix i.e. more than just the case sensitive part. I opened #1637 so this doesn't hold up any backports to 1.4 (probably should have done that first :sweat_smile: )

00vareladavid:
https://github.com/JuliaOpt/CPLEX.jl/issues/221

I think this issue supports my argument well. People expect `pkg> add Example` to always use the registered version. When you think about it, it is bizarre that the behavior depends on the working directory!

I don't believe the info message is enough. I've been bitten by this behavior myself.

StefanKarpinski:
Think about documenting this:

> `add Example` adds the registered `Example` package unless `Example` is a valid local path, in which case it adds whatever is in there.

It's embarrassing even writing that.

KristofferC:
Fine, I can accept writing `add ./Example` :)

00vareladavid:
Superseded by #2275

PR #1637, opened 2020-01-27
fix path resolution bug introduced in #1372

00vareladavid:


src/REPLMode/argument_parsers.jl
```diff
@@ -73,7 +73,7 @@ end
 # packages can be identified through: uuid, name, or name+uuid
 # additionally valid for add/develop are: local path, url
 function parse_package_identifier(word::AbstractString; add_or_develop=false)::PackageSpec
-    if add_or_develop && isdir_windows_workaround(expanduser(word))
+    if add_or_develop && (casesensitive_isdir(expanduser(word)) || word in (".", ".."))
```
KristofferC: Is the `word in (".", ".."))` stuff needed?

00vareladavid: Yes. It really wasn't working before, but people couldn't tell because Pkg was mixing paths and urls so freely:
```
julia> Pkg.casesensitive_isdir(".")
false

julia> Pkg.casesensitive_isdir("..")
false
```

The reason for being more strict about repo handling is that we can give better error messages. For example, we used to warn about malformed URLs when a user gave the path to a directory did not exist. Many people were confused by this but I don't think there was ever a formal issue

KristofferC: How about `../Foo`, will `casesensitive_isdir` work on that?

00vareladavid: I'm pretty sure it just does not work on these two special cases

KristofferC: Alright, strange. Maybe if we call `abspath` inside `casesensitive_isdir`

KristofferC:
```
julia> Pkg.casesensitive_isdir("../AppCompiler/")
false

julia> isdir("../AppCompiler/")
true
```

KristofferC: I pushed another commit that hopefully fixes this.

PR #1645, opened 2020-01-30
Use https://geo.pkg.julialang.org/ as the default Pkg server

DilumAluthge:
Instead of https://pkg.julialang.org/

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -20,7 +20,7 @@ function depots1()
 end
 
 function pkg_server()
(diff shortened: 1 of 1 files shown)


staticfloat:
I'm not certain that we're going to keep this name; Stefan, Avik and I have been debating the "right" way to do this, and this may change in the next couple of weeks.  It's entirely possible we'll just replace `pkg.julialang.org` with equivalent or better functionality.

KristofferC:
Yeah, let's chill a bit with this.

PR #1652, opened 2020-02-03
Update creating-packages.md

tawheeler:
The documentation does not mention `resolve`. The docs cover first-time package creation, but 99% of the time I'm editing an existing package. It seems that `resolve` is needed in those cases.

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -113,6 +113,11 @@ julia> HelloWorld.greet_alien()
 Hello aT157rHV
 ```
 
+When adding dependencies to an existing package, you may have to run `resolve` to update the manifest.
+```julia-repl
+(HelloWorld) pkg> resolve
+```
+
 ### Adding a build step to the package
 
 The build step is executed the first time a package is installed or when explicitly invoked with `build`.


00vareladavid:
Perhaps I misunderstand, but this doesn't seem accurate. Why do you need `resolve` when adding dependencies to an existing package? Can you clarify/expand on what the issue is?

tawheeler:
Sorry, this PR is borne out of frustration more than anything else --- and may not be worth merging.
I find the Pkg documentation is decent for showing how to create a new package but I very rarely get it to work when trying to modify an existing package. Most recently I tried to add LightGraphs as a dependency to a package. I tried multiple variations: `]` to enter Pkg mode, `activate PackageName` to activate my package, and `add LightGraphs` to add the dependency. After this I could load my package in the same Julia session and it would work. If I exit my Julia session it goes away. I am pretty sure running `resolve` at the end made it stick, but TBH I cannot confirm that 100%. 

In short, it would be very nice if the docs had a short recipe section that showed how to (1) add a dependency to an existing package:
```
] 
activate PackageName
add NewPackage
```
(and whatever is needed to make that persist)

to fix a package
```
]
activate PackageName
add NewPackage@0.1.0
```

and what to do if, as happens so often, you cannot do what you want because some existing dependency conflicts with what you want to do.

Maybe this would be better served as an issue.

PR #1656, opened 2020-02-05
WIP: bring back download indicator in status output

KristofferC:
Need to add / fixup the existing tests

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1595,10 +1595,11 @@ end
 
 function print_status(ctx::Context, old_ctx::Union{Nothing,Context}, header::Symbol,
                       uuids::Vector, names::Vector; manifest=true, diff=false)
+    not_installed_indicator = sprint((io, args) -> printstyled(io, args...; color=:red), "→", context=ctx.io)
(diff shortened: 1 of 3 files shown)


src/Operations.jl
```diff
@@ -1107,8 +1107,8 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}, new_git::Vector{UUID};
     update_manifest!(ctx, pkgs)
     new_apply = download_source(ctx, pkgs; readonly=true)
     download_artifacts(ctx, pkgs; platform=platform)
-    show_update(ctx)
     write_env(ctx.env) # write env before building
+    show_update(ctx)
```
00vareladavid: why after `write_env`? Wouldn't this make the `Updating` feedback misleading since we have already updated?

KristofferC: I think it would be weird to have the update print successfully and then have the `write_env` fail. Then you would be shown that stuff updated but it actually didn't. So it seems better to me to only print the updating after it actually succeeded.

Also, this made it easier to check if a package was installed in the `show_update` function since then the updating of the project files is already done.

PR #1662, opened 2020-02-06
`artifact"name"` should search the current directory when run from the REPL

staticfloat:
This enables more interactive usage of `@artifact_str` from the REPL.


StefanKarpinski:
Ah, this is so nice to have!

PR #1668, opened 2020-02-12
Update toml-files.md

timueh:
Added hint to add `Manifest.toml` to `.gitignore`. This avoids unexpected behavior.

### DIFF
--- a/docs/src/toml-files.md
+++ b/docs/src/toml-files.md
@@ -115,7 +115,7 @@ For the details, see [`Pkg.instantiate`](@ref).
 
 !!! note
     The `Manifest.toml` file is generated and maintained by Pkg and, in general, this file
-    should *never* be modified manually.
+    should *never* be modified manually. Add `Manifest.toml` to your `.gitignore` to avoid any unexpected behavior.
 
 
 ### `Manifest.toml` entries


fredrikekre:
What unexpected behavior? It is fine to commit the Manifest.toml.

timueh:
I had the `Manifest.toml` be commited to my repo, and then encountered a problem with `SpecialFunctions`: its version 0.10. requires Julia 1.3 or higher. If I now have a single committed `Manifest.toml`, in which I require `SpecialFunctions` 0.10, then this very same `Manifest.toml` will produce an error when testing for Julia versions less than 1.3.

If, instead, I do not commit the `Manifest.toml` to the repo, but add the correct dependencies in `[compat]` of `Project.toml`, then everything is fine.

fredrikekre:
But the manifest is ignored when you use a package. How did you use/run your package to run into such a failure?

KristofferC:
Probably when testing it on CI. As long as we don't have a `Pkg.test(;ignore_manifest)` I don't think commiting manifests is such a good idea.

timueh:
> Probably when testing it on CI.

Exactly. Looking at packages such as `JuMP` or `PowerModels` I saw that they added their `Manifest.toml` to the `.gitignore`. That's what got me thinking..

fingolfin:
I think it is sensible to warn about this, but perhaps the text should be a bit more nuanced? Like, 
> Usually it is not advisable to commit `Manifest.toml` to your source repository (although exceptions exists, such as XYZ), so you may wish to add it to your `.gitignore` file.

... where XYZ should obviously be filled in. Or, if nobody can think of such use cases, the text in parenthesis could go.

I can't think of a reason to do this for a regular package distributed to the global populace (which probably only attests to my lack of imagination). But it seems useful for "project" repositories, where you want to ensure the same config is used for multiple people. E.g. we are using this for software demos that we distribute to referees, were we don't want things to break because some component changed in a way we did not expect (I mean, in an ideal world, semantic versioning would protect us from issues... but this ain't no ideal world... ;-) ).

PR #1670, opened 2020-02-12
Mention that adding `#commit_id` is pinning the package

Datseris:
I had this question on Slack, on whether `add Pkg#master` pins to the branch or also to the commit and if not, how to pin to the commit. Maybe this extra sentence can help clarify things more.

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -81,7 +81,10 @@ we can explicitly track that branch (or commit) by appending `#branchname` (or `
 ```
 
 The status output now shows that we are tracking the `master` branch of `Example`.
-When updating packages, we will pull updates from that branch.
+When updating packages, we will pull updates from that branch. Notice that if instead
+of a branch name we specified an exact commit id, e.g.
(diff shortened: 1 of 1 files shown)


docs/src/managing-packages.md
```diff
@@ -81,7 +81,10 @@ we can explicitly track that branch (or commit) by appending `#branchname` (or `
 ```
 
 The status output now shows that we are tracking the `master` branch of `Example`.
-When updating packages, we will pull updates from that branch.
+When updating packages, we will pull updates from that branch. Notice that if instead
+of a branch name we specified an exact commit id, e.g.
+`add Example#277827f2b18e14a611d98144422eb29f61498104`, then we would be effectively
```
KristofferC, line 86: This commit SHA is excessively long. I can't just edit it myself because the repo-fork this PR is made from has been deleted.

Datseris: Ops. I'll try to fix it. Why don't you type the text you want here in the meantime, and I'll get it going.

Issue #1642, opened 2020-01-29
Fail to download registries from Pkg Server on Windows


KristofferC:
In the logs (e.g. https://ci.appveyor.com/project/JuliaLang/pkg-jl/builds/30415010)

```
    Cloning default registries into `C:\Users\appveyor\.julia`
┌ Warning: could not download https://pkg.julialang.org/registries
└ @ Pkg.Types C:\projects\pkg-jl\src\Types.jl:870
    Cloning registry from "https://github.com/JuliaRegistries/General.git"
      Added registry `General` to `C:\Users\appveyor\.julia\registries\General`
```

StefanKarpinski:
This has been working though, right? I wonder what changed. It's possible pkg.julialang.org was down for a bit. @vdayanand, did you happen to take that down for a bit?

KristofferC:
> This has been working though, right?

I'm not sure. I feel to not miss stuff like this we need to run some tests where the normal git fallbacks are disabled. Otherwise, we make a mistake but CI keeps working.

shihaoyan2018:
So,it is a bug? How can i use it on Windows,

johnnychen94:
FYI, I can successfully fix this issue locally with the following changes

```diff
[System.Net.ServicePointManager]::SecurityProtocol =
-                    [System.Net.SecurityProtocolType]::Tls12;
+                    [System.Net.SecurityProtocolType]::SystemDefault;
```

in https://github.com/JuliaLang/Pkg.jl/blob/master/src/PlatformEngines.jl#L372-L373

But looks like it would cause some other CI failures.

KristofferC:
@staticfloat is there a problem using Tls over Tls12?

staticfloat:
Yes; the majority of the internet won't talk to you if you use `Tls` or `Ssl3`; you _must_ use TLS 1.2+ if you want to, for instance talk to GitHub.  In the log that @johnnychen94  posted, note all the errors towards the bottom that say `Could not create SSL/TLS secure channel."`; that's the symptom of Powershell attempting to use an older protocol and the server failing to set things up.

[...]

PR #1671, opened 2020-02-12
Allow pkg server in scheme://netloc/path format

johnnychen94:
IIUC there're two formats supported now:

* `netloc`: `pkg.julialang.org`
* `scheme://netloc`: `https://pkg.julialang.org`

The `scheme://netloc/path` format, however, is widely used by mirror sites, e.g., `https://mirrors.ustc.edu.cn/julia/`. They can't simply create another domain for one of mirrors they host due to permission issues. (see also https://github.com/ustclug/discussions/issues/311)

This commit allows the `scheme://netloc/path` format as a package server
(either Pkg or Storage protocol) by disabling warnings

---

Before: (works normally except tons of warnings)

```julia
julia> using Pkg

julia> Pkg.add("Images")
   Updating registry at `~/Desktop/julia/registries/General`
┌ Warning: malformed Pkg server value
│   server = "http://juliapkg-test.lflab.cn/julia/"
└ @ Pkg.PlatformEngines /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/PlatformEngines.jl:601
┌ Warning: malformed Pkg server value
│   server = "http://juliapkg-test.lflab.cn/julia/"
└ @ Pkg.PlatformEngines /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/PlatformEngines.jl:601
^C┌ Warning: could not download http://juliapkg-test.lflab.cn/julia//registries
└ @ Pkg.Types /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/Types.jl:879
  Resolving package versions...
Downloading artifact: FFTW
┌ Warning: malformed Pkg server value
│   server = "http://juliapkg-test.lflab.cn/julia/"
└ @ Pkg.PlatformEngines /Users/sabae/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.4/Pkg/src/PlatformEngines.jl:601
```

After:

```julia
julia> Pkg.add("Images")
   Updating registry at `~/Desktop/julia/registries/General`
  Resolving package versions...
Downloading artifact: FFTW
#                                                                          2.2%
```

P.S. USTC don't want to host a daemon process such as `PkgServer.jl`, instead, they prefer the `gen_static.jl`+cron manner (although it would lose the compression features of Pkg protocol in the future).


StefanKarpinski:
This seems like a good idea (I'd thought about allowing this as well); I'll need to take a look and make sure that the implementation of `pkg_server` agrees and probably want to make sure that presence and absence of trailing slashes don't matter.

johnnychen94:
>  I'll need to take a look and make sure that the implementation of pkg_server agrees

I set up a [PkgServer.jl](https://github.com/JuliaPackaging/PkgServer.jl) with two upstreams: `http://mirrors.lflab.cn/julia`(a static content server built by `gen_static.jl`) and `https://pkg.julialang.org`. Looks like it works normally even without this PR.

```text
┌ Info: downloading resource
pkgserver_1   | │   server = "http://mirrors.lflab.cn/julia"
pkgserver_1   | └   resource = "/package/1520ce14-60c1-5f80-bbc7-55ef81b5835c/86d092c2599f1f7bb01668bf8eb3412f98d61e47"
pkgserver_1   | ┌ Info: downloading resource
pkgserver_1   | │   server = "https://pkg.julialang.org"
pkgserver_1   | └   resource = "/package/587475ba-b771-5e3f-ad9e-33799f191a9c/8134adbb0f10b0d22b22f8b4299d0d20509edc5f"
pkgserver_1   | ┌ Info: downloading resource
pkgserver_1   | │   server = "https://pkg.julialang.org"
pkgserver_1   | └   resource = "/package/91a5bcdd-55d7-5caf-9e0b-520d859cae80/fd11ab7aec59103217ecc5b5ccc34ce60e61b9ba"
```

src/Pkg.jl
```diff
@@ -23,7 +23,7 @@ function pkg_server()
     server = get(ENV, "JULIA_PKG_SERVER", "https://pkg.julialang.org")
     isempty(server) && return nothing
     startswith(server, r"\w+://") || (server = "https://$server")
-    return server
+    return rstrip(server, ['/', '\\'])
```
StefanKarpinski: This should only strip `/` not `\\`.

src/PlatformEngines.jl
```diff
@@ -592,16 +592,12 @@ end
 is_secure_url(url::AbstractString) =
     occursin(r"^(https://|\w+://(127\.0\.0\.1|localhost)(:\d+)?($|/))"i, url)
 
-function get_server_dir(url::AbstractString)
-    server = pkg_server()
+function get_server_dir(url::AbstractString, server=pkg_server())
     server === nothing && return
     startswith(url, server) || return
-    m = match(r"^\w+://([^\\/]+)$", server)
```
johnnychen94: > This should only strip `/` not `\\`.

I was wondering why there's a `\\` here. Can you explain it a bit?

StefanKarpinski: Because having a `/` or a `\\` in the server name makes it an invalid directory name—on Windows systems, it would put the files in a subdirectory, which is not what we want. Also, `\\` is just a weird thing to have in a URL, better to just not allow it.

johnnychen94:
Would it be possible that this change shipped with 1.4.0?

I ask this because Chinese users are really in need of the PkgServer feature. (Issue #1673 is only one of many, we've seen plenty of such issues in [Chinese discourse](https://discourse.juliacn.com/) recently)

In spite of the official servers `https://pkg.julialang.org`, which is not fast enough, I fully understand the frustrating paperwork and financial considerations setting up package servers in China mainland.  As I can see, the only way to increase user downloading experience is to ask mirror sites in universities to host them. But as I said, they (students) can't create a new subdomain owned by the university for permission issues.

P.S. I'm currently working on refactoring `gen_static.jl` so that we can ask USTC (https://github.com/ustclug/discussions/issues/311) to deploy a static content storage server. I hope it can catch up with the 1.4.0 release.

StefanKarpinski:
Sure, seems minor enough to backport.

staticfloat:
> The official servers https://pkg.julialang.org isn't fast and reliable enough in China. (e.g., I need 20 mins to do ]add Images Flux Plots.)

Do you mean including the `https://kr.pkg.julialang.org` server as well?  I thought that one worked significantly better?

In either case, this seems like a good enhancement.  :)

johnnychen94:
> Do you mean including the https://kr.pkg.julialang.org server as well? I thought that one worked significantly better?

https://kr.pkg.julialang.org works significantly better for me. There're still some people reporting it [not functioning well](https://discourse.juliacn.com/t/topic/2751/6), but without much feedback from them, I'm not sure if this is an issue. Perhaps with 1.4.0 released, more users would find it useful and discuss the usage of it. :P @staticfloat Thank you again for setting it up!

StefanKarpinski:
This might be a little tricky to backport to the 1.4 branch since the logic in `pkg_server` changed.

johnnychen94:
@staticfloat FYI, the reported failure seems to be caused by https://github.com/JuliaLang/Pkg.jl/issues/1642

PR #1677, opened 2020-02-15
backport: Allow pkg server in scheme://server/path format (#1671)

johnnychen94:
The tests in #1671 are removed since this branch doesn't have "Authentication Header Hooks" testset yet. I've tested it locally using the following servers:

* `http://juliapkg.lflab.cn`
* `https://mirrors.lflab.cn/julia` and `http://mirrors.lflab.cn/julia`


StefanKarpinski:
@KristofferC, @fredrikekre: what's the process here? Just merge into the release branch?

fredrikekre:
Seems fine. I did the backports manually, but I think Kristoffer uses a script. Doesn't matter, it should go the the release branch anyway.

PR #1680, opened 2020-02-16
remove tls security option

### LINKED ISSUE #1642: Fail to download registries from Pkg Server on Windows

johnnychen94:
After iterating over different options: `Tls`, `Tls11`, `Tls13` and `SystemDefault`(logs can be found in https://github.com/johnnychen94/Pkg.jl/pull/1). Seems like this fixes #1642  but needs Microsoft & network expert to identify if there's a problem

It would be great if this can be backported to `release-1.4`.

Reference: https://docs.microsoft.com/en-us/dotnet/api/system.net.securityprotocoltype?view=netframework-4.8

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -369,8 +369,6 @@ function probe_platform_engines!(;verbose::Bool = false)
         psh_download = (psh_path) -> begin
             return (url, path, hdrs...) -> begin
                 webclient_code = """
-                [System.Net.ServicePointManager]::SecurityProtocol =
-                    [System.Net.SecurityProtocolType]::Tls12;
                 \$webclient = (New-Object System.Net.Webclient);
                 \$webclient.UseDefaultCredentials = \$true;
                 \$webclient.Proxy.Credentials = \$webclient.Credentials;


staticfloat:
As I wrote in https://github.com/JuliaLang/Pkg.jl/issues/1642#issuecomment-586769589, this actually does break a lot of tests; it's bad that we have green checks here.  You can see the failures to download from GitHub in the logs above.

With this configuration you won't be able to download things from e.g. GitHub.  If you run into issues with TLS on Windows 7, you most likely need to install both the [TLS easy fix](https://support.microsoft.com/en-us/help/3140245/update-to-enable-tls-1-1-and-tls-1-2-as-a-default-secure-protocols-in) and [Windows Management Framework 3.0 or later](https://docs.microsoft.com/en-us/powershell/scripting/wmf/overview), as described in the [Platform-specific installation notes](https://julialang.org/downloads/platform/).

staticfloat:
One reason why the checkmarks might be green is that these workers may be running on Windows 10 with a recent enough Powershell that it _defaults_ to Tls 1.2+.

johnnychen94:
I'm collecting PkgServer usage reports from users in the Chinese community where one user gives me the dialog: https://discourse.juliacn.com/t/topic/2969/5

StefanKarpinski:
Can we just identify this failure better and give a more helpful error message?

PR #1686, opened 2020-02-18
write_env_usage: use proper TOML writer (fix #1681)

### LINKED ISSUE #1681: Pkg.gc not working on windows due to string interpolation
I reported a bug to julia, but I think should be listed here. 

https://github.com/JuliaLang/julia/issues/34781

StefanKarpinski:
The main fix here is to use TOML.print to generate the new content that is appended to the `manifest_usage.toml` log file, which should correctly escape the TOML key. I also moved the early exit test for the `source_file` not existing earlier so that we don't do any work in that case.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -377,20 +377,22 @@ function Context!(ctx::Context; kwargs...)
 end
 
 function write_env_usage(source_file::AbstractString, usage_filepath::AbstractString)
-    !ispath(logdir()) && mkpath(logdir())
-    usage_file = joinpath(logdir(), usage_filepath)
-    touch(usage_file)
-
     # Don't record ghost usage
     !isfile(source_file) && return
 
-    # Do not rewrite as do-block syntax (no longer precompilable)
-    io = open(usage_file, "a")
-    print(io, """
-    [[$(repr(source_file))]]
-    time = $(now())Z
-    """)
-    close(io)
+    # Ensure that log dir exists
+    !ispath(logdir()) && mkpath(logdir())
+
+    # Generate entire entry as a string first
+    entry = sprint() do io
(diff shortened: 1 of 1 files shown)


src/Types.jl
```diff
@@ -377,19 +377,22 @@ function Context!(ctx::Context; kwargs...)
...
     touch(usage_file)
 
-    # Don't record ghost usage
-    !isfile(source_file) && return
+    # Write entire new entry to a buffer first
+    buf = IOBuffer()
+    TOML.print(buf, Dict(source_file => [Dict("time" => now())]))
+    entry = String(take!(buf))
 
     # Do not rewrite as do-block syntax (no longer precompilable)
```
KristofferC: FWIW, we can remove this now since we now can precompile these things.

StefanKarpinski:
This fix probably also applies to 1.0.x, so I'm tagging that here, but I'm not sure how urgent backporting it actually is.

src/Types.jl
```diff
@@ -377,20 +377,22 @@ function Context!(ctx::Context; kwargs...)
...
-    io = open(usage_file, "a")
-    print(io, """
-    [[$(repr(source_file))]]
-    time = $(now())Z
-    """)
-    close(io)
+    # Ensure that log dir exists
+    !ispath(logdir()) && mkpath(logdir())
+
+    # Generate entire as a string first
```
feanor12: entire -> entry ?

src/Types.jl
```diff
@@ -377,20 +377,22 @@ function Context!(ctx::Context; kwargs...)
...
+    entry = sprint() do io
+        TOML.print(buf, Dict(source_file => [Dict("time" => now())]))
```
feanor12: buf -> io

PR #1688, opened 2020-02-19
Add docs explaining how to only test certain test sets

DilumAluthge:
We added the functionality in https://github.com/JuliaLang/Pkg.jl/pull/1226. But I think people don’t know about it, because the question is often asked in Discourse and Slack.

So this PR adds a section to the documentation explaining how you can set up your test suite to allow you to only run certain test sets.

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -248,6 +248,52 @@ The tests for a package can be run using `test`command:
    Testing Example tests passed
 ```
 
+Alternatively, you can use the `Pkg.test` function:
+```julia-repl
+julia> Pkg.test("Example")
+   Testing Example
+   Testing Example tests passed
+```
+
+### Testing only certain test sets
+
+The `Pkg.test` function accepts a keyword argument `test_args`. The arguments are passed to the test process, where they are available in the `ARGS` global variable. You can use this functionality to, for example, test only certain test sets.
+
+For example, suppose that you have a package named `HelloWorld`, and the contents of `test/runtests.jl` are:
+```julia
+
+using HelloWorld
+using Test
+
+@testset "HelloWorld.jl" begin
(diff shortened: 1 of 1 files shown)


crstnbr:
Thanks, I didn't know about the kwarg!

Just wondering, should the names of the `@testset`s perhaps automatically work with this method to only run specific test sets? That is without having to add an explicit if branches. AFAICS this would only require a minor modification of the `@testset` macro.

KristofferC:
This kinda feels like it should be in the docs for Test.jl (in maybe be made more full-fledged / automatic).

chriselrod:
> This kinda feels like it should be in the docs for Test.jl (in maybe be made more full-fledged / automatic).

I (and I'm sure many others) would love that, but what would be the preferred behavior in terms of handling nested testsets?
```julia
@testset "PkgName.jl" begin
    @testset "CategoryLevel1a" begin
        @testset "CategoryLevel2a" begin
            ...
        end
        @testset "CategoryLevel2b" begin
            ...
        end
    end
    @testset "CategoryLevel1b" begin
        @testset "CategoryLevel2a" begin
            ...
        end
        @testset "CategoryLevel2b" begin
            ...
        end
    end
end
```
Search each testset recursively for matching names, and when you get a match, run all testsets nested within it?
So that "PkgName.jl" is equivalent to "all"?

fredrikekre:
> This kinda feels like it should be in the docs for Test.jl (in maybe be made more full-fledged / automatic).

Was gonna say this too. The functionality that is here in Pkg is just for test-infrastructure packages to build upon IMO, and should probably be implemented/documented in those packages?

DilumAluthge:
I will close this in favor of adding support to `Test` and other testing framework stdlibs and/or packages.

Issue #1653, opened 2020-02-04
Corrupted path when developing packages on different drive than C on Windows


AnHeuermann:
When generating and developping a new package on Windows and you are not working on `C:` the path used by `pkg dev` will be corrupted and have "C:\" before the correct path.

Steps to reproduce the bug:
```julia
julia> cd("D:\\")
(v1.3) pkg> generate PackageName
Generating project PackageName:
    PackageName\Project.toml
    PackageName/src/PackageName.jl
julia> cd("PackageName")
(v1.3) pkg> dev .
 Resolving package versions...
ERROR: path C:\D:\PackageName for package PackageName no longer exists. Remove the package or `develop` it at a new path
```

Julia version: Version 1.3.1 (2019-12-30)
pkg version: 1.3
OS: Win10 64bit

StefanKarpinski:
Yeah, that's very suspicious. Looks like there's some code somewhere that's not calling `joinpath`.

visr:
Hmm, also hitting this on 1.4RC1. The info message shows the indentifier is correctly resolved to the path:
```julia
(@v1.4) pkg> generate MyPkg
 Generating  project MyPkg:
    MyPkg\Project.toml
    MyPkg\src/MyPkg.jl

(@v1.4) pkg> dev MyPkg
[ Info: Resolving package identifier `MyPkg` as a directory at `D:\MyPkg`.
Path `MyPkg` exists and looks like the correct package. Using existing path.
  Resolving package versions...
ERROR: expected package `MyPkg [e4d92ad3]` to exist at path `C:\D:\MyPkg`
```

Here is was working from the v1.4 project env which is on my C drive, as it will for most. However when working from a project on the D drive, this is fine:

```julia
(@v1.4) pkg> activate .
 Activating new environment at `D:\Project.toml`

() pkg> generate MyPkg2
 Generating  project MyPkg2:
    MyPkg2\Project.toml
    MyPkg2\src/MyPkg2.jl

() pkg> dev MyPkg2
[ Info: Resolving package identifier `MyPkg2` as a directory at `D:\MyPkg2`.
Path `MyPkg2` exists and looks like the correct package. Using existing path.
  Resolving package versions...
   Updating `D:\Project.toml`
  [6cc37cc3] + MyPkg2 v0.1.0 [`MyPkg2`]
   Updating `D:\Manifest.toml`
  [6cc37cc3] + MyPkg2 v0.1.0 [`MyPkg2`]
```

[...]

PR #1690, opened 2020-02-19
fix bug in Operations.source_path with paths relative to project

KristofferC:
This manifested itself that the registry got updated when doing `Pkg.instantiate` even though everything was installed if you had a relative path in the manifest and were not in the directory of the project file.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -35,9 +35,9 @@ end
 tracking_registered_version(pkg) =
     !is_stdlib(pkg.uuid) && pkg.path === nothing && pkg.repo.source === nothing
 
-function source_path(pkg::PackageSpec)
+function source_path(ctx, pkg::PackageSpec)
     return is_stdlib(pkg.uuid)      ? Types.stdlib_path(pkg.name) :
-        pkg.path        !== nothing ? pkg.path :
+        pkg.path        !== nothing ? joinpath(dirname(ctx.env.project_file), pkg.path) :
         pkg.repo.source !== nothing ? find_installed(pkg.name, pkg.uuid, pkg.tree_hash) :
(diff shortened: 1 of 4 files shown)


KristofferC:
Hm, I guess this is not always correct (looking at the test failure)...

src/Operations.jl
```diff
@@ -255,12 +255,12 @@ is_tracking_registry(pkg) = !is_tracking_path(pkg) && !is_tracking_repo(pkg)
...
 
 function collect_developed!(ctx::Context, pkg::PackageSpec, developed::Vector{PackageSpec})
-    source = project_rel_path(ctx, source_path(pkg))
+    source = project_rel_path(ctx, source_path(ctx, pkg))
     source_ctx = Context(env = EnvCache(projectfile_path(source)))
     pkgs = load_all_deps(source_ctx)
     for pkg in filter(is_tracking_path, pkgs)
         # normalize path
-        pkg.path = Types.relative_project_path(ctx, project_rel_path(source_ctx, source_path(pkg)))
+        pkg.path = Types.relative_project_path(ctx, project_rel_path(source_ctx, source_path(ctx, pkg)))
```
00vareladavid: it should be `source_path(source_ctx, pkgs)` here

00vareladavid:
`project_rel_path` currently handles this normalization. I think you have a good point, it should just be folded into `source_path`. I doubt there is a case where we need the unnormalized path.

There are some cases where you have to normalize against an env that is not the active env (like in the review comment).

src/Operations.jl
KristofferC: Thanks!

KristofferC:
Yeah, it seems less error prone to always work with an actual path and then only when you need to, you change that into a project relative one.

visr:
Did this fix #1653?

KristofferC:
Not sure

visr:
It doesn't, just checked.

PR #1692, opened 2020-02-22
ADDED JULIA_PKG_TELEMETRY ENV variable to control sending of telemetry.

Seelengrab:
This PR was created following a lengthy discussion on Slack with the purpose of separating the concepts of using the new Pkg servers (using JULIA_PKG_SERVER) from opting in to sending telemetry.

The JULIA_PKG_TELEMETRY environment variable indicates, if set to any value, that telemetry data will be sent to the Pkg servers.

This should probably be documented some place other than just this PR and I'd be happy to add the documentation, I'll just need some info where to put it.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -815,6 +815,7 @@ const CI_VARIABLES = [
 
 function get_telemetry_headers(url::AbstractString)
     headers = String[]
+    haskey(ENV, "JULIA_PKG_TELEMETRY") || return headers
     server_dir = get_server_dir(url)
     server_dir === nothing && return headers
     push!(headers, "Julia-Pkg-Protocol: 1.0")


StefanKarpinski:
There's already a way to opt out of telemetry at various levels:

1. Put `telemetry = false` in the `telemetry.toml` file to opt out entirely of sending anything;
2. Put `client_uuid = false` in the `telemetry.toml` file to opt out of sending anything UUID-based.

Adding more environment variables is not the solution. I'm looking into legal advice about this, but it's far from clear that there is any GDPR issue here in the first place because UUIDs are client-generated and there is no way of creating a data set mapping UUIDs back to identities.

Seelengrab:
Ah, of course! That change was suggested by kristoffer as a possible solution and I should have checked further. I missed the TOML variant since I'm not familiar with the code. 

To be clear, the original point was not about whether or not collecting telemetry should be done at all or anything like that, but rather that if people (in 1.4 at least) choose to use the new protocol mechanism that they'd be informed that usage statistics would be collected. Since no easy way to inform about that was found, we suggested that the easiest "fix" was to make the telemetry opt-in for 1.4 (irrespective of using the protocol itself) and check with legal counsel for how to best solve this for 1.5.

With that in mind, the same could be achieved by checking whether the telemetry entry has been set by the user:

```diff
diff --git a/src/PlatformEngines.jl b/src/PlatformEngines.jl
index 39d8486..2cd2dbc 100644
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -762,8 +762,8 @@ function load_telemetry_file(file::AbstractString)
             Dict(), true
         end
     end
-    # bail early if fully opted out
-    get(info, "telemetry", true) == false && return info
+    # check if user opted into telemetry
+    get(info, "telemetry", false) != false || return info
     # some validity checking helpers
     is_valid_uuid(x) = false
     is_valid_salt(x) = false
@@ -819,7 +819,7 @@ function get_telemetry_headers(url::AbstractString)
     server_dir === nothing && return headers
     push!(headers, "Julia-Pkg-Protocol: 1.0")
     info = load_telemetry_file(joinpath(server_dir, "telemetry.toml"))
-    get(info, "telemetry", true) == false && return headers
+    get(info, "telemetry", false) != false || return headers
     # general system information
     push!(headers, "Julia-Version: $VERSION")
     system = Pkg.BinaryPlatforms.triplet(Pkg.BinaryPlatforms.platform_key_abi())
```

The VS Code extension has something like that as the default with regular asking about whether or not one would like to send telemetry for crashes or problems. 

Another option would be to print something like the following during Pkg status/updates/etc.(?) when explicitly using the feature in 1.4:

```
You're using the new Pkg protocol. Since this is an experimental feature, some usage data (including but not limited to Julia-Version, Pkg-Protocol version, path and/or name of the currently active project, information about which CI service is used for the project (if available), and whether the julia session is interactive) is collected and sent as part of gaining insight into how julia is used.

In case you don't want to send any information beyond the Pkg-Protocol version used to the Pkg servers, you can opt-out of this telemetry by setting "telemetry = false" in the "telemetry.toml" located in the "servers" subdirectory of your Pkg depot.
```

StefanKarpinski:
We can also just put the consent on the download page for Julia itself. Documenting that opting into Pkg usage also by default opts into sending telemetry should also be sufficient for 1.4.

PR #1693, opened 2020-02-22
Pkg telemetry headers: GITHUB_ACTION => GITHUB_ACTIONS

StefanKarpinski:
The boolean variable that is set to true when running a GitHub action is plural. The singular is the ID of the action.


StefanKarpinski:
Not super crucial to backport this for 1.4.0 since we're not even collecting this data at the moment, but for 1.4.1 it would be good to fix.

PR #1696, opened 2020-02-24
Add current project as a Fixed package for the resolver and fix a bug when developed packages had cycles in paths

KristofferC:
@SimonDanisch posted an error message that didn't make sense to me:

```
(AbstractPlotting) pkg> add MakieGallery#master
ERROR: Unsatisfiable requirements detected for package FreeTypeAbstraction [663a7486]:
 FreeTypeAbstraction [663a7486] log:
 ├─possible versions are: 0.6.0 or uninstalled
 ├─FreeTypeAbstraction [663a7486] is fixed to version 0.6.0
 └─found to have no compatible versions left with AbstractPlotting [537997a7] 
   └─AbstractPlotting [537997a7] log:
     ├─possible versions are: 0.9.0-0.9.21 or uninstalled
     └─restricted to versions 0.9.16-0.9 by MakieGallery [dbd62bd0], leaving only versions 
```

The weirdness here is:

- Why is the resolver saying that the possible versions for `AbstractPlotting` is `0.9.0-0.9.21` when AbstractPlotting is the active project and should thus have a fixed version.
- Simon had added `FreeTypeAbstraction = "0.6"` to the project file but still the resolver said `found to have no compatible versions left with AbstractPlotting [537997a7] `.

I believe the reason for this weirdness is that we never explicitly say that the current project should be considered a fixed package to the resolver. The first commit tries to fix that and running with it makes the package add properly.

The second commit tries to address the fact that `collect_developed` can get into an infinite recursion if two packages are developed and depend on each other. I noticed this here because MakieGallery depends on AbstractPlotting.

However, when we have such a cyclic dependency, the project itself gets written to the Manifest (since it is a dependency). This is kinda weird but I will open a separate issue for that.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -276,6 +276,11 @@ end
 
 function collect_fixed!(ctx::Context, pkgs::Vector{PackageSpec}, names::Dict{UUID, String})
     deps_map = Dict{UUID,Vector{PackageSpec}}()
+    if ctx.env.pkg !== nothing
+        pkg = ctx.env.pkg
+        collect_project!(ctx, pkg, dirname(ctx.env.project_file), deps_map)
+        names[pkg.uuid] = pkg.name
+    end
     for pkg in pkgs
         path = project_rel_path(ctx, source_path(ctx, pkg))
         if !isdir(path)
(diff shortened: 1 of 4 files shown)


KristofferC:
Actually, the problem that the first commit fixes is also due to a cyclic dependency.

StefanKarpinski:
Good to fix this, but we may want to hard disallow cyclic dependencies in the future.

DilumAluthge:
> Good to fix this, but we may want to hard disallow cyclic dependencies in the future.

Yeah I thought it was clear that we do not support any cycles in the dependency graph.

Should we check for that explicitly?

What is the time/space complexity for searching a directed graph for any cycles? Perhaps we should do that as part of every Pkg operation.

SimonDanisch:
Btw, in my case it was only a test dependency, so not 100% cyclic, if that changes anything ;)

DilumAluthge:
> Btw, in my case it was only a test dependency, so not 100% cyclic, if that changes anything ;)

Hmm that’s interesting. Certainly better than actually having a cyclic dependency.

johnnychen94:
I get following error with the master branch of Pkg and Julia 1.5-DEV (not with Julia 1.4.0):

```bash
git clone https://github.com/JuliaGraphics/ColorTypes.jl.git
cd ColorTypes.jl
JULIA_DEPOT_PATH=~/julia_temp julia-latest --startup=no --project=. -e 'using Pkg; Pkg.add("Colors")'
```

```
Resolving package versions...
    Cloning default registries into `~/julia_temp`
      Added registry `General` to `~/julia_temp/registries/General`
  Installed Reexport ────────── v0.2.0
  Installed FixedPointNumbers ─ v0.8.0
  Installed Colors ──────────── v0.12.0
ERROR: KeyError: key "ColorTypes" not found
```

I assume this error is expected as a cyclic dependency? But perhaps a more illustrative error description is needed.

KristofferC:
Could you open a new issue with this, please? It looks like a bug indeed.

PR #1699, opened 2020-02-25
Any Pkg function that builds packages should accept the `verbose::Bool` keyword argument

DilumAluthge:
Currently, the only Pkg functions that accept the `verbose::Bool` keyword argument:
- `Pkg.build`
- `Pkg.instantiate`.

But I think that any function that builds packages should accept this keyword argument and pass it on to the `build` or `build_versions` function.

So this PR adds the `verbose::Bool` keyword argument to the following functions:
- `Pkg.add`
- `Pkg.developer`
- `Pkg.free`
- `Pkg.pin`
- `Pkg.up`

The keyword argument is optional. The default value is `verbose::Bool = false`.


KristofferC:
`verbose` should just be made a part of the `Context`https://github.com/JuliaLang/Pkg.jl/blob/f84d8117b4ce1c03b2cffda4a78d38340700c646/src/Types.jl#L319-L331 (we already have `graph_verbose` in there). Then there is no need to pass a `verbose` argument around, just just look inside `ctx`.

DilumAluthge:
> Does the REPL mode correctly forward --verbose to the function calls for all functions?

This PR doesn't make any changes to the REPL mode, primarily because I am not 100% sure how to add/modify functionality to the REPL mode.

DilumAluthge:
> `verbose` should just be made a part of the `Context`

Should I modify this PR to make `verbose` part of the `Context`? Or would it be better to make a separate PR?

00vareladavid:
I think it is fine in this PR :+1:

00vareladavid:
If you want to add the REPL mode, all you have to do is declare the flag like this:
```
        [:name => "verbose", :short_name => "v", :api => :verbose => true],
```
in `command_declarations` the same way that other commands do.

DilumAluthge:
> If you want to add the REPL mode, all you have to do is declare the flag like this:
> 
> ```
>         [:name => "verbose", :short_name => "v", :api => :verbose => true],
> ```
> 
> in `command_declarations` the same way that other commands do.

Will that get passed correctly into the constructor for `Context`?

00vareladavid:
Yes.  So `:name => "verbose", :short_name => "v"` means to parse `--verbose` and `-v` in the REPL and `:api => :verbose => true` means to convert the REPL flag to `API.fn(args; Dict(:verbose => true)...)`

then `Context!` will pick up the flag inside the API function and store it in the `Context` object

PR #1707, opened 2020-02-28
better merging of test project file

### LINKED ISSUE #1585: "cannot merge projects" error
Testing the package "DisplayAs" gives the error

```
ERROR: can not merge projects
Stacktrace:
 [1] pkgerror(::String) at /workspace/srcdir/usr/share/julia/stdlib/v1.4/Pkg/src/Types.jl:54
 [2] (::Pkg.Operations.var"#84#88"{Pkg.Operations.var"#93#94"{Bool,Cmd,Cmd,Nothing,Pkg.Types.Context,Array{String,1},String,Pkg.Types.PackageSpec},Pkg.Types.Context,Pkg.Types.PackageSpec,String,Nothing,String})(::String) at /workspace/srcdir/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1346
 [3] mktempdir(::Pkg.Operations.var"#84#88"{Pkg.Operations.var"#93#94"{Bool,Cmd,Cmd,Nothing,Pkg.Types.Context,Array{String,1},String,Pkg.Types.PackageSpec},Pkg.Types.Context,Pkg.Types.PackageSpec,String,Nothing,String}, ::String; prefix::String) at ./file.jl:673
 [4] mktempdir(::Function, ::String) at ./file.jl:671 (repeats 2 times)
```

This type of error message seems similar to an assertion error so is it a bug in Pkg? If not, the error message should probably be expanded to say what is actually wrong and how to fix it. And this package passes on 1.3 so what's up with that?

00vareladavid:
The current test project merging has two flaws:
1. Detecting whether two dependencies are actually the same. It uses dumb string comparison without any normalization.
2. It errors out when it detects two conflicting deps. This is just unnecessary.

This PR fixes both of those:
1. Determine the actual location we expect to load the source from and more intelligently determine if these two paths are the same (this is `samefile` in the code)
2. In the case where to deps conflicting, we should just warn, choose the dep specified by the main project, and move on (basically what @fredrikekre recommended in #1585 )

Additional note:
`samefile` should ideally just compare the filesystem node id, but the dep might not exist at this point, so we have to fallback to string comparisons in the worst case.

fix #1585


tkf:
> 2\. choose the dep specified by the main project, and move on

This is not always the best choice, as discussed in #1585 and #1714.

schlichtanders:
If there is no agreement on 2., couldn't we just have 1. for now?

It would already solve some setups, like e.g.  `dev ..`

IanButterworth:
Seems stale and things might have improved since. Plus workspaces avoid a lot of this?

PR #1710, opened 2020-03-01
WIP: Allow user to add additional manifest filenames

DilumAluthge:


KristofferC:
I don't get this. What does this solve? If I get someone's package I need to know what name they used for their manifest and add it to this list? What about when I want to do an analysis ona large number of packages?

DilumAluthge:
Yeah it does seem like a bad idea, doesn’t it.

For some reason it made sense to me two hours ago, but it longer does.

PR #1715, opened 2020-03-07
document command aliases

rfourquet:
This document "short name" aliases of commands
1. in the main help text, e.g:
```
update, up: update packages in manifest
```
2. in the help text of the command, e.g. 
```
[up|update] [-p|--project]  [opts] pkg[=uuid] [@version] ...
```

I don't insist on both points, one of them could be enough.
For 2), I'm not sure how to do this in a less ugly way. Maybe just a paragraph on itself saying "Alias: up". But this is not great when there are already 2 paragraphs for 2 distinct usages, like for help:
```
  help

  List available commands along with short descriptions.

  help cmd

  If cmd is a partial command, display help for all subcommands. If cmd is a full command, display help for cmd.
```
So maybe, just keep point 1).

### DIFF
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -26,11 +26,11 @@ julia is started with `--startup-file=yes`.
     :completions => complete_help,
     :description => "show this message",
     :help => md"""
-    help
+    [?|help]
 
 List available commands along with short descriptions.
 
-    help cmd
+    [?|help] cmd
 
 If `cmd` is a partial command, display help for all subcommands.
 If `cmd` is a full command, display help for `cmd`.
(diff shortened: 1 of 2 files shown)


src/REPLMode/command_declarations.jl
```diff
@@ -340,7 +340,7 @@ reached from any existing project.
     :api => API.undo,
     :description => "undo the latest change to the active project",
     :help => md"""
-    undo
+    [u|undo]
```
StefanKarpinski: It's a bit weird that "undo" and "redo" have such short aliases. It seems like they are rare enough to not really need aliases at all, let alone single-letter aliases.

KristofferC: Yeah, we could remove the aliases

rfourquet: So shall I remove these two aliases in this PR?

StefanKarpinski: Best to remove before documenting so, yes, that would be great.

fredrikekre:
https://github.com/JuliaLang/Pkg.jl/issues/1716

Issue #1716, opened 2020-03-13
Artifact error on CI


fredrikekre:
```
Artifact Creation: Error During Test at /home/travis/build/JuliaLang/Pkg.jl/test/artifacts.jl:90
  Got exception outside of a @test
  IOError: stat: permission denied (EACCES) for file "/tmp/jl_Jyzeli/artifacts/jl_Fwki2C/bar/bar"
  Stacktrace:
   [1] lstat(::String) at ./stat.jl:69
   [2] islink at ./stat.jl:314 [inlined]
   [3] rm(::String; force::Bool, recursive::Bool) at ./file.jl:260
   [4] rm(::String; force::Bool, recursive::Bool) at ./file.jl:278 (repeats 2 times)
   [5] create_artifact(::Main.PkgTests.ArtifactTests.var"#1#2"{Main.PkgTests.ArtifactTests.var"#13#25"}) at /home/travis/build/JuliaLang/Pkg.jl/src/Artifacts.jl:237
```

Seems to happen on Julia nightly only, see e.g. [this Linux run](https://travis-ci.org/github/JuliaLang/Pkg.jl/jobs/662001489#L840) and [this macOs run](https://travis-ci.org/github/JuliaLang/Pkg.jl/jobs/662001491#L772).

PR #1718, opened 2020-03-14
Ability to `Pkg.add` but not download artifacts or run build steps

DilumAluthge:
I’ve intentionally left this undocumented, because this is really only for internal use. E.g. I’m planning on proposing an additional CI step for the General registry that runs `Pkg.add` for every package. This will be much faster if I can skip artifacts and builds.

cc: @StefanKarpinski

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -108,7 +108,8 @@ add(pkgs::Vector{<:AbstractString}; kwargs...) =
     add([check_package_name(pkg, :add) for pkg in pkgs]; kwargs...)
 add(pkgs::Vector{PackageSpec}; kwargs...)      = add(Context(), pkgs; kwargs...)
 function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PRESERVE_TIERED,
-             platform::Platform=platform_key_abi(), kwargs...)
+             platform::Platform=platform_key_abi(),
+             artifacts::Bool=true, build::Bool=true, kwargs...)
     pkgs = deepcopy(pkgs)  # deepcopy for avoid mutating PackageSpec members
     Context!(ctx; kwargs...)
 
@@ -156,7 +157,9 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PR
         end
     end
 
-    Operations.add(ctx, pkgs, new_git; preserve=preserve, platform=platform)
+    Operations.add(ctx, pkgs, new_git;
+                   preserve=preserve, platform=platform,
+                   artifacts=artifacts, build=build)
     return
 end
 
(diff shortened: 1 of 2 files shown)


DilumAluthge:
AppVeyor failure seems unrelated.

DilumAluthge:
I think the Travis failure is unrelated. Pkg master seems to be failing on Julia master/nightly. See e.g. https://github.com/JuliaLang/Pkg.jl/pull/1719

DilumAluthge:
Presumably the Travis failure is https://github.com/JuliaLang/Pkg.jl/issues/1716

KristofferC:
Adding `.julia/artifacts` to the Travis cache should also skip having to re-download the artifacts. 

Is the goal here just to check the resolver? In that case, you could just call the function that does the resolving and check what it returns.

DilumAluthge:
> Adding `.julia/artifacts` to the Travis cache should also skip having to re-download the artifacts.

Yeah, but I figure the cache will be pretty huge. And whenever an artifact is updated, it’ll partially invalidate the cache.

Plus, even if we cache `.julia/artifacts`, we’ll still end up running the `Pkg.build` steps.

> Is the goal here just to check the resolver? In that case, you could just call the function that does the resolving and check what it returns.

I also want to check that the package (and its dependencies, recursively) are downloadable and installable. This makes sure e.g. that all the `git-tree-sha1` values in the registry are correct, etc.

DilumAluthge:
Also, from the Travis docs: ([link](https://docs.travis-ci.com/user/caching/#things-not-to-cache))
> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source

So I’m not sure if caching the artifacts would actually speed things up.

KristofferC:
> Large files that are quick to install but slow to download do not benefit from caching, as they take as long to download from the cache as from the original source

Oh, I see.

fredrikekre:
Why not skip downloading the source code as well? Basically a function or option to just run the resolver and output the `toml`s.

DilumAluthge:
> Why not skip downloading the source code as well? Basically a function or option to just run the resolver and output the `toml`s.

I figured that downloading the source code is relatively fast compared to artifacts and build steps. And I do want to check that the `git-tree-sha1` values are correct, and thus the package is actually installable.

If e.g. someone makes a manual PR, and they accidentally put the commit SHA instead of the tree SHA, this won’t get caught unless we actually download and install the package.

DilumAluthge:
> Basically a function or option to just run the resolver and output the tomls.

That being said, I do think that what you describe here could be a useful feature to have available.

DilumAluthge:
So... Imthe other option (for General registry CI) would be to use the GitHub API to get the list of file paths that were modified, and only do `Pkg.add` for packages that were modified by the pull request. Presumably, this will be faster than doing it for all packages. And in that case, it probably is fine to also test downloading of the artifacts.

DilumAluthge:
> Why not skip downloading the source code as well? Basically a function or option to just run the resolver and output the `toml`s.

@fredrikekre I ended up implementing exactly this suggestion in this PR: #1718

PR #1727, opened 2020-03-21
WIP: Search the active project for artifacts

### LINKED ISSUE #1726: Add support for Overrides.toml next to Project.toml/Manifest.toml
Would it be possible to add the following functionality to the artifacts story:

if there is an `Overrides.toml` next to the `Project.toml` that is the active project, then apply the information from that `Overrides.toml`. Essentially it would be a non-global option for `Overrides.toml`.

DilumAluthge:
Closes #1726 
The other half is located here: https://github.com/JuliaLang/julia/pull/35195

Needs tests.
Needs docs.
Needs compat annotation.
Needs changelog.

---

This pull request adds `/path_to_active_project/artifacts` to the end of the list of locations that we search when looking for artifacts.

h/t: @oxinabox for suggesting an expansion to the artifact search path
h/t: @fredrikekre for making an example of a depot stored inside a project
cc: @davidanthoff
cc: @StefanKarpinski 
Also contributed to the discussion: Kristoffer Carlsson, Elliot Saba, Sascha Mann
(I'm trying to avoid pinging too many people.)


StefanKarpinski:
This conversation offers some context:

https://docs.google.com/document/d/1Z63eqhkLYIg5iviwmZnq741zp5kIwGO1spf7zeSjAX8/edit

src/Artifacts.jl
```diff
@@ -45,7 +45,9 @@ current set of depot paths and the current artifact directory override via the m
 """
 function artifacts_dirs(args...)
     if ARTIFACTS_DIR_OVERRIDE[] === nothing
-        return [abspath(depot, "artifacts", args...) for depot in depots()]
+        depot_artifacts = [abspath(depot, "artifacts", args...) for depot in depots()] # search all depots
+        project_artifacts = [abspath(dirname(Base.active_project()), "artifacts", args...)] # search the active project
```
StefanKarpinski: Can `active_project()` be `nothing` in some circumstances?

DilumAluthge: I'm not sure, but it can't hurt to be prepared for that case.

DilumAluthge: Technically, `Base.ACTIVE_PROJECT[]` can be `nothing`.

StefanKarpinski: But what about `active_project()`. I think yes, but would need to be looked at.

DilumAluthge: Alright, so now we check to make sure that `active_project()` is not `nothing` and `active_project()` is not empty.

src/Artifacts.jl
```diff
@@ -45,7 +45,14 @@ current set of depot paths and the current artifact directory override via the m
...
     if ARTIFACTS_DIR_OVERRIDE[] === nothing
-        return [abspath(depot, "artifacts", args...) for depot in depots()]
+        depot_artifacts = [abspath(depot, "artifacts", args...) for depot in depots()] # search all depots
+        project = Base.active_project()
+        if project === nothing || isempty(project)
+            project_artifacts = String[]
+        else
+            project_artifacts = [abspath(dirname(Base.active_project()), "artifacts", args...)] # search the active project
+        end
+        return vcat(depot_artifacts, project_artifacts)
```
StefanKarpinski: Switching the order seems better—might as well look in the project first.

DilumAluthge: I thought about that, but here's the problem. If we put the project first, then, if the artifact needs to be downloaded, it will always be downloaded into the project.

DilumAluthge: But for most users, we want to download artifacts into the first depot.

DilumAluthge: I guess we could return them in this order when searching/reading:
- project first, then depots

But return them in this order when downloading/creating/writing:
- depots first, then project

But... this might make things more confusing?

tkf:
Does it make sense to use `.julia` as the parent directory (i.e., `/path_to_active_project/.julia/artifacts` and  `/path_to_active_project/.julia/packages`), to avoid cluttering the package directory?  I think there are other things you might want to make project-local. For example, it would also be nice if the system image is automatically loaded from, e.g., `/path_to_active_project/.julia/sys.so` https://github.com/JuliaLang/julia/issues/33973#issuecomment-601541817

DilumAluthge:
Personally I'd like to keep this as simple as possible. The motivation is to make a single self-contained tarball. The only things we need to accomplish that are artifacts and packages.

tkf:
How does using `/path_to_active_project/.julia/artifacts ` complicate the PRs?  It's just one more argument to `abspath`.  I'm just trying to point out that you might need other kinds of project-specific storage in the future.

DilumAluthge:
> How does using `/path_to_active_project/.julia/artifacts ` complicate the PRs? It's just one more argument to `abspath`. I'm just trying to point out that you might need other kinds of project-specific storage in the future.

Oh, I misunderstood. I thought you wanted me to add the "automatically load system image" feature to this PR.

DilumAluthge:
Yeah, doing `/path_to_active_project/.julia/artifacts` instead of `/path_to_active_project/artifacts` seems fine to me!

If we don't want to do a hidden folder, I suppose we could do `/path_to_active_project/julia/artifacts` instead.

DilumAluthge:
So, here are some candidates to vote on:

| Artifacts | Packages|
| -------- | ---------- |
| `/path_to_active_project/artifacts` | `/path_to_active_project/packages` |
| `/path_to_active_project/.julia/artifacts` | `/path_to_active_project/.julia/packages` |
| `/path_to_active_project/julia/artifacts` | `/path_to_active_project/julia/packages` |
| `/path_to_active_project/.julia_modules/artifacts` | `/path_to_active_project/.julia_modules/packages` |
| `/path_to_active_project/julia_modules/artifacts` | `/path_to_active_project/julia_modules/packages` |

tkf:
A non-hidden path like `/path_to_active_project/julia/*` sounds good to me, too. The original inspiration was from the default depot path.

DilumAluthge:
I was thinking maybe we do `julia_modules`, in a parallel to `node_modules`?

Or, since we are calling these "standalone bundles", maybe `julia_bundle`?

tkf:
I don't have a strong opinion on the name itself (other than using a single parent directory), but my preference is `julia > julia_bundle > julia_modules`.

DilumAluthge:
Hidden folder (dot) or regular folder (no dot)?

tkf:
I slightly prefer hidden directory `.julia` etc. more.

KristofferC:
It seems non-general to me to single out artifacts for this. Why not packages? To me, this would be better implemented as adding a new entry into `DEPOT_PATH` that resolves to the current active project. Then the same thing can be used for `packages` and precompile files. This would then not require any changes to Pkg.

DilumAluthge:
> Why not packages

See https://github.com/JuliaLang/julia/pull/35195

KristofferC:
My point is that if this is implemented correctly, there should be no need to change Pkg at all  because everything should fall out from just having an extra DEPOT_PATH entry which is already handled by the artifact system.

DilumAluthge:
> My point is that if this is implemented correctly, there should be no need to change Pkg at all (nor code loading like in that PR)  because everything should fall out from just having an extra DEPOT_PATH entry which is already handled by everything.

Ah, I see what you mean.

I’m not sure how this special entry would be implemented though.

DilumAluthge:
Specifically, how do you make sure that whenever the active project is changed, you change the corresponding depot entry?

fredrikekre:
> My point is that if this is implemented correctly, there should be no need to change Pkg at all (nor code loading like in that PR) because everything should fall out from just having an extra DEPOT_PATH entry which is already handled by everything.

This is why I keep suggesting making it a proper depot path, because the next guy is gonna ask to also bundle the `registries` directory, and then someone wants to include `config` etc.

I still think it is not too bad to make it explicit and put
```
push!(DEPOT_PATH, @__DIR__) # or push!(DEPOT_PATH, joinpath(@__DIR__, ".julia")) if you prefer
```
in some initialization script `init.jl` and then document that you run this project by
```
julia -L init.jl main.jl
```
or w/e.

If this for some reason has to be done automatically, then we should just add `"@"` at the end of the default `DEPOT_PATH` just like we have `"@"` at the front of the default `LOAD_PATH`. This would then expand to `joinpath(dirname(active_project()), ".julia")`.

> Specifically, how do you make sure that whenever the active project is changed, you change the corresponding depot entry?

Switching projects at runtime is a bit iffy anyway, but my suggestion above should work just as switching the load path entry `"@"` does.

src/Artifacts.jl
oxinabox: It may be interesting to look at DataDeps.jl's solution which is basically what you propose.
https://white.ucc.asn.au/DataDeps.jl/stable/z10-for-end-users/#The-Load-Path-1
Basically the search order starts looking in project, then continues to more and more general (it allows many locations).
the save order attempts to save in the same list of locations (since by design it can fail, and want to move on to next), but skips the project specific one at the top.

*however* that is largely irrelevent for you.
Search order doesn't matter because content addressing promises that no matter where you find a match, you know that match is good and is identical to anywhere else you might find a match.

staticfloat: With artifacts, search order doesn't matter for correctness, (since it's content-addressed, if you find it, it's the right thing) but it could matter for performance.  IMO you're more likely to have something sitting in your overall location (project-local deps will not be the norm, I don't think) so keeping it this way is better, IMO.

KristofferC:
Yeah, the "correct" implementation of this seems to me to add a special entry to `DEPOT_PATH` that resolves to the current project, similar to how things are handled with `LOAD_PATH`. Then we need a function like `Base.depot_path()` (equivalent to `Base.load_path())` and everyone uses that and then things agree with each other.

tkf:
`Base.depot_path()` sounds like the way to go.

> and then someone wants to include `config` etc.

IIUC, `startup.jl` file won't be included if the project-local depot is at the end of the list?  Though this probably just has to be clearly documented (or special-cased).

DilumAluthge:
When @StefanKarpinski, @davidanthoff, and I were having the conversation on Slack, my intent was purposefully not to create a full depot in the project. I very intentionally only want to look for artifacts and packages, because that is the bare minimum we need to make the "self-contained bundles".

If there is an overwhelming consensus that we should instead modify the way `Base.DEPOT_PATH` works, I’ll defer to that consensus. I think that change will be more complicated. For example, we will need to implement this new `Base.depot_path()` function. We will also need to go through every location in Base Julia and in Pkg that currently uses `Base.DEPOT_PATH` and replace that with a use of `Base.depot_path()`. Furthermore, adding a special entry like `@` will also break every user’s code that currently assumes that `Base.DEPOT_PATH` is a list of depot paths.

But personally, my vote is that we only implement the absolute bare minimum that we need to have self-contained bundles.

DilumAluthge:
To elaborate on that a little bit: the only things that I am suggesting that we allow people to store inside projects are artifacts and packages.

Artifacts are content-addressed with the hash. Packages are content-addressed with the package name plus the slug.

The artifacts and packages could exist anywhere. No matter what location we find them in, the content is exactly the same. So we are just adding another location to the search path.

This does not apply to the other things that people store inside depots (e.g. `startup.jl' files).

This is why I only want to expand the search for artifacts and packages.

DilumAluthge:
> Yeah, the "correct" implementation of this seems to me to add a special entry to `DEPOT_PATH` that resolves to the current project, similar to how things are handled with `LOAD_PATH`. Then we need a function like `Base.depot_path()` (equivalent to `Base.load_path())` and everyone uses that and then things agree with each other.

I have started a pull request that implements what you describe: https://github.com/JuliaLang/julia/pull/35207

I don't think that is the correct approach. But I figured I would start work on implementing it while we continue this discussion.

src/Artifacts.jl
StefanKarpinski: This is [another reason](https://github.com/JuliaLang/julia/pull/35207#issuecomment-602247884) why treating the project as a depot more generally is wrong: it should only be a place you look in when loading artifacts. If it's not there, you download in the normal fashion into the standard user depot.

StefanKarpinski: @staticfloat, I don't understand the performance consideration. Why would the system copy of an artifact be better than the project copy?

StefanKarpinski:
HARD DISAGREE with making the project a full depot. I'll quote my entire comment [here](https://github.com/JuliaLang/julia/pull/35207#issuecomment-602247884) since this seems to be where the discussion ended up taking place:

---

I can see the appeal of this since it allows the user to choose whether to look in a project for vendored dependencies (both packages and artifacts). However, I have a couple of issues with it:

1. I think we should look in the project for packages and artifacts first. After all, if the project has bothered to ship vendored resources, shouldn't we use the project's version of those resources?

2. I don't think we should be treating the project as a depot for all purposes. If `$project/registries` exists, should we look in registries that are included there? Seems like a bad idea. Similarly, if the project contains `$project/servers` should we let the authentication info included there apply to talking to authenticated servers and provide telemetry info? If there is an `$project/environments` directory, should we consider those to be named environments when this project happens to be active? These all seem like a bad idea.

An alternative would be to make the default `DEPOT_PATH` have `@` as its first entry (and replace `Pkg.depots1()` with `Pkg.user_depot()` which skips the `@` and is a better name anyway). We could then have a toggle for whether we should expand the `@` or drop it and expand it when looking for packages and artifacts and drop it when looking for anything else. That would allow people to opt-out of loading vendored packages and artifacts.

However, I think that unconditionally using vendored packages and artifacts when loading resources for a project would be fine and would be simpler. After all, what's the benefit of opting out? It's not like there's a security issue here: if the project wants to load code that it ships with, it can do that much more easily than by shipping a vendored artifact or package.

DilumAluthge:
I agree with all of that.

DilumAluthge:
Stefan's PR https://github.com/JuliaLang/julia/pull/35222 implements searching the project first for packages.

I'll update this PR to search the project first for artifacts.

Here's the one caveat: we are going to search the project first for artifacts, then we will search the other locations.

But when we download new artifacts, I don't think we should put them in the project by default. I think we should put them in the user depot (which is the current behavior). Is everyone fine with that? So we will change the search order for reading artifacts, but we won't chance the behavior for writing artifacts.

Because I think by default we should still write new artifacts to the user depot

src/Artifacts.jl
DilumAluthge: So, to be clear, the plan is that we will prepend the project to the front of the list of search paths when we are searching for an artifact. But when we are creating artifacts, we are going to use the user depot, correct?

StefanKarpinski:
Absolutely. The only role I see for the project package/artifact directories is as collection of content-addressed resources that might be needed when loading the project. Package & artifact installation by Pkg should still be to the normal place. A completely separate tool should be used to copy content-addressed resources into a project to make it standalone, with options to control whether to copy lazy artifacts or not and whether to copy artifacts for all platforms or just some.

DilumAluthge:
Alright, I’ve updated the PR. We now search the project first before searching other locations.

When writing artifacts, we write them to the user depot.

src/Artifacts.jl
staticfloat: > @staticfloat, I don't understand the performance consideration. Why would the system copy of an artifact be better than the project copy?

It's literally just saving a `stat()` or two.  It's not the performance of the binaries themselves, but merely how quickly you find them.  I don't think it's a big issue, even on Windows.

DilumAluthge:
Just FYI, searching the project first makes the code significantly more complicated than if we search the project last. This is because currently a lot of places in Pkg hardcode the assumption that the first artifact search path is also the place we would write new artifacts to.

PR #1734, opened 2020-03-24
fix issue #1732 Pkg.dependencies docstring not rendering underscore. 

### LINKED ISSUE #1732: Pkg.dependencies docstring not rendering underscore.
This is very minor issue.
The Pkg.dependecies docstring is:
https://github.com/JuliaLang/Pkg.jl/blob/702746380ba8a543f443669e9c98298c0f0eb84e/src/Pkg.jl#L300-L320

However, when I try to show it in REPL, it does not rendering underscore like:
image

I think this post is related
https://discourse.julialang.org/t/docstring-not-rendering-underscore/15818

AtsushiSakai:
To fix #1732 

Before:
SS 2020-03-24 at 20 57 01

After:
SS 2020-03-24 at 20 54 06

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -310,13 +310,13 @@ The result is a `Dict` that maps a package UUID to a `PackageInfo` struct repres
 
 | Field             | Description                                                |
 |:------------------|:-----------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


KristofferC:
Thanks

PR #1742, opened 2020-04-01
WIP/RFC: replacing PlatformEngines

StefanKarpinski:


src/Download.jl
```diff
@@ -0,0 +1,230 @@
...
+                @warn "invalid TOML" path=tree_info err
+            end
+            tree_info_hash == tree_hash && return path
+        end
+        rm(path, force=true, recursive=true)
+    end
+    tarball = get_tarball()
+    contents = Dict{String,String}()
+    symlinks = Dict{String,String}()
+    open(`gzcat $tarball`) do io
```
fingolfin, line 118: So this is hardcoded to only support gzipped files. Is the plan to extend this to use `CodecBzip2.jl` etc. to support more formats? Like, have a table which maps various file extensions to different transformations?

Is there also a `Zip.jl` which can deal with .zip files?

StefanKarpinski: I don't want to add tons of Julia dependencies here because this is going to get baked into the Julia sysimg, so the plan is TBD. I may try to make this portable using JLLs, which I'm less concerned about baking in since they're not in the sysimg and not evolving fast. I may also use gzip to bootstrap installing support for other compression formats.

StefanKarpinski:
Note: I'm factoring this into one or two layers of external packages for acquiring resources (registries, packages, artifacts). The layers I'm considering are:

1. ContentTrees — package for unpacking content-addressed trees with `.tree_info.toml` files in them. This will also include the ability to apply a diff to one content tree to get another.

2. FetchData — layer for managing the acquisition of data via http/s, git and in the future other methods as well. Unlike the current ad hoc approach to getting things from multiple sources, this will model things as a queue of resources to acquire with a sequence of methods to try for each resource, stopping when some method is successful. It will try to get some number of resources concurrently and display overall progress to the user.

I want to separate these layers out because I think they are each independently useful, but also so that I can test them thoroughly in isolation since they each have a fairly clear API that can be tested. With that done cleanly, we can use a vendored copy of each of these external packages in Pkg to implement the functionality there with far less code in Pkg itself.

StefanKarpinski:
Already further along elsewhere.

PR #1751, opened 2020-04-06
improve performance of TOML parsing by reading from memory instead of from file

KristofferC:
Reading a bunch of small things from a file tends to be much slower than just reading the whole file in one chunk and operating on memory (especially with Julia now locking the stream for every read). None of the files we parse are big enough that operating on a file makes sense.

Benchmark:

```jl
function parse_registry()
    path = joinpath(homedir(), ".julia/registries/General")
    for (root, dirs, files) in walkdir(path)
       for file in files
           if endswith(file, ".toml")
               TOML.parsefile(joinpath(root, file))
           end
       end
   end
end
``` 

Before:

```jl
696.417 ms (2111063 allocations: 157.45 MiB)
```

After:

``` jl
355.687 ms (2164815 allocations: 166.51 MiB)
```

### DIFF
--- a/ext/TOML/src/TOML.jl
+++ b/ext/TOML/src/TOML.jl
@@ -47,6 +47,6 @@ module TOML
     end
 
     "Parse file"
-    parsefile(filename::AbstractString) = open(parse, filename, "r")
+    parsefile(filename::AbstractString) = parse(IOBuffer(read(filename)))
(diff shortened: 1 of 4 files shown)


KristofferC:
Ah, this caught a too loose `try catch`.

PR #1750, opened 2020-04-06
Fix that version ranges in keys for one registry also applies to versions in another registry

KristofferC:
This "uncompresses" the version ranges that each registry provides using the known versions of the package in that registry. So for example, if the registry says

```
# Compat.toml
[0]
"Foo" = 0.1

# Versions.toml
v"0.3" => ...
v"0.4" => ...
```

this is then uncompressed to

```
v"0.3" => ("Foo" => 0.1, ...)
v"0.4" => ("Foo" => 0.1, ...)
```

TODO:
- Write a test
- ~Change the resolver tests to understand that they should now provide a version number (pretty annoying to fix...)~

cc @carlobaldassi 

Fixes #1434

PR #1754, opened 2020-04-07
Fix resolve tests (for new deps internal format)

carlobaldassi:
This is for #1750. The code is actually considerably simpler since the resolver internally already used a representation of the dependencies based in version numbers rather than version ranges.
It only solves the resolve tests, the "new.jl" test is still broken (I think it's unrelated to the resolver, if not let me know and I'll investigate).
@KristofferC

### DIFF
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -239,10 +239,8 @@ mutable struct Graph
             verbose::Bool = false
         )
         # make sure all versions of all packages know about julia uuid
-        for (uuid, _) in deps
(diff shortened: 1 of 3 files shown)


carlobaldassi:
I should add that when I test this on my machine the tests fail both on this branch and in the original one (something in the `copy_test_package` function in "test/utils.jl"), so I have commented all tests apart from resolve, and that one works.

KristofferC:
I will fix up the other tests, just needed to get the resolver tests working. Thanks you very much!

PR #1760, opened 2020-04-08
Document authors field in Project.toml

fingolfin:
Also fix a typo in one example that used `author` instead of `authors`


fingolfin:
Ping?

PR #1763, opened 2020-04-10
Ensure --depwarn=yes by default in Pkg.test

c42f:
If https://github.com/JuliaLang/julia/pull/35362 is merged, adding `--depwarn=yes` by default will be important for continuity. See @tkf's comment https://github.com/JuliaLang/julia/pull/35362#issuecomment-611818361 for more detail.

(Even without merging that PR in Base, merging this PR should be harmless: it's currently the default in Julia and users can still disable it with `julia_args`.)

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1282,7 +1282,7 @@ function gen_test_code(testfile::String;
         --color=$(Base.have_color === nothing ? "auto" : Base.have_color ? "yes" : "no")
         --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
         --check-bounds=yes
-        --depwarn=yes
+        --depwarn=$(Base.JLOptions().depwarn == 2 ? "error" : "yes")
         --inline=$(Bool(Base.JLOptions().can_inline) ? "yes" : "no")
         --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
         --track-allocation=$(("none", "user", "all")[Base.JLOptions().malloc_log + 1])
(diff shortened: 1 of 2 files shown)


src/Operations.jl
```diff
@@ -1282,6 +1282,7 @@ function gen_test_code(testfile::String;
         --color=$(Base.have_color === nothing ? "auto" : Base.have_color ? "yes" : "no")
         --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
         --check-bounds=yes
+        --depwarn=yes
```
tkf, line 1285: This makes `Pkg.test` not to propagate `--depwarn=error`. So I think we need to look at `JLOptions()` and try not to "downgrade" `depwarn` setting.

c42f: Hmm yes, on balance I think that's probably the most consistent with the other options. I added a tweak.

src/Operations.jl
```diff
@@ -1282,6 +1282,7 @@ function gen_test_code(testfile::String;
         --color=$(Base.have_color === nothing ? "auto" : Base.have_color ? "yes" : "no")
         --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
         --check-bounds=yes
+        --depwarn=$(Base.JLOptions().depwarn == 2 ? "error" : "yes")
```
fredrikekre, line 1285: If I have started Julia explicitly with `--depwarn=no` it would be nice to propagate that too, but maybe that info is not available as for other options?

KristofferC: Doesn't seem like it is available. Guess you need to just pass it explicitly.

c42f: Yep there's not much we can do about this without considerably complicating `Base.JLOptions()` with the ability to distinguish between default vs explicitly specified options.

kimikage:
Does this work in the default script of Travis CI? The `Base.JLOptions()` in "runtests.jl" says `depwarn = 0`.

KristofferC:
https://github.com/JuliaLang/julia/pull/35584 needs to be merged first

kimikage:
Oh, I see. I had no doubt in my mind that this was already merged to `JuliaLang/Julia`.

Issue #1570, opened 2019-12-30
Cannot `convert` an object of type Pkg.Types.VersionSpec to an object of type VersionNumber


timholy:
```juilia
(@v1.4) pkg> free Graphics ProfileView ImageMorphology
ERROR: MethodError: Cannot `convert` an object of type Pkg.Types.VersionSpec to an object of type VersionNumber
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:171
Stacktrace:
 [1] convert(::Type{Union{Nothing, VersionNumber}}, ::Pkg.Types.VersionSpec) at ./some.jl:34
 [2] Pkg.Types.PackageEntry(::String, ::Pkg.Types.VersionSpec, ::Nothing, ::Bool, ::Pkg.Types.GitRepo, ::Nothing, ::Dict{String,Base.UUID}, ::Nothing) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Pkg/src/Types.jl:222
 [3] Pkg.Types.PackageEntry(; name::String, version::Pkg.Types.VersionSpec, path::Nothing, pinned::Bool, repo::Pkg.Types.GitRepo, tree_hash::Nothing, deps::Dict{String,Base.UUID}, other::Nothing) at ./util.jl:736
 [4] update_manifest!(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:120
 [5] free(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Pkg/src/Operations.jl:1240
 [6] free(::Pkg.Types.Context, ::Array{Pkg.Types.PackageSpec,1}; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:287
 [7] free at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:268 [inlined]
 [8] #free#59 at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:265 [inlined]
[...]

PR #1781, opened 2020-04-18
prevent multiple copies of the same package to be considered when collecting recursively from paths

KristofferC:
The test here doesn't exactly replicate the error in https://github.com/JuliaLang/Pkg.jl/issues/1570 but it causes the assert to throw on master Pkg and this should be the root cause of the problem.


KristofferC:
cc @timholy

KristofferC:
The test that fails here makes no sense to me. Why would we want to have Example track the path from Unregistered just because it has a manifest with Example devved? cc @00vareladavid 

When you load and use TestSubgraphTrackingPath, you will use the normal Example but when you test it you will use the devved one. That can't be right?

timholy:
What do folks need from me to help move this investigation forward? I certainly don't know the answer to @KristofferC's questions, can I ping @00vareladavid?

PR #1783, opened 2020-04-19
Support subdir from the REPL.

### LINKED ISSUE #1759: Add a way to use the subdir functionality from the REPL
It is now possible to add packages that are not in the root of a repository by giving the `subdir` command to a `PackageSpec`. But just like we have the `#branch` syntax in the REPL it would be nice to have some syntax for adding a subdir from the REPL as well.

Just as a starting point, it could be something like

```
pkg> add https://github.com/JuliaLang/Package.jl/[subdir]
```

but I'm sure there are nicer/smarter ways.

GunnarFarneback:
Fixes #1759.

### DIFF
--- a/src/REPLMode/argument_parsers.jl
+++ b/src/REPLMode/argument_parsers.jl
@@ -46,12 +46,18 @@ function parse_package_args(args::Vector{PackageToken}; add_or_dev=false)::Vecto
     function apply_modifier!(pkg::PackageSpec, args::Vector{PackageToken})
         (isempty(args) || args[1] isa PackageIdentifier) && return
         modifier = popfirst!(args)
+        if modifier isa Subdir
+            pkg.repo.subdir = modifier.dir
+            (isempty(args) || args[1] isa PackageIdentifier) && return
+            modifier = popfirst!(args)
+        end
+
         if modifier isa VersionRange
             pkg.version = VersionSpec(modifier)
         elseif modifier isa Rev
             pkg.repo.rev = modifier.rev
-        else # modifier isa Subdir
-            pkg.repo.subdir = modifier.dir
+        else
+            pkgerror("Package name/uuid must precede subdir specifier `[$arg]`.")
         end
     end
 
@@ -66,7 +72,9 @@ function parse_package_args(args::Vector{PackageToken}; add_or_dev=false)::Vecto
         else
             arg isa VersionRange ?
                 pkgerror("Package name/uuid must precede version specifier `@$arg`.") :
-                pkgerror("Package name/uuid must precede revision specifier `#$(arg.rev)`.")
+            arg isa Rev ?
+                pkgerror("Package name/uuid must precede revision specifier `#$(arg.rev)`.") :
+                pkgerror("Package name/uuid must precede subdir specifier `[$arg]`.")
         end
     end
     return pkgs
--- a/src/REPLMode/argument_parsers.jl
+++ b/src/REPLMode/argument_parsers.jl
@@ -14,11 +14,16 @@ end
 struct Rev
     rev::String
 end
+
+struct Subdir
+    dir::String
+end
+
 const PackageIdentifier = String
-const PackageToken = Union{PackageIdentifier, VersionRange, Rev}
+const PackageToken = Union{PackageIdentifier, VersionRange, Rev, Subdir}
 
 const package_id_re =
-    r"((git|ssh|http(s)?)|(git@[\w\-\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)? | [^@\#\s]+\s*=\s*[^@\#\s]+ | \#\s*[^@\#\s]* | @\s*[^@\#\s]* | [^@\#\s]+"x
+    r"((git|ssh|http(s)?)|(git@[\w\-\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)? | [^@\#\s]+\s*=\s*[^@\#\s]+ | \#\s*[^@\#\s]* | @\s*[^@\#\s]* | \[[^\s]+\] | [^@\#\s\[]+"x
 
 function package_lex(qwords::Vector{QString})::Vector{String}
     words = String[]
@@ -33,6 +38,7 @@ end
 PackageToken(word::String)::PackageToken =
     first(word) == '@' ? VersionRange(word[2:end]) :
     first(word) == '#' ? Rev(word[2:end]) :
+    first(word) == '[' && last(word) == ']' ? Subdir(word[2:end-1]) :
     String(word)
 
 function parse_package_args(args::Vector{PackageToken}; add_or_dev=false)::Vector{PackageSpec}
@@ -42,8 +48,10 @@ function parse_package_args(args::Vector{PackageToken}; add_or_dev=false)::Vecto
         modifier = popfirst!(args)
         if modifier isa VersionRange
             pkg.version = VersionSpec(modifier)
-        else # modifier isa Rev
+        elseif modifier isa Rev
             pkg.repo.rev = modifier.rev
+        else # modifier isa Subdir
+            pkg.repo.subdir = modifier.dir
         end
     end
 
--- a/src/REPLMode/argument_parsers.jl
+++ b/src/REPLMode/argument_parsers.jl
@@ -23,7 +23,7 @@ const PackageIdentifier = String
 const PackageToken = Union{PackageIdentifier, VersionRange, Rev, Subdir}
 
 const package_id_re =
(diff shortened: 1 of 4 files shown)


GunnarFarneback:
Until #1784 is merged this fails with a username prompt:
```
(testproject) pkg> dev --local https://github.com/DilumAluthge/FooBar.git/[myfolder/Foo]
    Cloning git-repo `https://github.com/DilumAluthge/FooBar.git/`
                                   ERROR: failed to clone from https://github.com/DilumAluthge/FooBar.git/, error: GitError(Code:EUSER, Class:Callback, Aborting, user cancelled credential request.)                                           
```
It can be worked around with the not so pretty
```
(testproject) pkg> dev --local https://github.com/DilumAluthge/FooBar.git[myfolder/Foo]
    Cloning git-repo `https://github.com/DilumAluthge/FooBar.git`
  Resolving package versions...
Updating `/tmp/testproject/Project.toml`
  [ac29586c] + Foo v0.1.0 `dev/FooBar.git/myfolder/Foo`
Updating `/tmp/testproject/Manifest.toml`
  [ac29586c] + Foo v0.1.0 `dev/FooBar.git/myfolder/Foo`
```
which works but probably should strip `.git` from the target directory.

Add of subdir from path works:
```
(testproject) pkg> add /tmp/testproject/dev/FooBar.git/[otherfolder/subfolder/Bar]
    Cloning git-repo `/tmp/testproject/dev/FooBar.git`
   Updating git-repo `/tmp/testproject/dev/FooBar.git`
  Resolving package versions...
Updating `/tmp/testproject/Project.toml`
  [f73876a8] + Bar v0.1.0 `/tmp/testproject/dev/FooBar.git/[otherfolder/subfolder/Bar]#master`
Updating `/tmp/testproject/Manifest.toml`
  [f73876a8] + Bar v0.1.0 `/tmp/testproject/dev/FooBar.git/[otherfolder/subfolder/Bar]#master`
[...]
```

src/REPLMode/argument_parsers.jl
```diff
@@ -33,6 +38,7 @@ end
 PackageToken(word::String)::PackageToken =
     first(word) == '@' ? VersionRange(word[2:end]) :
     first(word) == '#' ? Rev(word[2:end]) :
+    first(word) == '[' && last(word) == ']' ? Subdir(word[2:end-1]) :
```
GunnarFarneback, line 41: What is the preferred unicode-correct way to do this?

KristofferC: `chop(word; head=1, tail=1)`

GunnarFarneback: Thanks.

GunnarFarneback:
This is ready for review.

fingolfin:
Above you give two examples, one formerly not working but now headed with "This is fine now"; then another with "Still works but not needed". But to me it looks as if the input is identical in both cases (the output differs), namely
```
(testproject) pkg> dev --local https://github.com/DilumAluthge/FooBar.git/[myfolder/Foo]
```
Am I missing something?

src/REPLMode/command_declarations.jl
```diff
@@ -103,7 +103,7 @@ that matches, or ranges thereof, such as `@1.2-3.4.5`. A git revision can be
 specified by `#branch` or `#commit`.
 
 If a local path is used as an argument to `add`, the path needs to be a git repository.
-The project will then track that git repository just like it would track a remote repository online.
+The project will then track that git repository just like it would track a remote repository online. If the package is not located at the top of the git repository, a subdirectory can be specified with `path/[subdir/path]`.
```
fingolfin, line 106: I am not sure if this project has any conventions for this, but perhaps avoid overlong lines?
They are annoying to read in the web UI and in diffs, and the rest of the comments seems to
avoid them too?
```suggestion
The project will then track that git repository just like it would track a remote repository online.
 If the package is not located at the top of the git repository, a subdirectory can be specified
 with `path/[subdir/path]`.
```

GunnarFarneback:
> Am I missing something?

There's no slash before `[myfolder/foo]` in the second example.

KristofferC:
Thanks for fixing this.

I guess the only thing that should be discussed is the `url[subdir]` syntax. I used it for the status printing and this is consistent with that but I think it is worth spending a few brain cycles on if that is the best. Perhaps it would be better with a single delimiter (like we have with `#master`)? Personally I am fine with `[]` though.

GunnarFarneback:
I agree that a single delimiter (`:` maybe?) could be more elegant but I also don't have a strong opinion about it. It's a bit of a fringe feature anyway since it's not needed for a normal `add` or `develop` from registry, nor for `develop` from path.

fredrikekre:
I like `:`.

DilumAluthge:
Can we add the triage label to this PR? @KristofferC @fredrikekre

DilumAluthge:
(To bikeshed the syntax, etc.)

fredrikekre:
`:` seems to be accepted, so would be good to switch to that and get this in 1.5.

DilumAluthge:
> `:` seems to be accepted, so would be good to switch to that and get this in 1.5.

Works for me :)

In that case, is there anything else that needs to be done before this PR can be merged?

GunnarFarneback:
Actually implementing `:` is still needed but shouldn't be too hard. I can look into that later today or tomorrow.

DilumAluthge:
@fredrikekre Will there be enough time to get this into 1.5? Keno said on Slack that today is the 1.5 feature freeze deadline. So we'd need to merge this into Pkg master, and also we'd need to make the PR to Julia master that updates the commit of Pkg.

fredrikekre:
Yea, no worries.

KristofferC:
Should also change the printing in Pkg.status, but that is quite trivial.

GunnarFarneback:
Well, it turns out that `:` is a bit more tricky than it might seem. Specifically, we have the infamous regexp (before adding subdir):
```
const package_id_re =
    r"((git|ssh|http(s)?)|(git@[\w\-\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)? | [^@\#\s]+\s*=\s*[^@\#\s]+ | \#\s*[^@\#\s]* | @\s*[^@\#\s]* | [^@\#\s]+"
```
This consists of 5 pieces:
```
URL:       ((git|ssh|http(s)?)|(git@[\w\-\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)?
NAME=UUID: [^@\#\s]+\s*=\s*[^@\#\s]+ 
#BRANCH:   \#\s*[^@\#\s]* 
@VERSION:  @\s*[^@\#\s]*
NAME:      [^@\#\s]+"
```
Now you can see that the `URL` regexp is fairly strict, in particular it needs to have `.git` at the end. However, Pkg has been able to handle more free-form URLs because they have been picked up by the `NAME` regexp and in later processing been identified as looking like URLs. This includes e.g. one of the tests in `new.jl`:
```
        api, args, opts = first(Pkg.pkg"add https://github.com/00vareladavid/Unregistered.jl#0.1.0")
```
This has worked okay because `#` is uncommon in URLs and `@` is not entirely critical.

However, if we add a `:SUBDIR` component to this, we have to exclude `:` from `NAME`. But `:` is highly critical in an URL and we will only be able to handle URLs that actually match the `URL` regexp. In particular the URL above from the tests turns into
```
name = "https"
subdir = "//github.com/00vareladavid/Unregistered.jl"
rev = "0.1.0"
```
and nothing good will come of that.

StefanKarpinski:
Seems reasonable to require subdir to start with one or more non-slash characters, no?

GunnarFarneback:
> Seems reasonable to require subdir to start with one or more non-slash characters, no?

Sure, but that's not enough to salvage a URL that doesn't match the `URL` regexp, is it?

src/REPLMode/argument_parsers.jl
```diff
@@ -14,11 +14,43 @@ end
...
+    dir::String
+end
+
 const PackageIdentifier = String
-const PackageToken = Union{PackageIdentifier, VersionRange, Rev}
+const PackageToken = Union{PackageIdentifier, VersionRange, Rev, Subdir}
+
+    # Match a git repository URL. This includes uses of `@` and `:` but
+    # requires that it has `.git` at the end.
+let url = raw"((git|ssh|http(s)?)|(git@[\w\-\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)(/)?",
```
KristofferC, line 27: Thanks for splitting this up a bit!

KristofferC:
Good to go?

GunnarFarneback:
I have done what I wanted and I don't think the Travis failure is related, so yes.

GunnarFarneback:
I'll push a minor comments only update.

PR #1784, opened 2020-04-19
Remove trailing slash from URL.

GunnarFarneback:
**Before**:
```
(@v1.5) pkg> add https://github.com/JuliaLang/Example.jl.git
```
adds `Example` as expected, whereas adding a trailing slash asks for username and password:
```
(@v1.5) pkg> add https://github.com/JuliaLang/Example.jl.git/
   Updating git-repo `https://github.com/JuliaLang/Example.jl.git/`
Username for 'https://github.com': 
Password for 'https://github.com': 
ERROR: failed to fetch from https://github.com/JuliaLang/Example.jl.git/, error: GitError(Code:EUSER, Class:Callback, Aborting, user cancelled credential request.)
```
**After**:
Both without and with trailing slash add `Example`.

If you think that this is the wrong way to solve the problem, please object.

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -85,6 +85,8 @@ function setprotocol!(;
 end
 
 function normalize_url(url::AbstractString)
+    # LibGit2 is fussy about trailing slash. Make sure there is none.
+    url = rstrip(url, '/')
     m = match(GIT_REGEX, url)
     m === nothing && return url
 
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -743,4 +743,14 @@ end
     end end
 end
 
+# PR #1784 - Remove trailing slash from URL.
+@testset "URL with trailing slash" begin
+    temp_pkg_dir() do project_path
+        with_temp_env() do
+            Pkg.add(Pkg.PackageSpec(url = "https://github.com/JuliaLang/Example.jl.git/"))
+            @test isinstalled("Example")
+        end
+    end
+end
+
 end # module


src/GitTools.jl
```diff
@@ -85,6 +85,10 @@ function setprotocol!(;
 end
 
 function normalize_url(url::AbstractString)
+    # GitLib2 is fussy about trailing slash. Make sure there is none.
```
fredrikekre, line 88:
```suggestion
    # libgit2 is fussy about trailing slash. Make sure there is none.
```

fredrikekre:
Does this also happen when just using `LibGit2.clone`? I guess we should report it upstream then.

GunnarFarneback:
Yes, it does.

src/GitTools.jl
```diff
@@ -85,6 +85,10 @@ function setprotocol!(;
 end
 
 function normalize_url(url::AbstractString)
+    # libgit2 is fussy about trailing slash. Make sure there is none.
+    while endswith(url, "/")
```
KristofferC, line 89: `rsplit(url, "/")[1]`?

KristofferC:
```
julia> a = "fooα/"
"fooα/"

julia> a[1:end-1]
ERROR: StringIndexError("fooα/", 5)
```

fredrikekre: `strip(url, '/')` removes them all

StefanKarpinski: I would suggest replacing all of this with `url = rstrip(url, '/')`

GunnarFarneback: Finally we have a winner.

test/pkg.jl
```diff
@@ -743,4 +743,14 @@ end
     end end
 end
 
+# PR #1784 - Remove trailing slash from URL.
+@testset "URL with trailing slash" begin
+    temp_pkg_dir() do project_path
+        with_temp_env() do
+            Pkg.add(Pkg.PackageSpec(url = "https://github.com/JuliaLang/Example.jl.git/"))
```
fredrikekre, line 750: This doesn't test the clone path though? But maybe it failed the download path too?

GunnarFarneback: Without the `rstrip` patch it stops with a username prompt at least.

GunnarFarneback: And as far as I can tell, `add` from URL clones.

fredrikekre: Yea you are right, since we need to checkout master.

PR #1786, opened 2020-04-20
Add uncaught precompile statements

ianshmean:
Investigating slow package loading on MacOS, I was using https://github.com/JuliaLang/julia/pull/34613 and I noticed a lot of inference being spent on these steps during the first time a package is loaded in a session. These calls were all high in the inference ranked list.

I've not fully tested this. 
But I'm not sure it's fixing the root issue, as I did test by manually running these before re-monitoring the calls generated during loading another package.. and that didn't prevent them from being recompiled


KristofferC:
I think these are in fact getting precompiled from the workload we have above. The problem is that they get invalidated so they have to be *re*-compiled.

Also, note that we can't put anonymous functions like this in the list since they might vary depending on OS / architecture. This is the reason we have a workload that we run instead of a list of statements.

ianshmean:
Ok, makes sense. I thought these were additional

PR #1794, opened 2020-04-24
add pinning commit mesage

Datseris:
Superseedes #1670 , cc @KristofferC  .

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -82,10 +82,13 @@ we can explicitly track that branch (or commit) by appending `#branchname` (or `
 
 The status output now shows that we are tracking the `master` branch of `Example`.
 When updating packages, updates are pulled from that branch.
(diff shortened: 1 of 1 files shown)


docs/src/managing-packages.md
```diff
@@ -82,6 +82,9 @@ we can explicitly track that branch (or commit) by appending `#branchname` (or `
 
 The status output now shows that we are tracking the `master` branch of `Example`.
 When updating packages, we will pull updates from that branch.
+Notice that if instead of a branch name we specified an exact commit id, e.g.
```
fredrikekre, line 85: !!! note
    If we would specify a commit id instead of a branch name, e.g.
    `add Example#025cf7e`, then we would effectively "pin" the package
    to that commit. This is because the commit id always point to the same
    thing unlike a branch, which may be updated.

Datseris:
I liked Frederick's suggestion of a Note, so I added it.

Issue #374, opened 2018-06-13
Garbage collector should remove unreachable clones as well


KristofferC:

StefanKarpinski:
Or maybe all clones? Since they're not usually needed.

00vareladavid:
Exactly what I was thinking. It hardly seems worth the effort if it is just a cache. It would be a step forward at least.

KristofferC:
Seems like a step backward for me heh. Trading some disk space for having to redownload a bunch of things you are likely currently using.

StefanKarpinski:
But we don't usually need to clone anything to install packages. What would the criteria for keeping a clone be? We could also do `git gc --aggresive` in each git repo.

fredrikekre:
In my experience `git gc --aggressive` often takes much more time compared to cloning the repo, so in that case it seems better to just remove the repo.

StefanKarpinski:
Wait do we have both `~/.julia/clones` and `~/.julia/.clones` now?

KristofferC:
We never had `.clones`, I wrote it wrong.

PR #1799, opened 2020-04-27
also clear out clones in gc

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/374

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -412,7 +412,7 @@ function gc(ctx::Context=Context(); collect_delay::Period=Day(7), kwargs...)
     all_manifest_files = filter(f -> endswith(f, "Manifest.toml"), all_index_files)
     all_artifacts_files = filter(f -> !endswith(f, "Manifest.toml"), all_index_files)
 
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -525,28 +538,33 @@ function gc(ctx::Context=Context(); collect_delay::Period=Day(7), kwargs...)
...
+
     for depot in depots()
         # We track orphaned objects on a per-depot basis, writing out our `orphaned.toml`
         # tracking file immediately, only pushing onto the overall `*_to_delete` lists if
         # the package has been orphaned for at least a period of `collect_delay`
         depot_orphaned_packages = String[]
         depot_orphaned_artifacts = String[]
+        depot_orphaned_repos = String[]
 
+        # ??: This code block is identical to one a bit above
```
staticfloat, line 567: Hmmm.  You're right.  I think there used to be differences, but they've since been abstracted away.  You can try deleting this block, I think.  I think it will just double-up everything in `depot_orphaned_packages`.

PR #1800, opened 2020-04-27
Clarify "Providing global configuration options" in docs

mbauman:
This sounded to me like a feature that I can specify at the Project.toml level or something.  Make it clearer it's an informal mandate and not a Pkg feature.


IanButterworth (approved):
Sounds reasonable to me

PR #1802, opened 2020-04-30
Document using environment with `julia` command

tbenst:


tbenst:
Incorporated feedback from Eric Forgy on slack

PR #1803, opened 2020-05-01
Disable SIGQUIT test on macOS.

fredrikekre:


fredrikekre:
Are you sure this is the test that failed? I will restart a couple of times just to make sure.

omus:
I'm not fully confident but the SIGQUIT test was mainly added due to the interesting behaviour it demonstrated. The SIGKILL one is the important one and luckily it seems very consistent.

fredrikekre:
Passed twice in a row.

PR #1806, opened 2020-05-03
Reduce footprint for invalidation

timholy:
This is a companion PR to https://github.com/JuliaLang/julia/pull/35714 and accounts for a subset of the gains reported there. The more specific implementations of `==` and `hash` avoid a `collect`;
that branch prevents many `collect`-induced invalidations via the `iteratorsize` trick, but this change is probably a good idea anyway.

The more important and subtle one is the one in `TOML._print`. This is used by a large number of methods in Pkg, so it's important to prevent it from being invalidated. Do to nospecialize markers, this method gets inferred for partially-specified types, and the `a` is typically just `AbstractDict`. However, `Iterators.Pairs` is also an AbstractDict, and the return type of `keys` cannot be inferred from the parameters of `AbstractDict{K,V}` because the `itr` field of a pair uses a later
parameter.

This short-circuits the whole problem by "hand-inlining" the result of `invoke(keys,  Tuple{AbstractDict}, a)` at the call site, i.e., just creating a `KeySet` directly.

That one-line change accounts for most of the improvement claimed in https://github.com/JuliaLang/julia/pull/35714; you can see that this accounts for more than 200 MethodInstances all by itself: https://gist.github.com/ianshmean/2825c2f9e118217c7855245e508c4f21#file-revise-invalidations-during-load-L115-L322

### DIFF
--- a/ext/TOML/src/print.jl
+++ b/ext/TOML/src/print.jl
@@ -57,7 +57,7 @@ function _print(io::IO, a::AbstractDict,
     sorted::Bool = false,
     by = identity,
 )
-    akeys = keys(a)
+    akeys = Base.KeySet(a)   # keys is non-inferrable due to Iterators.Pairs
     if sorted
         akeys = sort!(collect(akeys), by = by)
     end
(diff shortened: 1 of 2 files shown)


KristofferC:
Cool, thanks!

PR #1810, opened 2020-05-05
implement why, showing why a package is in the manifest

KristofferC:
bild

Don't know if the "Who required Package" is useful... I just took it from https://www.npmjs.com/package/npm-why.

We could also add some compat info here, for example, if a package is being upper bounded by another one but one step at a time.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -183,6 +183,19 @@ See also [`PackageSpec`](@ref), [`PackageMode`](@ref).
 const rm = API.rm
 
 """
+    Pkg.why(pkg::Union{String, Vector{String}})
+    Pkg.why(pkg::Union{PackageSpec, Vector{PackageSpec}})
+
+Show the reason why this package is in the manifest.
+The output is a the different way to reach the package
+through the dependency graph starting from the dependencies.
+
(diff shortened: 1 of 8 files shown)


PetrKryslUCSD:
In https://discourse.julialang.org/t/where-do-all-the-packages-come-from/43135/6 I outlined why one might want to have this capability.

cossio:
Bump. This would be very nice to have.

src/Pkg.jl
```diff
@@ -182,6 +182,19 @@ See also [`PackageSpec`](@ref), [`PackageMode`](@ref).
...
+"""
+    Pkg.why(pkg::Union{String, Vector{String}})
+    Pkg.why(pkg::Union{PackageSpec, Vector{PackageSpec}})
+
+Show the reason why this package is in the manifest.
+The output is a the different way to reach the package
+through the dependency graph starting from the dependencies.
+
+!!! compat "Julia 1.9"
+    This function requires at least Julia 1.6.
```
IanButterworth, line 194:
```suggestion
    This function requires at least Julia 1.9.
```

IanButterworth (approved):
Nice!

PR #1813, opened 2020-05-06
Pass on HashType while recursing.

GunnarFarneback:
What's the point of specifying a different hash type if it reverts to the default in the recursive calls?


GunnarFarneback:
Also changed the hard-coded 20 in the computation of the content size to the actual size of the hash.

StefanKarpinski:
Thanks for the bug fix!

PR #1817, opened 2020-05-08
TOML: dont overload Base.error

KristofferC:
This comment scared me:

https://github.com/JuliaLang/julia/blob/ddf7ce9a595b0c84fbed1a42e8c987a9fdcddaac/base/compiler/params.jl#L20-L21


StefanKarpinski:
This isn't even a justifiable way to extend the function in the first place. :100: to this change.

KristofferC:
> This isn't even a justifiable way to extend the function in the first place.

Yeah, that too :D

PR #1823, opened 2020-05-13
`pin` without arguments

heliosdrm:
Attempt to extend `pin`, to keep all direct dependencies in their current version.
Context: https://discourse.julialang.org/t/pin-all-packages-of-a-project/39308


heliosdrm:
I have tried to investigate the traceback of the error during the test, but I'm lost...

00vareladavid:
I believe this is superseded by #2432

PR #1824, opened 2020-05-13
Improve inference of call chains to TOML.print

timholy:
This is a better fix for #1806, one that resolves the inference problems. A big part of the problem is the dreaded 15276.

Addresses concerns in https://github.com/JuliaLang/www.julialang.org/pull/794#discussion_r424454133.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -629,7 +629,7 @@ function bind_artifact!(artifacts_toml::String, name::String, hash::SHA1;
             end
         end
     else
-        artifact_dict = Dict()
+        artifact_dict = TOML.DictType()
     end
 
     # Otherwise, the new piece of data we're going to write out is this dict:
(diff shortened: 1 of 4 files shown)


src/Artifacts.jl
```diff
@@ -673,8 +673,10 @@ function bind_artifact!(artifacts_toml::String, name::String, hash::SHA1;
     end
 
     # Spit it out onto disk
-    open(artifacts_toml, "w") do io
-        TOML.print(io, artifact_dict, sorted=true)
+    let artifact_dict = artifact_dict
```
KristofferC, line 676: I'm surprised this is needed, there is no assignment to `artifact_dict` so shouldn't the frontend be able to figure this out?

timholy: I'm not a hundred percent certain but the codegen shows some try/catch stuff that appears to complicate the analysis. In practice, adding this suddenly gave much nicer codegen.

PR #1827, opened 2020-05-15
Fix ~/nested/path expansion

### LINKED ISSUE #1826: Incorrect expansion of ~/nested/paths in Pkg REPL mode
With the following nested directory structure

```
❯ tree ~/foo
/Users/harry/foo
└── bar
    └── baz
```

If I type the following in the REPL in Pkg mode

```
(@v1.4) pkg> dev ~/foo/bar/baz
```

It autocompletes to

```
(@v1.4) pkg> dev /Users/harry/baz/
```

The path is correctly expanded by `Base.expanduser` though

```
julia> expanduser("~/foo/bar/baz")
"/Users/harry/foo/bar/baz"
```

I think the problem is in

https://github.com/JuliaLang/Pkg.jl/blob/a5df9271ac47415ddbc00f1d0ed1643100d803da/src/REPLMode/completions.jl#L41

and is fixed by

```julia
completions = [joinpath(dirname(s), x) for x in completions]
```

harryscholes:
Fixes #1826

### DIFF
--- a/src/REPLMode/completions.jl
+++ b/src/REPLMode/completions.jl
@@ -38,7 +38,7 @@ function complete_local_dir(s, i1, i2)
         if length(completions) == 1 && endswith(joinpath(homedir(), ""), first(completions))
             completions = [joinpath(s, "")]
         else
-            completions = [joinpath(homedir(), x) for x in completions]
+            completions = [joinpath(dirname(s), x) for x in completions]
         end
         return completions, i1:oldi2, true
     end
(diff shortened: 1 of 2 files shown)


src/REPLMode/completions.jl
```diff
@@ -38,7 +38,7 @@ function complete_local_dir(s, i1, i2)
         if length(completions) == 1 && endswith(joinpath(homedir(), ""), first(completions))
             completions = [joinpath(s, "")]
         else
-            completions = [joinpath(homedir(), x) for x in completions]
+            completions = [joinpath(dirname(s), x) for x in completions]
         end
         return completions, i1:oldi2, true
```
harryscholes, line 43: Should this now be `i1:i2`?

harryscholes:
Btw I've added a test and it passes

KristofferC:
Thanks!

PR #1832, opened 2020-05-20
Include in ParseError the filename and line-info

oxinabox:
# Without this PR
```julia
(@v1.6) pkg> gc
ERROR: Pkg.TOML.ParserError(119214, 119245, "expected a newline after a key")
```

# With this PR
```julia
julia> import Pkg; Pkg.pkg"gc"
[ Info: Precompiling Pkg [e9ea8145-4ac0-5781-a202-fc5c5b8eb05a]
ERROR: Pkg.TOML.ParserError(2594, 2594, "/Users/oxinabox/.julia/logs/manifest_usage.toml", "expected a newline after a key")
```

--- 
Now I can go and fix that file that has a missing newline

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,4 +1,5 @@
 name = "Pkg"
+uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 keywords = ["package management"]
 license = "MIT"
 desc = "The next-generation Julia package manager."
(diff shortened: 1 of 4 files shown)


KristofferC:
This is a good idea. For 1.6 I think we will try to include the parser in https://github.com/JuliaLang/julia/pull/36018 which has a bit better error messages:

bild

Issue #796, opened 2018-10-05
Sequester mutable state outside of package directories


staticfloat:
I think it would be desirable to have packages use something similar to `Pkg.package_state_dir(@__MODULE__)` or something as the default location where e.g. binaries, datasets, etc.. should be stored. This would be constructed as an overall per-environment state directory (overridable by an environment variable or environment config key perhaps), that then has hashed subdirectories similar to `~/.julia/packages` but explicitly including information to disambiguate julia OS, arch, calling ABI, GCC ABI and package options.  This has multiple benefits;

* Packages become more "immutable".  It would be lovely to be certain that the entire tree hash of a package directory inside of `~/.julia/packages` never changes.

* Packages get automatically pushed toward greater relocatability.  As recent experiments with PackageCompiler have shown, broad-spectrum usage of things like `@__FILE__` and `@__DIR__` should be discouraged anyway.  A common use case is for creating a scratch space for binaries (e.g. `<pkg dir>/deps/usr`), but others exist (downloading datasets, generating julia code, etc...  Forcing a runtime lookup based on `@__MODULE__` is already what we need to do, so this would dovetail nicely.

* Pkg3 package resolution could technically be arch/OS agnostic.  I'm imagining the nightmare scenario where Crazy Charlie has installed three copies of Julia 1.0, one with GCC 8 targeting x86_64, one with GCC 7 targeting x86_64 and one with GCC 6 targeting i686.  Technically, we could share the actual Julia package directories, but when run from a Julia with a particular `<arch>-<os>-<calling_abi>-<gcc_abi>` (increasingly inaccurately named) triplet, the result of `Pkg.package_state_dir(@__MODULE__)` could mutate accordingly.

* The storage directory for mutable state could be decoupled from the storage for julia code.  Imagine a heterogenous cluster with various CPUs and a shared package depot that is provided to all users; not only would the `.ji` files generated differ on each machine, but the downloaded binaries could differ as well (taking advantage of different ISAs).  This separation between Julia code and build/run-time content would solve both the "must provide binaries that work on every platform globally" problem and the "I don't have permissions to modify packages placed in this depot" problem

* This could make the "nuclear" state reset option a little easier for users; instead of nuking `~/.julia` entirely (as some still do to try and fix stale state problems) they could instead nuke `~/.julia/package_state` or whatever we default the location to.  This would essentially cause a "rebuild" of every package, as if it were freshly installed, without doing more drastic things like losing the set of installed packages.

[...]

PR #1833, opened 2020-05-21
Implement Scratch Spaces

staticfloat:
This implements functionality and tests for a new `Spaces`
subsystem in `Pkg`; analogous to the `Artifacts` added in 1.3, this
provides an abstraction for a mutable datastore that can be explicitly
lifecycled to an owning package, or shared among multiple packages.

Closes https://github.com/JuliaLang/Pkg.jl/issues/796


docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
+# [**9.** Scratchspaces](@id Scratchspaces)
+
+!!! compat "Julia 1.6"
+    Pkg's Scratchspaces functionality requires at least Julia 1.6.
+
+`Pkg` can manage and lifecycle scratchspaces of temporary or readily-recreatable data.
```
KristofferC, line 6: It's not obvious to me what "lifecycle" is. Is that like garbage collect?

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+
+end # module SpaceExample
+```
+
+Note that we initialize the `download_cache` within `__init__()` so that our packages are as relocatable as possible; we typically do not want to bake absolute paths into our precompiled files.
+This makes use of the `@get_space!()` macro, which is identical to the `get_space!()` method, except it automatically determines the UUID of the calling module, if possible.
+An equivalent (but more verbose) invocation is given here:
+```julia
+function __init__()
+    global download_cache = get_space!("downloaded_files"; pkg_uuid=Base.PkgId(@__MODULE__).uuid)
```
KristofferC, line 46: This feels quite a bit too internally to have in the docs. Couldn't `get_space!` take either a `Module` or a `UUID` and then dispatch accordingly?

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+end
+```
+
+If a user wishes to manually delete a scratchspace, the method `delete_space!(key; pkg_uuid)` is the natural analog to `get_space!()`, however in general users will not need to do so, the spaces will be garbage collected by `Pkg` automatically.
+
+For a full listing of docstrings and methods, see the [Spaces Reference](@ref) section.
+
+## Usecases
```
KristofferC, line 54:
```suggestion
## Use cases
```

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+
+Good usecases for a Pkg scratchspace include:
```
KristofferC, line 56:
```suggestion
Good use cases for a Pkg scratch space include:
```

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+!!! compat "Julia 1.6"
+    Pkg's Scratchspaces functionality requires at least Julia 1.6.
+
+`Pkg` can manage and lifecycle scratchspaces of temporary or readily-recreatable data.
+These spaces can contain datasets, text, binaries, or any other kind of data that would be convenient to store, but which is non-fatal to have garbage collected if it has not been accessed recently, or if the owning package has been uninstalled.
+As compared to [Artifacts](@ref), these containers of data are mutable and should be treated as ephemeral; all usage of scratchspaces should assume that the data stored within them could be gone by the next time your code is run.
+In the current implementation, scratchspaces are removed during Pkg garbage collection if the space has not been accessed for a period of time (see the `spaces_cleanup_period` keyword argument to [Pkg.gc](@ref)), or if the owning package has been removed.
+Users can also request a full wipe of all spaces to clean up unused disk space through `Pkg.Spaces.clear_spaces!()`.
+
+## API overview
```
KristofferC, line 12: I wonder if this should also mention the case when a scrach file is " shared among multiple packages."

KristofferC: Or is that done by just leaving out the `pkg_uuid` arg in `get_space!`?

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+
+* Generated data that depends on the characteristics of the host system.  Examples are compiled binaries, fontcache system font folder inspection output, generated CUDA bitcode files, etc...  Objects that would be difficult to compute off of the user's machine, and that can be recreated without user intervention are a great fit.
+
+* Directories that should be shared between multiple packages in a single depot.  The space keying mechanism (explained above) makes it simple to provide scratch space that can be shared between different versions of a package, or even between different packages.  This allows packages to provide a scratch space where other packages can easily find the generated data, however the typical race condition warnings apply here; always design your access patterns assuming another process could be reading or writing to this scratch space at any time.
+
+Bad usecases for a Pkg scratchspace include (but are not limited to):
```
KristofferC, line 64:
```suggestion
Bad use cases for a Pkg scratch space include (but are not limited to):
```

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+
+Yes!  Make use of the `key` parameter and Pkg's ability to look up the current version of your package at compile-time:
+
+```julia
+module VersionSpecificExample
+using Pkg, Pkg.Spaces
+
+# Helpers to get current package UUID and VersionNumber
+function get_uuid()
+    return Base.PkgId(@__MODULE__).uuid
```
KristofferC, line 98: Same with internal stuff here. I think we need an api that is not based on `PkgId`.

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+end
+function get_version(uuid)
+    ctx = Pkg.Types.Context()
```
KristofferC, line 101: This is also internals of `Pkg`.

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+    uuid, entry = first(filter(((u, e),) -> u == uuid, ctx.env.manifest))
+    return entry.version
+end
+
+# Get the current version at compile-time, that's fine it's not going to change. ;)
+const v = get_version(get_uuid())
```
KristofferC, line 107: If all the above was needed for this we should add an official API for it.

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+
+# This will be filled in by `__init__()`; it might change if we get deployed somewhere
+version_specific_space = ""
```
KristofferC, line 110: I wonder if it is in general better to advice people to use

```
version_specific_space = Ref{String}()
```

andb then 

```
version_specific_space[] = ...
```

in `__init__`?

That makes you able (for non isbitstypes) to actually check that the value has been set instead of not knowing whether you have the original `""` or it just got set to `""`.

docs/src/spaces.md
```diff
@@ -0,0 +1,155 @@
...
+        rm(artifact_dir)
+        cp(space_dir, artifact_dir)
+    end
+
+    # Archive artifact out to a tarball
+    mktempdir() do upload_dir
+        tarball_path = joinpath(upload_dir, "$(space_name).tar.gz")
+        tarball_hash = archive_artifact(hash, tarball_path)
+
+        # Upload tarball to a hosted site somewhere.  Note; this function does not exist:
```
KristofferC, line 141: Perhaps this should be written in a way that would make it easy to manually upload the tarball as well. Now with the `mktempdir` it would delete the directory when the scope exit so there is no point where you can manually do things.

docs/src/api.md
```diff
@@ -87,3 +87,19 @@ Pkg.Artifacts.ensure_all_artifacts_installed
...
+
+```@docs
+Pkg.Spaces.get_space!
+Pkg.Spaces.@get_space!
+Pkg.Spaces.delete_space!
+Pkg.Spaces.clear_spaces!
+Pkg.Spaces.with_spaces_directory
+Pkg.Spaces.spaces_dir
+Pkg.Spaces.space_path
+Pkg.Spaces.track_space_access
```
StefanKarpinski, line 104: I hate to be that guy, but these all read better to me with `scratch` in place of `space`. Space is quite generic. `get_scratch!`, `delete_scratch!`, `scratch_dir`, etc.

docs/src/spaces.md
staticfloat: Yes; lifecycle means how things get created and deleted.  Does this read better:

```
`Pkg` can manage and automatically garbage collect scratchspaces of temporary or readily-recreatable data.
```

docs/src/spaces.md
staticfloat: I'm confused; this is `get_space!()` taking in a `UUID` and dispatching accordingly?  We could of course have a method that takes in a `Module` as well.  Is that what you're suggesting?

docs/src/spaces.md
staticfloat: Yes, precisely.  That's what is meant by a "global" space.

docs/src/spaces.md
staticfloat: If this is too internalsy, then we need to add an API to Pkg that gives you the version of the currently-running package.  I've needed to use that in a few packages now.

docs/src/spaces.md
staticfloat: Hmmm, what situations are you thinking of that could allow for `version_specific_space` to _not_ be set?  It's within `__init__()`, so it should be guaranteed to either be set before any other (non-compile-time) code can run, or it will have thrown an error.

docs/src/spaces.md
staticfloat: ah, you mean literally take either a `Module` or `UUID`, I understand now.  I mis-parsed your sentence.  :)

staticfloat:
Thanks for the review guys, I think I have addressed all comments so far.

docs/src/scratch.md
```diff
@@ -0,0 +1,150 @@
...
+function get_dataset_dir()
+    dataset_dir = @get_scratch!("dataset")
+    if isempty(readdir(dataset_dir))
+        perform_expensive_dataset_generation(dataset_dir)
+    end
+    return dataset_dir
+end
+```
+
+> Can I create a scratchs pace that is not shared across versions of my package?
```
StefanKarpinski, line 87: “scratch space” although I like “scratch’s pace”

docs/make.jl
```diff
@@ -37,6 +37,7 @@ makedocs(
         "compatibility.md",
         "registries.md",
         "artifacts.md",
+        "spaces.md",
```
StefanKarpinski, line 40:
```suggestion
        "scratchspaces.md",
```

docs/src/scratch.md
```diff
@@ -0,0 +1,150 @@
+# [**9.** Scratch Spaces](@id Scratch-Spaces)
+
+!!! compat "Julia 1.6"
+    Pkg's Scratch Spaces functionality requires at least Julia 1.6.
+
+`Pkg` can manage and automatically garbage collet scratch spaces of temporary or readily-recreatable data.
```
StefanKarpinski, line 6:
```suggestion
`Pkg` can manage and automatically garbage collect scratch spaces of temporary or readily-recreatable data.
```

staticfloat:
This is good to go on my end.  :)

tkf:
Can this be used as a stable place for external "virtual environment" for _each_ Julia project?  I'm thinking to optionally create a conda environment or Python venv for each `Project.toml` path.  Would something like this

```julia
project_slug = Base.slug(Base._crc32c(Base.active_project()), 5)
scratch_name = "$other_things-$project_slug"
scratch_path = @get_scratch!(scratch_name)
```

work?

docs/src/scratchspaces.md
```diff
@@ -0,0 +1,150 @@
+# [**9.** Scratch Spaces](@id Scratch-Spaces)
+
+!!! compat "Julia 1.6"
+    Pkg's Scratch Spaces functionality requires at least Julia 1.6.
+
+`Pkg` can manage and automatically garbage collect scratch spaces of temporary or readily-recreatable data.
```
tkf, line 6: > temporary or readily-recreatable

This phrasing makes me a bit worried for using it for something like conda environments. But, IIUC, as long as the relevant `Project.toml` exists, the scratch space won't be GC'ed?

staticfloat: No, in the current implementation, if a package doesn't access it often enough, it will be reaped.

Perhaps this isn't how we want it to be done, and we instead want it to exist for as long as the owning Project.toml exists.  I'm open to either interpretation, but the use cases I have in mind are always such that it's okay that the underlying scratch space gets emptied every now and then. (They were originally called "caches", but we decided the name was too generic)

If you have something that should persist in a stronger way, I suggest just using a folder on-disk somewhere.

tkf: Thanks for the explanation. I guess it makes sense to eagerly GC things in auto-generated non-human-readable paths.

staticfloat: The default collection policy here is to wait 21 days after the last `@get_scratch!()` that targeted it, then move it to an orphanage and then clean it with everything else a week later, so on average it should take 3-4 weeks for something to get GC'ed.  I figured this was probably generous enough.

fingolfin: One thing that is not 100% clear to me from the Pkg.jl documentation: are garbage collections only triggered by the user explicitly asking for it, or also in other circumstances?

docs/src/scratchspaces.md
```diff
@@ -0,0 +1,150 @@
...
+
+For a full listing of docstrings and methods, see the [Scratch Space Reference](@ref) section.
+
+## Use cases
+
+Good use cases for a Pkg scratch space include:
+
+* Caching downloads of files that must be routinely accessed and modified by a package.  Files that must be modified are a bad fit for the immutable [Artifacts](@ref) abstraction, and files can always be re-downloaded if the cache is wiped by the user.
+
+* Generated data that depends on the characteristics of the host system.  Examples are compiled binaries, fontcache system font folder inspection output, generated CUDA bitcode files, etc...  Objects that would be difficult to compute off of the user's machine, and that can be recreated without user intervention are a great fit.
```
fingolfin, line 59: "compiled binaries" is the usecase I'd have in mind; but I am somewhat worried about the eager GC behavior, as recompilation can take a non-trivial amount of time. I guess I could print a message like "parts of this package were garbage collected and need to be regenerated, please wait" ?

staticfloat:
> Would something like this work?

Yes!  You can see that I actually do this in my compile time preferences package.

staticfloat:
I think the main sticking point on this PR is the GC time; I move that we discuss this on the next Pkg call (this coming Tuesday, at 2:15pm EST) and figure out what we do and don't want to support.

Possibilities include:

* Keep the current behavior; things get GC'ed after a set period of inactivity.
* Extend the timeline
* Allow for per-scratchspace control over when things get GC'ed by default.
* Only GC scratch spaces once the owning package has been removed (and in the case of no owning package; never)

docs/src/scratchspaces.md
staticfloat: Right now, it's only when users manually run `gc()`.  In the future, we may run `gc()` automatically.

StefanKarpinski:
> Only GC scratch spaces once the owning package has been removed

This seems like it might be the right default.

> (and in the case of no owning package; never)

When would there be no owning package?

staticfloat:
> When would there be no owning package?

In the event that someone uses `@get_scratch!()` from a module like `Main`, or `Base`.  E.g. in the REPL.

tkf:
> * Only GC scratch spaces once the owning package has been removed (and in the case of no owning package; never)

What is "owning package" in this context? For example, if

1. PkgA 1.0 creates a scratch space (which is an optional behavior)
2. PkgA 2.0 is installed in another project
3. PkgA 1.0 is removed

then, does PkgA own a scratch space?

Also, why not let projects own the scratch space? Wouldn't it work with above scenario? It'd mean `~/.julia/logs/manifest_usage.toml` becomes the ultimate GC root, I guess? I don't know the primary usecase of the log file, though.

staticfloat:
Right, so that means the scratch space would persist until ALL versions of PkgA are removed.

We don't want scratch spaces to be project-specific unless the user wants them to be; that would unnecessarily cripple them.  The files within the `logs` directory do end up being kind of like a GC root though.  :P

tkf:
> We don't want scratch spaces to be project-specific unless the user wants them to be; that would unnecessarily cripple them.

I don't know the implementation detail but, in principle, isn't it possible to reference the same scratch space from multiple projects?  With Julia as an analogy, I'm imagining something like

```julia
                                      # Analogy
scratch = Ref(:scratch)               # @get_scratch!(scratch_name) from PkgA
julia_logs = Dict()                   # ~/.julia/logs/manifest_usage.toml
julia_logs[:Project_A] = []           # PATH/TO/Project_A/Manifest.toml with PkgA
push!(julia_logs[:Project_A],         # PkgA uses/creates the scratch space
      :PkgA => scratch)
julia_logs[:Project_B] = []           # PATH/TO/Project_B/Manifest.toml with PkgA
push!(julia_logs[:Project_B],         # PkgA re-uses the scratch space
      :PkgA => scratch)
julia_logs[:Project_C] = []           # PATH/TO/Project_C/Manifest.toml with PkgA
# push!(julia_logs[:Project_C],       # (PkgA does not use the scratch space inside
#       :PkgA => scratch)             # Project C.)
scratch = nothing
```

So, at this point, reference graph is:

```
julia_logs -> Project_A -> scratch
julia_logs -> Project_B -> scratch
julia_logs -> Project_C
```

So, removing Project A and B would remove the scratch space:

```julia
pop!(julia_logs, :Project_A)          # Project A removed
GC.gc()                               # => `scratch` still exists
pop!(julia_logs, :Project_B)          # Project B removed
GC.gc()                               # => `scratch` can be CG'ed
```

This is in contrast with the model where `PkgA` directly owns the scratch space:

```julia
PkgA = [Ref(:scratch)]
julia_logs = Dict()
julia_logs[:Project_A] = [PkgA]
julia_logs[:Project_B] = [PkgA]
julia_logs[:Project_C] = [PkgA]

pop!(julia_logs, :Project_A)          # Project A removed
GC.gc()                               # => `scratch` still exists
pop!(julia_logs, :Project_B)          # Project B removed
GC.gc()                               # => `scratch` still exists due to :Project_C
```

staticfloat:
We discussed on the Pkg call how to make the lifecycle more reliable here, and we came to the decision that it would be best to just persist scratch spaces for as long as the owning manifest exists.  This makes it easy and should reduce fear of package authors, while still allowing old versions of packages whose scratch spaces are no longer used by newer versions of packages to get GC'ed.

Issue #616, opened 2018-08-11
Pkg.add should display nearest names


scls19fr:
Hello,

When doing `Pkg.add("NameOfPackage")`, if `NameOfPackage` is not found an error message such as
```
ERROR: The following package names could not be resolved:
 * NameOfPackage (not found in project, manifest or registry)
Please specify by known `name=uuid`.
```
is shown.

If a package name is not found, maybe `Pkg` could try to "help" users by listing name of some packages whom name is quite near to what user is looking for.

Computing string similarity using for example [dice coefficient](https://en.wikipedia.org/wiki/S%C3%B8rensen%E2%80%93Dice_coefficient) (see [various implementations](https://en.wikibooks.org/wiki/Algorithm_Implementation/Strings/Dice%27s_coefficient)) between user provided package name and name of each registered package could help.

Maybe comparison should be done after upper casing (or lower casing) both.
A threshold could probably be set.
Sorting by descending coefficient will be required, taking only (for example) five nearest names (however displaying them with correct case)

Kind regards

StefanKarpinski:
This is definitely a good feature to have. Isn't edit distance more commonly used for this?

scls19fr:
Several kind of methods for measuring strings similarity exist :
- Levenstein distance
- Jaro distance
- Jaro-Wrinkler distance
- Dice coefficient
- N-Gram similarity
- Cosine similarity
- Jaccard similarity
- Longest common subsequence
- Hamming distance

to name a few

https://en.wikipedia.org/wiki/String_metric can give a first idea about string metric but this paper gives probably a better overview
https://hpi.de/fileadmin/user_upload/fachgebiete/naumann/folien/SS13/DPDC/DPDC_12_Similarity.pdf 

I personally used Dice coefficient in https://github.com/scls19fr/arduino_libraries_search and was quite happy with that choice.

I noticed that several Julia packages that could help to calculate string similarity exist. 

To name a few :

- https://github.com/johnmyleswhite/SimilarityMetrics.jl (from @johnmyleswhite - unmaintained)
- https://github.com/JuliaStats/Distances.jl (Julia 0.7 - contributors @KristofferC @IainNZ @alyst @lindahua @juliohm @timholy ...)
- https://github.com/matthieugomez/StringDistances.jl (from @matthieugomez - Julia 0.7 - uses Distances.jl ) 

Maybe these contributors can give us some advises for choosing a "good distance" function for such an use case.

[...]

PR #1834, opened 2020-05-21
add package search

kczimm:
Trying to add a Pkg search feature after reading discussion in #616. Uses ~~an implementation of the Damerau-Levenshtein string distance~~ REPL.levsort to compare search names against those in the available registries.


kczimm:
I'm not sure how to test this with the REPL interface. I've only tested it by,
```
import Pkg
Pkg.search(["helloworld", "and", "other", "package", "names"])
```

oxinabox:
Test Pkg repl mode using `Pkg.pkg"search Exampol"`

kczimm:
Thanks, @oxinabox. I'll give it a try.

kczimm:
Okay, cool. That failed. I'll have to figure out where/how to hook into that functionality.

oxinabox:
Also rather than implementing own D.L distance can just use `REPL.levsort` which is a perfectly fine string metric.

kczimm:
Whoa, that's great. I didn't know that existed. I was just following the discussion in #616.

kczimm:
TODO: tests

kczimm:
So maybe this isn't that useful, as one can just tab-complete inside of pkg>. Any thoughts @oxinabox ?

StefanKarpinski:
No, I like it. We just need someone to review and then we can merge it, imo. Anyone strongly against having this?

src/API.jl
```diff
@@ -6,6 +6,7 @@ using UUIDs
 using Printf
 import Random
 using Dates
+using REPL
```
StefanKarpinski, line 9: Would be better to move the `levsort` somewhere common, but that may be beyond the scope of this PR.

oxinabox: Several packages in the wild depend on `REPL.levsort` so moving it would break them.
THe ones i am aware of are **NamedColors.jl** and **DataFrames.jl**

oxinabox: Maybe we should make it clear why we are depending on `REPL`
```suggestion
using REPL: levsort
```

KristofferC:
I think tab completion is good enough for simple things and for more complicated things it is better to use https://juliahub.com/ui/index.html

PR #1840, opened 2020-05-27
Clean up `jl_XXX-download.gz.sha256` files

staticfloat:
Because PlatformEngines assumes you may want to verify a tarball later, it spits out `.sha256` files to optimize for hashing in the future.  This is fine when you're downloading to a permanent location or a directory that was created by `mktempdir()`, but if your download location is just a file sitting in `/tmp` which you clean up manually afterward, you need to clean up that `.sha256` file as well.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -1226,7 +1226,9 @@ function download_verify_unpack(
         unpack(tarball_path, dest; verbose=verbose)
     finally
         if remove_tarball
-            Base.rm(tarball_path)
+            Base.rm(tarball)
+            # Remove cached tarball hash, if it exists.
+            Base.rm(string(tarball, ".sha256"); force=true)
         end
     end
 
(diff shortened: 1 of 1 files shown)


src/PlatformEngines.jl
```diff
@@ -1226,7 +1226,9 @@ function download_verify_unpack(
         unpack(tarball_path, dest; verbose=verbose)
     finally
         if remove_tarball
-            Base.rm(tarball_path)
+            Base.rm(tarball)
+            # Remove cached tarball hash, if it exists.
+            Base.rm(string(tarball, ".sha256"); force=true)
```
fredrikekre, line 1231: `tarball` does not seem to be defined, should it be
```suggestion
            Base.rm(tarball_path)
            # Remove cached tarball hash, if it exists.
            Base.rm(string(tarball_path, ".sha256"); force=true)
```
as the original code?

staticfloat: yeah, I don't know what happened.

PR #1841, opened 2020-05-28
some perf improvements

KristofferC:
Benchmark is `@btime Pkg.add("DifferentialEquations")` in the Pkg project.

Before: `215.759 ms (2557554 allocations: 189.00 MiB)`
After: `180.268 ms (1958939 allocations: 152.31 MiB)`

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -148,8 +148,6 @@ end
 ####################
 # Registry Loading #
 ####################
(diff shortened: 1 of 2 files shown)


KristofferC:
Added another commit taking it down to

```
  156.459 ms (1706811 allocations: 147.01 MiB)
```

Issue #1411, opened 2019-09-23
Build failures should probably be more fatal


KristofferC:
Right now, a build failure prints an error message but then nothing actually fails. The project and manifest files are still updated. The package can still be `import`ed. It doesn't really make sense, if a package fails to build then it is unusable and we shouldn't pretend that everything is fine.

There are two ways to handle this:

- Just remove the package from `packages` if it fails to build. On next `add` it will get downloaded again and another build attempt will be tried.
- Somehow keep the package around but indicate that it hasn't been built yet. On next `add` there is no need to download the package but we know that it hasn't been built so let's try that again. The problem here is that maybe the package got in a bad state from failing to build the first time.

pfitzseb:
The clean-slate approach seems much more robust *and* easier to implement, so I'd be in favour of that, fwiw.

StefanKarpinski:
This is how Pkg2 used to work, but people complained, so we changed it. They wanted the build directory to stay around so that they could go in and fix the build. That always struck me as a bad idea for anyone who’s not a fairly expert user, which was apparently correct (from some reports we've gotten from various places) but we never hear about it because non-expert users don’t post on GitHub. So I don't think the "clean slate" approach is quite right. So I think the approach of marking the package as not successfully built is what we'd have to go with, to give expert users a chance to go in and fix the build and non-expert users a clear, persistent indication that something is broken. Of course hopefully all of this is mooted by Binary Builder and Artifacts.

KristofferC:
> The approach of marking the package as not successfully built is what we'd have to go with, to give expert users a chance to go in and fix the build 

What I am mostly focusing on here is `Pkg.add` where you cannot go in and fix the build. For `dev` it arguably makes sense to leave the cloned folder in place.

pfitzseb:
Yeah, imho it doesn't really make any sense to try and fix build problems for `add`ed packages, especially since we're at least trying to make sure those are immutable.

StefanKarpinski:
> What I am mostly focusing on here is `Pkg.add` where you cannot go in and fix the build.

Why can't you? The `deps/` directory is writeable and is where the build step puts its results.

KristofferC:
You can't change the `build.jl` file and the directory is in a really weird place. No one is going to go in there and fix something, they'll dev it in that case.

StefanKarpinski:
Ok, so maybe suggest that they dev and try to fix the build if it fails?

fingolfin:
[...]

PR #1851, opened 2020-06-04
Make build errors fatal

simonbyrne:
Sort of addresses #1411.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -940,7 +940,7 @@ function build_versions(ctx::Context, uuids::Vector{UUID}; might_need_to_resolve
             else
                 "", ""
             end
-            @error "Error building `$(pkg.name)`$last_lines: \n$log_show$full_log_at"
+            error("Error building `$(pkg.name)`$last_lines: \n$log_show$full_log_at")
         end
     end
     return
--- a/test/new.jl
+++ b/test/new.jl
@@ -273,12 +273,6 @@ end
     end
 end
 
-@testset "build: fallback when no project file exists" begin
-    isolate() do
-        Pkg.add(name="ZMQ", version="0.6.3")
-    end
-end
-
 @testset "using a test/REQUIRE file" begin
     isolate() do
         Pkg.add(name="EnglishText", version="0.6.0")


simonbyrne:
Where would be a good place to add tests?

KristofferC:
Imo this should just be unconditionally true (like in pre 1.0)

ExpandingMan:
I'm not so sure... I'm in the middle of a situation right now where I have one package where the build failing is unavoidable, and the only way to possibly fix it is to try to do this in the build step of the package that's being installed.  This is of course *not* my preference of how to deal with this, but if this were always fatal, I think I'd really be up a creek.

KristofferC:
Use a try catch then? Like in all other cases with exceptions.

DilumAluthge:
> Imo this should just be unconditionally true (like in pre 1.0)

If this happens, I think it has the potential to break packages that currently assume that build failures are not fatal.

If we want to make this the default behavior, I think that we should first do a full PkgEval run to see how many packages it would break.

ExpandingMan:
How can I use try/catch if the thing I need to catch is in another packages build step?

The point is, if a packages build is breaking, and this fatally broke the build of everything else, it would be very hard to resolve it without forking the package that's breaking.

I don't have any specific suggestions here, obviously trying to hack a broken build after the fact is horrible anyway, I'm just pointing out a situation I'm encountering.

simonbyrne:
> I'm not so sure... I'm in the middle of a situation right now where I have one package where the build failing is unavoidable, and the only way to possibly fix it is to try to do this in the build step of the package that's being installed

Can you expand on how this happens?

ExpandingMan:
It's a situation with [Gurobi](https://github.com/JuliaOpt/Gurobi.jl) which has generally been a pain in the ass with these sort of things (referring to Gurobi itself, not the Julia wrapper).  I've wound up needing to let Gurobi.jl try to build itself, but if it fails I fetch Gurobi binaries and explicitly call `Pkg.build("Gurobi")` explicitly in my build step.

I suppose I should probably just make a PR to Gurobi.jl to make this unnecessary (which I will when I get a chance), but there seems to be a general point that if we make a build step failing be *necessarily* fatal, there may be circumstances in which it's almost impossible to circumvent.

That said, I'm *NOT* advocating that users should be encouraged to go through these sorts of awful hacks, it's definitely not ideal.  I think what we need is a clever, well-thought-out solution, but I'm afraid I can't offer any at the moment.

simonbyrne:
That's a good point: we do need a good pattern to allow for optional failures, especially when the dependencies can't be provided as artifacts.

KristofferC:
> If this happens, I think it has the potential to break packages that currently assume that build failures are not fatal.

What does this even mean? This is talking about when packages are getting installed, before any code is running.

ExpandingMan:
> What does this even mean? This is talking about when packages are getting installed / built, before any code is running.

I think the concern is about a failed build step from one package breaking the build steps of all other packages.  Now that I think more about it, maybe it's not so bad that these are completely independent, there just needs to be more transparency about what happened in all the build steps.

Maybe we all have different notions of what "fatal" means here?

KristofferC:
> I think the concern is about a failed build step from one package breaking the build steps of all other packages. 

That is not what happens. After all build steps are done, an exception is thrown if one of the packages failed to build.

KristofferC:
Or that is what is supposed to happen.. I haven't looked at the actual PR yet heh

StefanKarpinski:
Seems like there's an unresolved question here; @KristofferC has proposed that rather than threading yet another option through all the functions, that we should just make build failures unconditionally be errors, which a lot of people (myself included) feel is the way to go. The way to do it seems like it is to try to build everything, remember which ones didn't work and then raise a single large error at the end of the build phase indicating which builds failed with instructions on how to debug, e.g. how to dev those packages and manually try to build them.

simonbyrne:
Any reason why you wouldn't want to just bail on the first failure? You could end up in a case where 1 package build failure (e.g. Conda.jl) prevents all downstream packages from building (if their build steps depend on Conda.jl).

StefanKarpinski:
That seems fine to me as well.

simonbyrne:
Okay, changed to a fatal error.

simonbyrne:
So the problem is that previously this swallowed errors that ZMQ.jl failed to build:
https://travis-ci.org/github/JuliaLang/Pkg.jl/jobs/713035241#L266
and now it doesn't
https://travis-ci.org/github/JuliaLang/Pkg.jl/jobs/714635245#L292

The test that is failing is there to check compatibility with the old REQUIRE format. Unfortunately this is now broken due to the removal of a function from Base, and the older version of Compat.jl doesn't work (https://github.com/JuliaLang/Compat.jl/pull/671).

Any ideas on the best way to fix this?

simonbyrne:
I've removed it for now. The REQUIRE format is still being tested by Permutations.jl

simonbyrne:
bump?

StefanKarpinski:
Just to clarify, this now makes build failures strictly fatal, right? If so, I'm fine with it.

simonbyrne:
> Just to clarify, this now makes build failures strictly fatal, right? If so, I'm fine with it.

Yes

PR #1852, opened 2020-06-05
Package registration requires upper bounds

mbaz:
Currently, this document does not make it clear that package registration in the General Registry requires upper bounds. I think it's worth it to include that information here, since often packages are developed with registration in mind.

### DIFF
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -18,6 +18,9 @@ The format of the version specifier is described in detail below.
 !!! info
     There is currently no way to give compatibility from the Pkg REPL mode so for now, one has to manually edit the project file.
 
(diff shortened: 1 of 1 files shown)


DilumAluthge:
https://github.com/JuliaRegistries/General/blob/master/README.md#automatic-merging-of-pull-requests

mbaz:
@DilumAluthge Right, but my point is that the Pkg.jl docs should point to that document.

DilumAluthge:
Sure.

KristofferC:
Was going to update the link to https://github.com/JuliaRegistries/General/blob/master/README.md#automatic-merging-of-pull-requests but original repo is gone.

Issue #1855, opened 2020-06-09
Improve appearance of Conflict messsages


oxinabox:
People often complain about getting conflicts, and not knowing what to do from the logs.
I don't mind the conflicts myself, and i have learned how to read the logs to get the information i need to resolve it.

But I do think the logs are a bit of a wall of text when you see them.
So this issue proposed some thing to do to improve them.

While I am going to list a bunch of things, in order,
they don't all have to be done, nor done in this order.
But I only wanted to create 1 set of mockups

----

We start with a message like:

```
(@v1.4) pkg> pin CSV@0.6.2
ERROR: Unsatisfiable requirements detected for package CSV [336ed68f]:
 CSV [336ed68f] log:
 ├─possible versions are: [0.3.0-0.3.1, 0.4.0-0.4.3, 0.5.0-0.5.26, 0.6.0-0.6.2] or uninstalled
 ├─restricted to versions 0.6.2 by an explicit requirement, leaving only versions 0.6.2
 └─restricted by compatibility requirements with FilePathsBase [48062228] to versions: [0.3.0-0.3.1, 0.4.0-0.4.3, 0.5.0-0.5.13] or uninstalled — no versions left
   └─FilePathsBase [48062228] log:
     ├─possible versions are: [0.2.0, 0.3.0-0.3.1, 0.4.0, 0.5.0, 0.6.0-0.6.2, 0.7.0, 0.8.0, 0.9.0-0.9.3] or uninstalled
     └─restricted to versions 0.9.3 by an explicit requirement, leaving only versions 0.9.3
```

----

- [ ] Remove possible versions line, or Leaving only versions bit, Or Both (show)

This information is not really that helpful.
Having both is almost(but not quiet) redundant.
If we are getting rid of one but not both, i suggest the possible versions one.

One effect to bare in mind is humans are fairly good at spotting this kind of satisfiablity problem (its basically easier sudoku),
and actually working it out probably makes them happier about the error message (yay i solved a problem),
and front loads into there thoughts the packages they are going to have to workout how to change.

- [ ] make explict version restriction line, like compatiblity requriment line say "by ... restricted to ...", *not* "to ... by ...".

This just makes it visually easier to parse since the lines will all follow same structure.
Esp with colors as proposed further down.

These changes would take us to:
```
(@v1.4) pkg> pin CSV@0.6.2
ERROR: Unsatisfiable requirements detected for package CSV [336ed68f]:
 CVS@0.6.2 incompatible
[...]

PR #1863, opened 2020-06-11
Compress compat version in compat logs

oxinabox:
This does one of the suggestions from https://github.com/JuliaLang/Pkg.jl/issues/1855

I think it is a big improvement.

- [x] It needs tests.
- [x] Especially around the edge cases where it is broken into multiple discontigious runs

### Current Master
```julia
julia> using Pkg; pkg"add DataFrames@0.17 CSV@0.5"
#...
ERROR: Unsatisfiable requirements detected for package CSV [336ed68f]:
 CSV [336ed68f] log:
 ├─possible versions are: [0.3.0-0.3.1, 0.4.0-0.4.3, 0.5.0-0.5.26, 0.6.0-0.6.2] or uninstalled
 ├─restricted to versions 0.5 by an explicit requirement, leaving only versions 0.5.0-0.5.26
 └─restricted by compatibility requirements with DataFrames [a93c6f00] to versions: [0.3.0-0.3.1, 0.4.0-0.4.3] or uninstalled — no versions left
   └─DataFrames [a93c6f00] log:
     ├─possible versions are: [0.11.7, 0.12.0, 0.13.0-0.13.1, 0.14.0-0.14.1, 0.15.0-0.15.2, 0.16.0, 0.17.0-0.17.1, 0.18.0-0.18.4, 0.19.0-0.19.4, 0.20.0-0.20.2, 0.21.0-0.21.2] or uninstalled
     └─restricted to versions 0.17 by an explicit requirement, leaving only versions 0.17.0-0.17.1
```

### With this PR
```julia
julia> using Pkg; pkg"add DataFrames@0.17 CSV@0.5"
#...
ERROR: Unsatisfiable requirements detected for package CSV [336ed68f]:
 CSV [336ed68f] log:
 ├─possible versions are: 0.3.0-0.6.2 or uninstalled
 ├─restricted to versions 0.5 by an explicit requirement, leaving only versions 0.5.0-0.5.26
 └─restricted by compatibility requirements with DataFrames [a93c6f00] to versions: 0.3.0-0.4.3 or uninstalled — no versions left
   └─DataFrames [a93c6f00] log:
     ├─possible versions are: 0.11.7-0.21.2 or uninstalled
     └─restricted to versions 0.17 by an explicit requirement, leaving only versions 0.17.0-0.17.1
```

### DIFF
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -276,15 +276,15 @@ Base.show(io::IO, s::VersionSpec) = print(io, "VersionSpec(\"", s, "\")")
 
 
 """
-    compressed_versionspec(pool, subset=pool)
+    range_compressed_versionspec(pool, subset=pool)
 
  - `pool`: all versions that exist (a superset of `subset`)
  - `subset`: the subset of those versions that we want to include in the spec.
 
-Works out a compressed `VersionSpec` that permits everything in the `subset`,
-and nothing else from the `pool`.
+Finds a minimal collection of ranges as a `VersionSpec`, that permits everything in the
+`subset`, but does not permit anything else from the `pool`.
 """
-function compressed_versionspec(pool, subset=pool)
+function range_compressed_versionspec(pool, subset=pool)
     length(subset)==1 && return VersionSpec(only(subset))
     # PREM-OPT: we keep re-sorting these, probably not required.
     sort!(pool)
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -285,34 +285,30 @@ Works out a compressed `VersionSpec` that permits everything in the `subset`,
 and nothing else from the `pool`.
 """
 function compressed_versionspec(pool, subset=pool)
-    # PREM-OPT: we keep resorting these, probably not required.
+    length(subset)==1 && return VersionSpec(only(subset))
+    # PREM-OPT: we keep re-sorting these, probably not required.
     sort!(pool)
     sort!(subset)
-    # PREM-OPT: no need to store all the elements in contiguous_subsets, just can store version range of end-points
-    contiguous_subsets = Vector{VersionNumber}[]
-    local cur_subset
+    @assert subset ⊆ pool
 
-    pool_ii = 1
-    in_run = false
-    for s in subset
-        if !in_run
-            while s != pool[pool_ii]
-                pool_ii += 1
-                @assert(pool_ii <= length(pool), "Exhaused pool without finding element of subset. this is impossile")
+    contiguous_subsets = VersionRange[]
+
+    range_start = first(subset)
+    pool_ii = findfirst(isequal(range_start), pool) + 1  # skip-forward til we have started
+    for s in @view subset[2:end]
+        if s != pool[pool_ii]
+            range_end = pool[pool_ii-1]  # previous element was last in this range
+            push!(contiguous_subsets, VersionRange(range_start, range_end))
+            range_start = s  # start a new range
+            while (s != pool[pool_ii])  # advance til time to start
+                pool_ii+=1
             end
-            cur_subset = Vector{VersionNumber}()
-            push!(contiguous_subsets, cur_subset)
-            in_run = true
-        end
-        # in_run
-        if s == pool[pool_ii]
-            push!(cur_subset, s)
-        else
-            in_run=false
         end
         pool_ii+=1
     end
-    return VersionSpec(VersionRange.(first.(contiguous_subsets), last.(contiguous_subsets)))
+    push!(contiguous_subsets, VersionRange(range_start, last(subset)))
+
+    return VersionSpec(contiguous_subsets)
 end
 
 ###################
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -130,6 +130,8 @@ end
 VersionRange(b::VersionBound=VersionBound()) = VersionRange(b, b)
 VersionRange(t::Integer...)                  = VersionRange(VersionBound(t...))
 VersionRange(v::VersionNumber)               = VersionRange(VersionBound(v))
+VersionRange(lo::VersionNumber, hi::VersionNumber) = VersionRange(VersionBound(lo), VersionBound(hi))
+
 
 function VersionRange(s::AbstractString)
     m = match(r"^\s*v?((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*)(?:\s*-\s*v?((?:\d+(?:\.\d+)?(?:\.\d+)?)|\*))?\s*$", s)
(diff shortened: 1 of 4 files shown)


KristofferC:
Hm, wiring `possible versions are: 0.3.0-0.6.2` is pretty pointless to write since it is quite rare just the first and last version is of any particular use. The previous printing was compactified but still gave you all information of available versions.

oxinabox:
> Hm, wiring possible versions are: 0.3.0-0.6.2 is pretty pointless to write since it is quite rare just the first and last version is of any particular use. The previous printing was compactified but still gave you all information of available versions.

Our experience of resolving these seems to be very different.
I only ever need to know the range support, so i know how much i need to adjust version bounds.
What am I going to do with a list of all available versions?
I personally have never had a use for that list when I am resolving this.

I can find a list of all avaible version by looking at release notes on github (and intersecting with the range shown), which will also have release notes and will often tell me more about why compat was dropped etc.

oxinabox:
Or are you specifically talking about the "All possible versions" line?
rather than the other lines about restrictions?

I find that line almost entirely useless and was intending a follow up PR to discuss deleting it. 
It is on the list in #1855.
It does have some use to contextualize some of the other lines about restrictions.
So I am not sure deleting it is nessecarily the best thing to do, but compressing it seems like a net win to decrase how much noise it is adding

singularitti:
What will this PR print if you have a minor version skipped, e.g., `[0.6, 0.7, 0.8, 1.0]` and `[0.4, 0.6, 0.8]`? Will this lead users try to add `0.6, 0.7, 0.8, 0.9, 1.0` and `0.4, 0.5, 0.6, 0.7, 0.8` to their `[compat]` fields in `Project.toml`?

Could a compromised solution be like:
```julia
julia> using Pkg; pkg"add DataFrames@0.17 CSV@0.5"
#...
ERROR: Unsatisfiable requirements detected for package CSV [336ed68f]:
 CSV [336ed68f] log:
 ├─possible versions are: [0.3+, 0.4+, 0.5+, 0.6+] or uninstalled
 ├─restricted to versions 0.5 by an explicit requirement, leaving only versions 0.5.0-0.5.26
 └─restricted by compatibility requirements with DataFrames [a93c6f00] to versions: [0.3+, 0.4+] or uninstalled — no versions left
   └─DataFrames [a93c6f00] log:
     ├─possible versions are: [0.11.7, 0.12+, 0.13+, 0.14+, 0.15+, 0.16+, 0.17+, 0.18+, 0.19+, 0.20+, 0.21+] or uninstalled
     └─restricted to versions 0.17 by an explicit requirement, leaving only versions 0.17.0-0.17.1
```
What usually matters to me when writing `[compat]` is the major and minor versions, I don't care about patch versions.

DilumAluthge:
Personally I (like @KristofferC) find it helpful to have all of the versions listed out. I don't particular like this "compressed range" format- it hides too much information.

Just my personal preference.

DilumAluthge:
I think that e.g. adding colors and stuff like that will help make these messages more readable.

But personally I don't want to remove any information from the messages.

oxinabox:
> What will this PR print if you have a minor version skipped, e.g., `[0.6, 0.7, 0.8, 1.0]` or `[0.4, 0.6, 0.8]`? Will this lead users try to add `0.6, 0.7, 0.8, 0.9, 1.0` and `0.4, 0.5, 0.6, 0.7, 0.8` to their `[compat]` fields in `Project.toml`? I have seen packages skipping versions.


It is compressed relative to what exists so question can't be answered without that.
From the tests: if everything that exists is:
`[v"1.0.0", v"1.1.0", v"1.2.0", v"1.2.1", v"2.0.0", v"2.0.1", v"3.0.0", v"3.1.0"]`

and it is compatible with 1.x and 3.x but not 2.x, i.e. with:
`[v"1.0.0", v"1.1.0", v"1.2.0", v"1.2.1", v"2.0.0", v"2.0.1", v"3.0.0", v"3.1.0"]`

Then the shown versions will be: `1.0.0-1.2.1, 3.0.0-3.1.0`


---
> ...  I don't want to remove any information from the messages.

We need to balance the amount of information.
There is a trade-off between presenting every piece of information we know, vs keeping the amount of information presented from being overwelming.
We want to optimize the signal to noise ratio, presenting the useful facts that change what actions people do (to more efficent ones), without presenting too much excess that they ignore

@DilumAluthge  for my own understanding can you explain how you use the version numbers in resolving conflicts?
I think understanding your process  (which is clearly different from mine) will help understand what informat is signal, and what is noise.


My process is:
1. if it is a direct dependency, look at possible versions  and versions allowed under restrictions, and if I am not getting to the top then see about relaxing those requirements.
2. if it is further down stream, then use the bounds to workout what is being most restrictive and go to github for that package where I am going to look at Project.tomls, potentially on various tagged releases, so I can see what is allowed. I am potentially looking forward and back and at master to see what is allowed when. (This is also why it is so annoying to me to have the Project.toml and registry out of sync). From the Github page I am likely to then make a PR.

So I have no need to know about every single release in the Pkg screen, because I am about to go to github where i can look at those relases and immidately see even more information about them -- each releases exact compatibilities.

DilumAluthge:
> 2\. if it is further down stream, then use the bounds to workout what is being most restrictive and go to github for that package where I am going to look at Project.tomls, potentially on various tagged releases, so I can see what is allowed. I am potentially looking forward and back and at master to see what is allowed when. (This is also why it is so annoying to me to have the Project.toml and registry out of sync). From the Github page I am likely to then make a PR.
> 
> So I have no need to know about every single release in the Pkg screen, because I am about to go to github where i can look at those relases and immidately see even more information about them -- each releases exact compatibilities.

You can't do that. You have to go to the General registry, go to each package, open up Versions.toml and Compat.toml, etc.

The registry is the ultimate source of truth. You cannot rely on Git tags/GitHub releases, and you cannot rely on Project.toml files. The Pkg resolver uses the General registry only, so that's what I have to look at.

So if we don't print that information in the error message, it forces me to go to the General registry. It's much better in my opinion to just have these versions listed out in the error message.

@KristofferC What do you think?

oxinabox:
@dillumaluthge
Could you tell me your process?
Including, but not only what you use these numbers for

src/versions.jl
```diff
@@ -273,6 +275,42 @@ end
...
+"""
+    compressed_versionspec(pool, subset=pool)
+
+ - `pool`: all versions that exist (a superset of `subset`)
+ - `subset`: the subset of those versions that we want to include in the spec.
+
+Works out a compressed `VersionSpec` that permits everything in the `subset`,
+and nothing else from the `pool`.
+"""
+function compressed_versionspec(pool, subset=pool)
```
oxinabox, line 287: Pkg-Dev call wants to rename this because it is not quite the same as what compress means in the registry tools.
Maybe: `ranged_versionspec` ?

StefanKarpinski:
I'm ok with this since although the "possible versions" have less information, the relevant information is still included on the following "restricted to" and "restricted by" lines. Using the example in the PR body, yes, `0.3.0-0.6.2` as a list of possible versions is a bit vague, but then you immediately see what the important subsets of that are, namely `0.5.0-0.5.26` and `0.3.0-0.4.3`. Does it matter whether a `0.3.2` release exists? No, it does not. That is not relevant information. What matters is that these two ranges are disjoint. If there were a `0.3.2` version and it was compatible, then, yes, that would be relevant, but then again (a) these ranges wouldn't be printed the way they are and (b) we wouldn't be seeing this resolver error in the first place. In short, I'm in favor of this change as it makes it easier to see what's happening with a resolver error.

src/versions.jl
StefanKarpinski: `versionspec_range_string`?

KristofferC:
Ok, I'll know which people to cc then for debugging / figuring out resolver errors.

StefanKarpinski:
Stop being cranky 🙂

oxinabox:
Ok should be good to go now.

KristofferC:
Wow, I did not mean to close the PR... That was a finger slip. Sorry about that.

StefanKarpinski:
Let's give it a try and see how it goes 🤷‍♂️. It's hard to predict if something like this will be more or less confusing to people without testing it out on actual people.

PR #1864, opened 2020-06-12
README: howto build Julia with git checkout of Pkg

timholy:
Thanks to @fredrikekre.


KristofferC:
It's not clear to me where the cloned Pkg repo should live. In `stdlib`?

fredrikekre:
Yea, thats where it will be created by the build system.

timholy:
Seemed worth clarifying the location, thanks for noticing.

PR #1865, opened 2020-06-12
Add Color to the conflict log

oxinabox:
This adds color to the log, to enhance readability, per another point in  #1855 

## WIth this PR + #1863 
image

## This PR only:
image

## #1855 only:
image

## Current julia master
image

`pkgID`s are set using modulo of their  `hash`
Which I think is nice, because it avoids having global state to track which colors we have used, and means seeing the same package in different conflicts has a persistant recognizable hue.
On the downside there is the chance that package colors would conflict, so we could do it in order instead. Seems less fun.

TODO:
- [x] check this works right on terminals without 256 color
- [x] check this doesn't color things if we have run `julia --color=false`

### DIFF
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -582,7 +582,13 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
-color_string(c, str...) = string(Base.text_colors[c], str..., Base.text_colors[:default])
+function color_string(c, str...)
+    if stderr[:color]  #check if color has been disabled by `julia --color=no`
+        string(Base.text_colors[c], str..., Base.text_colors[:default])
+    else
+        string(str...)
+    end
+end
 pkgID_color(pkgID) = 17 + hash(pkgID) % 216  # Give each package a probably unique color
 logstr(pkgID, args...) = color_string(pkgID_color(pkgID), args...)
 logstr(pkgID) = logstr(pkgID, pkgID)
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -582,9 +582,6 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
-
-
-
 # system colors excluding whites/greys/blacks and error-red
 const CONFLICT_COLORS = [1:6; 10:14];
 pkgID_color(pkgID) = CONFLICT_COLORS[mod1(hash(pkgID), end)]
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -587,8 +587,18 @@ function color_string(c, str...)
     return sprint((io, args) -> printstyled(io, args...; color=c), str, context=stderr)
 end
 
-const CONTRAST_COLORS =  [21:51; 55:119; 124:142; 160:184; 196:220];
-pkgID_color(pkgID) = CONTRAST_COLORS[1 + hash(pkgID) % end]  # Give each package a probably unique color
+# system colors excluding greys, bright-yellow and dark-blue
+const CONFLICT_COLORS = [5,1,6,3,2,13,12,9,10];
+const pkgID_color = let  # Cycle through each color in turn when a new color is needed
+    pkgID2color = Dict{String, Int}()
+    cur_color_index = 0;
+    function pkgID_color(pkgID)
+        get!(pkgID2color, pkgID) do
+            cur_color_index+=1
+            CONFLICT_COLORS[mod1(cur_color_index, end)]
+        end
+    end
+end
 logstr(pkgID, args...) = color_string(pkgID_color(pkgID), args...)
 logstr(pkgID) = logstr(pkgID, pkgID)
 
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -582,6 +582,11 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
+color_string(c, str...) = string(Base.text_colors[c], str..., Base.text_colors[:default])
+pkgID_color(pkgID) = 17 + hash(pkgID) % 216  # Give each package a probably unique color
+logstr(pkgID, args...) = color_string(pkgID_color(pkgID), args...)
+logstr(pkgID) = logstr(pkgID, pkgID)
+
 function init_log!(data::GraphData)
     np = data.np
     pkgs = data.pkgs
@@ -592,11 +597,15 @@ function init_log!(data::GraphData)
         id = pkgID(p0, data)
         versions = pvers[p0]
         if isempty(versions)
-            msg = "$id has no known versions!" # This shouldn't happen?
+            msg = "$(logstr(id)) has no known versions!" # This shouldn't happen?
         else
-            msg = "possible versions are: $(range_compressed_versionspec(versions)) or uninstalled"
+            vspec = range_compressed_versionspec(versions)
+            vers = logstr(id, vspec)
+            msg = "possible versions are: $vers or uninstalled"
         end
-        first_entry = get!(rlog.pool, p) do; ResolveLogEntry(rlog.journal, p, "$id log:") end
+        first_entry = get!(rlog.pool, p) do
+            ResolveLogEntry(rlog.journal, p, "$(logstr(id)) log:")
+        end
 
         if p ≠ uuid_julia
             push!(first_entry, (nothing, msg))
@@ -609,7 +618,7 @@ end
 function log_event_fixed!(graph::Graph, fp::UUID, fx::Fixed)
     rlog = graph.data.rlog
     id = pkgID(fp, rlog)
-    msg = "$id is fixed to version $(fx.version)"
+    msg = "$(logstr(id)) is fixed to version $(logstr(id, fx.version))"
     entry = rlog.pool[fp]
     push!(entry, (nothing, msg))
     return entry
@@ -621,7 +630,7 @@ function log_event_req!(graph::Graph, rp::UUID, rvs::VersionSpec, reason)
     pdict = graph.data.pdict
     pvers = graph.data.pvers
     id = pkgID(rp, rlog)
-    msg = "restricted to versions $rvs by "
+    msg = "restricted to versions $(logstr(id, rvs)) by "
     if reason isa Symbol
         @assert reason === :explicit_requirement
         other_entry = nothing
@@ -632,14 +641,15 @@ function log_event_req!(graph::Graph, rp::UUID, rvs::VersionSpec, reason)
             msg *= "julia compatibility requirements"
             other_entry = nothing # don't propagate the log
         else
-            other_id = pkgID(other_p, rlog)
-            msg *= "$other_id"
+            msg *= logstr(pkgID(other_p, rlog))
         end
     end
     rp0 = pdict[rp]
     @assert !gconstr[rp0][end]
     if any(gconstr[rp0])
-        msg *= ", leaving only versions $(range_compressed_versionspec(pvers[rp0], pvers[rp0][gconstr[rp0][1:(end-1)]]))"
+        vspec = range_compressed_versionspec(pvers[rp0], pvers[rp0][gconstr[rp0][1:(end-1)]])
+        vers = logstr(id, vspec)
+        msg *= ", leaving only versions $vers"
     else
         msg *= " — no versions left"
     end
@@ -660,9 +670,13 @@ function log_event_implicit_req!(graph::Graph, p1::Int, vmask::BitVector, p0::In
     pkgs = graph.data.pkgs
     pvers = graph.data.pvers
 
+    p = pkgs[p1]
+    id = pkgID(p, rlog)
+
     function vs_string(p0::Int, vmask::BitVector)
         if any(vmask[1:(end-1)])
-            vns = string(range_compressed_versionspec(pvers[p0], pvers[p0][vmask[1:(end-1)]]))
+            vspec = range_compressed_versionspec(pvers[p0], pvers[p0][vmask[1:(end-1)]])
+            vns = logstr(id, vspec)
             vmask[end] && (vns *= " or uninstalled")
         else
             @assert vmask[end]
@@ -671,8 +685,6 @@ function log_event_implicit_req!(graph::Graph, p1::Int, vmask::BitVector, p0::In
         return vns
     end
 
-    p = pkgs[p1]
-    id = pkgID(p, rlog)
     other_p, other_entry = pkgs[p0], rlog.pool[pkgs[p0]]
     other_id = pkgID(other_p, rlog)
     if any(vmask)
@@ -681,8 +693,7 @@ function log_event_implicit_req!(graph::Graph, p1::Int, vmask::BitVector, p0::In
             msg *= "julia compatibility requirements "
             other_entry = nothing # don't propagate the log
         else
-            other_id = pkgID(other_p, rlog)
-            msg *= "compatibility requirements with $other_id "
+            msg *= "compatibility requirements with $(logstr(other_id)) "
         end
         msg *= "to versions: $(vs_string(p1, vmask))"
         if vmask ≠ gconstr[p1]
@@ -698,8 +709,7 @@ function log_event_implicit_req!(graph::Graph, p1::Int, vmask::BitVector, p0::In
             msg *= "julia"
             other_entry = nothing # don't propagate the log
         else
-            other_id = pkgID(other_p, rlog)
-            msg *= "$other_id "
+            msg *= logstr(other_id)
         end
     end
     entry = rlog.pool[p]
@@ -718,7 +728,8 @@ function log_event_pruned!(graph::Graph, p0::Int, s0::Int)
     if s0 == spp[p0]
         msg = "determined to be unneeded during graph pruning"
     else
-        msg = "fixed during graph pruning to its only remaining available version, $(pvers[p0][s0])"
+        ver = logstr(id, pvers[p0][s0])
+        msg = "fixed during graph pruning to its only remaining available version, $ver"
     end
     entry = rlog.pool[p]
     push!(entry, (nothing, msg))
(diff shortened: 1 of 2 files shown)


Project.toml
```diff
@@ -1,5 +1,5 @@
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78a"
```
oxinabox, line 2:
```suggestion
uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```

src/Resolve/graphtype.jl
```diff
@@ -274,10 +274,10 @@ mutable struct Graph
             uuid = pkgs[p0]
             vnmap = get(Dict, compat[pkgs[p0]], vn)
             for (name,vs) in vnmap
-                haskey(n2u, name) || error("Unknown package $name found in the compatibility requirements of $(pkgID(pkgs[p0], uuid_to_name))")
+                haskey(n2u, name) || error("Unknown package $name found in the compatibility requirements of $(logstr_pkgID(pkgs[p0], uuid_to_name))")
                 uuid = n2u[name]
                 p1 = pdict[uuid]
-                p1 == p0 && error("Package $(pkgID(pkgs[p0], uuid_to_name)) version $vn has a dependency with itself")
+                p1 == p0 && error("Package $(logstr_pkgID(pkgs[p0], uuid_to_name)) version $vn has a dependency with itself")
```
oxinabox, line 280: This isn't a log message this is an error so probably should not change it 
```suggestion
                p1 == p0 && error("Package $(pkgID(pkgs[p0], uuid_to_name)) version $vn has a dependency with itself")
```

src/Resolve/graphtype.jl
```diff
@@ -274,10 +274,10 @@ mutable struct Graph
             uuid = pkgs[p0]
             vnmap = get(Dict, compat[pkgs[p0]], vn)
             for (name,vs) in vnmap
-                haskey(n2u, name) || error("Unknown package $name found in the compatibility requirements of $(pkgID(pkgs[p0], uuid_to_name))")
+                haskey(n2u, name) || error("Unknown package $name found in the compatibility requirements of $(logstr_pkgID(pkgs[p0], uuid_to_name))")
```
oxinabox, line 277: This isn't a log message this is an error so probably should not change it 
```suggestion
                haskey(n2u, name) || error("Unknown package $name found in the compatibility requirements of $(pkgID(pkgs[p0], uuid_to_name))")
```

src/Resolve/graphtype.jl
```diff
@@ -400,7 +400,7 @@ function _add_reqs!(graph::Graph, reqs::Requires, reason)
     pvers = graph.data.pvers
 
     for (rp,rvs) in reqs
-        haskey(pdict, rp) || error("unknown required package $(pkgID(rp, graph))")
+        haskey(pdict, rp) || error("unknown required package $(logstr_pkgID(rp, graph))")
```
oxinabox, line 403: This isn't a log message this is an error so probably should not change it 
```suggestion
        haskey(pdict, rp) || error("unknown required package $(pkgID(rp, graph))")
```

src/Resolve/graphtype.jl
```diff
@@ -432,7 +432,7 @@ function _add_fixed!(graph::Graph, fixed::Dict{UUID,Fixed})
     vdict = graph.data.vdict
 
     for (fp,fx) in fixed
-        haskey(pdict, fp) || error("unknown fixed package $(pkgID(fp, graph))")
+        haskey(pdict, fp) || error("unknown fixed package $(logstr_pkgID(fp, graph))")
```
oxinabox, line 435: This isn't a log message this is an error so probably should not change it 

```suggestion
        haskey(pdict, fp) || error("unknown fixed package $(pkgID(fp, graph))")
```

src/Resolve/graphtype.jl
```diff
@@ -720,12 +730,13 @@ function log_event_greedysolved!(graph::Graph, p0::Int, s0::Int)
...
+    id = logstr_pkgID(p, rlog)
     if s0 == spp[p0]
         msg = "determined to be unneeded by the solver"
     else
         if s0 == spp[p0] - 1
-            msg = "set by the solver to its maximum version: $(pvers[p0][s0])"
+            ver = logstr_version(pvers[p0][s0])
+            msg = "set by the solver to its maximum version: $ver"
         else
             msg = "set by the solver to the maximum version compatible with the constraints: $(pvers[p0][s0])"
```
oxinabox, line 741: Missed this one
```suggestion
            ver = logstr_version(pvers[p0][s0])
            msg = "set by the solver to the maximum version compatible with the constraints: $ver"
```

oxinabox:
After some discussion on Slack, have concluded that version number should be matched to the package they are restrictions upon.
Because it helps notice that the restriction CategoicalArray places is a version range for  CSV, not for Tables or CategoricalArrays

src/Resolve/graphtype.jl
```diff
@@ -570,7 +570,13 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
-color_string(c, str...) = string(Base.text_colors[c], str..., Base.text_colors[:default])
+function color_string(c, str...)
+    if stderr[:color]  #check if color has been disabled by `julia --color=no`
```
KristofferC: The `color` attribute on the io is what determines color support AFAIU.

Also, might be better to use `sprint` with `printstyled` instead of relying on internals (like `text_color`).

oxinabox: I initially tried to use `sprint` + `printstyled` but it doesn't work.
I guess related to the properties of the IO Object it uses.

image

Maybe I can create a different one that does support it?

> The color attribute on the io is what determines color support AFAIU.

Yes. but I do not have the `io` object, because it wasn't passed into all the logging functions.
So I used `stderr` which supports the this attribute also and is where the logs eventually get written. (I checked it with julia run with and without `--color=no`, it works)

oxinabox:
Bump

StefanKarpinski:
Unclear what the state of this is or what the next steps are. Are you done with it? Do you want someone to review?

oxinabox:
> Unclear what the state of this is or what the next steps are. Are you done with it? Do you want someone to review?

I am basically done with this.  
I do want someone to review, yes.

src/Resolve/graphtype.jl
KristofferC: >  I guess related to the properties of the IO Object it uses.

You need to pass `context` to `sprint` as well.

src/Resolve/graphtype.jl
```diff
@@ -570,6 +570,19 @@ function wipe_snapshots!(graph::Graph)
...
+function color_string(c, str...)
+    if get(stderr, :color, false)  #check if color is allowed
+        string(Base.text_colors[c], str..., Base.text_colors[:default])
+    else
+        string(str...)
+    end
+end
+
+const CONTRAST_COLORS =  [21:51; 55:119; 124:142; 160:184; 196:220];
+pkgID_color(pkgID) = CONTRAST_COLORS[1 + hash(pkgID) % end]  # Give each package a probably unique color
```
KristofferC, line 582: I think we should limit to the system colors since that's what we do in Base Julia for error messages. Likely means that we would use a counter to pick the color since hashing would too often give packages with the same color.

src/Resolve/graphtype.jl
```diff
@@ -570,6 +570,19 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
+function color_string(c, str...)
+    if get(stderr, :color, false)  #check if color is allowed
```
KristofferC, line 574: Use `sprint` with the `context` option.

https://github.com/JuliaLang/Pkg.jl/blob/f80d44345f61d0c1a02650f33322f9a330aa77d1/src/Operations.jl#L1661

src/Resolve/graphtype.jl
oxinabox: this is sad, they are such ugly colors by default.

oxinabox: So the system colors:
image

 - 1 (red): No, that is for warnings/errors
 - 2 (green): No, that is for success
 - 3 (off yellow): yes
 - 4 (blue): No, too low contrast with black
 - 5 (purple): yes
 - 6 (light blue): yes
 - 7 (light grey): No, too low contrast with white
 - 8 (dark grey): No too low contrast with black
 - 9 (orange): Yes
 - 10 (light green): yes
 - 11 (yellow): No, too low contrast with white
 - 12 (indigo): No, too low contrast with black
 - 13 (magenta): yes
 - 14 (cyan): no too low contrast with white
 - 15 (light grey): no too low contrast with white
 - 16 (black): no too low contrast with black.


 So the colors we have are:
 - 3 (off yellow): yes
 - 5 (purple): yes
 - 6 (light blue): yes
 - 9 (orange): Yes
 - 10 (light green): yes
 - 13 (magenta): yes

6 colors is probably enough, I guess

Color contrast being something I am distinctly aware of, being that I have pretty bad vision.
And the other colors are hard for me to read.
Though I don't have cataracts or another condition where contrast *really* starts to matter.

oxinabox: I guess if have to do this we can at least prioritize so best colors come first.
Which would be:
```
[5,9,6,3,10,13,    4,1,2,12, 11]
```
(Still avoiding the greys)

src/Resolve/graphtype.jl
oxinabox: I can't use that because i don't have a `ctx` object, or an `IO` object
and I can't (in this case) use `stderr` in its place, not sure why but it just doesn't.

I could do:
```suggestion
    if Base.have_color  #check if color is allowed
```

src/Resolve/graphtype.jl
oxinabox: Ok I had that a bit wrong amended list will be in the next PR

oxinabox:
Ok cycling though system colors but excluding the very worse ones (greys, bright yellow and dark blue).


An interesting thing (to me) is that the first color on the list actually never shows up, i guess because we write a message that we chose to never show)

ITerm2
Screenshot 2020-06-25 at 14 22 40

Terminal.app
Screenshot 2020-06-25 at 14 26 26

It is livable.
I think then that is all comments addressed and this should be good to go

KristofferC:
>  i guess because we write a message that we chose to never show

Is there any way to delay the coloring until the message is actually shown?

>  So maybe we should go back to just hashing.

I think it will be quite confusing if two packages share the same color.

oxinabox:
> Is there any way to delay the coloring until the message is actually shown?

Not without more significantly overhalling the logging code thoughout this.
I can imagine how it would be done,
to make it not work with strings but instread with some structs that have the semantic information like package ID etc,
and then by rending that all right at the end.
But that is a long way beyond what i have time for right now.
I think overall it would be a good change.

> I think it will be quite confusing if two packages share the same color.

It would be a bit confusing, i agree, but not terrible.

______

Let me break out options down:

Overall 3 options:
 - O1: Don't make any change
 - O2: Make this PR that works with strings (i.e. merge this or something like it)
 - O3: Make a new PR that works with structured logs (but that will not be by me, any time in next 3 months)

On how to chose colors  we have 2 options.

 - C1.  hashing
 - C2.  cycling

Under color list we have 2 options:
 - L1. System 16ish
 - L2. extended: 255ish


---
Some specific comments

## O3 + C2:  This gives best confidance of avoiding same color twice.
(Even better if combined with L2, but not needful).
And the fact that this option is true is an argument for doing O2 + C2 so that we can later change to O3, and not also have to write cycling code. On the other hand doing O3 is so much work that writing cycling code is hardly worth it

## C1 vs C2
Hashing (C1) gives us what is expected to be uniform distribution over colors.
Cycling (C2) is almost certainly not uniform, just chaotic, but the distribution is some characteristic of our ecosystem.
It's probably just about uniform, though i wouldn't be surprised if it e.g. avoided odd or even indexs.

Hashing does have the nice property that a given package always has the same color, which I think is one of the subtle things that humans subconciously notice really quickly.
Further it will make it more readable as one solves some conflicts but some still remain it will stay visually similar.
C2 (cycling), doesn't really even garentee that the same conflict with even a single change will remain looking at all similar (at least not with O2).

It's cute that  C2 is basically chaotic, even though C1 has the statistical randomness properties.

## L2 vs L1:

System Colors (L1), has the advantage that they are easily reconfigured via themes.
Extended colors L2, has the advantage that the much larger range means that the chance of collision (from C2 or C1) is much lower.

---

I think my preferences are: **O2, C1, L2**.
though it is no longer as strong a preference as it once was, @KristofferC  has convinced me of the virtue of L1.

Except O2, I just can't do O1 or O3.

oxinabox:
Conclusion of Pkg-Dev call was to use hashing, because it is better than cycling if we are not doing structured.
and if making it structured then changing from hashing to cycling is the smaller part of that PR

src/Resolve/graphtype.jl
```diff
@@ -570,6 +570,26 @@ function wipe_snapshots!(graph::Graph)
...
 end
 
+
+function color_string(c, str...)
+    return sprint((io, args) -> printstyled(io, args...; color=c), str, context=stderr)
+end
+
+# system colors excluding greys, bright-yellow and dark-blue
+const CONFLICT_COLORS = [5,1,6,3,2,13,12,9,10];
+const pkgID_color = let  # Cycle through each color in turn when a new color is needed
```
KristofferC, line 580: Inconsistent capitalization with e.g. the global variable just above it. It could also go inside the `let`?

src/Resolve/graphtype.jl
```diff
@@ -570,6 +570,26 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
+
+function color_string(c, str...)
```
KristofferC, line 574: Just inline this since it is only used in one place.

src/Resolve/graphtype.jl
```diff
@@ -570,6 +570,26 @@ function wipe_snapshots!(graph::Graph)
...
+end
+
+# system colors excluding greys, bright-yellow and dark-blue
+const CONFLICT_COLORS = [5,1,6,3,2,13,12,9,10];
+const pkgID_color = let  # Cycle through each color in turn when a new color is needed
+    pkgID2color = Dict{String, Int}()
+    cur_color_index = 0;
+    function pkgID_color(pkgID)
+        get!(pkgID2color, pkgID) do
+            cur_color_index+=1
```
KristofferC, line 585:
```suggestion
            cur_color_index += 1
```

src/Resolve/graphtype.jl
KristofferC: This could also just be put inside the `Graph` struct? No need for it to be global?

oxinabox:
Ok, and we should be good to go now.
This is what latest version looks like:

image

src/Resolve/graphtype.jl
```diff
@@ -582,6 +582,21 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
+
+
+
```
StefanKarpinski, line 587:
```suggestion
```

test/new.jl
```diff
@@ -2417,7 +2417,7 @@ end
             Pkg.add(Pkg.PackageSpec(uuid=exuuid, version=v"0.5.3"))
         catch e
             @test e isa ResolverError
-            @test occursin("possible versions are: 0.5.1 or uninstalled", e.msg)
+            @test occursin(r"possible versions are: .*0.5.1.* or uninstalled", e.msg)
```
StefanKarpinski, line 2420:
```suggestion
            @test occursin(r"possible versions are: .*\b0.5.1\b.* or uninstalled", e.msg)
```
I wonder if there's a way to make this a bit more precise? Maybe something like this, requiring word boundaries?

oxinabox: Could be even more precise and workout the right regex for color escape sequences.
but the additional difficulty reading might offset the surity added to the tests
the  `\b` might strike the right balance

StefanKarpinski: That was my thinking as well. Just make sure that `0.5.1` is in there as a word.

src/Resolve/graphtype.jl
oxinabox, line 587:
```suggestion
```

test/new.jl
martinholters: Also should be `0\.5\.1`.

fredrikekre:
Need to sanitize the test from ANSI escape codes (or we could just run tests with `--color=no`).

test/new.jl
```diff
@@ -2417,7 +2417,8 @@ end
             Pkg.add(Pkg.PackageSpec(uuid=exuuid, version=v"0.5.3"))
         catch e
             @test e isa ResolverError
-            @test occursin("possible versions are: 0.5.1 or uninstalled", e.msg)
+            # `\S*` in regex below will allow for ANSI color escape codes in the logs
+            @test occursin(r"possible versions are: \S*0.5.1\S* or uninstalled", e.msg)
```
martinholters, line 2421:
```suggestion
            @test occursin(r"possible versions are: \S*0\.5\.1\S* or uninstalled", e.msg)
```

oxinabox: Sensible

oxinabox:
bump

test/new.jl
```diff
@@ -2417,7 +2417,8 @@ end
             Pkg.add(Pkg.PackageSpec(uuid=exuuid, version=v"0.5.3"))
         catch e
             @test e isa ResolverError
-            @test occursin("possible versions are: 0.5.1 or uninstalled", e.msg)
+            # `\S*` in regex below will allow for ANSI color escape codes in the logs
+            @test occursin(r"possible versions are: \S*0\.5\.1\S* or uninstalled", e.msg)
```
StefanKarpinski, line 2421:
```suggestion
            @test occursin(r"possible versions are: [^\d\s]*0\.5\.1[^\d\s]* or uninstalled", e.msg)
```
This should ensure that the leading/trailing "junk" is not digits and not whitespace.

oxinabox: That "junk" will contain digits.
It is something like: `"\e[34m` and `\e[39m"`

oxinabox: CI agrees that it will contain digits and failed.
I have reverted your commit.

oxinabox:
Is this done, or does @StefanKarpinski  want to push another commit into into the PR that breaks the tests?
😁 😂

StefanKarpinski:
Don't tempt me—I've got a lot more broken commits where that came from 😈

PR #1869, opened 2020-06-16
Yet more inference/invalidation fixes

timholy:
This is the Pkg part of what I found in my investigations of invalidations due to extending `==`, c.f. https://github.com/JuliaLang/julia/pull/36282. This also requires https://github.com/JuliaLang/julia/pull/36280, so if that's not on nightly yet this wiil fail tests. (It also requires https://github.com/JuliaLang/julia/pull/36188, but I imagine that's on nightly by now.)

This will be easiest to review ignoring whitespace changes, due to some big `let` blocks from circumventing more cases of the infamous 15276.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -218,12 +218,8 @@ function set_maximum_version_registry!(ctx::Context, pkg::PackageSpec)
         pathvers = keys(load_versions(ctx, path; include_yanked=false))
         union!(pkgversions, pathvers)
     end
-    if length(pkgversions) == 0
-        pkg.version = VersionNumber(0)
-    else
-        max_version = maximum(pkgversions)
-        pkg.version = VersionNumber(max_version.major, max_version.minor, max_version.patch, max_version.prerelease, ("",))
-    end
+    max_version = maximum(pkgversions; init=VersionNumber(0))
+    pkg.version = VersionNumber(max_version.major, max_version.minor, max_version.patch, max_version.prerelease, ("",))
 end
 
 function collect_project!(ctx::Context, pkg::PackageSpec, path::String,
@@ -694,7 +690,7 @@ function download_source(ctx::Context, pkgs::Vector{PackageSpec},
     end
 
     widths = [textwidth(pkg.name) for (pkg, _) in pkgs_to_install]
-    max_name = length(widths) == 0 ? 0 : maximum(widths)
+    max_name = maximum(widths; init=0)
 
     ########################################
     # Install from archives asynchronously #
@@ -906,7 +902,7 @@ function build_versions(ctx::Context, uuids::Vector{UUID}; might_need_to_resolve
     # toposort builds by dependencies
     order = dependency_order_uuids(ctx, map(first, builds))
     sort!(builds, by = build -> order[first(build)])
-    max_name = isempty(builds) ? 0 : maximum(textwidth.([build[2] for build in builds]))
+    max_name = maximum(build->textwidth(build[2]), builds; init=0)
     # build each package versions in a child process
     for (uuid, name, source_path, version) in builds
         pkg = PackageSpec(;uuid=uuid, name=name, version=version)
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -699,7 +699,7 @@ function download_source(ctx::Context, pkgs::Vector{PackageSpec},
     ########################################
     # Install from archives asynchronously #
     ########################################
-    jobs = Channel(ctx.num_concurrent_downloads);
+    jobs = Channel{eltype(pkgs_to_install)}(ctx.num_concurrent_downloads);
     results = Channel(ctx.num_concurrent_downloads);
     @async begin
         for pkg in pkgs_to_install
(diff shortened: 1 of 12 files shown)


src/Types.jl
```diff
@@ -1240,13 +1244,13 @@ function find_registered!(ctx::Context,
...
         reg_abspath = abspath(registry.path)
         data = read_registry(joinpath(registry.path, "Registry.toml"))
-        for (_uuid, pkgdata) in data["packages"]
-              uuid = UUID(_uuid)
-              name = pkgdata["name"]
-              path = joinpath(reg_abspath, pkgdata["path"])
-              push!(get!(ctx.env.uuids, name, UUID[]), uuid)
-              push!(get!(ctx.env.paths, uuid, String[]), path)
-              push!(get!(ctx.env.names, uuid, String[]), name)
+        for (_uuid, pkgdata) in data["packages"]   # what types do these have?
```
timholy, line 1247: Anyone know what types `_uuid` and `pkgdata` are?

timholy:
Ah, does Pkg get deployed to older Julia versions? Or just, sometimes developed on older Julia versions? Should I add version checks?

KristofferC:
> Should I add version checks?

Depends on how much extra work there is. We might need to lose 1.4 and 1.5 compat soon anyway when we start using https://github.com/JuliaLang/julia/pull/36018.

src/Types.jl
KristofferC: `_uuid` should be `String` and `pkgdata` a `Dict{String, Any}` I think.

timholy:
It wouldn't be much work, but your TOML-parser sounds like critical infrastructure to me. So perhaps this can just hang out until you're ready to commit to that and Julia 1.6.

src/Operations.jl
```diff
@@ -1749,8 +1749,8 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}=PackageSpec[];
         old_ctx = Context(;env=env_diff)
     end
     # display
-    filter_uuids = [pkg.uuid for pkg in pkgs if pkg.uuid !== nothing]
-    filter_names = [pkg.name for pkg in pkgs if pkg.name !== nothing]
+    filter_uuids = [pkg.uuid::UUID for pkg in pkgs if pkg.uuid !== nothing]
```
KristofferC, line 1752:
```suggestion
    filter_uuids = UUID[pkg.uuid for pkg in pkgs if pkg.uuid !== nothing]
```

maybe? Same below.

KristofferC:
I think we should drop 1.4 and merge this. We should probably then add a matrix entry to the Travis runner so that we also run with `JULIA_PKG_SERVER=""` to make sure that it doesn't regress. Also, we can probably get rid of AV in favor of Windows on Travis.

src/Operations.jl
timholy: The version here avoids one runtime dispatch and one invalidation-worthy call to `convert` (from inside the `setindex!` call):
```julia
julia> f(x) = Int[x]
f (generic function with 1 method)

julia> g(x) = [x::Int]
g (generic function with 1 method)

julia> code_typed(f, (Integer,))
1-element Array{Any,1}:
 CodeInfo(
1 ─ %1 = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{Int64,1}, svec(Any, Int64), 0, :(:ccall), Array{Int64,1}, 1, 1))::Array{Int64,1}
│        Base.setindex!(%1, x, 1)::Any
└──      return %1
) => Array{Int64,1}

julia> code_typed(g, (Integer,))
1-element Array{Any,1}:
 CodeInfo(
1 ─      Core.typeassert(x, Main.Int)::Int64
│   %2 = π (x, Int64)
│   %3 = $(Expr(:foreigncall, :(:jl_alloc_array_1d), Array{Int64,1}, svec(Any, Int64), 0, :(:ccall), Array{Int64,1}, 1, 1))::Array{Int64,1}
│        Base.arraysize(%3, 1)::Int64
└──      goto #3 if not true
2 ─      Base.arrayset(false, %3, %2, 1)::Array{Int64,1}
3 ┄      goto #4
4 ─      return %3
) => Array{Int64,1}
```

KristofferC: Alright, thanks for exaplining.

KristofferC:
I'll push to this branch for the stuff I mentioned in the last comment

timholy:
Are you fixing the conflict or shall I? I don't want to cause more trouble by trying to be helpful.

timholy:
Looks like this should pass. I therefore took the liberty of reorganizing the commits into more logical form. Feel free to merge or squash, whatever you prefer.

timholy:
Looks good to go. Might want to increase the amount of time on windows, since one of the two jobs timed out and the other was fairly close to the limit.

KristofferC:
I'll put back AV then, it seems faster there.

timholy:
BTW, might be worth mentioning that after I reconfigured the commits, I'm pretty sure the only two that need `nightly` are the `unique` and `maximum` commits. Let me know if you want me to do anything with that observation.

PR #1871, opened 2020-06-19
telemetry: print legal notice first time talking to each pkg server

StefanKarpinski:
* print legal notice first time talking to each pkg server

* generate and send HyperLogLog estimator values

* use ~/.julia/servers/telemetry.toml for defaults

This allows, for example, a user to do

    echo 'telemetry = false' >> ~/.julia/servers/telemetry.toml

and opt out of all telemetry for all servers they might connect to. If they
only want to opt out of sending any telemetry that is based on a client
UUID, then they can do

	echo 'client_uuid = false' >> ~/.julia/servers/telemetry.toml

This can be done anywhere in the depot path, so if a sysadmin wants to set
up a shared system so that all users default to sending no telemetry, they
can do that. A user can opt back into telemetry by Putting an empty default
`telemetry.toml` file in their own depot.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -874,9 +874,7 @@ function get_telemetry_headers(url::AbstractString, notify::Bool=true)
     if info["client_uuid"] != false
         push!(headers, "Julia-Client-UUID: $(info["client_uuid"])")
         if info["secret_salt"] != false
-            salt_hash = hash_data("salt", info["client_uuid"], info["secret_salt"])
             project_hash = hash_data("project", Base.active_project(), info["secret_salt"])
-            push!(headers, "Julia-Salt-Hash: $salt_hash")
             push!(headers, "Julia-Project-Hash: $project_hash")
         end
     end
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -845,6 +845,7 @@ const CI_VARIABLES = [
 ]
 
 const telemetry_file_lock = ReentrantLock()
+const telemetry_notice_printed = Ref(false)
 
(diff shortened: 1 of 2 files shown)


src/PlatformEngines.jl
```diff
@@ -822,13 +844,28 @@ const CI_VARIABLES = [
...
+    notify &= !ispath(telemetry_file)
+    info = lock(telemetry_file_lock) do
+        load_telemetry_file(telemetry_file)
+    end
     get(info, "telemetry", true) == false && return headers
+    # legal (GDPR/CCPA) message about telemetry
+    if notify && !telemetry_notice_printed[]
+        telemetry_notice_printed[] = true
+        @info """
+        LEGAL NOTICE: package operations send anonymous data about your install to $server (your current package server), including the operating system and Julia versison you are running and a random client UUID. See https://julialang.org/legal/data/ for more details about what data is sent, what it is used for, how long it is retained, and how to opt out of sending this information.
```
fredrikekre, line 866: Worth referencing `Pkg.telemetryinfo()`?

StefanKarpinski: I was going to put that at the page that’s linked and go through each value that’s returned, what it means, why we collect it, and how to opt out of it. But maybe showing here would be helpful?

fredrikekre: Yea why not? It gives you a very quick access to the data that is sent and then you might not even bother visiting the website. `Pkg.telemetryinfo()` will also be more up-to-date compared to the website, and more accurate, since that is exactly what is sent whereas the website might not correspond to the same Julia version etc.

---

We should add what each value means to the output of `Pkg.telemetryinfo()` (or to its docstring).

Issue #1377, opened 2019-09-11
Proposal: Pkg & Storage Protocols


StefanKarpinski:
In order to install and update packages, a Julia client needs to get and update verious resources, including: registries, package versions, and artifacts (introduced in Julia 1.3). The current version of Pkg acquires these resources by a variety of mechanisms, over a variety of protocols, including git, SSH, HTTP and HTTPS—usually talking directly to proprietary services like GitHub, GitLab, BitBucket, and sometimes servers run by unknown parties. This proposal aims to replace this variety of protocols, mechanisms and proprietary services with a standard unified protocol, such that the Pkg client, by default, gets all resources over HTTPS from a single open source service run by the Julia community. This service for serving packages will additionally be backed by multiple independent storage services which interface with proprietary origin services (GitHub, etc.) and guarantee persitent availabilty of resources into the future.

The current approach of leaning on GitHub for resource hosting and storage has been pragmatic and has allowed Julia's package ecosystem to grow rapidly, but it has some significant drawbacks:

1. **Vanishing resources.** It is increasingly common for people to delete registered Julia packages, after which point no one can install them anymore. If someone happens to have a current fork of a deleted package, that can be made the new official repository for the package, but this has rarely happened. An even worse problem exists with artifacts since they tend not to be kept in version control and are much more likely to be served from "random" web servers at a fixed URL with content changing over time. Artifact publishers are unlikely to retain all past versions of artifacts, so old versions of packages that depend on specific artifact content will not be reproducible in the future unless we do something to ensure that they are kept around after the publisher has stopped hosting them.

2. **Lack of insight.** Currently, the Julia community has no way of knowing how many people are using Julia or what the relative popularity of different packages and operating systems is. On the other hand, GitHub—a commercial, proprietary service—has this information but does not make it available to the Julia community. We are, of course, using GitHub to host our ecosystem for free, so we can't complain, but it seems unfortunate that a commercial entity has this vauable information while the open source community remains in the dark. The Julia community really could use insight into who is using Julia and how, so that we can prioritize packages and platforms, and give real numbers when people ask "how many people are using Julia?"

[...]

PR #1876, opened 2020-06-22
[docs] add a section to pkg server feature

johnnychen94:
I'm very positive that this isn't complete, but just propose a section for this. When I tell others how to set/switch pkg server, I don't really want to point them to the epic #1377 design. The contents mainly come from questions when we advertise the pkg server feature in the Chinese community.

I haven't tried a private pkg server with authentication, so I didn't document this part. Also, feel free to rephrase or drop these words as I'm not a native English speaker.

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -298,3 +298,56 @@ To run a typical garbage collection with default arguments, simply use the `gc` 
 ```
 
 Note that only packages in `~/.julia/packages` are deleted.
+
+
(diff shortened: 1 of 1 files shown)


johnnychen94:
Do you think it worths mentioning the 3rd party mirrors in China? https://github.com/johnnychen94/PkgServerClient.jl gathers all that I know of.

KristofferC:
This is an improvement so let's get it in.

PR #1884, opened 2020-06-24
A few fixes

### LINKED ISSUE #1769: `free` updates all packages
Feems like this should use the same tiered resolver strategy as `add´.

### LINKED ISSUE #1882: Pkg doesn't prune manifest correctly
I keep having situations where my Manifest contains a bunch of orphan packages that are never pruned. Don't have a MWE now but it is really annoying so opening this.

KristofferC:
- fix manifest not getting pruned sometimes, fixes #1882
- fix pin and free using tiered resolver, fixes #1769
- fix getting the correct deps when using multiple registries


fredrikekre:
Good to go?

KristofferC:
Yep

PR #1885, opened 2020-06-25
Make tree hash computation non-fatal when installing artifacts.

fredrikekre:


src/Artifacts.jl
```diff
@@ -786,7 +786,9 @@ function download_artifact(
             msg *= "  Expected git-tree-sha1:   $(bytes2hex(tree_hash.bytes))\n"
             msg *= "  Calculated git-tree-sha1: $(bytes2hex(calc_hash.bytes))"
             @error(msg)
-            return false
+            # Tree hash calculation is still broken on some systems, e.g. Pkg.jl#1860,
+            # so we return true here and only raise the warning on the lines above.
+            # return false
```
oxinabox, line 791: Don't we need to actually have the return?
```suggestion
            # return false
            return true
```

johnnychen94:
My first impression is that this PR takes the opposite direction to https://github.com/JuliaLang/Pkg.jl/issues/1411?

KristofferC:
> My first impression is that this PR takes the opposite direction to #1411? It's a little non-intuitive to me why Artifacts downloading failure should be non-fatal while build failure should.

This is not a download failure. It happens when there is a bug/inconcistency in how we compute the tree hash. The sha hash is already verified so we already know we got the right file.

src/Artifacts.jl
fredrikekre: That would result in an early return in case where more code is added later in this function, we just want to keep going, not explicitly return here.

StefanKarpinski:
It feels a little drastic to effectively skip checking tree hashes everywhere because of a corner case where someone was mounting an ntfs file system on Linux, which seems likely to be uncommon. Are you certain that this check cannot be reached without the download itself having being hash checked?

fredrikekre:
https://github.com/JuliaLang/Pkg.jl/blob/36de8a7eac80e11036e17cb3bdc4f8eeaec24776/src/Artifacts.jl#L773 verifies the sha256 of the downloaded tarball so a failed tree-hash computation should only occur if (i) the computation itself is wrong (as the ntfs case) or (ii) the unpacking process got something wrong.

Given that this fatal error essentially makes Julia useless on such systems I think it is fine to only use a logging `@error` here.

StefanKarpinski:
Yeah, ok. In 1.6 we'll do the tree hashing on the tarball which won't be susceptible to these issues. Trying to do the hashing on every janky filesystem in existence clearly doesn't work out well.

staticfloat:
I'd rather merge this: https://github.com/JuliaLang/Pkg.jl/pull/1573

StefanKarpinski:
Whatever fixes it is fine. This does feel like a game of whack-a-mole, but then again using `Tar.tree_hash` will fix it everywhere so I'm not too concerned about that.

staticfloat:
Okay, I'm fine with this fix, but some tests need updating.

fredrikekre:
It is on the release-1.5 branch.

fritzratna:
@fredrikekre  I downloaded julia 1.5.2 in a NFS mount directory. I tried installing the package NLPModels and I am still getting a tree hash error. I thought the issue was solved for the release of Julia 1.5?

fredrikekre:
Yea, the fix was not complete, you have to do something like https://discourse.julialang.org/t/unable-to-automatically-install-compilersupportlibraries/36669/32

PR #1891, opened 2020-06-29
no-arg activate: if already deactivated, activate current project

StefanKarpinski:
* home project: remove all uses of this

* Previously there was no convenient way to activate the current project from the Pkg REPL or API. This changes `Pkg.activate()` so that if there is no active project to deactivate, it instead activates `Base.current_project()`. This means that we can now say that `julia --project` is exactly equivlent to `julia` followed by `pkg> activate`.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -886,9 +886,11 @@ end
 function activate(;temp=false,shared=false)
     shared && pkgerror("Must give a name for a shared environment")
     temp && return activate(mktempdir())
-    Base.ACTIVE_PROJECT[] = nothing
+    Base.ACTIVE_PROJECT[] =
+        Base.ACTIVE_PROJECT[] === nothing ? Base.current_project() : nothing
     p = Base.active_project()
-    p === nothing || printpkgstyle(Context(), :Activating, "environment at $(pathrepr(p))")
+    p === nothing ||
+        printpkgstyle(Context(), :Activating, "environment at $(pathrepr(p))")
     add_snapshot_to_undo()
     return nothing
 end
--- a/test/utils.jl
+++ b/test/utils.jl
@@ -13,7 +13,6 @@ const LOADED_DEPOT = joinpath(@__DIR__, "loaded_depot")
 function isolate(fn::Function; loaded_depot=false, kwargs...)
     old_load_path = copy(LOAD_PATH)
     old_depot_path = copy(DEPOT_PATH)
-    old_home_project = Base.HOME_PROJECT[]
     old_active_project = Base.ACTIVE_PROJECT[]
     old_working_directory = pwd()
     old_general_registry_url = Pkg.Types.DEFAULT_REGISTRIES[1].url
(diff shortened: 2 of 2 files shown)


StefanKarpinski:
well that may have broken some things

StefanKarpinski:
Marking as 1.6 mostly as a reminder to myself to get this passing tests at some point before then.

danielwe:
Coming here from #3547 and just wanted to confirm that this is the intended behavior for repeated no-arg calls to `activate`:

* First call: discard and forget any active environment set via `JULIA_PROJECT` or the `--project` parameter at startup, or via subsequent `activate` calls with argument. Unless `LOAD_PATH` has been modified, this activates the base `@v1.x` environment.
* Second call: activate `current_project()`, which is determined based on the current working directory (as if `--project=@.` were passed at startup)
* Subsequent calls: toggle between the latter two

I confess I'd be surprised by this behavior, specifically having the environment I provided on the command line forgotten while an unrelated non-base environment is brought into orbit. Wouldn't it make more sense to toggle back to the previously active environment rather than `current_project()`?

Socob:
I agree that it’s very confusing to make the behavior of `activate` without arguments depend on the current state. Further, after all this time, it’d be yet another breaking change in the behavior of an existing function.

If all that’s needed is a convenient way to activate the current project, this could simply be added as a new flag, i. e. `Pkg.activate(current=true)` and `] activate --current`, or as a separate function, `Pkg.activate_current()` and `] activate_current`.

StefanKarpinski:
> Further, after all this time, it’d be yet another breaking change in the behavior of an existing function.

SemVer doesn't apply to interactive functionality, so that argument does not prevent changing this. It's worth taking into consideration, but we are free to alter interactive interfaces like Pkg and the REPL.

Socob:
I wasn’t aware that Pkg was considered an interactive interface! Is this stated anywhere? I’m certainly using it in a non-interactive way! What is the stable interface to accomplish the same things?

StefanKarpinski:
If you're talking about the API then that should remain compatible, but the Pkg REPL certainly is interactive. I don't really feel strongly about this issue, btw, just seemed like this would be convenient.

Socob:
I was referring to the function `Pkg.activate` (see above), since this is what you were changing in this PR. (As it’s defined in `API.jl`, I’m assuming that this is part of the API you mentioned). The behavior of this function/API was already broken once by the changes in JuliaLang/julia#36434 (see #3127), apparently without much/any announcement (at least on the Pkg.jl side) or changes to the documentation (#3547, JuliaLang/julia#50527). I agree that it would be nice to have an easy way to activate the current project (moreso on the interactive side, since using the API, you can already just do `Pkg.activate(Base.current_project())`), but that the API shouldn’t be broken for no reason.

(Side note: In principle, I do think it was a good idea to get rid of the “home project” concept, since it just added unnecessary complexity and confusion.)

PR #1897, opened 2020-07-03
Fix invalidations from loading OrderedCollections

timholy:
Somehow, I think I convinced myself early on that the OrderedCollections invalidations were unfixable. That is most assuredly false, and this PR gets rid of them all. That seems pretty useful, because if memory serves more than half the Julia ecosystem loads OrderedCollections.

The major source of invalidations come from `print_status`; I don't know why I didn't fix this properly before. (Note that the double type-specification in each comprehension seems necessary, if we want to get rid of all the vulnerabilities, since there's a `Union` involved---the `Tuple{T,S,S}[comp...]` ensures we get the right element type and `[(x,y,z)::Tuple{T,S,S}...` eliminates a call to `convert`.) Additionally, `TOML._print` calls itself, with apparently-noninferrable types, so we force runtime dispatch to sever the resulting self-backedges. An alternative fix would be to specify the arg types more precisely than `AbstractDict` or `AbstractArray{<:AbstractDict}`, i.e., if https://github.com/JuliaLang/Pkg.jl/blob/20f9b9e14ab59d81991b5984a2f9eb87d38b6230/ext/TOML/src/print.jl#L78 means that it's really a `Dict{String,Any}` and https://github.com/JuliaLang/Pkg.jl/blob/20f9b9e14ab59d81991b5984a2f9eb87d38b6230/ext/TOML/src/print.jl#L93 means it's a `Vector{Dict{String,Any}}` then it might be better to just enforce that.

### DIFF
--- a/ext/TOML/src/print.jl
+++ b/ext/TOML/src/print.jl
@@ -87,8 +87,9 @@ function _print(io::IO, a::AbstractDict,
                 printkey(io, ks)
                 Base.print(io,"]\n")
             end
-            _print(io, value, ks,
-                indent = indent + header, first_block = header, sorted = sorted, by = by)
+            # Use runtime dispatch here since the type of value seems not to be enforced other than as AbstractDict
+            invoke(_print, Tuple{typeof(io), typeof(value), typeof(ks)},
+                  io, value, ks; indent = indent + header, first_block = header, sorted = sorted, by = by)
             pop!(ks)
         elseif is_array_of_tables(value)
             # print array of tables
@@ -101,7 +102,8 @@ function _print(io::IO, a::AbstractDict,
                 printkey(io, ks)
                 Base.print(io,"]]\n")
                 !isa(v, AbstractDict) && error("array should contain only tables")
-                _print(io, v, ks, indent = indent + 1, sorted = sorted, by = by)
+                invoke(_print, Tuple{typeof(io), typeof(v), typeof(ks)},
+                       io, v, ks; indent = indent + 1, sorted = sorted, by = by)
             end
             pop!(ks)
         end
--- a/ext/TOML/src/print.jl
+++ b/ext/TOML/src/print.jl
@@ -88,8 +88,7 @@ function _print(io::IO, a::AbstractDict,
                 Base.print(io,"]\n")
             end
             # Use runtime dispatch here since the type of value seems not to be enforced other than as AbstractDict
-            invoke(_print, Tuple{typeof(io), typeof(value), typeof(ks)},
-                  io, value, ks; indent = indent + header, first_block = header, sorted = sorted, by = by)
+            Base.invokelatest(_print, io, value, ks; indent = indent + header, first_block = header, sorted = sorted, by = by)
             pop!(ks)
         elseif is_array_of_tables(value)
             # print array of tables
@@ -102,8 +101,7 @@ function _print(io::IO, a::AbstractDict,
                 printkey(io, ks)
                 Base.print(io,"]]\n")
                 !isa(v, AbstractDict) && error("array should contain only tables")
-                invoke(_print, Tuple{typeof(io), typeof(v), typeof(ks)},
-                       io, v, ks; indent = indent + 1, sorted = sorted, by = by)
+                Base.invokelatest(_print, io, v, ks; indent = indent + 1, sorted = sorted, by = by)
             end
             pop!(ks)
         end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1645,14 +1645,14 @@ function diff_array(old_ctx::Union{Context,Nothing}, new_ctx::Context; manifest=
     end
     # load deps
     new = manifest ? load_manifest_deps(new_ctx) : load_direct_deps(new_ctx)
+    T, S = Union{UUID,Nothing}, Union{PackageSpec,Nothing}
     if old_ctx === nothing
-        return [(pkg.uuid, nothing, pkg) for pkg in new]
+        return Tuple{T,S,S}[(pkg.uuid, nothing, pkg)::Tuple{T,S,S} for pkg in new]
     end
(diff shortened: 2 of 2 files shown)


timholy:
Interesting. With `invoke` I got
```julia

julia> t = Template(plugins=[Git(;ssh=true), GitHubActions(), TravisCI(), Codecov(), GitHubPages()]);

julia> t("SomeNewPkg")
[ Info: Running prehooks
[ Info: Running hooks
ERROR: invoke: argument type error
Stacktrace:
 [1] _print(::IOStream, ::Dict{String,Any}, ::Array{String,1}; indent::Int64, first_block::Bool, sorted::Bool, by::Function) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/ext/TOML/src/print.jl:91
 [2] _print at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/ext/TOML/src/print.jl:60 [inlined] (repeats 2 times)
 [3] #print#10 at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/ext/TOML/src/print.jl:113 [inlined]
 [4] print at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/ext/TOML/src/print.jl:113 [inlined]
 [5] #44 at /home/tim/.julia/packages/PkgTemplates/pOq4d/src/plugins/project_file.jl:24 [inlined]
 [6] open(::PkgTemplates.var"#44#45"{Dict{String,Any}}, ::String, ::Vararg{String,N} where N; kwargs::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}) at ./io.jl:330
 [7] open(::Function, ::String, ::String) at ./io.jl:328
 [8] hook(::ProjectFile, ::Template, ::String) at /home/tim/.julia/packages/PkgTemplates/pOq4d/src/plugins/project_file.jl:24
 [9] (::PkgTemplates.var"#17#20"{typeof(PkgTemplates.hook),Template,String})(::ProjectFile) at /home/tim/.julia/packages/PkgTemplates/pOq4d/src/template.jl:132
 [10] foreach(::PkgTemplates.var"#17#20"{typeof(PkgTemplates.hook),Template,String}, ::Array{PkgTemplates.Plugin,1}) at ./abstractarray.jl:2018
 [11] (::PkgTemplates.var"#16#19"{Template,String})(::typeof(PkgTemplates.hook)) at /home/tim/.julia/packages/PkgTemplates/pOq4d/src/template.jl:131
 [12] foreach(::PkgTemplates.var"#16#19"{Template,String}, ::Tuple{typeof(PkgTemplates.prehook),typeof(PkgTemplates.hook),typeof(PkgTemplates.posthook)}) at ./abstractarray.jl:2018
 [13] (::Template)(::String) at /home/tim/.julia/packages/PkgTemplates/pOq4d/src/template.jl:129
 [14] top-level scope at REPL[3]:1
```

but switching it to `Base.invokelatest` worked fine. I'm not quite sure what's up with this, but perhaps `invoke` really shouldn't be used with runtime-defined types?

c42f:
Hi Tim, I'm very curious now that the experimental https://github.com/JuliaLang/julia/pull/35844 is merged - can it be used as a general tool (wrapped around pkg repl mode entry points, for example) to solve this without needing these kind of very specific-looking workarounds scattered through the code?

The obvious caveat with the frozen-world idea is that by default it would make Pkg act as a non-extensible library with respect to user types and methods, separate from the user's session. Oh, and it's available only on 1.6.

timholy:
I agree that's worth thinking about, and I'm glad we have that functionality, but the frozen-world issue does seem like a disadvantage.

My view is that for "library-quality" code that has to deal with type-instability (e.g., the compiler, `Base.show_unquoted`, etc.), it's not unreasonable that it should be written to a standard that doesn't have lots of unnecessary inference problems. In other words, I've been taking the perspective is that the invalidation hunt is just revealing places where we should have improved the quality of engineering anyway.

*Keeping* it that way will be a challenge, of course, which is why I'm eager to get people taking https://github.com/JuliaLang/julia/issues/36393 a bit more seriously.

c42f:
> the frozen-world issue does seem like a disadvantage

It's a matter of perspective. I think the litmus test is thus: "would you be happy if the `pkg>` repl mode ran in a separate process from user code?"

If yes, then Pkg is not expecting users to extend it in a meaningful way, and frozen-world makes a lot of sense for this use case. Actually it's an advantage because users can't unintentionally break things.

> or perhaps you were referring to the invokelatest part of this

Somewhat, but I also feel the need for "unnecessary-looking" type annotations is a sign of trouble, in that it contributes to two things:
* Fragility of avoiding these coming back in future. You already mentioned this.
* "Rigidity" - people being afraid of changing tricky-looking code because it doesn't directly express the intentions of the original programmer. Luckily `git blame` comes to the rescue here, to some extent.

Not to say I object to these changes specifically — I trust these help in the short/medium term, and they're minimal enough for sure.

timholy:
I definitely see the merits of your perspective.

>  I think the litmus test is thus: "would you be happy if the pkg> repl mode ran in a separate process from user code?"

Easy answer: if I can Revise it, I'm happy---if not, not. I'm actually not entirely certain of whether your recent work affects Revise-ability? Core.Compiler is walled off, yet I can still Revise stuff. As long as you can Revise it, then not being able to inadvertently break it is really cool.

> the need for "unnecessary-looking" type annotations

Here it's because of https://github.com/JuliaLang/julia/issues/36454. In other places they might look unnecessary, the reasoning explained in https://github.com/JuliaLang/julia/pull/36470#discussion_r447215752 is a common source. As we deliberately scale back the thoroughness of our inference this will affect both performance and vulnerability to invalidation.

c42f:
> if I can Revise it, I'm happy---if not, not

You can. (XRef for other people following this conversation: I put a complete description of how this works over at https://github.com/JuliaLang/julia/pull/35844#issuecomment-655240174.)

PR #1906, opened 2020-07-06
Telemetry CI variables: add CI_SERVER, JULIA_PKGEVAL, JULIA_REGISTRYCI_AUTOMERGE, and PKGEVAL

DilumAluthge:
`CI_SERVER` is set by [GitLab](https://docs.gitlab.com/ce/ci/variables/predefined_variables.html).

`PKGEVAL` and `JULIA_PKGEVAL` are set by [NewPkgEval](https://github.com/JuliaComputing/NewPkgEval.jl/blob/master/src/run.jl).

`JULIA_REGISTRYCI_AUTOMERGE` is set by [RegistryCI](https://github.com/JuliaRegistries/RegistryCI.jl/pull/257).

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -847,7 +847,6 @@ const CI_VARIABLES = [
     "JULIA_CI",
     "JULIA_PKGEVAL",
     "JULIA_REGISTRYCI_AUTOMERGE",
-    "PKGEVAL",
     "TF_BUILD",
     "TRAVIS",
 ]
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -839,11 +839,15 @@ end
 const CI_VARIABLES = [
     "APPVEYOR",
     "CI",
+    "CI_SERVER",
     "CIRCLECI",
     "CONTINUOUS_INTEGRATION",
     "GITHUB_ACTIONS",
     "GITLAB_CI",
     "JULIA_CI",
+    "JULIA_PKGEVAL",
+    "JULIA_REGISTRYCI_AUTOMERGE",
+    "PKGEVAL",
     "TF_BUILD",
     "TRAVIS",
 ]


DilumAluthge:
Once this is merged, we need to update the list of environment variables on https://julialang.org/legal/data/

KristofferC:
> PKGEVAL and JULIA_PKGEVAL are set by NewPkgEval.

This is not needed since it already sets `CI`.

fredrikekre:
> CI_SERVER is set by GitLab.

They also use `CI` so I don't understand why `CI_SERVER` is also needed.

DilumAluthge:
> > JULIA_REGISTRYCI_AUTOMERGE is set by RegistryCI.
> 
> Why does this need a dedicated env variable? Just set `CI` or something like that?

Seems fine to me. I'll change RegistryCI to just use `CI`.

DilumAluthge:
> > PKGEVAL and JULIA_PKGEVAL are set by NewPkgEval.
> 
> This is not really needed since it already sets `CI`.

For PkgEval, I was thinking that it would be good to separate PkgEval from other CI, for infrastructure planning purposes? My understanding is that PkgEval runs on a physical machine we control, not in a datacenter. So e.g. putting additional PkgServers in Amazon or Microsoft datacenters wouldn't help PkgEval at all.

It's possible that the PkgEval traffic won't be big enough of a contributor for this to matter.

DilumAluthge:
> > CI_SERVER is set by GitLab.
> 
> They also use `CI` so I don't understand why `CI_SERVER` is also needed.

Similar to above, my thinking was that it would be good to get an idea of which specific CI services are being used. Since `CI_SERVER' is set by GitLab but does not seem to be set by other CI services, this would help us figure out how much GitLab CI traffic we have, and then we can decide whether or not it makes sense to put a dedicated PkgServer inside the datacenters most commonly used by GitLab.

Again, it's possible that there isn't enough GitLab CI traffic to make a difference. But this would give us an idea at least of how much volume we have on GitLab CI.

StefanKarpinski:
I do think it's good to get a sense of what kind of CI traffic is coming, given that there's no real privacy concern here.

KristofferC:
Yeah, I guess being able to easily see if Github CI completely nukes the server is useful. 

However, there should be no need for both  "JULIA_PKGEVAL", and "PKGEVAL", or?

StefanKarpinski:
Agree, let's just do `JULIA_PKGEVAL` and move to having that as the standard indicator for PkgEval.

fredrikekre:
> For PkgEval, I was thinking that it would be good to separate PkgEval from other CI, for infrastructure planning purposes?

But we already know where those machines are so no need to find out through this data. But perhaps it is useful anyway?

StefanKarpinski:
Other people could be running PkgEval and we wouldn't know about it.

fredrikekre:
Why do we need to know that though?

StefanKarpinski:
We don't entirely need it but it's useful to know what people are doing that's causing automated CI systems to pull content from public package servers. It's also possible that someone would have a setup that sets some variables but not others for whatever reason and then it wouldn't get classified as being CI traffic.

PR #1916, opened 2020-07-09
BinaryPlatforms: Recognize MacOS(:aarch64)

Keno:
Also normalize arm64 -> aarch64, c.f.
https://lists.gnu.org/archive/html/config-patches/2020-07/msg00002.html

### DIFF
--- a/src/BinaryPlatforms.jl
+++ b/src/BinaryPlatforms.jl
@@ -140,7 +140,7 @@ struct MacOS <: Platform
                    libc::Union{Nothing,Symbol}=nothing,
                    call_abi::Union{Nothing,Symbol}=nothing,
                    compiler_abi::CompilerABI=CompilerABI())
-        if arch !== :x86_64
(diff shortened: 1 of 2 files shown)


giordano:
`MacOS()` still defaults to x86_64, kept for backward compatibility I guess?

staticfloat:
Yes, it's going to be a while before that's changed, I think.  :P

giordano:
Sure, I was not thinking about changing the default, but rather the fact other platforms with multiple possible architecture like `Linux` and `Windows` don't have a default constructor without the specification of the architecture.  I assume removing the `MacOS()` method would be breaking though.

staticfloat:
Indeed; hindsight is 20:20.

That being said, if we think it would be better, I have no qualms about breaking `build_tarballs.jl` files.  It's a very easy fix, and Yggdrasil contains a high percentage of what we care about.

PR #1918, opened 2020-07-09
Deprecate `Pkg.generate`

DilumAluthge:
cc: @StefanKarpinski

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -2,6 +2,7 @@
 
 generate(path::String; kwargs...) = generate(Context(), path; kwargs...)
 function generate(ctx::Context, path::String; kwargs...)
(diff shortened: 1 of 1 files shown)


KristofferC:
It's used everywhere in the tests so cannot do it exactly like this.

StefanKarpinski:
How about just having the deprecation warning in the command-line interface? I feel a lot less bad about breaking interactive stuff like that. That way the tests don't have to change and any scripted usage of this keeps working.

DilumAluthge:
Sure. How do you do a depwarn for the REPL mode only?

DilumAluthge:
Alright, here is a PR to deprecate only the REPL mode: https://github.com/JuliaLang/Pkg.jl/pull/1923

DilumAluthge:
Closing in favor of #1923

PR #1923, opened 2020-07-10
Deprecate the REPL command `] generate`

DilumAluthge:
This PR deprecates the REPL command `] generate` but leaves `Pkg.generate` untouched.

@StefanKarpinski @KristofferC

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -1,5 +1,10 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
+function generate_deprecated(varargs...; kwargs...)
+    Base.depwarn("Pkg.generate is deprecated. Please use PkgTemplates.jl instead.", Core.Typeof(generate).name.mt.name)
+    return generate(varargs...; kwargs...)
+end
+
 generate(path::String; kwargs...) = generate(Context(), path; kwargs...)
 function generate(ctx::Context, path::String; kwargs...)
     Context!(ctx; kwargs...)
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -1,7 +1,7 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
 function generate_deprecated(varargs...; kwargs...)
-    Base.depwarn("Pkg.generate is deprecated. Please use PkgTemplates.jl instead.", Core.Typeof(generate).name.mt.name)
+    Base.depwarn("Pkg.generate is deprecated. Please use PkgTemplates.jl instead.", Symbol("Pkg.generate"))
     return generate(varargs...; kwargs...)
 end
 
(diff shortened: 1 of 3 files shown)


src/generate.jl
```diff
@@ -1,5 +1,10 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
+function generate_deprecated(varargs...; kwargs...)
+    Base.depwarn("Pkg.generate is deprecated. Please use PkgTemplates.jl instead.", Core.Typeof(generate).name.mt.name)
```
KristofferC, line 4: What's `Core.Typeof(generate).name.mt.name`?

DilumAluthge: It's how the deprecate macro gets the symbol for the function being deprecated.

But we are not using the deprecate macro. We are calling depwarn directly. So we can just hardcode the symbol.

DilumAluthge: @KristofferC Take a look at the new code. Can this be resolved?

StefanKarpinski:
This looks good to me. @KristofferC, you good with this now? If so, let's squash-merge it.

DilumAluthge:
Bump @StefanKarpinski @KristofferC

anandijain:
😢 bring it back please

StefanKarpinski:
Why?

anandijain:
It's a very weird thing to be told, "Oh you want to create a package, go get this package PkgTemplates, we don't let people make packages with just Julia."

Many modern languages (cf Rust) include minimal package generation as a part of core tooling. Additionally, I personally dislike all the hoops that PkgTemplates makes me jump through. I use `Pkg.generate()` and `]generate` frequently.

IMO it feels strange to have a split between the API and functions available to user in Pkg

StefanKarpinski:
We kept getting complaints that `generate` was broken or insufficient and no one here wanted to maintain it or make it do all the things that people wanted it to do for them. Would you be willing to figure out what a built-in bare bones `generate` needs to support and make sure it implements those things?

pazner:
I'm also a frequent user of `] generate` and was disappointed to see that it was deprecated.

> We kept getting complaints that `generate` was broken or insufficient and no one here wanted to maintain it or make it do all the things that people wanted it to do for them. Would you be willing to figure out what a built-in bare bones `generate` needs to support and make sure it implements those things?

I think the current implementation is already good.

anandijain:
Yeah keep it how it is and just stress it is only for minimal package generation. if anybody asks for more then give them PkgTemplates

StefanKarpinski:
That would be fine if someone wants to revert this and fix those related issues.

pazner:
> That would be fine if someone wants to revert this and fix those related issues.

I took a stab at this. Let me know if you think #2492 would do the trick.

Issue #1925, opened 2020-07-14
Subdir information lost by develop


GunnarFarneback:
As reported in https://github.com/JuliaLang/Pkg.jl/issues/1251#issuecomment-657915380, adding a subdir package by URL and then developing it by name does not work correctly.
```
(@v1.5) pkg> add https://github.com/DilumAluthge/FooBar.git:myfolder/Foo
   Updating git-repo `https://github.com/DilumAluthge/FooBar.git`
  Resolving package versions...
Updating `/tmp/foo16/environments/v1.5/Project.toml`
  [ac29586c] + Foo v0.1.0 `https://github.com/DilumAluthge/FooBar.git:myfolder/Foo#master`
Updating `/tmp/foo16/environments/v1.5/Manifest.toml`
  [ac29586c] + Foo v0.1.0 `https://github.com/DilumAluthge/FooBar.git:myfolder/Foo#master`

(@v1.5) pkg> dev Foo
    Cloning git-repo `https://github.com/DilumAluthge/FooBar.git`
ERROR: could not find project file in package at `https://github.com/DilumAluthge/FooBar.git` maybe `subdir` needs to be specified
```

GunnarFarneback:
Should be fixed by #4271.

PR #1926, opened 2020-07-14
fix computing the correct path when devving a package with a subdir

KristofferC:
Mostly moves around some code so that information is available at an earlier stage.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/1925

### DIFF
--- a/test/subdir.jl
+++ b/test/subdir.jl
@@ -321,4 +321,16 @@ end
     end
 end
 
+@testset "dev a package added by subdir" begin
+    isolate(loaded_depot=false) do
+        Pkg.add(url="https://github.com/DilumAluthge/FooBar.git", subdir="myfolder/Foo")
+        Pkg.develop("Foo")
+        @test isinstalled("Foo")
+        Pkg.develop("Foo")
+        @test isinstalled("Foo")
+        Pkg.develop("Foo")
+        @test isinstalled("Foo")
+    end
+end
+
 end # module
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -436,19 +436,36 @@ function devpath(ctx::Context, name::AbstractString, shared::Bool)
 end
 
 function handle_repo_develop!(ctx::Context, pkg::PackageSpec, shared::Bool)
-    # First, check if we can compute the path easily (which requires a given local path or name)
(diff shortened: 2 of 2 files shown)


GunnarFarneback:
For what it's worth, in my experiments the more conservative patch
```
diff --git a/src/Types.jl b/src/Types.jl
index 0b7f5675..ca3afbf3 100644
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -469,6 +469,7 @@ function handle_repo_develop!(ctx::Context, pkg::PackageSpec, shared::Bool)
             entry = manifest_info(ctx, uuid)
             if entry !== nothing
                 pkg.repo.source = entry.repo.source
+                pkg.repo.subdir = entry.repo.subdir
             end
         end
     end
@@ -512,13 +513,13 @@ function handle_repo_develop!(ctx::Context, pkg::PackageSpec, shared::Bool)
         end
         new = true
     end
+    if pkg.repo.subdir !== nothing
+        dev_path = joinpath(dev_path, pkg.repo.subdir)
+    end
     if !has_uuid(pkg)
         resolve_projectfile!(ctx, pkg, dev_path)
     end
     pkg.path = shared ? dev_path : relative_project_path(ctx, dev_path)
-    if pkg.repo.subdir !== nothing
-        pkg.path = joinpath(pkg.path, pkg.repo.subdir)
-    end
 
     return new
 end
```
seemed to be sufficient.

KristofferC:
Does that work if you do `Pkg.dev("Foo")` twice in a row?

test/subdir.jl
```diff
@@ -321,4 +321,16 @@ end
     end
 end
 
+@testset "dev a package added by subdir" begin
+    isolate(loaded_depot=false) do
+        Pkg.add(url="https://github.com/DilumAluthge/FooBar.git", subdir="myfolder/Foo")
```
GunnarFarneback, line 326: I know this is not the only occurrence but it seems unfortunate to depend on an external repository, especially when the testset above has a local subdir repository set up.

KristofferC: Yeah. I can update it after the tests pass.

GunnarFarneback:
Right you are, it doesn't.

PR #1930, opened 2020-07-29
Fix PkgServer download failures

staticfloat:
Add missing `project` back in, accidentally dropped in 2acb2f9ca7f8b908954d6fdde074b65e30b8395f

This pretty much broke downloading PkgServer resources, so we should update the Pkg on master ASAP.


KristofferC:
Hm, CI didn't catch this because of fallbacks?

fredrikekre:
There is a bit too much try-catch happening I think.

fredrikekre:
For example, why is this `try` necessary https://github.com/JuliaLang/Pkg.jl/blob/13456b944fd21ec180111c6c29d06b610d47d088/src/Operations.jl#L534-L535 when there is already a `try` around the possibly-failing part: https://github.com/JuliaLang/Pkg.jl/blob/master/src/PlatformEngines.jl#L952-L953

PR #1934, opened 2020-07-31
Add running tests notice

omus:
Adds a notice to indicate that tests are starting. There are two motivations behind this change:

1. Adds a clear demarcation between resolving test dependencies and starting tests (was more relevant back with Julia 1.3)
2. Provides an easy string to search for in the test output logs to find the start of test execution


omus:
Am I free to merge this?

StefanKarpinski:
Not sure about the CI failure, but if that is unrelated, then yes.

PR #1940, opened 2020-08-04
make Pkg.precompile fatal

simonbyrne:
Similar to #1851

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -773,10 +773,7 @@ function precompile(ctx::Context)
             break
         end
         if stale
-            try
-                Base.compilecache(pkg, sourcepath)
-            catch
-            end
+            Base.compilecache(pkg, sourcepath)
         end
     end
     nothing


simonbyrne:
Any thoughts @KristofferC @StefanKarpinski?

KristofferC:
Do you want it to throw as soon as possible (like with the current implementation) or try all packages and only throw in the end?

simonbyrne:
Throw as soon as possible would be best, I think.

KristofferC:
Alright, then lgtm.

StefanKarpinski:
Similar to how I wanted build errors to just be fatal, why wouldn't precompile errors also just be fatal?

StefanKarpinski:
If precompilation is broken for a package, shouldn't the recourse be either turning it off for that package or fixing it? This just seems like it enables people to have registered packages for which precompilation is broken but which nevertheless don't opt out of precompilation.

simonbyrne:
I agree with that. WIll change to fail by default

simonbyrne:
updated.

PR #1947, opened 2020-08-07
Add `Pkg.API.get_uuid()` and `Pkg.API.get_version()`

staticfloat:
These are helper methods to aid in packages wanting to inspect their own
(or their dependencies') versions at runtime.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1050,4 +1050,31 @@ Package(name::AbstractString) = PackageSpec(name)
 Package(name::AbstractString, uuid::UUID) = PackageSpec(name, uuid)
 Package(name::AbstractString, uuid::UUID, version::VersionTypes) = PackageSpec(name, uuid, version)
 
+
+"""
+    get_uuid(m::Module)
+
+Given a `Module`, find the `UUID` of the parent package to that `Module`.  If none can be
+found (e.g. `Main`, `Base`, anonymous modules, etc...) returns `nothing`.
+"""
+function get_uuid(m::Module)
+    return Base.PkgId(m).uuid
+end
+
+"""
+    get_version(pkg)
+
+Given a `UUID` identifying a package, returns the version of that package within the
+current project.  If the package cannot be found by the given UUID, throws an
+`ArgumentError`.
+"""
+function get_version(uuid::UUID)
+    ctx = Pkg.Types.Context()
+    matching_pkgs = filter(((u, e),) -> u == uuid, ctx.env.manifest)
+    if isempty(matching_pkgs)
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -1050,4 +1050,31 @@ Package(name::AbstractString) = PackageSpec(name)
...
+end
+
+"""
+    get_version(pkg)
+
+Given a `UUID` identifying a package, returns the version of that package within the
+current project.  If the package cannot be found by the given UUID, throws an
+`ArgumentError`.
+"""
+function get_version(uuid::UUID)
```
fredrikekre, line 1071: This is not always correct. I don't think we can support this until the version number of the loaded package is recorded in base.

KristofferC:
Just pointing out that `API` is an internal module.

src/API.jl
KristofferC: We at least likely want this to be a precompiled query and not a runtime one (for relocatability).

staticfloat: > This is not always correct. 

How would this not be correct?  `ctx.env.manifest` would change as the user does `activate` and such, right?

> We at least likely want this to be a precompiled query and not a runtime one (for relocatability).

If a user does `const version = Pkg.API.get_version(Pkg.API.get_uuid(@__MODULE__))` at the top level of their package, that should work, right?

fredrikekre: > ctx.env.manifest would change as the user does activate and such, right?

Yea but that is the point, it can change after you have already loaded the package.

staticfloat:
> Just pointing out that `API` is an internal module.

Do you think it would be better to export it from API and instead have users refer to it as `Pkg.get_version()` and `Pkg.get_uuid()`?

src/API.jl
staticfloat: Ah, I see.  Perhaps that is solvable by documenting to users that they should just store the results at compile time?

staticfloat: The primary usecase I have for this is for users easily getting the version of their own package, yes.  It's brittle to hardcode it in the source; people forget to update it, and it seems wasteful to import an entire TOML parser to try and figure out the UUID/version from the `Project.toml`.

staticfloat:
I'll just suggest to users to parse their own TOMLs.  :)

fredrikekre:
>  it seems wasteful to import an entire TOML parser to try and figure out the UUID/version from the Project.toml.

Well, that should be faster than parsing the full environment as is done in `Context()` anywya.

> I'll just suggest to users to parse their own TOMLs. :)

I believe someone made a small package for this. Maybe it was @omus ?

PR #1951, opened 2020-08-09
RFC: add `add_artifact!` function

### LINKED ISSUE #1950: better API for generating artifact from tarball URL
It would be nice to have a function like
```julia
add_artifact!(artifact_toml::String, name::String, tarball_url::String; platform=nothing, lazy=false)
```
in Pkg.Artifacts, which would automatically download a tarball from `url`, generate all the needed hashes, and add an artifact with the given `name` to `artifact_toml`, optionally specifying a specific platform and whether the artifact should be lazy. BinaryBuilder already does all this, but I think it would be a pretty nice QOL improvement for shipping artifacts independent of BinaryBuilder.

simeonschaub:
This is just an initial prototype and doesn't have tests and docs yet. Is this the right way to do this? Should there be an option for multiple `tarball_url`s as well?

Closes #1950

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -12,7 +12,8 @@ using SHA
 
 export create_artifact, artifact_exists, artifact_path, remove_artifact, verify_artifact,
        artifact_meta, artifact_hash, bind_artifact!, unbind_artifact!, download_artifact,
-       find_artifacts_toml, ensure_artifact_installed, @artifact_str, archive_artifact
+       find_artifacts_toml, ensure_artifact_installed, @artifact_str, archive_artifact,
+       add_artifact!
 
 # keep in sync with Base.project_names and Base.manifest_names
 const artifact_names = ("JuliaArtifacts.toml", "Artifacts.toml")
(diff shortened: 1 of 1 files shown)


simeonschaub:
@staticfloat Could you take a quick look at this? Just to know whether this is the right direction.

staticfloat:
Can you explain a bit what the purpose of this function is?  Is this to make it easier to download a pre-existing tarball and bind it as an artifact?

simeonschaub:
Yes exactly! I found it quite cumbersome with the current API to just ship a tarball that's already hosted somewhere as an artifact, especially when you want to update it periodically. This aims to provide this functionality, so the package developer doesn't have to worry about calculating all the correct checksums manually.

StefanKarpinski:
I wonder if we shouldn't have an external `ArtifactUtils.jl` package that has extra artifact-related functionality.

giordano:
An external package is also fine. I believe many people (mis)use BinaryBuilder to create JLL packages because that's the easiest way to create an Artifacts.toml

simeonschaub:
Is that something that should live under the JuliaLang or JuliaPackaging umbrella, with the goal of factoring out some of BinaryBuilder's functionality into such a package or should I just go ahead and put this into a package of my own?

jlperla:
This was pointed out to me.  Looks like exactly what I needed, but also happy to use a separate package if it is existed.

simeonschaub:
@giordano Should I just go ahead and start a package for this, or do you have grander plans in terms of factoring out parts of BinaryBuilder?

giordano:
No, we don't have any plan about this at the moment

simeonschaub:
@jlperla I started a package here: https://github.com/simeonschaub/ArtifactUtils.jl Should also (hopefully) be registered soon.

PR #1953, opened 2020-08-10
Propagate Threads.nthreads() to Pkg.test subprocess.

fredrikekre:


src/Operations.jl
```diff
@@ -1323,6 +1323,7 @@ function gen_test_code(testfile::String;
         --inline=$(Bool(Base.JLOptions().can_inline) ? "yes" : "no")
         --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
         --track-allocation=$(("none", "user", "all")[Base.JLOptions().malloc_log + 1])
+        --threads=$(Threads.nthreads())
```
fredrikekre, line 1326: Maybe this should be added to `Base.julia_cmd()` actually?

fredrikekre: No.

PR #1955, opened 2020-08-11
Add test for Pkg.build() failure

staticfloat:
Test for https://github.com/JuliaLang/Pkg.jl/pull/1851


test/new.jl
```diff
@@ -1767,6 +1767,13 @@ end
         @test args == [Pkg.PackageSpec(;name="Foo"), Pkg.PackageSpec(;name="Bar")]
         @test opts == Dict(:verbose => true)
     end
+    
+    # Test package that fails build
+    isolate(loaded_depot=true) do; mktempdir() do tempdir
+        package_path = copy_test_package(tempdir, "FailBuild")
+        Pkg.activate(package_path)
+        @test_throws ErrorException Pkg.build()
```
KristofferC, line 1775: Maybe this should be a `PkgError` instead.

staticfloat: Isn't the error that gets thrown dependent on what happens inside of `deps/build.jl`?

KristofferC: No, that's a separate process. It will always throw an ErrorException right now (since `error` is used).

staticfloat: Ah, of course.  I thought you were hinting at my `error()` in `build.jl`, so I got confused.

Issue #1949, opened 2020-08-08
]up only downgrades packages


kescobo:
As mentioned on Slack, I have a weird situation where `]up` is doing nothing but downgrading packages, though it doesn't seem like there's any incompatibility problems:

```
(ECHOAnalysis) pkg> up
   Updating registry at `~/.julia/registries/General`
   Updating git-repo `https://github.com/JuliaRegistries/General.git`
Updating `~/repos/scratch/ECHOAnalysis.jl/Project.toml`
  [3bd8f0ae] ↓ Microbiome v0.6.4 ⇒ v0.6.1
Updating `~/repos/scratch/ECHOAnalysis.jl/Manifest.toml`
  [b4f34e82] - Distances v0.9.0
  [3bd8f0ae] ↓ Microbiome v0.6.4 ⇒ v0.6.1
  [3cdcf5f2] ↓ RecipesBase v1.0.2 ⇒ v0.8.0
  [348f2d5d] ↓ SpatialEcology v0.9.5 ⇒ v0.7.0
  [2913bbd2] ↓ StatsBase v0.33.0 ⇒ v0.32.2
```

This is just a personal package and the issue is easily solved by adding Microbiome 0.6.4 to the deps, so it's not urgent for me, but seems like an odd bug. I'm wondering if it might be a downstream consequence of the move to the BioJulia registry and then back to General, coupled with something I did here: https://github.com/JuliaRegistries/General/pull/10324

Steps to reproduce:

```
$ git clone git@github.com:Klepac-Ceraj-Lab/ECHOAnalysis.jl.git # current commit 4f3e59b
Cloning into 'ECHOAnalysis.jl'...
remote: Enumerating objects: 233, done.
remote: Counting objects: 100% (233/233), done.
remote: Compressing objects: 100% (118/118), done.
remote: Total 1840 (delta 118), reused 193 (delta 83), pack-reused 1607
Receiving objects: 100% (1840/1840), 12.91 MiB | 7.24 MiB/s, done.
Resolving deltas: 100% (1134/1134), done.
$ cd ECHOAnalysis.jl 
$ julia --project
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.0 (2020-08-01)
[...]

PR #1957, opened 2020-08-12
some microoptimizations for the resolver

KristofferC:
These are results for `Pkg.update` in an environment where I added a bunch of big packages (Plots, Makie, DifferentialEquations, ApproxFun etc).

Master:

```
ENV["JULIA_PKGRESOLVE_ACCURACY"] = 1
  966.591 ms (6843000 allocations: 1.01 GiB)

ENV["JULIA_PKGRESOLVE_ACCURACY"] = 2
  2.124 s (11577823 allocations: 3.01 GiB)
```

PR:

```
ENV["JULIA_PKGRESOLVE_ACCURACY"] = 1
  602.627 ms (5432062 allocations: 301.72 MiB)

ENV["JULIA_PKGRESOLVE_ACCURACY"] = 2
   1.308 s (7230777 allocations: 368.51 MiB)
```

The performance improvements here only matter when the greedy solver fails.

The reason for this is that https://github.com/JuliaLang/Pkg.jl/issues/1949 suggests that we might want to bump the resolver accuracy but it would be a bit sad if we lost performance from doing that. The optimized resolver is now pretty close for accuracy 2, as the previous one with accuracy 1.
 
cc @carlobaldassi

### DIFF
--- a/src/Resolve/maxsum.jl
+++ b/src/Resolve/maxsum.jl
@@ -245,7 +245,7 @@ function iterate!(graph::Graph, msgs::Messages, perm::NodePerm)
         graph.ignored[p0] && continue
         maxdiff0 = update!(p0, graph, msgs)
         if maxdiff0 isa Unsat
-            return Unsat
+            return maxdiff0
         end
         maxdiff = max(maxdiff, maxdiff0)
     end
--- a/src/Resolve/maxsum.jl
+++ b/src/Resolve/maxsum.jl
@@ -8,9 +8,9 @@ end
 
 # Some parameters to drive the decimation process
 mutable struct MaxSumParams
-    dec_interval # number of iterations between decimations
-    dec_fraction # fraction of nodes to decimate at every decimation
-                 # step
+    dec_interval::Int # number of iterations between decimations
+    dec_fraction::Float64 # fraction of nodes to decimate at every decimation
+                          # step
 
     function MaxSumParams()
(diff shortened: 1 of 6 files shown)


src/Resolve/maxsum.jl
```diff
@@ -346,6 +358,10 @@ function try_simplify_graph_soft!(graph, sources)
     return true
 end
 
+struct Unsat
+    p0::Int
```
KristofferC, line 362: This was mostly to avoid using `try catch` as control flow.

carlobaldassi:
This all looks good to me, thanks.
As for increasing the default accuracy, it might be that with the growing ecosystem and larger and more complicated dependency graphs it's going to be necessary. I should think if there is any way we can come up with a better heuristic to set it, but at the moment nothing particularly clever comes to mind.

src/Resolve/maxsum.jl
```diff
@@ -235,6 +244,9 @@ function iterate!(graph::Graph, msgs::Messages, perm::NodePerm)
     for p0 in perm
         graph.ignored[p0] && continue
         maxdiff0 = update!(p0, graph, msgs)
+        if maxdiff0 isa Unsat
+            return Unsat
```
carlobaldassi, line 248: Shouldn't this return `maxdiff0`, so as to include the same `p0` field inside the `Unsat` object?

KristofferC: eeh, yes. Seems we are missing some test coverage here..

Issue #1778, opened 2020-04-17
KeyError when adding package introduces a cyclic dependency


johnnychen94:
I get following error with the master branch of Pkg (596998b5a45ecceaab8d48cf316515ba63142494) and Julia 1.5-DEV.634 (not with Julia 1.4.0):

```bash
git clone https://github.com/JuliaGraphics/ColorTypes.jl.git
cd ColorTypes.jl
JULIA_DEPOT_PATH=~/julia_temp julia-latest --startup=no --project=. -e 'using Pkg; Pkg.add("Colors")'
```

```
Resolving package versions...
    Cloning default registries into `~/julia_temp`
######################################################################## 100.0%
      Added registry `General` to `~/julia_temp/registries/General`
  Installed Reexport ────────── v0.2.0
  Installed FixedPointNumbers ─ v0.8.0
  Installed Colors ──────────── v0.12.0
ERROR: KeyError: key "ColorTypes" not found
```

I assume an error is more or less expected as this introduces a cyclic dependency (`ColorTypes` is a dependency of `Colors`), but perhaps a more illustrative error description is needed.

The background is that I want to do package test of `Colors` in `ColorTypes` CI using the following travis script to make sure the dev version of `ColorTypes` doesn't break `Colors`:

```yaml
script:
  - julia --project --color=yes -e 'using Pkg;
                                    Pkg.instantiate();
                                    Pkg.test(coverage=true);
                                    Pkg.add("Colors");
                                    Pkg.test("Colors");'
```

KristofferC:
Yeah, the problem is that we here have a dependency that depends on the package itself which thus means that the manifest is not "closed". 

It might make more sense to write the Travis stuff as

```
julia --project --color=yes -e 'using Pkg;
                                    #Pkg.test(coverage=true);
                                    Pkg.activate(mktempdir())
                                    Pkg.develop([PackageSpec(name="Colors"), PackageSpec(path=".")])
                                    Pkg.test("Colors");'
```

but we should figure out what to do here as well.

PR #1958, opened 2020-08-12
add project to manifest in case a package depends on it, so that manifest is always complete

KristofferC:
Fixes: https://github.com/JuliaLang/Pkg.jl/issues/1778

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -116,11 +116,18 @@ end
 function update_manifest!(ctx::Context, pkgs::Vector{PackageSpec}, deps_map)
     manifest = ctx.env.manifest
     empty!(manifest)
+    if ctx.env.pkg !== nothing
+        pkgs = [pkgs; ctx.env.pkg]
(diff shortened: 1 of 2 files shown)


src/Operations.jl
```diff
@@ -116,11 +116,18 @@ end
 function update_manifest!(ctx::Context, pkgs::Vector{PackageSpec}, deps_map)
     manifest = ctx.env.manifest
     empty!(manifest)
+    if ctx.env.pkg !== nothing
+        pkgs = [pkgs; ctx.env.pkg]
```
fredrikekre, line 120: Should this filter out if it already is in the manifest?

KristofferC: Hm, probably but not sure

fredrikekre: Seems to disappear in some pruning, and they should (?) be equivalent since if it exists in the manifest Pkg must have put it there and then it should match.

PR #1963, opened 2020-08-16
Correctly forward the value of `--depwarn` to the test subprocess

DilumAluthge:
Before this PR:
| Value provided to `--depwarn` by user | Value passed to test subprocess |
| --------------------------------------- | --------------------------------- |
| `error`                                                     | `error` |
| `yes`                                                       | `yes` |
| `no`                                                        | `yes` |

After this PR:
| Value provided to `--depwarn` by user | Value passed to test subprocess |
| --------------------------------------- | --------------------------------- |
| `error`                                                     | `error` |
| `yes`                                                       | `yes` |
| `no`                                                        | `no` |

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1313,13 +1313,20 @@ function gen_test_code(testfile::String;
         append!(empty!(ARGS), $(repr(test_args.exec)))
         include($(repr(testfile)))
         """
+    if Base.JLOptions().depwarn == 2
+        depwarn = "error"
+    elseif Base.JLOptions().depwarn == 1
+        depwarn = "yes"       
+    else
+        depwarn = "no"
+    end
     return ```
         $(Base.julia_cmd())
         --code-coverage=$(coverage ? "user" : "none")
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Since this is a bugfix, would it be possible to backport this PR to Julia 1.5.1?

fredrikekre:
The current behaviour is expected, cf. https://github.com/JuliaLang/Pkg.jl/pull/1763. I have meant to fix JLOptions but forgot about it. I will put it on my list.

DilumAluthge:
> The current behaviour is expected, cf. #1763. I have meant to fix JLOptions but forgot about it. I will put it on my list.

I see. Once you fix JLOptions, will explicitly passing `--depwarn=no` be passed to the subprocess?

fredrikekre:
Yea. The problem right now is that we can't see the difference between `--depwarn=no` and not specifying at all.

DilumAluthge:
Ok cool, I'll close this in favor of the fix you describe.

PR #1968, opened 2020-08-18
Improve consistency of Dict types & construction

timholy:
This defines some aliases and otherwise increases consistency in Dict types. Note that in some cases the final result is no different, but this avoids an unnecessary conversion by creating the Dict to be of the correct type at the outset. Essentially the Pkg equivalent of https://github.com/JuliaLang/julia/pull/37088.

### DIFF
--- a/test/artifacts.jl
+++ b/test/artifacts.jl
@@ -551,7 +551,7 @@ end
         end
 
         # Force Pkg to reload what it knows about artifact overrides
-        Pkg.Artifacts.load_overrides(;force=true)
+        @inferred Union{Nothing,Dict{Symbol,Any}} Pkg.Artifacts.load_overrides(;force=true)
 
         # Verify that the hash-based override worked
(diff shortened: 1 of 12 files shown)


KristofferC:
Some of the artifact code might move to Base (become a new stdlib) in 1.6 so we should make sure that improvements here are also transferred to that.

cc @staticfloat

test/artifacts.jl
```diff
@@ -551,7 +551,7 @@ end
         end
 
         # Force Pkg to reload what it knows about artifact overrides
-        Pkg.Artifacts.load_overrides(;force=true)
+        @inferred(Pkg.Artifacts.load_overrides(;force=true))
```
KristofferC, line 554: Looks like this test fails.

KristofferC:
Something is weird with the Pkg server so the CI that runs with them seems to fail.

PR #1969, opened 2020-08-18
Redesign the HashType handling for better inference

timholy:
Encoding the HashType as a keyword argument tends to cause inference to lose any type parameters due to https://github.com/JuliaLang/julia/issues/36454 and its interaction with `NamedTuple`. Making it a positional argument fixes this problem.

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -1767,7 +1767,7 @@ end
         @test args == [Pkg.PackageSpec(;name="Foo"), Pkg.PackageSpec(;name="Bar")]
         @test opts == Dict(:verbose => true)
     end
-    
+
     # Test package that fails build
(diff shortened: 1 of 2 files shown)


test/new.jl
```diff
@@ -2294,7 +2294,7 @@ end
     end
 end
 
-tree_hash(root::AbstractString) = bytes2hex(Pkg.GitTools.tree_hash(root))
+tree_hash(root::AbstractString) = bytes2hex(@inferred Pkg.GitTools.tree_hash(root))
```
KristofferC, line 2297: Perhaps just making a separate test to test that it infers is slightly better than calling `@inferred` over and over (many times with the same argument).

PR #1980, opened 2020-08-23
Improve inference for paths leading to `similar`

timholy:
The companion PR to https://github.com/JuliaLang/julia/pull/37163

The message in the first commit mentions this, but perhaps worth repeating: using `Vector` rather than `Array` in certain places allows inference to know that `size(A)` is a `Tuple{Int}`, and that ends up fixing some otherwise hard-to-solve problems in inference with the `similar` pipeline.


timholy:
Too hasty, that commit hadn't hit nightly yet. Should be OK tomorrow.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/1984 made a merge conflict with this, sorry.

timholy:
No worries.

timholy:
Given @staticfloat's upcoming Artifacts stdlib, should I modify this in any way?

PR #1981, opened 2020-08-24
Correctly pass on the ctx::Context when adding default registries.

fredrikekre:


KristofferC:
Is this method really needed?

https://github.com/JuliaLang/Pkg.jl/blob/8273c21111522be48d0407dacab18df7ab81d6ff/src/Types.jl#L974-L975

PR #1997, opened 2020-08-31
Change develop error message to state it doesn't support rev

iamed2:
The original error message made me think that I was passing arguments incorrectly or there was some internal bug; turns out that I was just not supposed to use `develop` like that.

I'm not sure if there are any valid uses of `develop` for which this error message is incorrect.


KristofferC (approved):

iamed2:
The test failure here seems to be an unrelated GitHub issue

PR #1998, opened 2020-08-31
Re-add AbstractString method to platform_key_abi

Keno:
BinaryBuilder tries to call it with a SubString. That seems like a reasonable thing to do from
a user API perspective:
```
ERROR: MethodError: no method matching platform_key_abi(::SubString{String})
Closest candidates are:
  platform_key_abi(::String) at /home/keno/julia/usr/share/julia/stdlib/v1.6/Pkg/src/BinaryPlatforms.jl:402
  platform_key_abi() at /home/keno/julia/usr/share/julia/stdlib/v1.6/Pkg/src/BinaryPlatforms.jl:732
Stacktrace:
 [1] CompilerShard(name::SubString{String}, version::VersionNumber, host::SubString{String}, archive_type::Symbol; target::SubString{String})
   @ BinaryBuilderBase ~/.julia/dev/BinaryBuilderBase/src/Rootfs.jl:31
 [2] CompilerShard(art_name::String)
   @ BinaryBuilderBase ~/.julia/dev/BinaryBuilderBase/src/Rootfs.jl:83
 [3] all_compiler_shards()
   @ BinaryBuilderBase ~/.julia/dev/BinaryBuilderBase/src/Rootfs.jl:100
 [4] top-level scope
   @ REPL[3]:1
```

Add back that method, but simply have it convert to string, since there
are no performance concerns.

### DIFF
--- a/src/BinaryPlatforms.jl
+++ b/src/BinaryPlatforms.jl
@@ -513,6 +513,7 @@ function platform_key_abi(machine::String)
     @warn("Platform `$(machine)` is not an officially supported platform$msg")
     return UnknownPlatform()
 end
+platform_key_abi(machine::AbstractString) = platform_key_abi(String(machine))
 
 
 # Define show() for these Platform objects for two reasons:


staticfloat:
@timholy can you help me to understand why it's undesirable to have `AbstractString`s in these APIs?  I have other use cases that are more and more passing things like `SubStrings` through them, and of course I can reify them all to be `Strings`, but it would be helpful to know why just having an `AbstractString` is bad.  Is it because of the backedges that the compiler has to process if a user defines a new type that is `<: AbstractString`?

KristofferC:
AFAIU, the issue is that in the presence of imprecise type information, inference will infer these functions with non-concrete types. That makes them susceptible to invalidations. If the functions are "big" and they call many other functions this can cause significant latency. If instead, the functions are restricted to concrete types, this "cuts off" the non-concrete type inference chain. It is therefore better to structure code such that you have small "conversion" functions that convert the inputs to known concrete types and then have the "core" function accept that concrete type. Note that this is exactly what this PR does. Instead of adding `AbstractString` to the core function of `platform_key_abi` a conversion shim is provided.

staticfloat:
So I should do things like this: https://github.com/JuliaLang/Pkg.jl/commit/fd702864ebf4072f45de3a6b4d7cdfb02c03928c

KristofferC:
Or just take a `SubString{String}` as an argument as well (using a Union). Don't think there will be other cases?

timholy:
Sorry to be slow to notice this.

> @timholy can you help me to understand why it's undesirable to have AbstractStrings in these APIs?

There are two problems:

1. First, from a compile-time perspective, usually there is no good reason to compile "big" methods for "trivial" differences in argument types. For example, say you have a 200-line method (of which there are quite a few in Pkg) that takes an `AbstractString` input; does it really make sense to infer it three times--once for `String`, once for `SubString{String}`, and once for `AbstractString`--if the differences in what it will do with these three input types are trivial and have no impact on runtime performance? Much better to compile small methods that normalize the input types and compile the big method only once. cf https://docs.julialang.org/en/latest/manual/style-guide/#Handle-excess-argument-diversity-in-the-caller-1

2. Lots of Pkg and Artifacts code looks like this: `foo(dict[key])` where `dict::Dict{String,Any}`. `foo` is declared `foo(::AbstractString)`, and so inference will figure out that if the call to `foo` succeeds it must be the case that its input is an `AbstractString`. But inference on `AbstractString` is a mess, leading to crappy performance and huge invalidation risk. Most immediately, any package defining a new `AbstractString` subtype ends up invalidating all of Pkg's internal code. But the risk extends well beyond that, because:
```julia
julia> Base.return_types(length, (AbstractString,))
3-element Vector{Any}:
 Any
 Int64
 Any

julia> Base.return_types(startswith, (AbstractString,))
1-element Vector{Any}:
 Base.Fix2{typeof(startswith), _A} where _A
```
and on Julia 1.5
```julia
julia> Base.return_types(iterate, (AbstractString,))
4-element Array{Any,1}:
 Union{Nothing, Tuple{Char,Int64}}
 Any
 Union{Nothing, Tuple{Any,Any}}
 Any
```
and so anything that *uses* the result on methods like these will also open up invalidation issues. For example,  `i < length(str) + 1` is suddenly a risk for a package that defines new `+` methods if you don't know that `length(str)` returns an `Int`.

Invalidation risk is described in [this blog post](https://julialang.org/blog/2020/08/invalidations/). I've been on a mission to improve inference throughout Julia, and the big sources of poorly inferred calls have been Pkg, REPL, and Distributed. These propagate down through their Base dependencies, although I've severed many of the "dangerous" links now. But string-handling seems to be the big remaining vulnerability, and in particular the entire code-loading path can still be invalidated by packages that add new string types or methods.

timholy:
> So I should do things like this: fd70286

No, that doesn't help:
```julia
julia> Base.return_types(string, (AbstractString,))
4-element Vector{Any}:
 String
 String
 String
 AbstractString
```
If you want to be absolutely, positively certain, do this instead: `String(str)::String`. That's because
```julia
julia> Base.return_types(String, (AbstractString,))
3-element Vector{Any}:
 String
 String
 String
```
That's 3 methods, the current limit of what Julia will inspect to see if all methods return the same type; if a package adds a fourth `String(str::MyNewStringType)` method then `String(str)` will no longer be inferred to return a `String`. Hence the final type-assert is a form of insurance.

Issue #1902, opened 2020-07-05
Use a rolling identifier for Pkg Telemetry


dlakelan:
Instead of generating a UUID when Pkg is first run and continuing to send this UUID with every Pkg telemetry exchange, Pkg should generate a UUID (using a cryptographic RNG) and then send a hash of the UUID concatenated with div(round(Int,time()),60*60*24*60)

This is a counter that increases every 60 days. So every 60 day period provides a snapshot where all user ids remains static, but an individual can not be tracked across long durations of time unless their (secret) UUID is known.

example:
```
uu = UUIDs.uuid4() # store this
counter = div(round(Int,time()),60*60*24*60)
idtosend = reduce(*,map(x->string(x,base=16),SHA.sha3_256("$uu - $counter")))
```

dlakelan:
After discussion on discourse, and giving some example attack vectors, it seems to me it's important that the history can't be reconstructable when the secret is compromised. So I propose to change the implementation here to instead generate and store a UUID using a crypto RNG (such as /dev/urandom), and a date on which that UUID was generated, and then at the next interaction with the Pkg server, if the UUID is more than x days old, generate and store a new UUID before the Pkg interaction... at this point if the UUID is discovered on the computer, it can't be tied directly to previous UUIDs that were in use on this computer.

This prevents the long term tracking, even if the local computer's "secret" is compromised (say by a discovery request in a lawsuit or by an investigation by a malicious authoritarian government seeking to punish dissidents who release or contribute to free software enabling say cryptography or the like)

dlakelan:
A further refinement based on further discussions on discourse... (see eg: https://discourse.julialang.org/t/pkg-jl-telemetry-should-be-opt-in/42209/366?u=dlakelan and associated conversation).

Client side:

1) Julia generates a UUID which is static, and a 128 bit salt which changes.
2) Telemetry sends the first 48 bits of sha_256(salt * UUID * salt) (or some similar crypto transform). 48 bits is enough that most of these are unique and only a very small number correspond to collisions between two different people. However there are 2^(128-48) = 1e24 different salt values that could have generated a given identifier, and therefore finding one of them is not at all informative.
3) By default Julia chooses say probability 0.25/day of changing the salt. This gives a few days window of correlatable information on the default install. The duration of this default forms a continuously variable "degree of opt-in/out"
[...]

Issue #1901, opened 2020-07-05
Allow the package server to opt out of telemetry


GunnarFarneback:
Consider the use case that you have a package server that is internal to your organization (e.g. company) and you have no interest in recording telemetry for your internal users. Then it would be good if the **package server** could opt of telemetry. Although the benefit of eliminating the telemetry HTTP headers from the communication on the internal network is fairly minor, the real value of this option would be that your local users wouldn't need to ponder what
```
[ Info: LEGAL NOTICE: package operations send anonymous data about your system to http://your.local.package.server
(your current package server), including the operating system and Julia versions you are using, and a random
client UUID. Running `Pkg.telemetryinfo()` will show exactly what data is sent. See
https://julialang.org/legal/data/ for more details about what this data is used for, how long it is retained,
and how to opt out of sending it.
```
means to them as there is no telemetry sent.

Proposal:
* The package server has a resource `/meta/telemetry_enabled`, which returns some suitable representation of `true` if it wants to have telemetry sent.
* When `Pkg` first encounters a new package server it first queries this resource and if the answer is no, turns off telemetry and doesn't print the legal notice.

DilumAluthge:
I like this idea. I'm thinking of deploying a Pkg server internally. No point in sending telemetry/metadata headers (or printing a privacy notice) if the Pkg server doesn't want it.

StefanKarpinski @staticfloat Would this be a feasible feature to add?

DilumAluthge:
I like this idea. I'm thinking of deploying a Pkg server internally. No point in sending telemetry/metadata headers (or printing a privacy notice) if the Pkg server doesn't want it.

@StefanKarpinski @staticfloat Would this be a feasible feature to add?

DilumAluthge:
I like this idea. I'm thinking of deploying a Pkg server internally. No point in sending telemetry/metadata headers (or printing a privacy notice) if the Pkg server doesn't want it.

@StefanKarpinski @staticfloat Would this be a feasible feature to add?

StefanKarpinski:
There's no notice or option on release-1.5, so this isn't an issue for that release. It's unclear at this point what 1.6 will have, although we may have an opt-in/out nag in interactive mode, asking people to either opt-out of any future prompts or opt-in to sending a client UUID. In that case, it could make sense for the server to be able to control whether users get prompted with this nag or not.

One possible design is to generate the client UUID on the server side and prompt the client if they're willing to save it on the client side to be sent again in the future or not.

[...]

Issue #1900, opened 2020-07-05
Disabling telemetry for a specific set of installations


alhirzel:
With the telemetry changes (https://discourse.julialang.org/t/pkg-jl-telemetry-should-be-opt-in/42209), I feel there will be legitimate concern when Julia is used in commercial environments, especially when internal packages on an internal registry use external/public dependencies. **To help manage these concerns, it would be best to build-in the capability to disable this telemetry across a set of PCs.** Currently, it may be possible (albeit kludgy) to do with the existing documentation for the `telemetry.toml` file (https://julialang.org/legal/data/) and the `/etc/skel` directory for Linux and equivalent profile templates for Windows. My request is to make this easier.

One suggestion that is present in the Discourse is respect for the `DO_NOT_TRACK` environment variable (https://discourse.julialang.org/t/pkg-jl-telemetry-should-be-opt-in/42209/172).

Another possibility is clarifying the documentation currently provided on the referenced website (https://julialang.org/legal/data/) with a how-to for disabling telemetry for a given network. Ideally, this would look something like "redirect the `telemetry.pkg.julialang.org` domain to a black hole". As-suggested, this would require a change to the existing design so that telemetry data posts to a different domain.

DilumAluthge:
I don't believe that this is necessary.

If you are in a commercial environment with security considerations, you can always set up your own [Pkg server](https://github.com/JuliaPackaging/PkgServer.jl) and use it by setting the environment variable:
```
export JULIA_PKG_SERVER="my.custom.pkg.server.mycompany.example.com
```

If you don't feel like setting up your own Pkg server, you can opt-out of all Pkg servers by setting the environment variable:
```
export JULIA_PKG_SERVER=""
```
If you opt-out of all Pkg servers, then all packages, artifacts, etc. will be downloaded from GitHub, GitLab, etc.

Therefore, because two very good options already exist, I don't think that this issue is necessary. In particular, I don't think that this issue should block Julia 1.5.

alhirzel:
Hey, thanks for your reply! I agree that growing the scope of an internal registry is one way to disable telemetry reporting to public pkg servers, but this will increase overhead and reduce flexibility for commercial users (risk of negative perception as presented in https://github.com/JuliaLang/julia/issues/36548).

[...]

Issue #1899, opened 2020-07-05
Disabling telemetry for a specific package


alhirzel:
With the telemetry changes coming in 1.5 (https://discourse.julialang.org/t/pkg-jl-telemetry-should-be-opt-in/42209/188), I feel there will be legitimate concern when Julia is used in commercial environments, especially when packages are internally published using a registry with external/public dependencies. I would like to see an API which is prominently documented that can be queried from a package to determine whether telemetry is enabled or disabled. Perhaps this could be accomplished by documenting the `telemetry_status` function and providing an example -  https://github.com/JuliaLang/Pkg.jl/pull/1895/commits/1f8ead24c18500a9b53f72436057772324ba1b88#diff-47db2d66769667be0982a7b56891b4a8R1005  /  #1895.

If I were at a company which wanted to ensure telemetry was never enabled (for example when distributing software to a customer who requires that it is disabled), I would be checking this flag at the start of e.g. an install script. Even better would be to have this flag available at the `Project.toml` level so that a specific project is guaranteed to never report its own telemetry (or telemetry for descendant packages) when `]update`ing, `]resolve`ing or `]instantiate`ing.

I understand that the telemetry functionality is only within the package manager, which means that e.g. compiled binaries will not do any reporting. I see such binary distribution as a future way of solving this problem. The binary distribution process is not currently mature enough to enable this, however.

DilumAluthge:
> If I were at a company which wanted to ensure telemetry was never enabled (e.g. when distributing software to a customer who requires that it is disabled), I would be checking this flag at the start of an install script.

If you want to ensure that telemetry is never enabled, then you can either set up your own Pkg server, or you can completely opt out of all Pkg servers by setting `export JULIA_PKG_SERVER=""`. See my comment on https://github.com/JuliaLang/Pkg.jl/issues/1900#issuecomment-653927692 for more details.

alhirzel:
Thank you for the good info, both here and there! Do you feel that the `telemetry_status` function should not be documented and exposed for public use? :)

[...]

PR #1999, opened 2020-09-01
1.6: remove telemetry; do not send client UUIDs (#1909)

KristofferC:
Closes https://github.com/JuliaLang/Pkg.jl/issues/1902
Closes https://github.com/JuliaLang/Pkg.jl/issues/1901
Closes https://github.com/JuliaLang/Pkg.jl/issues/1900
Closes https://github.com/JuliaLang/Pkg.jl/issues/1899


StefanKarpinski:
Fine by me. We may want to have a nag prompt for people to allow more telemetry, but in that case I think that should probably collect more complete info since people are explicitly allowing it. Might as well do this for now.

PR #2001, opened 2020-09-01
Move Artifacts and BinaryPlatforms out of Pkg

staticfloat:
This PR guts the `Artifacts` and `BinaryPlatforms` modules, replacing them with imports from Base and stdlibs.  This will obviously not pass until https://github.com/JuliaLang/julia/pull/37320 is merged.


staticfloat:
This will fail until we merge https://github.com/JuliaLang/julia/pull/37320, but on my local machine, when run with that version it passes tests.

StefanKarpinski:
YES!!

timholy:
Even the name of this branch is awesome.

timholy:
How much of `BinaryPlatforms_compat.jl`. It causes problems:

```julia
julia> methods(Base.BinaryPlatforms.libgfortran_version, (Base.BinaryPlatforms.AbstractPlatform,))
# 5 methods for generic function "libgfortran_version":
[1] libgfortran_version(p::Pkg.BinaryPlatforms.Linux) in Pkg.BinaryPlatforms at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.9/Pkg/src/BinaryPlatforms_compat.jl:80
[2] libgfortran_version(p::Pkg.BinaryPlatforms.Windows) in Pkg.BinaryPlatforms at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.9/Pkg/src/BinaryPlatforms_compat.jl:80
[3] libgfortran_version(p::Pkg.BinaryPlatforms.MacOS) in Pkg.BinaryPlatforms at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.9/Pkg/src/BinaryPlatforms_compat.jl:80
[4] libgfortran_version(p::Pkg.BinaryPlatforms.FreeBSD) in Pkg.BinaryPlatforms at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.9/Pkg/src/BinaryPlatforms_compat.jl:80
[5] libgfortran_version(p::Base.BinaryPlatforms.AbstractPlatform) in Base.BinaryPlatforms at binaryplatforms.jl:453
```

5 is more than the limit of 3 for successful abstract inference. It means that calls like `Base.BinaryPlatforms.libgfortran_version` get inferred as returning `Any`, even though all 5 methods return a `Union{Nothing, VersionNumber}`. This in turn leads to about 3000 invalidations from loading a package like InlineStrings.jl, including invalidation of `Base.require` which is *always* bad news.

staticfloat:
I think we can try deleting it, but we'll probably need to do a PkgEval run to see how many people are actually still using these.

A quick [julia hub code search](https://juliahub.com/ui/Search?q=Pkg.BinaryPlatforms&type=code&f=.%2a%28%3F%21_jll%29.%2a) shows that nobody other than JLL packages are attempting to use `Pkg.BinaryPlatforms`, but that may not be 100% reliable, as for instance, [BinaryProvider](https://github.com/JuliaPackaging/BinaryProvider.jl/blob/master/src/BinaryProvider.jl#L4) uses it but doesn't show up in that search.  I'm pretty sure this change would break BinaryProvider, so it would be good to know who that would break.

Alternatively, we can just simplify this code to be less dispatch-heavy.  Those methods you've noticed at the end are the same for all of these "compat" types, so we can just simply hoist the definition out of the loop and provide a union type as the argument:

```
const PlatformUnion = Union{Linux,MacOS,Windows,FreeBSD}
for f in (:libgfortran_version, :libstdcxx_version, :platform_name, :wordsize, :platform_dlext, :tags, :triplet)
    @eval begin
        $(f)(p::PlatformUnion) = $(f)(p.p)
    end
end
```

giordano:
> I'm pretty sure this change would break BinaryProvider, so it would be good to know who that would break.

We can cap compatibility of BinaryProvider in the registry.  It's already quite broken in latest releases.

timholy:
I made a failed run at this in #3017. I just don't know enough about the requirements to do this well.

timholy:
Thanks for the suggestion @staticfloat, that seemed to work well (#3020).

PR #2015, opened 2020-09-11
Better registry verification.

fredrikekre:


staticfloat:
Is there any way to travel one of these code paths for coverage's sake?  Purposefully mangle a Registry in a test somewhere perhaps?

fredrikekre:
Yea I am also unsure if we can't be this aggressive about deleting things. Will this render Pkg unusable on Windows computers that decides to delete folders randomly? But I guess it usually works the second time around since users report that deleting the registry and adding again solves the problem...

Issue #1578, opened 2020-01-02
precompile misses many packages


timholy:
I ran with `@show pkg stale` in `Pkg.API.precompile` and verified that many packages that need precompilation are not getting it. For example, I can do the following:

```julia
pkg> precompile
pkg> dev ColorTypes
pkg> precompile  # ColorTypes properly precompiles, but none of its dependencies do
pkg> precompile  # this doesn't help
julia> using Images
[ Info: Precompiling Images [916415d5-f1e6-5110-898d-aaa5f9f070e0]
```
The latter should happen with the `precompile` statement, not the `using` statement.

KristofferC:
Hm, the code there should mimic how loading itself determines if it needs to precompile but maybe it has bitrotted a bit?

KristofferC:
When you say

> `# ColorTypes properly precompiles, but none of its dependencies do`

You mean FixedPointNumbers doesn't precompile again? Did it have to?

Also, is Images.jl in the current project or available through the global project?

A bit more detailed reproduction steps would be great here.

timholy:
Sorry, I should have said "dependents," meant the packages that depend on ColorTypes. 

Here's more info. Quite a few of the packages that show up during the `precompile` step are not yet public. (You'll be interested in `MethodCompiler` when I get around to finishing it, it's an alternative to https://github.com/JuliaDebug/JuliaInterpreter.jl/pull/309. The goal is to allows you to compile individual frames. It compiles in builtins, ccalls, and llvmcalls, but breaks long inference chains by replacing all other `:call` exprs in lowered code with `do_call(f, args)`, where `do_call` is not specialized on `f`. Hence you basically compile each method as a standalone object.)

```julia
julia> versioninfo()
Julia Version 1.3.1-pre.20
Commit 90620af6b9* (2019-12-19 17:59 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-6.0.1 (ORCJIT, skylake)
Environment:
  JULIA_CPU_THREADS = 4

julia> using Images
[ Info: Precompiling Images [916415d5-f1e6-5110-898d-aaa5f9f070e0]

(v1.3) pkg> st ColorTypes
    Status `~/.julia/environments/v1.3/Project.toml`
[...]

PR #2018, opened 2020-09-13
Parallelize Pkg.precompile

ianshmean:
Pkg on julia master:
```
(dev) pkg> st
Status `~/Documents/dev/Project.toml`
  [0c46a032] DifferentialEquations v6.15.0
  [91a5bcdd] Plots v1.6.3

julia> @time Pkg.precompile()
Precompiling project...
[ Info: Precompiling DifferentialEquations [0c46a032-eb83-5123-abaf-570d42b7fbaa]
312.061599 seconds (133.69 k allocations: 9.246 MiB)
```

This PR (same project, with `.julia/compiled` emptied out): 
```
julia> @time Pkg.precompile()
Precompiling project...
[ Info: Precompiling x264_jll [1270edf5-f2f9-52d2-97e9-ab00b5d0237a]
...
[ Info: Precompiling DifferentialEquations [0c46a032-eb83-5123-abaf-570d42b7fbaa]
 87.851553 seconds (4.25 M allocations: 216.406 MiB, 0.13% gc time)
```

The approach taken here async queues up all the precomp jobs for all deps in the manifest, and each watches for when its deps are all precomped before starting.

Thanks to @oxinabox for conceptualization of this approach.

## Performance considerations
- ~~The process could launch an unlimited number of julia instances (via `compilecache`), so we might want to add a hard limit.~~
- CPU load is initially high while there are many deps with few deps of their own, and falls off as the larger packages initiate. I don't see much room for further parallelization optimization
Screen Shot 2020-09-13 at 2 20 23 AM

- Memory pressure on my system seemed unaffected in this test case
Screen Shot 2020-09-13 at 2 20 38 AM

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -909,10 +909,10 @@ function precompile(ctx::Context)
         push!(pkg_dep_uuid_lists, collect(keys(ctx.env.project.deps)))
     end    
     
-    precomp_events = Dict{Base.UUID,Base.Event}()
+    was_processed = Dict{Base.UUID,Base.Event}()
     was_recompiled = Dict{Base.UUID,Bool}()
     for pkgid in pkgids
-        precomp_events[pkgid.uuid] = Base.Event()
+        was_processed[pkgid.uuid] = Base.Event()
         was_recompiled[pkgid.uuid] = false
     end
     
@@ -934,11 +934,11 @@ function precompile(ctx::Context)
         occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String)) && continue
         
         @async begin
-            for dep_uuid in pkg_dep_uuid_lists[i]
-                wait(precomp_events[dep_uuid])
+            for dep_uuid in pkg_dep_uuid_lists[i] # wait for deps to finish
+                wait(was_processed[dep_uuid])
             end
             if errored # early termination of precompilation session
-                notify(precomp_events[pkg.uuid])
+                notify(was_processed[pkg.uuid])
                 return
             end
             
@@ -953,11 +953,11 @@ function precompile(ctx::Context)
                     errored = true
                     throw(err)
                 finally
-                    notify(precomp_events[pkg.uuid])
+                    notify(was_processed[pkg.uuid])
                     Base.release(parallel_limiter)
                 end
             else
-                notify(precomp_events[pkg.uuid])
+                notify(was_processed[pkg.uuid])
             end
         end  
     end
--- a/src/API.jl
+++ b/src/API.jl
@@ -942,6 +942,7 @@ function precompile(ctx::Context)
                 return
             end
             
+            # skip stale checking and force compilation if any dep was recompiled in this session
             any_dep_recompiled = any(map(dep_uuid->was_recompiled[dep_uuid], pkg_dep_uuid_lists[i]))
             if any_dep_recompiled || is_stale(paths, sourcepath)
                 Base.acquire(parallel_limiter)
--- a/src/API.jl
+++ b/src/API.jl
@@ -915,7 +915,7 @@ function precompile(ctx::Context)
         precomp_events[pkgid.uuid] = Base.Event()
         was_recompiled[pkgid.uuid] = false
     end
-    # TODO: since we are a complete list, but not topologically sorted, handling of recursion will be completely at random
+    
     errored = false
     @sync for (i, pkg) in pairs(pkgids)
         paths = Base.find_all_in_cache_path(pkg)
@@ -937,7 +937,6 @@ function precompile(ctx::Context)
             for path_to_try in paths::Vector{String}
                 staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache()) #|| any(deps_recompiled)
                 staledeps === true && continue
-                # TODO: else, this returns a list of packages that may be loaded to make this valid (the topological list)
                 stale = false
                 break
             end
--- a/src/API.jl
+++ b/src/API.jl
@@ -894,9 +894,6 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
-    
-    is_stdlib_from_name(name::String) = name in values(stdlibs())
-    
     printpkgstyle(ctx, :Precompiling, "project...")
     
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
@@ -904,17 +901,17 @@ function precompile(ctx::Context)
     
     man = Pkg.Types.read_manifest(ctx.env.manifest_file)
     pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    pkg_dep_lists = [collect(keys(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    filter!.(!is_stdlib_from_name, pkg_dep_lists)
+    pkg_dep_uuid_lists = [collect(values(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
+    filter!.(!is_stdlib, pkg_dep_uuid_lists)
 
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
-        push!(pkg_dep_lists, collect(keys(ctx.env.project.deps)))
+        push!(pkg_dep_uuid_lists, collect(keys(ctx.env.project.deps)))
     end    
     
-    precomp_events = Dict{String,Base.Event}()
+    precomp_events = Dict{Base.UUID,Base.Event}()
     for pkgid in pkgids
-        precomp_events[pkgid.name] = Base.Event()
+        precomp_events[pkgid.uuid] = Base.Event()
     end
     
     precomp_tasks = Task[]
@@ -936,15 +933,15 @@ function precompile(ctx::Context)
         end
         if stale
             t = @async begin
-                length(pkg_dep_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_lists[i]))
+                length(pkg_dep_uuid_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_uuid_lists[i]))
                 Base.acquire(parallel_limiter)
                 Base.compilecache(pkg, sourcepath)
-                notify(precomp_events[pkg.name])
+                notify(precomp_events[pkg.uuid])
                 Base.release(parallel_limiter)
             end  
             push!(precomp_tasks, t) 
         else
-            notify(precomp_events[pkg.name])
+            notify(precomp_events[pkg.uuid])
         end
     end
     wait.(precomp_tasks)
--- a/src/API.jl
+++ b/src/API.jl
@@ -895,14 +895,26 @@ end
 precompile() = precompile(Context())
 function precompile(ctx::Context)
     printpkgstyle(ctx, :Precompiling, "project...")
+    
+    man = Pkg.Types.read_manifest(ctx.env.manifest_file)
+    pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
+    pkg_dep_lists = [collect(keys(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
+    filter!.(!Pkg.Operations.is_stdlib, pkg_dep_lists)
+    
+    perm = sortperm(length.(pkg_dep_lists), rev=true) # sort so that we queue pkgs with fewer deps first
+    pkgids = pkgids[perm]
+    pkg_dep_lists = pkg_dep_lists[perm]
 
-    pkgids = [Base.PkgId(uuid, name) for (name, uuid) in ctx.env.project.deps if !is_stdlib(uuid)]
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
-    end
+        push!(pkg_dep_lists, collect(keys(ctx.env.project.deps)))
+    end    
+    
+    precomp_list = String[]
+    precomp_tasks = Task[]
 
     # TODO: since we are a complete list, but not topologically sorted, handling of recursion will be completely at random
-    for pkg in pkgids
+    for (i, pkg) in pairs(pkgids)
         paths = Base.find_all_in_cache_path(pkg)
         sourcepath = Base.locate_package(pkg)
         sourcepath === nothing && continue
@@ -917,9 +929,21 @@ function precompile(ctx::Context)
             break
         end
         if stale
-            Base.compilecache(pkg, sourcepath)
+            t = @async begin
+                if length(pkg_dep_lists[i]) > 0 # if pkg has deps
+                    while length(precomp_list) == 0 || !all(map(x->in(x, precomp_list), pkg_dep_lists[i])) # if all deps not precomped
+                        sleep(0.001)
+                    end
+                end
+                Base.compilecache(pkg, sourcepath)
+                push!(precomp_list, pkg.name) 
+            end  
+            push!(precomp_tasks, t) 
+        else
+            push!(precomp_list, pkg.name)
         end
     end
+    wait.(precomp_tasks)
     nothing
 end
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -915,19 +915,24 @@ function precompile(ctx::Context)
         precomp_events[pkgid.uuid] = Base.Event()
         was_recompiled[pkgid.uuid] = false
     end
-    
-    precomp_tasks = Task[]
     # TODO: since we are a complete list, but not topologically sorted, handling of recursion will be completely at random
-    for (i, pkg) in pairs(pkgids)
+    errored = false
+    @sync for (i, pkg) in pairs(pkgids)
         paths = Base.find_all_in_cache_path(pkg)
         sourcepath = Base.locate_package(pkg)
         sourcepath === nothing && continue
         # Heuristic for when precompilation is disabled
         occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String)) && continue
         
-        t = @async begin
-            length(pkg_dep_uuid_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_uuid_lists[i]))
-            any_dep_recompiled = any(map(x->was_recompiled[x], pkg_dep_uuid_lists[i]))
+        @async begin
+            for dep_uuid in pkg_dep_uuid_lists[i]
+                wait(precomp_events[dep_uuid])
+            end
+            if errored # early termination of precompilation session
+                notify(precomp_events[pkg.uuid])
+                return
+            end
+            any_dep_recompiled = any(map(dep_uuid->was_recompiled[dep_uuid], pkg_dep_uuid_lists[i]))
             stale = true
             for path_to_try in paths::Vector{String}
                 staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache()) #|| any(deps_recompiled)
@@ -938,17 +943,21 @@ function precompile(ctx::Context)
             end
             if any_dep_recompiled || stale
                 Base.acquire(parallel_limiter)
-                Base.compilecache(pkg, sourcepath)
-                was_recompiled[pkg.uuid] = true
-                notify(precomp_events[pkg.uuid])
-                Base.release(parallel_limiter)
+                try
+                    Base.compilecache(pkg, sourcepath)
+                    was_recompiled[pkg.uuid] = true
+                catch err
+                    errored = true
+                    throw(err)
+                finally
+                    notify(precomp_events[pkg.uuid])
+                    Base.release(parallel_limiter)
+                end
             else
                 notify(precomp_events[pkg.uuid])
             end
         end  
-        push!(precomp_tasks, t) 
     end
-    wait.(precomp_tasks)
     nothing
 end
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -894,6 +894,9 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
+    
+    is_stdlib_from_name(name::String) = name in values(stdlibs())
+    
     printpkgstyle(ctx, :Precompiling, "project...")
     
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
@@ -902,7 +905,7 @@ function precompile(ctx::Context)
     man = Pkg.Types.read_manifest(ctx.env.manifest_file)
     pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
     pkg_dep_lists = [collect(keys(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    filter!.(!Pkg.Operations.is_stdlib, pkg_dep_lists)
+    filter!.(!is_stdlib_from_name, pkg_dep_lists)
 
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
--- a/src/API.jl
+++ b/src/API.jl
@@ -896,6 +896,9 @@ precompile() = precompile(Context())
 function precompile(ctx::Context)
     printpkgstyle(ctx, :Precompiling, "project...")
     
+    num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
+    parallel_limiter = Channel{Bool}(num_tasks)
+    
     man = Pkg.Types.read_manifest(ctx.env.manifest_file)
     pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
     pkg_dep_lists = [collect(keys(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
@@ -931,8 +934,10 @@ function precompile(ctx::Context)
                         sleep(0.001)
                     end
                 end
+                put!(parallel_limiter, true)
                 Base.compilecache(pkg, sourcepath)
                 push!(precomp_list, pkg.name) 
+                take!(parallel_limiter)
             end  
             push!(precomp_tasks, t) 
         else
--- a/src/API.jl
+++ b/src/API.jl
@@ -900,7 +900,7 @@ function precompile(ctx::Context)
     printpkgstyle(ctx, :Precompiling, "project...")
     
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
-    parallel_limiter = Channel{Bool}(num_tasks)
+    parallel_limiter = Base.Semaphore(num_tasks)
     
     man = Pkg.Types.read_manifest(ctx.env.manifest_file)
     pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
@@ -937,10 +937,10 @@ function precompile(ctx::Context)
         if stale
             t = @async begin
                 length(pkg_dep_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_lists[i]))
-                put!(parallel_limiter, true)
+                Base.acquire(parallel_limiter)
                 Base.compilecache(pkg, sourcepath)
                 notify(precomp_events[pkg.name])
-                take!(parallel_limiter)
+                Base.release(parallel_limiter)
             end  
             push!(precomp_tasks, t) 
         else
--- a/src/API.jl
+++ b/src/API.jl
@@ -903,19 +903,20 @@ function precompile(ctx::Context)
     pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
     pkg_dep_uuid_lists = [collect(values(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
     filter!.(!is_stdlib, pkg_dep_uuid_lists)
-
+    
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
         push!(pkg_dep_uuid_lists, collect(keys(ctx.env.project.deps)))
     end    
     
     precomp_events = Dict{Base.UUID,Base.Event}()
+    was_recompiled = Dict{Base.UUID,Bool}()
     for pkgid in pkgids
         precomp_events[pkgid.uuid] = Base.Event()
+        was_recompiled[pkgid.uuid] = false
     end
     
     precomp_tasks = Task[]
-
     # TODO: since we are a complete list, but not topologically sorted, handling of recursion will be completely at random
     for (i, pkg) in pairs(pkgids)
         paths = Base.find_all_in_cache_path(pkg)
@@ -923,26 +924,29 @@ function precompile(ctx::Context)
         sourcepath === nothing && continue
         # Heuristic for when precompilation is disabled
         occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String)) && continue
-        stale = true
-        for path_to_try in paths::Vector{String}
-            staledeps = Base.stale_cachefile(sourcepath, path_to_try)
-            staledeps === true && continue
-            # TODO: else, this returns a list of packages that may be loaded to make this valid (the topological list)
-            stale = false
-            break
-        end
-        if stale
-            t = @async begin
-                length(pkg_dep_uuid_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_uuid_lists[i]))
+        
+        t = @async begin
+            length(pkg_dep_uuid_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_uuid_lists[i]))
+            any_dep_recompiled = any(map(x->was_recompiled[x], pkg_dep_uuid_lists[i]))
+            stale = true
+            for path_to_try in paths::Vector{String}
+                staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache()) #|| any(deps_recompiled)
+                staledeps === true && continue
+                # TODO: else, this returns a list of packages that may be loaded to make this valid (the topological list)
+                stale = false
+                break
+            end
+            if any_dep_recompiled || stale
                 Base.acquire(parallel_limiter)
                 Base.compilecache(pkg, sourcepath)
+                was_recompiled[pkg.uuid] = true
                 notify(precomp_events[pkg.uuid])
                 Base.release(parallel_limiter)
-            end  
-            push!(precomp_tasks, t) 
-        else
-            notify(precomp_events[pkg.uuid])
-        end
+            else
+                notify(precomp_events[pkg.uuid])
+            end
+        end  
+        push!(precomp_tasks, t) 
     end
     wait.(precomp_tasks)
     nothing
--- a/src/API.jl
+++ b/src/API.jl
@@ -900,10 +900,6 @@ function precompile(ctx::Context)
     pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
     pkg_dep_lists = [collect(keys(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
     filter!.(!Pkg.Operations.is_stdlib, pkg_dep_lists)
-    
-    perm = sortperm(length.(pkg_dep_lists), rev=true) # sort so that we queue pkgs with fewer deps first
-    pkgids = pkgids[perm]
-    pkg_dep_lists = pkg_dep_lists[perm]
 
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
--- a/src/API.jl
+++ b/src/API.jl
@@ -937,15 +937,16 @@ function precompile(ctx::Context)
             for dep_uuid in pkg_dep_uuid_lists[i] # wait for deps to finish
                 wait(was_processed[dep_uuid])
             end
-            if errored # early termination of precompilation session
-                notify(was_processed[pkg.uuid])
-                return
-            end
             
             # skip stale checking and force compilation if any dep was recompiled in this session
             any_dep_recompiled = any(map(dep_uuid->was_recompiled[dep_uuid], pkg_dep_uuid_lists[i]))
-            if any_dep_recompiled || is_stale(paths, sourcepath)
+            if !errored && (any_dep_recompiled || is_stale(paths, sourcepath))
                 Base.acquire(parallel_limiter)
+                if errored # catch things queued before error occurred
+                    notify(was_processed[pkg.uuid])
+                    Base.release(parallel_limiter)
+                    return
+                end
                 try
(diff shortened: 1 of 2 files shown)


ianshmean:
One more datapoint. Adding `Images` nicely highlights the speed up when deps aren't closely interlinked.. it's only a few seconds slower than above 

julia master
```
(dev) pkg> st
Status `~/Documents/dev/Project.toml`
  [0c46a032] DifferentialEquations v6.15.0
  [916415d5] Images v0.22.4
  [91a5bcdd] Plots v1.6.3

julia> @time Pkg.precompile()
Precompiling project...
[ Info: Precompiling DifferentialEquations [0c46a032-eb83-5123-abaf-570d42b7fbaa]
[ Info: Precompiling Images [916415d5-f1e6-5110-898d-aaa5f9f070e0]
323.380337 seconds (145.01 k allocations: 10.265 MiB)

```
this PR
```
92.614622 seconds (4.43 M allocations: 229.642 MiB, 0.23% gc time)
```

src/API.jl
```diff
@@ -917,9 +925,21 @@ function precompile(ctx::Context)
             break
         end
         if stale
-            Base.compilecache(pkg, sourcepath)
+            t = @async begin
+                if length(pkg_dep_lists[i]) > 0 # if pkg has deps
+                    while length(precomp_list) == 0 || !all(map(x->in(x, precomp_list), pkg_dep_lists[i])) # if all deps not precomped
+                        sleep(0.001)
+                    end
```
staticfloat, line 932: Spinning in loops like this is probably adding precious milliseconds to your precompile time.  Yes, it's a tiny fraction of the overall time, but you could use `Event()`'s to signify that a dep is done being precompiled.  You build a `Dict{PkgId,Event}()`, then in each `Task`, iterate over all deps and `wait()` upon their deps.  Once you're done precompiling your module, you `notify()` your event and you're done!  No more spinning.

src/API.jl
```diff
@@ -917,9 +925,21 @@ function precompile(ctx::Context)
             break
         end
         if stale
-            Base.compilecache(pkg, sourcepath)
+            t = @async begin
```
staticfloat, line 928: You should probably limit the parallelization here somewhat.  I suggest limiting to `Sys.CPU_THREADS + 1` by default, and limit with an environment variable or something.  `num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))` perhaps?

src/API.jl
```diff
@@ -917,9 +925,21 @@ function precompile(ctx::Context)
...
+                if length(pkg_dep_lists[i]) > 0 # if pkg has deps
+                    while length(precomp_list) == 0 || !all(map(x->in(x, precomp_list), pkg_dep_lists[i])) # if all deps not precomped
+                        sleep(0.001)
+                    end
+                end
+                Base.compilecache(pkg, sourcepath)
```
staticfloat, line 934: To limit parallelization, I suggest launching a million tasks like this, but then creating a `Channel(num_tasks)`, then having each task `put!()` something into it just before this call to `compilecache`, then when it's finished, you `take!()` something back out.  This will create, essentially, an `N`-parallel critical section, and allow `N` tasks to be running that section at once, while all others are blocked, waiting for the channel to free up space.

staticfloat (commented):
This is awesome!

src/API.jl
ianshmean: Nice. Ok

src/API.jl
ianshmean: Sounds good

src/API.jl
ianshmean: Oh nice. Yeah that's better

ianshmean:
Thanks @staticfloat. Original example sped up from 87 seconds, to 80, with much fewer allocations 👍🏻
```
80.721224 seconds (854.44 k allocations: 76.073 MiB, 0.11% gc time)
```

src/Types.jl
```diff
@@ -372,6 +372,7 @@ function stdlibs()
     return STDLIB[]
 end
 is_stdlib(uuid::UUID) = uuid in keys(stdlibs())
+is_stdlib(name::String) = name in values(stdlibs())
```
oxinabox, line 375: This seems bad for when we have packages bwith same names as stdlibs.
Like TOML.jl?

fredrikekre: Yea this shouldn't be necessary since we know the UUIDs of everything.

src/API.jl
oxinabox: Would a `Base.Semaphore` be more idiomatic?
Conceptual it is for this, right?

It would be thrrad-safe also.
Which might introduce some overhead?
Idk how hard it is to instantiate a `Semaphore` with a nonthreadsafe `Condition`

src/Types.jl
ianshmean: But we don't directly know the uuids of deps of deps in the manifest. That's a string name list. Perhaps the proper thing is to convert those to PkgId lists, but I thought that if this was ok, it would be faster. 

Either way, you have to match a string name value. This way just skips the uuid comparison.

ianshmean: Perhaps it shouldn't be a separate function though.. I could internalize it, to prevent promoting use.

ianshmean: Now a local method 👍🏻

fredrikekre: > But we don't directly know the uuids of deps of deps in the manifest.

Yes we do? Everything is there.

src/API.jl
```diff
@@ -894,15 +894,33 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
+    
+    is_stdlib_from_name(name::String) = name in values(stdlibs())
+    
     printpkgstyle(ctx, :Precompiling, "project...")
+    
+    num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
+    parallel_limiter = Channel{Bool}(num_tasks)
```
oxinabox, line 903: Would a `Base.Semaphore` be more idiomatic?
Conceptual it is for this, right?

It would be thrrad-safe also.
Which might introduce some overhead?
(Idk how hard it is to instantiate a `Semaphore` with a nonthreadsafe `Condition`, i think it can be done?)
But that overhead is probably tiny, and it would save on allocations.

ianshmean: Yeah, seems to be a better fit, although I actually see a slight increase in allocations
```
 80.416509 seconds (1.07 M allocations: 88.681 MiB, 0.04% gc time)
```
I guess using the right tooling is the best approach, and if unnecessary allocations are occurring they can be fixed in base.
I'll push the update, but let me know if you think the allocations increase is worth a revert

ianshmean:
Ping for review @fredrikekre @KristofferC @StefanKarpinski

src/API.jl
```diff
@@ -894,15 +894,33 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
+    
+    is_stdlib_from_name(name::String) = name in values(stdlibs())
```
KristofferC, line 898: This needs to go away. Check the uuid.

src/API.jl
```diff
@@ -894,15 +894,33 @@ end
...
+    
     printpkgstyle(ctx, :Precompiling, "project...")
+    
+    num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
```
KristofferC, line 902: Seems kinda excessive to introduce an env variable for this. Its so specific.

src/API.jl
KristofferC: What happens if this throws?

src/Types.jl
ianshmean: I think I'm being slow.. How do we tell the difference between a dep in the manifest entries that's a stdlib vs. a same-named package? The deps list is a list of names, not uuids
```
[[AbstractAlgebra]]
deps = ["InteractiveUtils", "LinearAlgebra", "Markdown", "Random", "SparseArrays", "Test"]
git-tree-sha1 = "8fa03ecf25341ff3e8fb301dba3f41c6fe09952e"
uuid = "c3fe647b-3220-5bb0-a1ea-a7954cac585d"
version = "0.10.0"
```

KristofferC (commented):
This is cool. I don't really know enough about what compilecache does to know if this is fully sound.

Also, before complicating this function further it might be best to thoroughly investigate https://github.com/JuliaLang/Pkg.jl/issues/1578 first.

src/Types.jl
staticfloat: Those names are required to be keys within the `Manifest`; so if someone else had a `deps = ["AbstractAlgebra"]` you'd look at the entry you see above you right there and pull out the `uuid`.  A `Manifest` can only have a single mapping for each name, so there's a direct correspondence.

That being said, I think this is a mostly academic distinction, as for stdlibs with packages that have the same name, I'm pretty sure we're going to ensure they have the same UUID as well, so that earlier versions of Julia can transparently deal with newer packages in an intelligent way.

fredrikekre: > A Manifest can only have a single mapping for each name, so there's a direct correspondence.

No, there can be multiple mappings. It is just in the common case where all names are unique that the deps are a list like that, otherwise deps are listed as `name = uuid`. E.g.
```
[[JSON]]
git-tree-sha1 = "81690084b6198a2e1da36fcfda16eeca9f9f24e4"
uuid = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
version = "0.21.1"

    [JSON.deps]
    Mmap = "a63ad114-7e13-5084-954f-fe012c677804"
    Parsers = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
    Unicode = "4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5"

[[Parsers]]
path = "Parsers"
uuid = "27091c14-f214-46fa-a0da-49c36ffec0b3"
version = "0.1.0"
[[Parsers]]
deps = ["Dates", "Test"]
git-tree-sha1 = "8077624b3c450b15c087944363606a6ba12f925e"
uuid = "69de0a69-1ddd-5017-9359-2bf0b02dc9f0"
version = "1.0.10"
```

src/API.jl
staticfloat: I'm not sure how else to gate this.  Suggestions?  There are some concerns that with a large core count this could accidentally OOM.

KristofferC: How much memory does each worker use approximately. Isn't this the case for every parallel workload?

src/Types.jl
ianshmean: Ok, I get why what I'm doing is wrong, but not sure how to fix it. The keys in the result of `Pkg.Types.read_manifest(ctx.env.manifest_file)` are the UUIDs. I don't see an efficient way to convert the deps list into a list of UUIDs.

ianshmean: I got it now.. Commit incoming

staticfloat: Ah, thanks for that Fredrik, I didn't know that, or forgot that I ever knew it, I guess.  :P

src/API.jl
staticfloat: > Isn't this the case for every parallel workload that uses memory?

Yes, but most other workloads allow tuning (e.g. via `-t`).

> I guess we should look at nthreads but everyone runs with that equal to 1.

There's that, and also Lyndon's comment above that this is more like a multiprocessing thing than a multithreading thing.  I also agree that I shouldn't have to limit my computation's thread count to limit the precompilation, and vice versa.

src/API.jl
ianshmean: Fixed. Didn't quite get how to do it initially

ianshmean:
@KristofferC Regarding #1578, I couldn't quite figure out why `Pkg.precompile` would've stopped re-compiling stale dependents.. I _think_ it might be that part of the `stale_cachefile` heuristic only works if the package is loaded in Main, which doesn't happen during the `Pkg.compilecache` process, but that's a bit of a guess.

However I did find a fix that I just added here.

Basically, I add an additional check for if any dep was recompiled in this session, and if so, recompile the package

Behavior before the last commit:
I touch the ColorTypes source, then:
```
(dev) pkg> st
Status `~/Documents/dev/Project.toml`
  [3da002f7] ColorTypes v0.10.9 `~/.julia/dev/ColorTypes`
  [916415d5] Images v0.22.4

julia> Pkg.precompile()
Precompiling project...
[ Info: Precompiling ColorTypes [3da002f7-5984-5a60-b8a6-cbb66c0b333f]

julia> using Images
[ Info: Precompiling Images [916415d5-f1e6-5110-898d-aaa5f9f070e0]

julia> 
```

Now:
```
julia> ENV["JULIA_DEBUG"] = "Base"

julia> Pkg.precompile()
Precompiling project...
┌ Debug: Rejecting stale cache file /Users/ian/.julia/compiled/v1.6/ColorTypes/db21U_vWD4f.ji (mtime 1.60004722644721e9) because file /Users/ian/.julia/dev/ColorTypes/src/ColorTypes.jl (mtime 1.600047296964872e9) has changed
└ @ Base loading.jl:1431
[ Info: Precompiling ColorTypes [3da002f7-5984-5a60-b8a6-cbb66c0b333f]
[ Info: Precompiling Colors [5ae59095-9a9b-59fe-a467-6f913c188581]
[ Info: Precompiling Graphics [a2bd30eb-e257-5431-a919-1863eab51364]
[ Info: Precompiling ColorVectorSpace [c3611d14-8923-5661-9e6a-0046d554d3a4]
[ Info: Precompiling ImageCore [a09fc81d-aa75-5fe9-8630-4744c3626534]
[ Info: Precompiling ImageDistances [51556ac3-7006-55f5-8cb3-34580c88182d]
[ Info: Precompiling ImageAxes [2803e5a7-5153-5ecf-9a86-9b4c37f5f5ac]
[ Info: Precompiling ImageShow [4e3cecfd-b093-5904-9786-8bbb286a6a31]
[ Info: Precompiling ImageMorphology [787d08f9-d448-5407-9aad-5290dd7ab264]
[ Info: Precompiling ImageTransformations [02fcd773-0e25-5acc-982a-7f6622650795]
[ Info: Precompiling ImageMetadata [bc367c6b-8a6b-528e-b4bd-a4b897500b49]
[ Info: Precompiling ImageContrastAdjustment [f332f351-ec65-5f6a-b3d1-319c6670881a]
[ Info: Precompiling ImageFiltering [6a3955dd-da59-5b1f-98d4-e7296123deb5]
[ Info: Precompiling ImageQualityIndexes [2996bd0c-7a13-11e9-2da2-2f5ce47296a9]
[ Info: Precompiling Images [916415d5-f1e6-5110-898d-aaa5f9f070e0]

julia> using Images

julia> 
```
Note that the precompillation of the following packages doesn't generate the debug info here, because they were forced to compile by the new catch.

cc. @timholy given https://github.com/JuliaLang/Pkg.jl/issues/1578

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
+    for pkgid in pkgids
+        precomp_events[pkgid.uuid] = Base.Event()
+        was_recompiled[pkgid.uuid] = false
     end
-
+    
+    precomp_tasks = Task[]
     # TODO: since we are a complete list, but not topologically sorted, handling of recursion will be completely at random
-    for pkg in pkgids
+    for (i, pkg) in pairs(pkgids)
```
tkf, line 921: Why not use `@sync` so that we can get rid of `precomp_tasks = Task[]`?

```suggestion
    @sync for (i, pkg) in pairs(pkgids)
```

tkf, line 921: Why not use `@sync` so that we can get rid of `precomp_tasks = Task[]`?

```suggestion
    @sync for (i, pkg) in pairs(pkgids)
```

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
+    
+    precomp_events = Dict{Base.UUID,Base.Event}()
+    was_recompiled = Dict{Base.UUID,Bool}()
+    for pkgid in pkgids
+        precomp_events[pkgid.uuid] = Base.Event()
+        was_recompiled[pkgid.uuid] = false
     end
-
+    
+    precomp_tasks = Task[]
```
tkf, line 919:
```suggestion
```

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
+                was_recompiled[pkg.uuid] = true
+                notify(precomp_events[pkg.uuid])
+                Base.release(parallel_limiter)
+            else
+                notify(precomp_events[pkg.uuid])
+            end
+        end  
+        push!(precomp_tasks, t) 
     end
+    wait.(precomp_tasks)
```
tkf, line 951:
```suggestion
    end
```

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
+            length(pkg_dep_uuid_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_uuid_lists[i]))
+            any_dep_recompiled = any(map(x->was_recompiled[x], pkg_dep_uuid_lists[i]))
+            stale = true
+            for path_to_try in paths::Vector{String}
+                staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache()) #|| any(deps_recompiled)
+                staledeps === true && continue
+                # TODO: else, this returns a list of packages that may be loaded to make this valid (the topological list)
+                stale = false
+                break
+            end
```
tkf, line 938: I feel it'd be better to have a function in `Base` that does this to reduce coupling to the implementation details in `Base` and `Pkg`.

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
+            if any_dep_recompiled || stale
+                Base.acquire(parallel_limiter)
+                Base.compilecache(pkg, sourcepath)
+                was_recompiled[pkg.uuid] = true
+                notify(precomp_events[pkg.uuid])
+                Base.release(parallel_limiter)
+            else
+                notify(precomp_events[pkg.uuid])
+            end
```
tkf, line 947: IIUC, the only relevant logic here is `Base.compilecache(pkg, sourcepath)` and everything else is for implementing `future-like system. My personal preference for doing something like this is to create a task pool and future separately from the core logic. Dealing semaphore and event at the application layer is way too low-level for me as I'd need to worry about deadlocks all the time. Anyway, it's just a comment. If Pkg devs are cool with it I guess there is no need to do something else.

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
-            # TODO: else, this returns a list of packages that may be loaded to make this valid (the topological list)
-            stale = false
-            break
-        end
-        if stale
-            Base.compilecache(pkg, sourcepath)
-        end
+        
+        t = @async begin
+            length(pkg_dep_uuid_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_uuid_lists[i]))
```
tkf, line 929: `wait.(...)` creates unnecessary `Vector{Nothing}`. You can do `foreach(wait, ...)` but at this point it's probably better to jsut write a `for` loop:

```suggestion
            for x in pkg_dep_uuid_lists[i]
                wait(precomp_events[x])
            end
```

tkf (commented):
I left some nitpicks but it LGTM! Not that I'm an expert on any of this, though. Anyway, thanks a lot for doing this!

tkf:
> why `Pkg.precompile` would've stopped re-compiling stale dependents

Indeed, old `Pkg.precompile` is not enough. Here is a repro that fools old `Pkg.precompile`: https://discourse.julialang.org/t/rfc-speeding-up-code-loading-when-using-multiple-processes-by-2x/37583/4

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
+            any_dep_recompiled = any(map(x->was_recompiled[x], pkg_dep_uuid_lists[i]))
+            stale = true
+            for path_to_try in paths::Vector{String}
+                staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache()) #|| any(deps_recompiled)
+                staledeps === true && continue
+                # TODO: else, this returns a list of packages that may be loaded to make this valid (the topological list)
```
tkf, line 935: IIUC, this TODO comment is not accurate anymore since by construction the dependencies are all in the non-stale state at this point.

cc @KristofferC

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
-            stale = false
-            break
-        end
-        if stale
-            Base.compilecache(pkg, sourcepath)
-        end
+        
+        t = @async begin
+            length(pkg_dep_uuid_lists[i]) > 0 && wait.(map(x->precomp_events[x], pkg_dep_uuid_lists[i]))
+            any_dep_recompiled = any(map(x->was_recompiled[x], pkg_dep_uuid_lists[i]))
```
tkf, line 930: I'm guessing you'd need `any_dep_recompiled` here since `stale_cachefile` treats packages that are already loaded as non-stale? If so, maybe adding some comments can be useful? (Though I think having a `stale_cachefile`-like function that ignores `Base.loaded_modules` would be much nicer.)

src/API.jl
KristofferC: I think Jameson added this comment.

src/API.jl
```diff
@@ -895,31 +895,60 @@ end
...
+                staledeps === true && continue
+                # TODO: else, this returns a list of packages that may be loaded to make this valid (the topological list)
+                stale = false
+                break
+            end
+            if any_dep_recompiled || stale
+                Base.acquire(parallel_limiter)
+                Base.compilecache(pkg, sourcepath)
+                was_recompiled[pkg.uuid] = true
+                notify(precomp_events[pkg.uuid])
```
KristofferC, line 943: This implementation of parallelism feels very low-level. `Semaphore`, `Event`, `notify`, `acquire`, `release` etc. Why not just start n tasks that take work from a channel?

src/API.jl
KristofferC: I'd rather have it as a normal argument to the `precompile` function then. This is exactly what we already have to limit parallelism in the asynchronous package downloader.

Looking at it, funnily enough we do have an env variable for the package downloader but that seems like it was added as a workaround for something:

https://github.com/JuliaLang/Pkg.jl/blob/ede7b07f7bbd8b275b5d80660c19a79b82330ab6/src/Types.jl#L329-L331

KristofferC: Although if we want to run this automatically when a package is updated, there is no chance to give this argument.

Perhaps there should be a `.julia/config/PkgConfig.toml` where things like this could be set?

ianshmean:
## Error handling
This wasn't done properly, so I fixed it and now if I throw an error into ZygoteRules (middle of the pack dep), the precompile session will terminate early, but any packages currently finishing precomp won't be forcibly terminated.
```
julia> @time Pkg.precompile()
Precompiling project...
...
[ Info: Precompiling QuadGK [1fd47b50-473d-5c70-9696-f719f8f3bcdc]
[ Info: Precompiling ZygoteRules [700de1a5-db45-46bc-99cf-38207098b444]
[ Info: Precompiling ForwardDiff [f6369f11-7733-5829-9624-2563aa707210]
ERROR: LoadError: dummy error
Stacktrace:
 [1] error(s::String)
   @ Base ./error.jl:33
 [2] top-level scope
   @ ~/.julia/dev/ZygoteRules/src/ZygoteRules.jl:3
 [3] include
   @ ./Base.jl:377 [inlined]
 [4] include_package_for_output(input::String, depot_path::Vector{String}, dl_load_path::Vector{String}, load_path::Vector{String}, concrete_deps::Vector{Pair{Base.PkgId,UInt64}}, uuid_tuple::Tuple{UInt64,UInt64}, source::Nothing)
   @ Base ./loading.jl:1109
 [5] top-level scope
   @ none:1
 [6] eval
   @ ./boot.jl:344 [inlined]
 [7] eval(x::Expr)
   @ Base.MainInclude ./client.jl:446
 [8] top-level scope
   @ none:1
in expression starting at /Users/ian/.julia/dev/ZygoteRules/src/ZygoteRules.jl:1
[ Info: Precompiling Latexify [23fbe1c1-3f47-55db-b15f-69d7ec21a316]
[ Info: Precompiling Bzip2_jll [6e34b625-4abd-537c-b88f-471c36dfa7a0]
[ Info: Precompiling SortingAlgorithms [a2af1166-a08f-5f64-846c-94a0d3cef48c]
[ Info: Precompiling SimpleTraits [699a6c99-e7fa-54fc-8d76-47d257e15c1d]
[ Info: Precompiling LabelledArrays [2ee39098-c373-598a-b85f-a56591580800]
ERROR: TaskFailedException

    nested task error: Failed to precompile ZygoteRules [700de1a5-db45-46bc-99cf-38207098b444] to /Users/ian/.julia/compiled/v1.6/ZygoteRules/LmjSI_vWD4f.ji.
    Stacktrace:
     [1] macro expansion
       @ ~/Documents/GitHub/Pkg.jl/src/API.jl:951 [inlined]
     [2] (::Pkg.API.var"#199#205"{Dict{Base.UUID,Bool},Dict{Base.UUID,Base.Event},Vector{Vector{Base.UUID}},Base.Semaphore,Base.PkgId,Int64,String,Vector{String}})()
       @ Pkg.API ./task.jl:389
    
    caused by: Failed to precompile ZygoteRules [700de1a5-db45-46bc-99cf-38207098b444] to /Users/ian/.julia/compiled/v1.6/ZygoteRules/LmjSI_vWD4f.ji.
    Stacktrace:
     [1] error(s::String)
       @ Base ./error.jl:33
     [2] compilecache(pkg::Base.PkgId, path::String)
       @ Base ./loading.jl:1240
     [3] macro expansion
       @ ~/Documents/GitHub/Pkg.jl/src/API.jl:947 [inlined]
     [4] (::Pkg.API.var"#199#205"{Dict{Base.UUID,Bool},Dict{Base.UUID,Base.Event},Vector{Vector{Base.UUID}},Base.Semaphore,Base.PkgId,Int64,String,Vector{String}})()
       @ Pkg.API ./task.jl:389
Stacktrace:
 [1] sync_end(c::Channel{Any})
   @ Base ./task.jl:347
 [2] macro expansion
   @ ./task.jl:366 [inlined]
 [3] precompile(ctx::Pkg.Types.Context)
   @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:920
 [4] precompile
   @ ~/Documents/GitHub/Pkg.jl/src/API.jl:895 [inlined]
 [5] top-level scope
   @ ./timing.jl:174 [inlined]
 [6] top-level scope
   @ ./REPL[3]:0
```

src/API.jl
```diff
@@ -915,7 +915,7 @@ function precompile(ctx::Context)
         precomp_events[pkgid.uuid] = Base.Event()
         was_recompiled[pkgid.uuid] = false
     end
-    # TODO: since we are a complete list, but not topologically sorted, handling of recursion will be completely at random
+
```
ianshmean, line 918: @KristofferC I looked at the git blame and it seems these todos were yours. I think they're both dealt with in this PR, so removed them

src/API.jl
ianshmean: Thanks. Now adapted

src/API.jl
baggepinnen: @staticfloat aren't you describing a (counting) `Semaphore`?

staticfloat: Yep, pretty much.

src/API.jl
ianshmean: Yeah, I don't understand quite why `stale_cachefile` isn't identifying packages that have recompiled deps as stale.. Judging by #1578 I'm not sure anyone does yet. I just changed the code to skip the `stale_cachefile` checking entirely if `any_dep_recompiled == true`. It would likely always be faster than using `stale_cachefile` but there is an edge case that would still warrant fixing `stale_cachefile` though; if a package that others depend on was recompiled externally from this session

src/API.jl
ianshmean: Missed this thread. I git blamed them and that they were added by you @KristofferC so I removed both

src/API.jl
ianshmean: I tried out a channel based approach, but it still requires the notify system. It works but seems slightly more complicated https://github.com/JuliaLang/Pkg.jl/compare/master...ianshmean:ib/parallel_precomp_chanelbased

src/API.jl
ianshmean: This PR now has a Semafore approach, and I tried out a channel-based approach here, which doesn't seem simpler https://github.com/JuliaLang/Pkg.jl/compare/master...ianshmean:ib/parallel_precomp_chanelbased

What should we move forward with?

ianshmean: cc @tkf just to bring the conversation to a single thread

src/API.jl
ianshmean: Turns out that edge case isn't an issue. More detail here https://github.com/JuliaLang/Pkg.jl/issues/1578#issuecomment-692360500

src/API.jl
KristofferC: Anyway, let's go with this for now. Can always tweak it later.

src/API.jl
KristofferC: They are not mine, they come from:

https://github.com/JuliaLang/julia/pull/28402/files#diff-f472857fede6d42cd50c2de4cdb414c3R465

but Pkg was developed in Base and in a separate repo at the same time so when I synced them, they show up as mine.

ianshmean: Ah ok

KristofferC (approved):
Can merge when ready. But I'm gonna need help with the future maintenance of this code.

ianshmean:
Just a small tweak to the error handling, but should be good to go now. Happy to help maintain once this starts getting used

src/API.jl
tkf: I'd actually implement `precompile` differently by using a recursive function that returns `was_recompiled::Bool` and wrapping the recursive calls with tasks. This way, we don't need to implement a future-like construct (i.e., `was_processed` + `was_recompiled`). The error handling would probably be more straightforward this way. Resource control is probably still easier with semaphore (unless we have an easy-to-use task pool and future in Base or stdlib) although I wish there were `Base.acquire(f, semaphore)`.

But this is the kind of thing the trade-off is not completely clear until you have a concrete implementation. So, I think it's reasonable to defer this to future refactoring.

staticfloat:
Thanks Ian!  This is great!

src/API.jl
tkf: Yeah, I noticed that this comment is committed by @KristofferC in #626

https://github.com/JuliaLang/Pkg.jl/pull/626/files#diff-47db2d66769667be0982a7b56891b4a8R483

But maybe @KristofferC and @vtjnash discussed it in slack or sth? It'd be nice if @vtjnash can chime in so that we are not removing something important.

src/API.jl
```diff
@@ -895,30 +895,72 @@ end
...
+                    notify(was_processed[pkg.uuid])
+                    Base.release(parallel_limiter)
+                    return
+                end
+                try
+                    Base.compilecache(pkg, sourcepath)
+                    was_recompiled[pkg.uuid] = true
+                catch err
+                    errored = true
+                    throw(err)
```
tkf, line 955: I couldn't bring https://github.com/JuliaLang/julia/pull/37374#issuecomment-689256691 up again in this PR in time, but did somebody considered the case for OS-specific packages? That is to say, if we have

```julia
module DownstreamPkg
if Sys.iswindows()
    using SomeWindowsOnlyPackage
end
...
end
```

we don't need to compile `SomeWindowsOnlyPackage` (or propagate the error from compiling `SomeWindowsOnlyPackage` in non-Windows OS) when compiling `DownstreamPkg` in non-Windows OS.

I think we should report errors only from the importable packages (i.e., the ones in Project.toml). Of course, it'd be much nicer to integrate precompilation machinery in Base itself so that this kind of hack is not necessary.

KristofferC: Is that what makes it hard to parallelize in Base, because you don't know your actual dependencies up front?

src/API.jl
vtjnash: I don't recall anything about this

src/API.jl
ianshmean: I see. Perhaps the docstring for `Pkg.precompile` could be changed to

```
help?> Pkg.precompile
  Pkg.precompile()

  Parallelized precompilation of all the dependencies in the manifest of the project. 
  If you want to only precompile the packages that are actually used, instead load the packages to trigger standard precompilation.
...
```

ianshmean: I'll let others decide if this is necessary, but it proved simple to implement an option to revert behavior https://github.com/JuliaLang/Pkg.jl/pull/2021

tkf: > Is that what makes it hard to parallelize in Base, because you don't know your actual dependencies up front?

Yeah.  But it's a problem in Pkg, too (even if you can parse Manifest).  I think the problem is that there is no database for "true" dependency tree since you can hide `using` in `if` branch.

Ultimately, only the precompilation subprocess "knows" the dependencies while it's precompiling a package. So, I think the "correct" way of doing this to use some kind of bidirectional RPC so that the precompilation subprocess can request the precompilation of its dependencies to the parent process. But that's very tedious. I think ignoring compilation error from non-direct dependencies is a decent approach.

(I feel like I should try implementing it now that I complained so much :rofl:)

fredrikekre:
Trying this out:
```
(@v1) pkg> precompile
Precompiling project...
[ Info: Precompiling MacroTools [1914dd2f-81c6-5fcd-8719-6d5c9610ff09]
[...]
[ Info: Precompiling PackageCompiler [9b87118b-4619-50d2-8e1e-99f35a4d4d9d]
ERROR: LoadError: LoadError: Cannot locate artifact 'x86_64-w64-mingw32' in '/home/fredrik/.julia/packages/PackageCompiler/vsMJE/Artifacts.toml'
Stacktrace:
```
Why did that happen? PackageCompiler that does something bad?

ianshmean:
Does the same thing happen if you just `using PackageCompiler`?

fredrikekre:
Yea, strange. Perhaps this is not connected to this but to @staticfloat move to artifacts stdlib ?

fredrikekre:
Looks like stdout/stderr is not supressed though:
```
pkg> precompile
Precompiling project...
[...]
[ Info: Precompiling Pluto [c3e4b0f8-55cb-11ea-2926-15256bba5781]
┌ Info: 
│ 
│     Welcome to Pluto v0.11.14 🎈
│     Start a notebook server using:
│ 
│   julia> Pluto.run()
│ 
│     Have a look at the FAQ:
│     https://github.com/fonsp/Pluto.jl/wiki
└
```

ianshmean:
The same thing happened in the old `Pkg.precompile` Pluto just puts a welcome message directly in the module rather than `__init__()`. I made the recommendation to remove it already https://github.com/fonsp/Pluto.jl/pull/389#issuecomment-692743741

src/API.jl
ianshmean: Neat idea on non-direct dependencies being allowed to fail.. I implemented it here https://github.com/JuliaLang/Pkg.jl/pull/2021#issuecomment-693095161

haampie:
I've tried this on a machine with 128 physical cores (AMD EPYC 7742 2.25GHz), 128 GB memory, with everything in-memory (so, julia itself and JULIA_DEPOTH_PATH too):

`] add LightGraphs, Optim, JuMP, PyCall, Images, Flux, Plots, Pluto, DifferentialEquations, DataFrames, Turing`

Before:

```julia
julia> @time Pkg.precompile()
442.760569 seconds (209.17 k allocations: 16.029 MiB)
```
After:
```julia
julia> @time Pkg.precompile()
61.723726 seconds (2.18 M allocations: 185.585 MiB, 0.06% gc time)
```

that's >7x speedup. Unfortunately not a 128x speedup :p but I guess parallelism is limited.

src/API.jl
```diff
@@ -895,30 +895,72 @@ end
...
 function precompile(ctx::Context)
     printpkgstyle(ctx, :Precompiling, "project...")
-
-    pkgids = [Base.PkgId(uuid, name) for (name, uuid) in ctx.env.project.deps if !is_stdlib(uuid)]
+    
+    num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
+    parallel_limiter = Base.Semaphore(num_tasks)
+    
+    man = Pkg.Types.read_manifest(ctx.env.manifest_file)
+    pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
```
fredrikekre, line 903: I know this filtering was here before, but why is it necessary to filter out stdlibs?
```
(tmp.9L2rmMdEXO) pkg> st
Status `/tmp/tmp.9L2rmMdEXO/Project.toml`
  [37e2e46d] LinearAlgebra

(tmp.9L2rmMdEXO) pkg> precompile
Precompiling project...

julia> using LinearAlgebra
[ Info: Precompiling LinearAlgebra [37e2e46d-f89d-539d-b4ee-838fcccc9c8e]
```

staticfloat:
@haampie can you give us a feel for the memory requirements?  Did the peak memory usage go up significantly?

src/API.jl
fredrikekre: Wouldn't it be enought to wrap in 
```
redirect_stderr(devnull)
```
here? That should propagate to the subprocess, no?

ianshmean: That's implemented in https://github.com/JuliaLang/Pkg.jl/pull/2021 but there was a question about `redirect_stderr` concurrency here https://github.com/JuliaLang/Pkg.jl/pull/2021#discussion_r489098311 so https://github.com/JuliaLang/julia/pull/37596 was opened

haampie:
I just checked it with `htop`. Initially it has 9.2G used, when I run Pkg.precompile it jumps to 16.7G, but not for long. Initially many processes spin up, I'm not sure if `htop` is showing the peak memory or an average.

giordano:
How many threads do you see busy?  The scheduler should use up to 257 parallel tasks on that machine, but I guess it won't use most of them

src/API.jl
ianshmean: Aren't stdlib's always going to be precompiled already, and if you're dev-ing them they'd need to have their uuid removed, so wouldn't identify as stdlibs in that check?

haampie:
Screenshot from 2020-09-17 02-14-57

6s later
Screenshot from 2020-09-17 02-15-03

35s later
Screenshot from 2020-09-17 02-15-32

ianshmean:
Seems like a pretty extreme test, and reassuring results. Thanks @haampie!

src/API.jl
fredrikekre: Right, forgot to say that this is when you compile Julia without them in the sysimg. Perhaps we can instead filter based on if the package is already loaded? That should work for both regular packages and stdlibs. If it is a stdlib that is in the sysimg it doesn't need to precompile, and if it is a regular package that is already loaded in the session it is probably just precompiled from the `using`?

ianshmean: But what if they're loaded, and in need of recompiling? Perhaps the filter just isn't needed at all?

fredrikekre: Yea I am not sure what happens if you try to precompile stdlibs that are loaded though? Since no precompiles files exist, will that spend time on precompiling them anyway? At least we can add the filter I suggested to the stdlibs.

tkf: You can update a package after importing it. In this case, I think `Pkg.precompile` should re-compile it and all its downstream packages. So, I don't think it's a good strategy to check if it's loaded or not. Maybe it's better to have an API `Base.in_sysimage(pkg)`?

ianshmean: That sounds good. Do you have a design of how that might work?

tkf: Maybe the easiest way to do it is to copy `keys(loaded_modules)` in `Base.__init__` to some constant?

fredrikekre: > You can update a package after importing it.

Yea but not really with stdlibs. If we for stdlibs check if they are loaded or not, then the only thing we potentially miss is if someone has (i) loaded the stdlib and (ii) changed version of it which is a really niche thing to do, so I think it is fine.

ianshmean: Before I saw your last message @fredrikekre I got @tkf's last suggestion working. Should I PR that?

fredrikekre: IMO that seems overly complicated for a corner case that will not ever happen really.

ianshmean: Ok, got the alternative working with this. I'll convert the relevant part of the code to work off `PkgId[]` rather than `UUID[]` to make this cleaner. But it works as a prototype
```
function is_stdlib_and_loaded(uuid::UUID)
    return Pkg.Operations.is_stdlib(uuid) && (uuid in map(x->key(x).uuid, module_keys))
end
```

ianshmean: https://github.com/JuliaLang/Pkg.jl/pull/2021 updated with this now (the PkgId version)

tkf: I thought the stdlib check was for some kind of optimization (launching `julia` takes a few hundreds of ms even if you don't do anything). So, I think the correct predicate here is "is it in sysimage?" than "is it a stdlib?" since non-stdlib packages can be in sysimage and there is no point in calling `compilecache` for them.  This also covers the exotic situation where some stdlibs are not in sysimage.

I think `is_stdlib_and_loaded` is better than nothing. But I feel it's a bit half-way solution if my assumption (the stdlib check was optimization) is correct.

fredrikekre: Thats true, I didn't think about regular packages in the sysimg. But perhaps https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r490555795 is a good enough approximation of that? It seems pretty strange to (i) load a dependency, (ii) update its version, (iii) pkg> precompile, (iv) restart Julia and expect everything to be precompiled?

ianshmean: Although #2021 is looking good, I do like the properness of `in_sysimage`. It explains exactly why we'd always want to skip. I'll prepare a PR

tkf: > It seems pretty strange to (i) load a dependency, (ii) update its version, (iii) pkg> precompile, (iv) restart Julia and expect everything to be precompiled?

@fredrikekre Hmm... That was my expectation, actually. I generally expect `pkg> $cmd` and `shell> jlpkg $cmd` to be identical (when a project is not activated). Anyway, what do you think about #2021 + https://github.com/JuliaLang/julia/pull/37652? I think `in_sysimage` is simple enough and nice to have.

PR #2021, opened 2020-09-15
Allow failures of indirect dependencies during parallel precompillation

ianshmean:
Edit: The mechanism below has now been removed from this PR, given a better approach of allowing failures for indirect dependencies

----

Builds on #2018 to add the option to revert to the old-style `Pkg.precompile()` behavior if the user just wants/needs to precompile dependencies that are loaded.

```
julia> Pkg.precompile(parallel=false)
Precompiling project...
[ Info: Precompiling DifferentialEquations [0c46a032-eb83-5123-abaf-570d42b7fbaa]
```

```
julia> Pkg.precompile()
Precompiling project...
[ Info: Precompiling DocStringExtensions [ffbed154-4ef7-542d-bbb7-c09d3a79fcae]
[ Info: Precompiling PoissonRandom [e409e4f3-bfea-5376-8464-e040bb5c01ab]
...
...
[ Info: Precompiling DifferentialEquations [0c46a032-eb83-5123-abaf-570d42b7fbaa]
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -904,11 +904,6 @@ function precompile(ctx::Context)
         return true
     end
     
-    # stdlibs that aren't loaded might need precompiling if not part of sysimage
-    function is_stdlib_and_loaded(pkg::Base.PkgId)
-        return Pkg.Operations.is_stdlib(pkg.uuid) && Base.root_module_exists(pkg)
-    end
-    
     printpkgstyle(ctx, :Precompiling, "project...")
     
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
@@ -916,22 +911,22 @@ function precompile(ctx::Context)
     
     direct_deps = [
         Base.PkgId(uuid, name) 
-        for (name, uuid) in ctx.env.project.deps if !is_stdlib_and_loaded(Base.PkgId(uuid, name))
+        for (name, uuid) in ctx.env.project.deps if !Base.in_sysimage(Base.PkgId(uuid, name))
     ]
     
     man = Pkg.Types.read_manifest(ctx.env.manifest_file)
     deps_pair_or_nothing = Iterators.map(man) do dep
         pkg = Base.PkgId(first(dep), last(dep).name)
-        is_stdlib_and_loaded(pkg) && return nothing
+        Base.in_sysimage(pkg) && return nothing
         deps = [Base.PkgId(last(x), first(x)) for x in last(dep).deps]
-        return pkg => filter!(!is_stdlib_and_loaded, deps)
+        return pkg => filter!(!Base.in_sysimage, deps)
     end
     depsmap = Dict(Iterators.filter(!isnothing, deps_pair_or_nothing)) #flat map of each dep and its deps
     
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         depsmap[Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name)] = [
             Base.PkgId(last(x), first(x)) 
-            for x in ctx.env.project.deps if !is_stdlib_and_loaded(Base.PkgId(last(x), first(x)))
+            for x in ctx.env.project.deps if !Base.in_sysimage(Base.PkgId(last(x), first(x)))
         ]
     end    
     
--- a/src/API.jl
+++ b/src/API.jl
@@ -896,8 +896,9 @@ precompile() = precompile(Context())
 function precompile(ctx::Context)
     
     function is_stale(paths, sourcepath)
+        toml_c = Base.TOMLCache()
         for path_to_try in paths::Vector{String}
-            staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache())
+            staledeps = Base.stale_cachefile(sourcepath, path_to_try, toml_c)
             staledeps ? continue : return false
         end
         return true
--- a/src/API.jl
+++ b/src/API.jl
@@ -937,9 +937,9 @@ function precompile(ctx::Context)
     
     was_processed = Dict{Base.PkgId,Base.Event}()
     was_recompiled = Dict{Base.PkgId,Bool}()
-    for dep in depsmap
-        was_processed[first(dep)] = Base.Event()
-        was_recompiled[first(dep)] = false
+    for pkgid in keys(depsmap)
+        was_processed[pkgid] = Base.Event()
+        was_recompiled[pkgid] = false
     end
     
     errored = false
--- a/src/API.jl
+++ b/src/API.jl
@@ -938,9 +938,10 @@ function precompile(ctx::Context)
     end
     
     errored = false
+    toml_c = Base.TOMLCache()
     @sync for (pkg, deps) in depsmap
         paths = Base.find_all_in_cache_path(pkg)
-        sourcepath = Base.locate_package(pkg)
+        sourcepath = Base.locate_package(pkg, toml_c)
         sourcepath === nothing && continue
         # Heuristic for when precompilation is disabled
         occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String)) && continue
--- a/src/API.jl
+++ b/src/API.jl
@@ -895,8 +895,7 @@ end
 precompile() = precompile(Context())
 function precompile(ctx::Context)
     
-    function is_stale(paths, sourcepath)
-        toml_c = Base.TOMLCache()
+    function is_stale(paths, sourcepath, toml_c)
         for path_to_try in paths::Vector{String}
             staledeps = Base.stale_cachefile(sourcepath, path_to_try, toml_c)
             staledeps === true ? continue : return false
@@ -937,9 +936,12 @@ function precompile(ctx::Context)
         was_recompiled[pkgid] = false
     end
     
+    last_dep_preprocessed = Base.Event()
     errored = false
     toml_c = Base.TOMLCache()
+    i = 0
     @sync for (pkg, deps) in depsmap
+        i += 1
         paths = Base.find_all_in_cache_path(pkg)
         sourcepath = Base.locate_package(pkg, toml_c)
         sourcepath === nothing && continue
@@ -949,14 +951,18 @@ function precompile(ctx::Context)
             continue
         end
         
+        # pre-process all first to make tomlcache read-only during async
+        i == length(depsmap) && notify(last_dep_preprocessed) 
+        
         @async begin
+            wait(last_dep_preprocessed) # wait for tomlcache to be ready
             for dep in deps # wait for deps to finish
                 wait(was_processed[dep])
             end
             
             # skip stale checking and force compilation if any dep was recompiled in this session
             any_dep_recompiled = any(map(dep->was_recompiled[dep], deps))
-            if !errored && (any_dep_recompiled || is_stale(paths, sourcepath))
+            if !errored && (any_dep_recompiled || is_stale(paths, sourcepath, toml_c))
                 Base.acquire(parallel_limiter)
                 if errored # catch things queued before error occurred
                     notify(was_processed[pkg])
@@ -981,7 +987,7 @@ function precompile(ctx::Context)
             else
                 notify(was_processed[pkg])
             end
-        end  
+        end
     end
     nothing
 end
--- a/src/API.jl
+++ b/src/API.jl
@@ -944,7 +944,10 @@ function precompile(ctx::Context)
         sourcepath = Base.locate_package(pkg, toml_c)
         sourcepath === nothing && continue
         # Heuristic for when precompilation is disabled
-        occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String)) && continue
+        if occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String))
+            notify(was_processed[pkg])
+            continue
+        end
         
         @async begin
             for dep in deps # wait for deps to finish
--- a/src/API.jl
+++ b/src/API.jl
@@ -892,21 +892,26 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
     Operations.build(ctx, pkgs, verbose)
 end
 
-precompile() = precompile(Context())
-function precompile(ctx::Context)
+precompile(;kwargs...) = precompile(Context(); kwargs...)
+function precompile(ctx::Context; parallel=true)
     printpkgstyle(ctx, :Precompiling, "project...")
     
-    num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
+    num_tasks = parallel ? parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1))) : 1
     parallel_limiter = Base.Semaphore(num_tasks)
     
-    man = Pkg.Types.read_manifest(ctx.env.manifest_file)
-    pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    pkg_dep_uuid_lists = [collect(values(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    filter!.(!is_stdlib, pkg_dep_uuid_lists)
+    if parallel
+        man = Pkg.Types.read_manifest(ctx.env.manifest_file)
+        pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
+        pkg_dep_uuid_lists = [collect(values(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
+        filter!.(!is_stdlib, pkg_dep_uuid_lists)
+    else
+        pkgids = [Base.PkgId(uuid, name) for (name, uuid) in ctx.env.project.deps if !is_stdlib(uuid)]
+        pkg_dep_uuid_lists = fill(Base.UUID[], length(pkgids)) # give empty arrays because we're not waiting for deps
+    end
     
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
-        push!(pkg_dep_uuid_lists, collect(keys(ctx.env.project.deps)))
+        push!(pkg_dep_uuid_lists, parallel ? collect(keys(ctx.env.project.deps)) : Base.UUID[])
     end    
     
     was_processed = Dict{Base.UUID,Base.Event}()
--- a/src/API.jl
+++ b/src/API.jl
@@ -957,7 +957,7 @@ function precompile(ctx::Context)
                         errored = true
                         throw(err) 
                     else
-                        @warn "Precompilation failed for sub-dependency $pkg"
+                        @warn "Precompilation failed for indirect dependency $pkg"
                     end
                 finally
                     notify(was_processed[pkg.uuid])
--- a/src/API.jl
+++ b/src/API.jl
@@ -894,29 +894,6 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
-    printpkgstyle(ctx, :Precompiling, "project...")
-    
-    num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
-    parallel_limiter = Base.Semaphore(num_tasks)
-    
-    toplevel_pkgids = [Base.PkgId(uuid, name) for (name, uuid) in ctx.env.project.deps if !is_stdlib(uuid)]
-    
-    man = Pkg.Types.read_manifest(ctx.env.manifest_file)
-    pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    pkg_dep_uuid_lists = [collect(values(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    filter!.(!is_stdlib, pkg_dep_uuid_lists)
-    
-    if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
-        push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
-        push!(pkg_dep_uuid_lists, collect(keys(ctx.env.project.deps)))
-    end    
-    
-    was_processed = Dict{Base.UUID,Base.Event}()
-    was_recompiled = Dict{Base.UUID,Bool}()
-    for pkgid in pkgids
-        was_processed[pkgid.uuid] = Base.Event()
-        was_recompiled[pkgid.uuid] = false
-    end
     
     function is_stale(paths, sourcepath)
         for path_to_try in paths::Vector{String}
@@ -925,6 +902,34 @@ function precompile(ctx::Context)
             return false
         end
         return true
+    end
+    
+    # stdlibs that aren't loaded might need precompiling if not part of sysimage
+    function is_stdlib_and_loaded(pkg::Base.PkgId)
+        return Pkg.Operations.is_stdlib(pkg.uuid) && Base.root_module_exists(pkg)
+    end
+    
+    printpkgstyle(ctx, :Precompiling, "project...")
+    
+    num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
+    parallel_limiter = Base.Semaphore(num_tasks)
+    
+    toplevel_pkgids = [Base.PkgId(uuid, name) for (name, uuid) in ctx.env.project.deps if !is_stdlib_and_loaded(Base.PkgId(uuid, name))]
+    
+    man = Pkg.Types.read_manifest(ctx.env.manifest_file)
+    pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !is_stdlib_and_loaded(Base.PkgId(first(dep), last(dep).name))]
+    pkg_dep_lists = [[Base.PkgId(last(x), first(x)) for x in last(dep).deps if !is_stdlib_and_loaded(Base.PkgId(last(x), first(x)))] for dep in man if !is_stdlib_and_loaded(Base.PkgId(first(dep), last(dep).name))]
+    
+    if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
+        push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
+        push!(pkg_dep_lists, [Base.PkgId(last(x), first(x)) for x in ctx.env.project.deps if !is_stdlib_and_loaded(Base.PkgId(last(x), first(x)))])
+    end    
+    
+    was_processed = Dict{Base.PkgId,Base.Event}()
+    was_recompiled = Dict{Base.PkgId,Bool}()
+    for pkgid in pkgids
+        was_processed[pkgid] = Base.Event()
+        was_recompiled[pkgid] = false
     end
     
     errored = false
@@ -936,21 +941,21 @@ function precompile(ctx::Context)
         occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String)) && continue
         
         @async begin
-            for dep_uuid in pkg_dep_uuid_lists[i] # wait for deps to finish
-                wait(was_processed[dep_uuid])
+            for dep in pkg_dep_lists[i] # wait for deps to finish
+                wait(was_processed[dep])
             end
             
             # skip stale checking and force compilation if any dep was recompiled in this session
-            any_dep_recompiled = any(map(dep_uuid->was_recompiled[dep_uuid], pkg_dep_uuid_lists[i]))
+            any_dep_recompiled = any(map(dep->was_recompiled[dep], pkg_dep_lists[i]))
             if !errored && (any_dep_recompiled || is_stale(paths, sourcepath))
                 Base.acquire(parallel_limiter)
                 if errored # catch things queued before error occurred
-                    notify(was_processed[pkg.uuid])
+                    notify(was_processed[pkg])
                     Base.release(parallel_limiter)
                     return
                 end
                 try
-                    was_recompiled[pkg.uuid] = true
+                    was_recompiled[pkg] = true
                     Base.compilecache(pkg, sourcepath, false) # don't print errors given we control
                 catch err
                     if pkg in toplevel_pkgids # only throw errors for top-level
@@ -960,11 +965,11 @@ function precompile(ctx::Context)
                         @warn "Precompilation failed for indirect dependency $pkg"
                     end
                 finally
-                    notify(was_processed[pkg.uuid])
+                    notify(was_processed[pkg])
                     Base.release(parallel_limiter)
                 end
             else
-                notify(was_processed[pkg.uuid])
+                notify(was_processed[pkg])
             end
         end  
     end
--- a/src/API.jl
+++ b/src/API.jl
@@ -959,11 +959,12 @@ function precompile(ctx::Context)
                     Base.release(parallel_limiter)
                     return
                 end
+                is_direct_dep =  pkg in direct_deps
                 try
                     was_recompiled[pkg] = true
-                    Base.compilecache(pkg, sourcepath, false) # don't print errors given we control
+                    Base.compilecache(pkg, sourcepath, is_direct_dep) # don't print errors from indirect deps
                 catch err
-                    if dep in direct_deps # only throw errors for direct dependencies (in Project)
+                    if is_direct_dep # only throw errors for direct dependencies (in Project)
                         errored = true
                         throw(err) 
                     else
--- a/src/API.jl
+++ b/src/API.jl
@@ -893,7 +893,7 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
 end
 
 function _is_stale(paths::Vector{String}, sourcepath::String, toml_c::Base.TOMLCache)
-    for path_to_try in paths::Vector{String}
+    for path_to_try in paths
         staledeps = Base.stale_cachefile(sourcepath, path_to_try, toml_c)
         staledeps === true ? continue : return false
     end
--- a/src/API.jl
+++ b/src/API.jl
@@ -899,13 +899,15 @@ function precompile(ctx::Context; parallel=true)
     num_tasks = parallel ? parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1))) : 1
     parallel_limiter = Base.Semaphore(num_tasks)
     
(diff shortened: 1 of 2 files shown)


ianshmean:
As a means to not block failures by packages that weren't going to be loaded anyway, sub-dependenncy compilation failures only warn now (main deps still fail).

Idea from @tkf here https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r489029683

Note that in this case, ColorTypes is a non-deved package that I hacked an error into

```
(dev) pkg> st
Status `~/Documents/GitHub/dev/Project.toml`
  [916415d5] Images v0.22.4

(dev) pkg> precompile
Precompiling project...
[ Info: Precompiling FileIO [5789e2e9-d7fb-5bc7-8068-2c6fae9b9549]
[ Info: Precompiling UnPack [3a884ed6-31ef-47d7-9d2a-63182c4928ed]
[ Info: Precompiling OffsetArrays [6fe1bfb0-de20-5000-8ca7-80f57d26f881]
[ Info: Precompiling IndirectArrays [9b13fd28-a010-5f03-acff-a1bbcff69959]
[ Info: Precompiling Compat [34da2185-b29b-5c13-b0c7-acf172513d20]
[ Info: Precompiling DataAPI [9a962f9c-6df0-11e9-0e5d-c546b8b5ee8a]
[ Info: Precompiling WoodburyMatrices [efce3f68-66dc-5838-9240-27a6d6f5f9b6]
[ Info: Precompiling MacroTools [1914dd2f-81c6-5fcd-8719-6d5c9610ff09]
[ Info: Precompiling IntelOpenMP_jll [1d5cc7b8-4909-519e-a0f8-d0f5ad9712d0]
[ Info: Precompiling RangeArrays [b3c3ace0-ae52-54e7-9d0b-2c1406fd6b9d]
[ Info: Precompiling StaticArrays [90137ffa-7385-5640-81b9-e52037218182]
[ Info: Precompiling AbstractFFTs [621f4979-c628-5d54-868e-fcf4e3e8185c]
[ Info: Precompiling FFTW_jll [f5851436-0d7a-5f13-b9de-f02708fd171a]
[ Info: Precompiling ComputationalResources [ed09eef8-17a6-5b46-8889-db040fac31e3]
[ Info: Precompiling Distances [b4f34e82-e78d-54a5-968a-f98e89d6e8f7]
[ Info: Precompiling IdentityRanges [bbac6d45-d8f3-5730-bfe4-7a449cd117ca]
[ Info: Precompiling AxisAlgorithms [13072b0f-2c55-5437-9ae7-d433b7a33950]
[ Info: Precompiling Missings [e1d29d7a-bbdc-5cf2-9ac0-f12de2c33e28]
[ Info: Precompiling EllipsisNotation [da5c29d0-fa7d-589e-88eb-ea29b0a81949]
[ Info: Precompiling Requires [ae029012-a4dd-5104-9daa-d747884805df]
[ Info: Precompiling MappedArrays [dbb5928d-eab1-5f90-85c2-b9b0edb7c900]
[ Info: Precompiling CompilerSupportLibraries_jll [e66e0078-7015-5450-92f7-15fbd957f2ae]
[ Info: Precompiling NaNMath [77ba4419-2d1f-58cd-9bb1-8ffee604a2e3]
[ Info: Precompiling MKL_jll [856f044c-d86e-5d09-b602-aeab76dc8ba7]
[ Info: Precompiling OrderedCollections [bac558e1-5e72-5ebc-8fee-abe8a469f55d]
[ Info: Precompiling TiledIteration [06e1c1a7-607b-532d-9fad-de7d9aa2abac]
[ Info: Precompiling IntervalSets [8197267c-284f-5f27-9208-e0e47529a953]
[ Info: Precompiling Ratios [c84ed2f1-dad5-54f0-aa8e-dbefe2724439]
[ Info: Precompiling CustomUnitRanges [dc8bdbbb-1ca9-579f-8c36-e416f6a65cce]
[ Info: Precompiling IterTools [c8e1da08-722c-5040-9ed9-7db0dc04731e]
[ Info: Precompiling FixedPointNumbers [53c48c17-4a7d-5ca2-90c5-79b7896eea93]
[ Info: Precompiling PaddedViews [5432bcbf-9aad-5242-b902-cca2824c8663]
[ Info: Precompiling Parameters [d96e819e-fc66-5662-9728-84c9c7592b0a]
[ Info: Precompiling DataStructures [864edb3b-99cc-5e75-8d2d-829cb0a9cfe8]
[ Info: Precompiling CatIndices [aafaddc9-749c-510e-ac4f-586e18779b91]
[ Info: Precompiling OpenSpecFun_jll [efe28fd5-8261-553b-a9e1-b2916fc3738e]
[ Info: Precompiling MosaicViews [e94cdb99-869f-56ef-bcf0-1ae2bcbe0389]
[ Info: Precompiling FFTW [7a1cc6ca-52ef-59f5-83cd-3a7055c09341]
[ Info: Precompiling SimpleTraits [699a6c99-e7fa-54fc-8d76-47d257e15c1d]
[ Info: Precompiling AxisArrays [39de3d68-74b9-583c-8d2d-e117c070f3a9]
[ Info: Precompiling SpecialFunctions [276daf66-3868-5448-9aa4-cd146d93841b]
[ Info: Precompiling ColorTypes [3da002f7-5984-5a60-b8a6-cbb66c0b333f]
[ Info: Precompiling FFTViews [4f61f5a4-77b1-5117-aa51-3ab5ef4ef0cd]
[ Info: Precompiling SortingAlgorithms [a2af1166-a08f-5f64-846c-94a0d3cef48c]
[ Info: Precompiling CoordinateTransformations [150eb455-5306-5404-9cee-2592286d6298]
[ Info: Precompiling Rotations [6038ab10-8711-5258-84ad-4b1120ba62dc]
[ Info: Precompiling Interpolations [a98d9a8b-a2ab-59e6-89dd-64a1c18fca59]
┌ Warning: Precompilation failed for sub-dependency ColorTypes [3da002f7-5984-5a60-b8a6-cbb66c0b333f]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling Colors [5ae59095-9a9b-59fe-a467-6f913c188581]
[ Info: Precompiling StatsBase [2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91]
┌ Warning: Precompilation failed for sub-dependency Colors [5ae59095-9a9b-59fe-a467-6f913c188581]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling Graphics [a2bd30eb-e257-5431-a919-1863eab51364]
[ Info: Precompiling ColorVectorSpace [c3611d14-8923-5661-9e6a-0046d554d3a4]
┌ Warning: Precompilation failed for sub-dependency Graphics [a2bd30eb-e257-5431-a919-1863eab51364]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling ImageCore [a09fc81d-aa75-5fe9-8630-4744c3626534]
┌ Warning: Precompilation failed for sub-dependency ColorVectorSpace [c3611d14-8923-5661-9e6a-0046d554d3a4]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
┌ Warning: Precompilation failed for sub-dependency ImageCore [a09fc81d-aa75-5fe9-8630-4744c3626534]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling ImageAxes [2803e5a7-5153-5ecf-9a86-9b4c37f5f5ac]
[ Info: Precompiling ImageDistances [51556ac3-7006-55f5-8cb3-34580c88182d]
[ Info: Precompiling ImageMorphology [787d08f9-d448-5407-9aad-5290dd7ab264]
[ Info: Precompiling ImageTransformations [02fcd773-0e25-5acc-982a-7f6622650795]
[ Info: Precompiling ImageShow [4e3cecfd-b093-5904-9786-8bbb286a6a31]
┌ Warning: Precompilation failed for sub-dependency ImageMorphology [787d08f9-d448-5407-9aad-5290dd7ab264]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
┌ Warning: Precompilation failed for sub-dependency ImageDistances [51556ac3-7006-55f5-8cb3-34580c88182d]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
┌ Warning: Precompilation failed for sub-dependency ImageTransformations [02fcd773-0e25-5acc-982a-7f6622650795]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling ImageContrastAdjustment [f332f351-ec65-5f6a-b3d1-319c6670881a]
┌ Warning: Precompilation failed for sub-dependency ImageAxes [2803e5a7-5153-5ecf-9a86-9b4c37f5f5ac]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling ImageMetadata [bc367c6b-8a6b-528e-b4bd-a4b897500b49]
┌ Warning: Precompilation failed for sub-dependency ImageShow [4e3cecfd-b093-5904-9786-8bbb286a6a31]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
┌ Warning: Precompilation failed for sub-dependency ImageContrastAdjustment [f332f351-ec65-5f6a-b3d1-319c6670881a]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
┌ Warning: Precompilation failed for sub-dependency ImageMetadata [bc367c6b-8a6b-528e-b4bd-a4b897500b49]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling ImageFiltering [6a3955dd-da59-5b1f-98d4-e7296123deb5]
┌ Warning: Precompilation failed for sub-dependency ImageFiltering [6a3955dd-da59-5b1f-98d4-e7296123deb5]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling ImageQualityIndexes [2996bd0c-7a13-11e9-2da2-2f5ce47296a9]
┌ Warning: Precompilation failed for sub-dependency ImageQualityIndexes [2996bd0c-7a13-11e9-2da2-2f5ce47296a9]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:967
[ Info: Precompiling Images [916415d5-f1e6-5110-898d-aaa5f9f070e0]
ERROR: TaskFailedException

    nested task error: Failed to precompile Images [916415d5-f1e6-5110-898d-aaa5f9f070e0] to /Users/ian/.julia/compiled/v1.6/Images/H8Vxc_vWD4f.ji.
    Stacktrace:
     [1] macro expansion
       @ ~/Documents/GitHub/Pkg.jl/src/API.jl:965 [inlined]
     [2] (::Pkg.API.var"#203#213"{Pkg.API.var"#is_stale#212", Dict{Base.UUID, Bool}, Dict{Base.UUID, Base.Event}, Vector{Base.PkgId}, Base.Semaphore, Base.PkgId, Int64, String, Vector{String}})()
       @ Pkg.API ./task.jl:392
    
    caused by: Failed to precompile Images [916415d5-f1e6-5110-898d-aaa5f9f070e0] to /Users/ian/.julia/compiled/v1.6/Images/H8Vxc_vWD4f.ji.
    Stacktrace:
     [1] error(s::String)
       @ Base ./error.jl:33
     [2] compilecache(pkg::Base.PkgId, path::String)
       @ Base ./loading.jl:1262
     [3] #205
       @ ~/Documents/GitHub/Pkg.jl/src/API.jl:960 [inlined]
     [4] redirect_stderr(f::Pkg.API.var"#205#215"{Base.PkgId, String}, stream::Base.DevNull)
       @ Base ./stream.jl:1224
     [5] macro expansion
       @ ~/Documents/GitHub/Pkg.jl/src/API.jl:959 [inlined]
     [6] (::Pkg.API.var"#203#213"{Pkg.API.var"#is_stale#212", Dict{Base.UUID, Bool}, Dict{Base.UUID, Base.Event}, Vector{Base.PkgId}, Base.Semaphore, Base.PkgId, Int64, String, Vector{String}})()
       @ Pkg.API ./task.jl:392
Stacktrace:
  [1] sync_end(c::Channel{Any})
    @ Base ./task.jl:350
  [2] macro expansion
    @ ./task.jl:369 [inlined]
  [3] precompile(ctx::Pkg.Types.Context; parallel::Bool)
    @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:936
  [4] precompile
    @ ~/Documents/GitHub/Pkg.jl/src/API.jl:897 [inlined]
  [5] #precompile#195
    @ ~/Documents/GitHub/Pkg.jl/src/API.jl:895 [inlined]
  [6] precompile()
    @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:895
  [7] do_cmd!(command::Pkg.REPLMode.Command, repl::REPL.LineEditREPL)
    @ Pkg.REPLMode ~/Documents/GitHub/Pkg.jl/src/REPLMode/REPLMode.jl:400
  [8] do_cmd(repl::REPL.LineEditREPL, input::String; do_rethrow::Bool)
    @ Pkg.REPLMode ~/Documents/GitHub/Pkg.jl/src/REPLMode/REPLMode.jl:381
  [9] do_cmd
    @ ~/Documents/GitHub/Pkg.jl/src/REPLMode/REPLMode.jl:376 [inlined]
 [10] (::Pkg.REPLMode.var"#24#27"{REPL.LineEditREPL, REPL.LineEdit.Prompt})(s::REPL.LineEdit.MIState, buf::IOBuffer, ok::Bool)
    @ Pkg.REPLMode ~/Documents/GitHub/Pkg.jl/src/REPLMode/REPLMode.jl:545
 [11] #invokelatest#2
    @ ./essentials.jl:707 [inlined]
 [12] invokelatest
    @ ./essentials.jl:706 [inlined]
 [13] run_interface(terminal::REPL.Terminals.TextTerminal, m::REPL.LineEdit.ModalInterface, s::REPL.LineEdit.MIState)
    @ REPL.LineEdit /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/LineEdit.jl:2434
 [14] run_frontend(repl::REPL.LineEditREPL, backend::REPL.REPLBackendRef)
    @ REPL /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/REPL/src/REPL.jl:1124
 [15] (::REPL.var"#44#49"{REPL.LineEditREPL, REPL.REPLBackendRef})()
    @ REPL ./task.jl:392
```

src/API.jl
```diff
@@ -948,11 +955,17 @@ function precompile(ctx::Context)
                     return
                 end
                 try
-                    Base.compilecache(pkg, sourcepath)
                     was_recompiled[pkg.uuid] = true
+                    redirect_stderr(Base.devnull) do
```
tkf, line 959: Is `redirect_stderr` "concurrent-safe"?  I don't think `redirect_stderr` is safe to use with `@async` since it has to modify the global state including, IIUC, C-level file descriptor. (Unless it does magic that I don't know.)

Is this to make `Pkg.precompile` work well with <https://github.com/JuliaLang/julia/pull/37374>? If so, I think it'd be better to change the API of `compilecache` to something like `compilecache(pkg::PkgId, path::String, show_log_deps::Bool = true)`.

src/API.jl
```diff
@@ -948,11 +955,17 @@ function precompile(ctx::Context)
...
+                        Base.compilecache(pkg, sourcepath)
+                    end
                 catch err
-                    errored = true
-                    throw(err)
+                    if pkg in toplevel_pkgids # only throw errors for top-level
+                        errored = true
+                        throw(err) 
+                    else
+                        @warn "Precompilation failed for sub-dependency $pkg"
```
tkf, line 967: Thanks for the patch! Yes, it does what I was thinking.

src/API.jl
ianshmean: No I didn't do this for compatibility with that PR, I did it because the try catch wasn't catching the error message from the julia sub process within compilecache, so even though the error wasn't throwing, stderr was making the log messy. 

Seems like the suggested change to compilecache would be a good idea. Should I PR it, or you?

tkf: It'd be great if you can go ahead and do the PR. I don't have the time to do it right now.

ianshmean: Just opened that PR https://github.com/JuliaLang/julia/pull/37596
This PR now uses that method arg to redirect stderr to devnull directly in the core pipeline

ianshmean:
Requires https://github.com/JuliaLang/julia/pull/37596

tkf:
Quoting the time-complexity consideration I commented in https://github.com/JuliaLang/julia/pull/37596#issuecomment-693587853

> It means that `Pkg.precompile` has quadratic time-complexity when an error happens. I think it's reasonable to optimize for the happy path for now given that doing this "correctly" is presumably much more challenging https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r489029683. Also, we can still use `using TheProblematicPackage` to get precompilation error in linear time.

ianshmean:
Do we think this and the julia PR is good as it stands?

src/API.jl
```diff
@@ -892,21 +892,28 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
...
+    toplevel_pkgids = [Base.PkgId(uuid, name) for (name, uuid) in ctx.env.project.deps if !is_stdlib(uuid)]
+    
+    if parallel
+        man = Pkg.Types.read_manifest(ctx.env.manifest_file)
+        pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
+        pkg_dep_uuid_lists = [collect(values(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
+        filter!.(!is_stdlib, pkg_dep_uuid_lists)
+    else
+        pkgids = toplevel_pkgids
+        pkg_dep_uuid_lists = fill(Base.UUID[], length(pkgids)) # give empty arrays because we're not waiting for deps
```
tkf, line 911: Wouldn't this trick break when a top-level package depends on other top-level packages?

If we are to do this, I think an easier option is to bring back the old `precompile` as a separate function and call it when `parallel == false`.  If we reuse the `is_stale` function, I guess it's not a very big function?

ianshmean: Ah, yes, you're right. I also wondered whether the `parallel=false` functionality was even needed now, with allowing failures for sub-depepdencies (is that a good name for them??)?
The user can always just precompile the classical way via code-load

tkf: > whether the `parallel=false` functionality was even needed now

This is exactly what I'm just wondering now, too.  We have `JULIA_NUM_PRECOMPILE_TASKS` to control the number of tasks so I think it'd be better to not implement this right now.

ianshmean: Ok, I'll remove it. It can always be added back if needed

src/API.jl
```diff
@@ -894,35 +894,42 @@ end
...
     parallel_limiter = Base.Semaphore(num_tasks)
     
+    toplevel_pkgids = [Base.PkgId(uuid, name) for (name, uuid) in ctx.env.project.deps if !is_stdlib_and_loaded(Base.PkgId(uuid, name))]
+    
     man = Pkg.Types.read_manifest(ctx.env.manifest_file)
-    pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    pkg_dep_uuid_lists = [collect(values(last(dep).deps)) for dep in man if !Pkg.Operations.is_stdlib(first(dep))]
-    filter!.(!is_stdlib, pkg_dep_uuid_lists)
+    pkgids = [Base.PkgId(first(dep), last(dep).name) for dep in man if !is_stdlib_and_loaded(Base.PkgId(first(dep), last(dep).name))]
+    pkg_dep_lists = [[Base.PkgId(last(x), first(x)) for x in last(dep).deps if !is_stdlib_and_loaded(Base.PkgId(last(x), first(x)))] for dep in man if !is_stdlib_and_loaded(Base.PkgId(first(dep), last(dep).name))]
```
ianshmean, line 921: This has got a bit too long.. is there a more concise way to generate a vector of vector of deps of deps as PkgIds?

tkf: Please use 92 width limit. FYI, JuliaFormatter.jl would format this to

```julia
    pkg_dep_lists = [
        [
            Base.PkgId(last(x), first(x))
            for
            x in last(dep).deps if !is_stdlib_and_loaded(Base.PkgId(last(x), first(x)))
        ]
        for dep in man if !is_stdlib_and_loaded(Base.PkgId(first(dep), last(dep).name))
    ]
```

which is reasonable, I think.

Also, I just noticed that it's crucial that the orders/indices of `pkgids` and `pkg_dep_lists` are consistent.  I think it'd be better to merge them in a single data structure like a vector-of-pairs or a dictionary (so that it's impossible to break such "consistency" by construction):

```julia
    depsmap = Dict(
        Base.PkgId(first(dep), last(dep).name) => [
            Base.PkgId(last(x), first(x))
            for
            x in last(dep).deps if !is_stdlib_and_loaded(Base.PkgId(last(x), first(x)))
        ]
        for dep in man if !is_stdlib_and_loaded(Base.PkgId(first(dep), last(dep).name))
    )
```

To me it's a bit hard to read because of repeated `Base.PkgId(...)`. I'd probably go with something like

```julia
    deps_pair_or_nothing = Iterators.map(man) do dep
        pkg = Base.PkgId(first(dep), last(dep).name)
        is_stdlib_and_loaded(pkg) && return nothing
        deps = (Base.PkgId(last(x), first(x)) for x in last(dep).deps)
        return pkg => filter(!is_stdlib_and_loaded, deps)
    end
    depsmap = Dict(Iterators.filter(!isnothing, deps_pair_or_nothing))
```

src/API.jl
```diff
@@ -894,35 +894,42 @@ end
...
+            staledeps === true && continue
+            return false
+        end
+        return true
+    end
+    
+    # stdlibs that aren't loaded might need precompiling if not part of sysimage
+    function is_stdlib_and_loaded(pkg::Base.PkgId)
+        return Pkg.Operations.is_stdlib(pkg.uuid) && Base.root_module_exists(pkg)
+    end
```
tkf, line 910: ref https://github.com/JuliaLang/Pkg.jl/pull/2018#discussion_r489705760

src/API.jl
ianshmean: Thanks. That last form hits this, which I'm not sure how to handle
```
ERROR: MethodError: no method matching filter(::Base.var"#70#71"{Pkg.API.var"#is_stdlib_and_loaded#210"}, ::Base.Generator{Dict{String, Base.UUID}, Pkg.API.var"#204#220"})
Closest candidates are:
  filter(::Any, ::Tuple{Any, Any, Any, Any, Any, Any, Any, Any, Any, Any, Any, Any, Any, Any, Any, Any, Vararg{Any, N}} where N) at tuple.jl:311
  filter(::Any, ::Tuple) at tuple.jl:308
  filter(::Any, ::Array{T, N}) where {T, N} at array.jl:2469
```

tkf: OK, so `filter(f, generator)` does not work... I think this should be fixed in `Base`. Meanwhile, I'd try something like this

```julia
        deps = [Base.PkgId(last(x), first(x)) for x in last(dep).deps]
        return pkg => filter!(!is_stdlib_and_loaded, deps)
```

ianshmean: Thanks. Fixed now

src/API.jl
```diff
@@ -914,39 +914,51 @@ function precompile(ctx::Context)
...
     end    
     
     was_processed = Dict{Base.PkgId,Base.Event}()
     was_recompiled = Dict{Base.PkgId,Bool}()
-    for pkgid in pkgids
-        was_processed[pkgid] = Base.Event()
-        was_recompiled[pkgid] = false
+    for dep in depsmap
+        was_processed[first(dep)] = Base.Event()
+        was_recompiled[first(dep)] = false
```
tkf, line 942: I suggest

```suggestion
    for pkgid in keys(depsmap)
        was_processed[pkgid] = Base.Event()
        was_recompiled[pkgid] = false
```

src/API.jl
```diff
@@ -894,71 +894,94 @@ end
...
+    for pkgid in keys(depsmap)
+        was_processed[pkgid] = Base.Event()
+        was_recompiled[pkgid] = false
     end
     
     errored = false
-    @sync for (i, pkg) in pairs(pkgids)
+    @sync for deppair in depsmap
+        pkg = first(deppair)
+        deps = last(deppair)
```
tkf, line 948:
```suggestion
    @sync for (pkg, deps) in depsmap
```

src/API.jl
```diff
@@ -894,71 +894,94 @@ end
...
+        is_stdlib_and_loaded(pkg) && return nothing
+        deps = [Base.PkgId(last(x), first(x)) for x in last(dep).deps]
+        return pkg => filter!(!is_stdlib_and_loaded, deps)
+    end
+    depsmap = Dict(Iterators.filter(!isnothing, deps_pair_or_nothing)) #flat map of each dep and its deps
     
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
-        push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
-        push!(pkg_dep_uuid_lists, collect(keys(ctx.env.project.deps)))
+        depsmap[Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name)  ] = [
```
tkf, line 932:
```suggestion
        depsmap[Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name)] = [
```

ianshmean:
Now also requires https://github.com/JuliaLang/julia/pull/37652

ianshmean:
@staticfloat it would be good to get this in before the next Pkg version so that the new `Pkg.precompile` method lands properly. 

Just waiting for https://github.com/JuliaLang/julia/pull/37652 and JuliaLang/julia#37596

ianshmean:
This is how an (unexpected) error from FFTW.jl looks, when precompiling an environment with Images.jl
https://gist.github.com/ianshmean/424d79ad08cb399195127c874de7debb

ianshmean:
Both julia PRs are merged, but I was probably to eager to re-trigger CI. I guess the nightly version CI picks up won't be ready for a while(?)

src/API.jl
```diff
@@ -894,71 +894,88 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
+    
+    function is_stale(paths, sourcepath)
+        for path_to_try in paths::Vector{String}
+            staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache())
+            staledeps === true && continue
+            return false
```
fredrikekre, line 902:
```suggestion
            staledeps ? continue : return false
```

src/API.jl
```diff
@@ -894,71 +894,88 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
+    
+    function is_stale(paths, sourcepath)
+        for path_to_try in paths::Vector{String}
+            staledeps = Base.stale_cachefile(sourcepath, path_to_try, Base.TOMLCache())
```
fredrikekre, line 900: Perhaps hoist the creation of the TOMLCache?
```suggestion
        toml_c = Base.TOMLCache()
        for path_to_try in paths::Vector{String}
            staledeps = Base.stale_cachefile(sourcepath, path_to_try, toml_c)
```

src/Pkg.jl
```diff
@@ -126,7 +126,10 @@ const add = API.add
 """
     Pkg.precompile()
 
-Precompile all the dependencies of the project.
+Precompile all dependencies of the project via a depth-first depdendency tree-based 
+parallel precompilation of all dependencies in the project manifest. 
+Errors will only throw when precompiling the top-level dependencies, given that 
+not all manifest dependencies may be loaded by the top-level dependencies on the given system.
```
fredrikekre, line 132: Perhaps make this a note-admonition?
```suggestion
!!! note
    Errors will only throw when precompiling the top-level dependencies, given that 
    not all manifest dependencies may be loaded by the top-level dependencies on the given system.
```

src/Pkg.jl
```diff
@@ -126,7 +126,10 @@ const add = API.add
 """
     Pkg.precompile()
 
-Precompile all the dependencies of the project.
+Precompile all dependencies of the project via a depth-first depdendency tree-based 
+parallel precompilation of all dependencies in the project manifest.
```
fredrikekre, line 130: Perhaps this is too detailed?
```suggestion
Precompile all the dependencies of the project in parallel.
```

src/API.jl
ianshmean: Could it be hoisted even higher? Generated once at the start of the precompile job?

fredrikekre: Maybe? I wasn't sure about that since you spawn multiple tasks.

ianshmean: True. I didn't realize it was an empty object for preallocation. I thought it was a snapshot of some form of system state for reference. I guess I don't know enough to answer your original question/suggestion

fredrikekre: The suggestion in the diff is safe, but I don't think you can move it out further since then multiple tasks will use the same object.

ianshmean:
That CI failure seems to be a connection glitch https://travis-ci.com/github/JuliaLang/Pkg.jl/jobs/389278321#L419

ianshmean:
Passed now. Just the docs that auto-cancelled with the first fail

src/API.jl
```diff
@@ -894,71 +894,88 @@ end
 
 precompile() = precompile(Context())
 function precompile(ctx::Context)
+    
+    function is_stale(paths, sourcepath)
+        toml_c = Base.TOMLCache()
+        for path_to_try in paths::Vector{String}
+            staledeps = Base.stale_cachefile(sourcepath, path_to_try, toml_c)
+            staledeps ? continue : return false
```
tkf, line 902: `stale_cachefile` can return non-boolean: https://github.com/JuliaLang/julia/blob/c0d06d3219751d28237cb32d9d6f7e2ef45916cd/base/loading.jl#L1428-L1429

```suggestion
            staledeps === true ? continue : return false
```

tkf (changes requested):

src/API.jl
ianshmean: ah yes, thanks

tkf (approved):

src/API.jl
fredrikekre: Sorry for the bad review, I just thought `x === true` looked weird.

src/API.jl
```diff
@@ -894,71 +894,89 @@ end
...
     errored = false
-    @sync for (i, pkg) in pairs(pkgids)
+    toml_c = Base.TOMLCache()
+    @sync for (pkg, deps) in depsmap
         paths = Base.find_all_in_cache_path(pkg)
-        sourcepath = Base.locate_package(pkg)
+        sourcepath = Base.locate_package(pkg, toml_c)
         sourcepath === nothing && continue
         # Heuristic for when precompilation is disabled
         occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String)) && continue
```
tkf, line 947: I wonder if you need to notify the downstream even if you don't recompile the package? Isn't it going to cause deadlocks if you don't do this?

```suggestion
        if occursin(r"\b__precompile__\(\s*false\s*\)", read(sourcepath, String))
            notify(was_processed[pkg])
            continue
        end
```

ianshmean: Yes, you're right. Thanks

src/API.jl
```diff
@@ -894,71 +894,89 @@ end
...
 function precompile(ctx::Context)
+    
+    function is_stale(paths, sourcepath)
+        toml_c = Base.TOMLCache()
+        for path_to_try in paths::Vector{String}
+            staledeps = Base.stale_cachefile(sourcepath, path_to_try, toml_c)
+            staledeps === true ? continue : return false
+        end
+        return true
+    end
```
tkf, line 905: I suggest to move `is_stale` out of `precompile`. As of https://github.com/JuliaLang/Pkg.jl/pull/2021/commits/88b5a71873e641b3e23074de25d811265f98c805 we have

https://github.com/JuliaLang/Pkg.jl/blob/88b5a71873e641b3e23074de25d811265f98c805/src/API.jl#L941-L944

I think it makes `is_stale` a closure that has `toml_c` as a boxed value as in this toy example:

```julia
julia> function f()
           g() = toml_c = 1
           toml_c = 2
           g
       end;

julia> f().toml_c
Core.Box(2)
```

Since `is_stale` is used in `@async`, it would mean that all tasks share the same `Base.TOMLCache()`.

(We *could* avoid this by renaming `toml_c` to something else. But I think closures with `@async`/`@spawn` are very fragile and hard to maintain in general. (Although obviously closures are super useful.) So, I think defining it as a function is the best approach here.)

If it's not ideal to put `is_stale` in `Pkg.API` namespace, maybe use `_is_stale` or define it in another submodule?

ianshmean: Indeed! Done. Thanks

ianshmean:
Some optimizations have been added to `TOMLCache` handling to speed up the no-op case, based on slack discussion with @staticfloat and @KristofferC 

A no-op, now looks like this (with Pkg dev-ed and not in the sysimage, so likely to be fast first time (?)):

```
julia> @time Pkg.precompile()
Precompiling project...
  1.194856 seconds (1.05 M allocations: 74.634 MiB, 2.22% gc time)

julia> @time Pkg.precompile()
Precompiling project...
  0.688262 seconds (851.08 k allocations: 63.627 MiB, 1.56% gc time)

julia> @time Pkg.precompile()
Precompiling project...
  0.625681 seconds (851.08 k allocations: 63.627 MiB, 3.42% gc time)
```

ianshmean:
I think this is good to go now, along with https://github.com/JuliaLang/Pkg.jl/pull/2033 to allow opting-in to make it automatic after the common pkg calls

src/API.jl
```diff
@@ -892,75 +892,94 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
...
-                wait(was_processed[dep_uuid])
+            for dep in deps # wait for deps to finish
+                wait(was_processed[dep])
             end
             
             # skip stale checking and force compilation if any dep was recompiled in this session
-            any_dep_recompiled = any(map(dep_uuid->was_recompiled[dep_uuid], pkg_dep_uuid_lists[i]))
-            if !errored && (any_dep_recompiled || is_stale(paths, sourcepath))
+            any_dep_recompiled = any(map(dep->was_recompiled[dep], deps))
+            if !errored && (any_dep_recompiled || _is_stale(paths, sourcepath, toml_c))
```
tkf, line 957: If we are going to rely on `Base.TOMLCache` to be "`async`-safe" (which is, BTW, pretty easy to break; e.g., a `@debug` can ruin this property), I don't think it's necessary to move `_is_stale` out. It's actually where closure is useful. Although there is nothing bad in factoring out `_is_stale` this way.

src/API.jl
```diff
@@ -892,75 +892,94 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
     Operations.build(ctx, pkgs, verbose)
 end
 
+function _is_stale(paths::Vector{String}, sourcepath::String, toml_c::Base.TOMLCache)
+    for path_to_try in paths::Vector{String}
```
tkf, line 896:
```suggestion
    for path_to_try in paths
```

src/API.jl
ianshmean: @staticfloat did some probing of the list of toml files that were accessed when no tomlcaches were provided, and it was just the env manifest and project files, so the thinking was that it was async safe as it's read-only after the first use in the for-loop. Would that be broken by `@debug`?

ianshmean: It'd be good to resolve this, and merge, to fix #2035

@fredrikekre @KristofferC

tkf: `parsed_toml(cache::TOMLCache, project_file::String)` reuses `TOML.Parser`: https://github.com/JuliaLang/julia/blob/2a36f83ce9bdfd17c572472f60b830aedcf181f0/base/loading.jl#L174-L175

So, for example, `TOML.parse(::TOML.Parser)` has to be async-safe. It probably is and likely be so forever but relying on "likely" is not the best approach especially in concurrent programming. For example, maybe somebody will implement `TOML.parse(::IO)` at some point and the parsing can happen in a streaming fashion.

(It's also a bit concerning that the I/O `read(project_file, String)` is happening in `get!` but that's probably handled by the `.age` field in `Dict`.)

Anyway, I don't think my comment should block this PR. We can fix it in `Base` if we want.

ianshmean: Ok, merging this as-is seems like the thing to do for now

Issue #1717, opened 2020-03-14
pkg> fsck


fredrikekre:
See https://github.com/JuliaLang/Pkg.jl/issues/1198#issuecomment-529224374

KristofferC:
Ref https://github.com/KristofferC/PkgFsck.jl

KristofferC:
Will close as dup of https://github.com/JuliaLang/Pkg.jl/issues/1683 because it is more or less the same thing.

PR #2022, opened 2020-09-16
Place Pkg.build logs in Pkg's scratchspace.

fredrikekre:
This puts build logs in Pkg's scratchspace and sets the building package as the owner. This makes sure that the build log is deleted when the package is garbage collected.

The reason for this is that I am implementing https://github.com/JuliaLang/Pkg.jl/issues/1717 and want to verify the content, but currently we pollute the package directory with the build log.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -4,7 +4,7 @@ module Operations
 
 using UUIDs
 using Random: randstring
-import LibGit2
+import LibGit2, Dates, TOML
 
 import REPL
 using REPL.TerminalMenus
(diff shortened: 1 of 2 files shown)


KristofferC:
> but currently we pollute the package directory with the build log.

I think this is a good idea, but don't packages that have a build step usually put a lot of other garbage there, other than the log file?

fredrikekre:
True, but we should push them to use Scratch or Artifacts for that, right?

staticfloat:
Unfortunate that we are further tying `Scratch` internals together with `Pkg` internals, but I'm fine with it.  :0

staticfloat (approved):

PR #2024, opened 2020-09-17
Add hook to augment `Platform` objects for Artifact selection

staticfloat:
This, combined with the recent overhaul of BinaryPlatforms to give
extended tags to `Platform` objects, enables packages to run arbitrary
(ideally, small) pieces of code to better inform artifact selection.

`Platform`-specific Artifacts can now be tagged with arbitrary metadata,
examples of which are `libc` version, processor microarchitecture, or
CUDA version.  Because we do not want to have to teach Julia how to
autodetect its compatibility with all possible tags, we instead add the
ability for a package to perform the autodetection, and run that code
after the package is installed, but before its artifacts are downloaded.

Cooperating packages will create a `.pkg/platform_augmentation_hook.jl`
file within their package, and that file will return a function as its
final value.  This code will be evaluated within an anonymous module,
and within the Package's own environment, allowing the hook to import
functionality if need be (this should be avoided as much as possible, so
as to reduce the possibility of conflicting versions of dependencies
causing an issue at install time).

### DIFF
--- a/docs/src/artifacts.md
+++ b/docs/src/artifacts.md
@@ -227,3 +227,77 @@ Note that while that hash was previously overridden, it is no longer, and theref
 Most methods that are affected by overrides have the ability to ignore overrides by setting `honor_overrides=false` as a keyword argument within them.
 For UUID/name based overrides to work, `Artifacts.toml` files must be loaded with the knowledge of the UUID of the loading package.
 This is deduced automatically by the `artifacts""` string macro, however if you are for some reason manually using the `Pkg.Artifacts` API within your package and you wish to honor overrides, you must provide the package UUID to API calls like `artifact_meta()` and `ensure_artifact_installed()` via the `pkg_uuid` keyword argument.
+
+## Extending Platform Selection
+
+!!! compat "Julia 1.6"
+    Pkg's extended platform selection requires at least Julia 1.6, and is considered experimental.
+
+New in Julia 1.6, `Platform` objects can have extended attributes applied to them, allowing artifacts to be tagged with things such as CUDA driver version compatibility, microarchitectural compatibility, julia version compatibility and more!
+Note that this feature is considered experimental and may change in the future.
+If you as a package developer find yourself needing this feature, please get in contact with us so it can evolve for the benefit of the whole ecosystem.
+In order to support artifact selection at `Pkg.add()` time, `Pkg` will run the specially-named file `<project_root>/.pkg/select_artifacts.jl`, passing the current platform triplet as the first arugment.
+This artifact selection script should print a `TOML`-serialized dictionary representing the artifacts that this package needs according to the given platform, and performing any inspection of the system as necessary to auto-detect platform capabilities, if they are not explicitly provided by the given platform triplet.
+The format of the dictionary should match that returned from `Artifacts.select_downloadable_artifacts()`, and indeed most packages should simply call that function with an augmented `Platform` object.
+An example artifact selection hook definition might look like the following, split across two files:
+
+```julia
+# .pkg/platform_augmentation.jl
+using Libdl, Base.BinaryPlatforms
+function augment_platform!(p::Platform)
+    # If this platform object already has a `cuda` tag set, don't augment
+    if haskey(p, "cuda")
+        return p
+    end
+
+    # Open libcuda explicitly, so it gets `dlclose()`'ed after we're done
+    dlopen("libcuda") do lib
+        # find symbol to ask for driver version; if we can't find it, just silently continue
+        cuDriverGetVersion = dlsym(lib, "cuDriverGetVersion"; throw_error=false)
+        if cuDriverGetVersion !== nothing
+            # Interrogate CUDA driver for driver version:
+            driverVersion = Ref{Cint}()
+            ccall(cuDriverGetVersion, UInt32, (Ptr{Cint},), driverVersion)
+
+            # Store only the major version
+            p["cuda"] = div(driverVersion, 1000)
+        end
+    end
+
+    # Return possibly-altered `Platform` object
+    return p
+end
+```
+
+```julia
+using TOML, Artifacts, Base.BinaryPlatforms
+include("./platform_augmentation.jl")
+artifacts_toml = joinpath(dirname(@__DIR__), "Artifacts.toml")
+
+# Get "target triplet" from ARGS, if given (defaulting to the host triplet otherwise)
+target_triplet = get(ARGS, 1, Base.BinaryPlatforms.host_triplet())
+
+# Augment this platform object with any special tags we require
+platform = augment_platform!(HostPlatform(parse(Platform, target_triplet)))
+
+# Select all downloadable artifacts that match that platform
+artifacts = select_downloadable_artifacts(artifacts_toml; platform)
+
+# Output the result to `stdout` as a TOML dictionary
+TOML.print(stdout, artifacts)
+```
+
+In this hook definition, our platform augmentation routine opens a system library (`libcuda`), searches it for a symbol to give us the CUDA driver version, then embeds the major version of that version number into the `cuda` property of the `Platform` object we are augmenting.
+While it is not critical for this code to actually attempt to close the loaded library (as it will most likely be opened again by the CUDA package immediately after the package operations are completed) it is best practice to make hooks as lightweight and transparent as possible, as they may be used by other Pkg utilities in the future.
+In your own package, you should also use augmented platform objects when using the `@artifact_str` macro, as follows:
+
+```julia
+include("../.pkg/platform_augmentation.jl")
+
+function __init__()
+    p = augment_platform!(HostPlatform())
+    global my_artifact_dir = @artifact_str("MyArtifact", p)
+end
+```
+
+This ensures that the same artifact is used by your code as Pkg attempted to install.
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -10,7 +10,8 @@ import REPL
 using REPL.TerminalMenus
 using ..Types, ..Resolve, ..PlatformEngines, ..GitTools, ..MiniProgressBars
 import ..depots, ..depots1, ..devdir, ..set_readonly, ..Types.PackageEntry
-import ..Artifacts: ensure_all_artifacts_installed, artifact_names, extract_all_hashes, artifact_exists
+import ..Artifacts: ensure_artifact_installed, artifact_names, extract_all_hashes,
+                    artifact_exists, select_downloadable_artifacts
 using Base.BinaryPlatforms
 import ...Pkg
 import ...Pkg: pkg_server, Registry, pathrepr, can_fancyprint, printpkgstyle
@@ -586,12 +587,36 @@ function install_git(
     end
 end
 
+function collect_artifacts(pkg_root::String; platform::AbstractPlatform=HostPlatform())
+    # Check to see if this package has an (Julia)Artifacts.toml
+    artifacts_tomls = Tuple{String,Dict}[]
+    for f in artifact_names
+        artifacts_toml = joinpath(pkg_root, f)
+        if isfile(artifacts_toml)
+            selector_path = joinpath(pkg_root, ".pkg", "select_artifacts.jl")
+
+            # If there is a dynamic artifact selector, run that in an appropriate sandbox to select artifacts
+            if isfile(selector_path)
+                select_cmd = Cmd(`$(gen_build_code(selector_path)) $(triplet(platform))`)
+                addenv(select_cmd, "JULIA_LOAD_PATH" => "@stdlib")
+                meta_toml = String(read(select_cmd))
+                push!(artifacts_tomls, (artifacts_toml, TOML.parse(meta_toml)))
+            else
+                # Otherwise, use the standard selector from `Artifacts`
+                artifacts = select_downloadable_artifacts(artifacts_toml; platform)
+                push!(artifacts_tomls, (artifacts_toml, artifacts))
+            end
+            break
+        end
+    end
+    return artifacts_tomls
+end
+
+
 function download_artifacts(env::EnvCache, pkgs::Vector{PackageSpec};
                             platform::AbstractPlatform=HostPlatform(),
                             julia_version = VERSION,
                             verbose::Bool=false)
-    # Filter out packages that have no source_path()
-    # pkg_roots = String[p for p in source_path.((env.project_file,), pkgs) if p !== nothing]  # this runs up against inference limits?
     pkg_roots = String[]
     for pkg in pkgs
         p = source_path(env.project_file, pkg, julia_version)
@@ -603,34 +628,22 @@ end
 function download_artifacts(pkg_roots::Vector{String};
                             platform::AbstractPlatform=HostPlatform(),
                             verbose::Bool=false)
-    # List of Artifacts.toml files that we're going to download from
-    artifacts_tomls = String[]
-
-    for path in pkg_roots
-        # Check to see if this package has an (Julia)Artifacts.toml
-        for f in artifact_names
-            artifacts_toml = joinpath(path, f)
-            if isfile(artifacts_toml)
-                push!(artifacts_tomls, artifacts_toml)
-                break
+    for pkg_root in pkg_roots
+        for (artifacts_toml, artifacts) in collect_artifacts(pkg_root; platform)
+            # For each Artifacts.toml, install each artifact we've collected from it
+            for name in keys(artifacts)
+                ensure_artifact_installed(name, artifacts[name], artifacts_toml;
+                                            verbose, quiet_download=!(stderr isa Base.TTY))
             end
-        end
-    end
-
-    if !isempty(artifacts_tomls)
-        for artifacts_toml in artifacts_tomls
-            ensure_all_artifacts_installed(artifacts_toml; platform=platform, verbose=verbose, quiet_download=!(stderr isa Base.TTY))
             write_env_usage(artifacts_toml, "artifact_usage.toml")
         end
     end
 end
 
 function check_artifacts_downloaded(pkg_root::String; platform::AbstractPlatform=HostPlatform())
-    for f in artifact_names
-        artifacts_toml = joinpath(pkg_root, f)
-        if isfile(artifacts_toml)
-            hashes = extract_all_hashes(artifacts_toml)
-            if !all(artifact_exists.(hashes))
+    for (artifacts_toml, artifacts) in collect_artifacts(pkg_root; platform)
+        for name in keys(artifacts)
+            if !artifact_exists(Base.SHA1(artifacts[name]["git-tree-sha1"]))
                 return false
             end
             break
@@ -824,7 +837,6 @@ function dependency_order_uuids(env::EnvCache, uuids::Vector{UUID})::Dict{UUID,I
     seen = UUID[]
     k = 0
     function visit(uuid::UUID)
-        is_stdlib(uuid) && return
         uuid in seen &&
             return @warn("Dependency graph not a DAG, linearizing anyway")
         haskey(order, uuid) && return
--- a/.gitignore
+++ b/.gitignore
@@ -9,4 +9,5 @@ docs/build
 /docs/src/repl.md
 .DS_Store
 /test/registries/*
+/test/registry_depot/*
 /test/loaded_depot/*
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -467,8 +467,19 @@ be provided to properly support overrides from `Overrides.toml` entries in depot
 
 If `include_lazy` is set to `true`, then lazy packages will be installed as well.
 
+This function is deprecated and should be replaced with the following snippet:
+
+    artifacts = select_downloadable_artifacts(artifacts_toml; platform, include_lazy)
+    for name in keys(artifacts)
+        ensure_artifact_installed(name, artifacts[name], artifacts_toml; platform=platform)
+    end
+
 !!! compat "Julia 1.3"
     This function requires at least Julia 1.3.
+
+!!! warning
+    This function is deprecated in Julia 1.6 and will be removed in a future version.
+    Use `select_downloadable_artifacts()` and `ensure_artifact_installed()` instead.
 """
 function ensure_all_artifacts_installed(artifacts_toml::String;
                                         platform::AbstractPlatform = HostPlatform(),
@@ -476,6 +487,8 @@ function ensure_all_artifacts_installed(artifacts_toml::String;
                                         include_lazy::Bool = false,
                                         verbose::Bool = false,
                                         quiet_download::Bool = false)
+    # This function should not be called anymore; use `select_downloadable_artifacts()` directly.
+    Base.depwarn("`ensure_all_artifacts_installed()` is deprecated; iterate over `select_downloadable_artifacts()` output with `ensure_artifact_installed()`.", :ensure_all_artifacts_installed)
     # Collect all artifacts we're supposed to install
     artifacts = select_downloadable_artifacts(artifacts_toml; platform, include_lazy, pkg_uuid)
     for name in keys(artifacts)
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -18,7 +18,8 @@ using SHA
 
 export create_artifact, artifact_exists, artifact_path, remove_artifact, verify_artifact,
        artifact_meta, artifact_hash, bind_artifact!, unbind_artifact!, download_artifact,
-       find_artifacts_toml, ensure_artifact_installed, @artifact_str, archive_artifact
+       find_artifacts_toml, ensure_artifact_installed, @artifact_str, archive_artifact,
+       select_downloadable_artifacts
 
 """
     create_artifact(f::Function)
@@ -475,25 +476,11 @@ function ensure_all_artifacts_installed(artifacts_toml::String;
                                         include_lazy::Bool = false,
                                         verbose::Bool = false,
                                         quiet_download::Bool = false)
-    if !isfile(artifacts_toml)
-        return
-    end
-    artifact_dict = load_artifacts_toml(artifacts_toml; pkg_uuid=pkg_uuid)
-
-    for name in keys(artifact_dict)
-        # Get the metadata about this name for the requested platform
-        meta = artifact_meta(name, artifact_dict, artifacts_toml; platform=platform)
-
-        # If there are no instances of this name for the desired platform, skip it
-        meta === nothing && continue
-
-        # If this mapping doesn't have a `download` stanza or is lazy, skip it
-        if !haskey(meta, "download") || (get(meta, "lazy", false) && !include_lazy)
-            continue
-        end
-
+    # Collect all artifacts we're supposed to install
+    artifacts = select_downloadable_artifacts(artifacts_toml; platform, include_lazy, pkg_uuid)
+    for name in keys(artifacts)
         # Otherwise, let's try and install it!
-        ensure_artifact_installed(name, meta, artifacts_toml; platform=platform,
+        ensure_artifact_installed(name, artifacts[name], artifacts_toml; platform=platform,
                                   verbose=verbose, quiet_download=quiet_download)
     end
 end
--- a/test/artifacts.jl
+++ b/test/artifacts.jl
@@ -355,18 +355,22 @@ end
 
         # Install artifacts such that `c_simple` is not installed properly
         # because of the platform we requested, but `socrates` is.
-        ensure_all_artifacts_installed(artifacts_toml; platform=Platform("powerpc64le", "linux"))
+        missing_platform = Platform("powerpc64le", "linux")
+        artifacts = select_downloadable_artifacts(artifacts_toml; platform=missing_platform)
+        for name in keys(artifacts)
+            ensure_artifact_installed(name, artifacts[name], artifacts_toml; platform=missing_platform)
+        end
 
         # Test that c_simple doesn't even show up
-        c_simple_hash = artifact_hash("c_simple", artifacts_toml; platform=Platform("powerpc64le", "linux"))
+        c_simple_hash = artifact_hash("c_simple", artifacts_toml; platform=missing_platform)
         @test c_simple_hash == nothing
 
         # Test that socrates shows up, but is not installed
-        socrates_hash = artifact_hash("socrates", artifacts_toml; platform=Platform("powerpc64le", "linux"))
+        socrates_hash = artifact_hash("socrates", artifacts_toml; platform=missing_platform)
         @test !artifact_exists(socrates_hash)
 
         # Test that collapse_the_symlink is installed
-        cts_hash = artifact_hash("collapse_the_symlink", artifacts_toml; platform=Platform("powerpc64le", "linux"))
+        cts_hash = artifact_hash("collapse_the_symlink", artifacts_toml; platform=missing_platform)
         @test artifact_exists(cts_hash)
     end
 
--- a/test/artifacts.jl
+++ b/test/artifacts.jl
@@ -1,7 +1,7 @@
 module ArtifactTests
 import ..Pkg # ensure we are using the correct Pkg
 
-using Test, Random, Pkg.Artifacts, Pkg.BinaryPlatforms, Pkg.PlatformEngines
+using Test, Random, Pkg.Artifacts, Base.BinaryPlatforms, Pkg.PlatformEngines
 import Pkg.Artifacts: pack_platform!, unpack_platform, with_artifacts_directory, ensure_all_artifacts_installed, extract_all_hashes
 using TOML, Dates
 import Base: SHA1
@@ -368,6 +368,112 @@ end
         # Test that collapse_the_symlink is installed
         cts_hash = artifact_hash("collapse_the_symlink", artifacts_toml; platform=Platform("powerpc64le", "linux"))
         @test artifact_exists(cts_hash)
+    end
+
+    # Ensure that platform augmentation hooks work.  We will switch between two arbitrary artifacts for this,
+    # by inspecting an environment variable in our package hook.
+    engaged_hash = SHA1("a5f8161ca1ab2e94fedd3578586fe06d7906177c")
+    engaged_url = "https://github.com/JuliaBinaryWrappers/HelloWorldGo_jll.jl/releases/download/HelloWorldGo-v1.0.4+0/HelloWorldGo.v1.0.4.aarch64-linux-musl.tar.gz"
+    engaged_sha256 = "9b66d6b02a370d0170a8c217a872cd1f3d53de267d4e63c22a40b49f04367f8a"
+    disengaged_hash = SHA1("ea8ea92ecd57aa602d254ca6c637309642202768")
+    disengaged_url = "https://github.com/JuliaBinaryWrappers/HelloWorldGo_jll.jl/releases/download/HelloWorldGo-v1.0.4+0/HelloWorldGo.v1.0.4.i686-w64-mingw32.tar.gz"
+    disengaged_sha256 = "5c96a327fc6f0dc71d533373bc6cc6719a1e477c72319b800f29abf1b1e7d812"
+
+    function generate_flooblegrank_artifacts(ap_path)
+        # Bind both "engaged" and "disengaged" variants of our `gooblebox` artifact to generate an Artifacts.toml file
+        artifacts_toml = joinpath(ap_path, "Artifacts.toml")
+        engaged_platform = HostPlatform()
+        engaged_platform["flooblecrank"] = "engaged"
+        Pkg.Artifacts.bind_artifact!(
+            artifacts_toml,
+            "gooblebox",
+            engaged_hash;
+            download_info = [(engaged_url, engaged_sha256)],
+            platform = engaged_platform,
+        )
+        disengaged_platform = HostPlatform()
+        disengaged_platform["flooblecrank"] = "disengaged"
+        Pkg.Artifacts.bind_artifact!(
+            artifacts_toml,
+            "gooblebox",
+            disengaged_hash;
+            download_info = [(disengaged_url, disengaged_sha256)],
+            platform = disengaged_platform,
+        )
+    end
+
+    for flooblecrank_status in ("engaged", "disengaged")
+        # Ensure that they're both missing at first so tests can fail properly
+        temp_pkg_dir() do project_path
+            copy_test_package(project_path, "AugmentedPlatform")
+            ap_path = joinpath(project_path, "AugmentedPlatform")
+            generate_flooblegrank_artifacts(ap_path)
+
+            Pkg.activate(ap_path)
+            add_this_pkg()
+            if flooblecrank_status == "engaged"
+                right_hash = engaged_hash
+                wrong_hash = disengaged_hash
+            else
+                right_hash = disengaged_hash
+                wrong_hash = engaged_hash
+            end
+
+            @test !isdir(artifact_path(right_hash))
+            @test !isdir(artifact_path(wrong_hash))
+
+            # Instantiate with this environment variable, to install the proper one
+            withenv("FLOOBLECRANK" => flooblecrank_status) do
+                Pkg.instantiate()
+                @test isdir(artifact_path(right_hash))
+                @test !isdir(artifact_path(wrong_hash))
+            end
+
+            # Manual test that artifact is installed by instantiate()
+            artifacts_toml = joinpath(ap_path, "Artifacts.toml")
+            p = HostPlatform()
+            p["flooblecrank"] = flooblecrank_status
+            flooblecrank_hash = artifact_hash("gooblebox", artifacts_toml; platform=p)
+            @test flooblecrank_hash == right_hash
+            @test artifact_exists(flooblecrank_hash)
+
+            # Test that if we load the package, it knows how to find its own artifact,
+            # because it feeds the right `Platform` object through to `@artifact_str()`
+            cmd = setenv(`$(Base.julia_cmd()) --color=yes --project=$(ap_path) -e 'using AugmentedPlatform; print(get_artifact_dir("gooblebox"))'`,
+                         "JULIA_DEPOT_PATH" => join(Base.DEPOT_PATH, Sys.iswindows() ? ";" : ":"),
+                         "FLOOBLECRANK" => flooblecrank_status)
+            using_output = chomp(String(read(cmd)))
+            @test success(cmd)
+            @test artifact_path(right_hash) == using_output
+
+            mkpath("/tmp/foo/$(flooblecrank_status)")
+            rm("/tmp/foo/$(flooblecrank_status)"; recursive=true, force=true)
+            cp(project_path, "/tmp/foo/$(flooblecrank_status)")
+            cp(Base.DEPOT_PATH[1], "/tmp/foo/$(flooblecrank_status)/depot")
+        end
+    end
+
+    # Also run a test of "cross-installation"
+    temp_pkg_dir() do project_path
+        copy_test_package(project_path, "AugmentedPlatform")
+        ap_path = joinpath(project_path, "AugmentedPlatform")
+        generate_flooblegrank_artifacts(ap_path)
+
+        Pkg.activate(ap_path)
+        add_this_pkg()
+        @test !isdir(artifact_path(engaged_hash))
+        @test !isdir(artifact_path(disengaged_hash))
+
+        # Instantiate with the environment variable set, but with an explicit
+        # tag set in the platform object, which overrides.
+        withenv("FLOOBLECRANK" => "disengaged") do
+            p = HostPlatform()
+            p["flooblecrank"] = "engaged"
+            Pkg.instantiate(;platform=p)
+
+            @test isdir(artifact_path(engaged_hash))
+            @test !isdir(artifact_path(disengaged_hash))
+        end
     end
 end
 
--- a/test/test_packages/AugmentedPlatform/.pkg/platform_augmentation.jl
+++ b/test/test_packages/AugmentedPlatform/.pkg/platform_augmentation.jl
@@ -1,0 +1,18 @@
+using Base.BinaryPlatforms
+
+# Our platform augmentation simply inspects the environment
+function augment_platform!(p::Platform)
+    # If the `flooblecrank` tag is already set, don't auto-detect!
+    if haskey(p, "flooblecrank")
+        return p
+    end
+
+    # If the tag is not set, autodetect through magic (in this case, checking environment variables)
+    flooblecrank_status = get(ENV, "FLOOBLECRANK", "disengaged")
+    if flooblecrank_status == "engaged"
+        p["flooblecrank"] = "engaged"
+    else
+        p["flooblecrank"] = "disengaged"
+    end
+    return p
+end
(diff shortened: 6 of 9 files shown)


lobingera:
"enables packages to run arbitrary pieces of code" - i might miss the context, but wasn't this a big, big attack surface for security problems?

fredrikekre:
Packaged can already run arbitrary pieces of code.

lobingera:
> Packaged can already run arbitrary pieces of code.

at install time?

KristofferC:
Yes, `deps/build.jl`.

KristofferC:
Maybe you could elaborate a bit on this more. Say I want to create a PackageCompiler type of application. I then need to bundle the artifacts I want to use. Do I need to run some code in the packages now to know what artifacts they want to use? It feels kinda unfortunate to get away from the declarativeness of Artifacts but maybe I am misunderstanding.

lobingera:
me? I had the impression, that all the movement to Artifacts and taking this out of Pkg to juiia core was building a infrastructure in which things like deps/build.jl are not needed anymore and all installation is 'executing' only datafiles (.toml) anymore.

KristofferC:
Sorry, I was talking to @staticfloat about this implementation.

maleadt:
> Do I need to run some code in the packages now to know what artifacts they want to use?

For CUDA apps, that would be an improvement. PackageCompiler currently downloads them all (we do our own artifact handling), so you end up with a 12GB `artifacts` folder.

@staticfloat Will there be a way to re-trigger artifact resolution, or will that be covered by `Pkg.resolve`? CUDA compatibility can change when the user upgrades his driver, and I assume we'll need to detect that case in CUDA.jl's `__init__` and prompt the user to do something.

KristofferC:
>  (we do our own artifact handling), so you end up with a 12GB artifacts folder.

PackageCompiler does the absolute simplest possible thing when it comes to artifacts so there's no surprise things could be improved there.

With this, how much do I need to worry about the state of the machine I build the sysimage on? What would a good interface for `create_app` be when it comes to artifact handling?

staticfloat:
Wow, what a bunch of great questions.  You guys really strike to the heart of the matter.

## A Bit of Elaboration

First off, this whole feature was basically designed for Tim, so let's talk in the context of CUDA, but I've done my best to make it generic.  The problem statement is that we have this nice, declarative syntax for platform-specific Artifacts, but they are not extensible.  If there are compatibility constraints outside of the traditional OS, CPU architecture, libc, etc... then it's completely up to the package to download the correct Artifact.  That's currently possible by using lazy artifacts (e.g. tell Pkg to not try and download at `Pkg.add()` time, and only download the artifacts upon first use) but in the spirit of giving more information to Pkg and PackageCompiler and friends, we want to add more information to the `Platform` object to allow it to carry compatibility information about things we never dreamed of.

First, we rewrote `BinaryPlatforms` to have a new `Platform` object that is a full-blown key-value store with all sorts of bells and whistles around how to compare.

<details>
<summary>`Platform` object usage example</summary>

```julia
using Base.BinaryPlatforms

# HostPlatform() is how you get the Platform for the currently-running machine.
P = HostPlatform()

# This results in "x86_64-linux-gnu-cxx11-julia_version+1.6.0"
@show triplet(P)

# Here's a pretend CUDA artifact identified by platform in the `Artifacts.toml` file
cuda_artifact = Platform("x86_64", "linux"; cuda = v"11")

# This results in "x86_64-linux-gnu-cuda+11.0.0"
@show triplet(cuda_artifact)

# Note that these match, because on every tag that is currently specified
# (which is `os` ("linux"), `arch` ("x86_64"), and `libc` ("glibc")) they match:
@show platforms_match(P, cuda_artifact)

# If, however, we were to add an incompatible `cuda` constraint to the
# `HostPlatform`, they will no longer match.
P["cuda"] = string(v"10")
@show platforms_match(P, cuda_artifact)
```
</details>

Of course, we could have solved this by having the base Julia `Platform` object know about every single possible tag under the sun, but that's clearly not a great solution, as the manner in which we detect CUDA version might change over time, etc... The purpose of this PR is to make it possible for packages to add tags to the `Platform` object so that the internals of `Pkg` and `Artifacts` can continue to blindly apply `platforms_match()` to find the appropriate artifact to pull out of an `Artifacts.toml`.

A hook definition will look something like this:
```julia
using Base.BinaryPlatforms
include("../src/utils/cuda_detector.jl")

function augment_platform!(p::Platform)
    p["cuda"] = do_cuda_version_detection_magic()
    return p
end 
```

This PR will cause `Pkg` to apply `augment_platform!()` to `Platform` objects before it sends them to the `Artifacts` installation code, and [this PR to Base](https://github.com/JuliaLang/julia/pull/37621) does the same but for the `@artifact_str` macro, so that when you use `artifact"cuda"`, you don't have to manually insert the same platform augmentation into your `Platform` objects, it just does it automatically.

# Questions

> Say I want to create a PackageCompiler type of application. I then need to bundle the artifacts I want to use. Do I need to run some code in the packages now to know what artifacts they want to use? It feels kinda unfortunate to get away from the declarativeness of Artifacts but maybe I am misunderstanding.

In fact, I argue this _increases_ the amount of (effective) declarativeness, as it gives a hope for packages like CUDA to inform you which version of CUDA artifact the current machine would use.  That may still be useless for PackageCompiler in some cases (what if you deploy to machine with a different CUDA version?) but it should certainly not make the matter worse for anyone.  For PackageCompiler, if you use `Pkg.add()` to install things, you'll get smart artifact selection for free; if you want to "cross-add" packages, you can pass a specific `Platform` object in, and that `Platform` object will get modified by the hooks as it installs artifacts.

> Will there be a way to re-trigger artifact resolution, or will that be covered by Pkg.resolve? CUDA compatibility can change when the user upgrades his driver, and I assume we'll need to detect that case in CUDA.jl's __init__ and prompt the user to do something.

First off, to be clear, the code in this PR is only really concerned with behavior at `Pkg.add()` time, so if this were the only thing we merged, and your code suddenly started to access a different artifact than what was installed at `Pkg.add()` time, it would just download that new (missing) artifact on first use.  This would simply be the typical lazy artifact behavior.  However, I have done a lot of work recently to move things to compile-time, including artifact selection, whenever possible.  And so in my PR to base, I made sure that using a platform augmentation hook [specifically does not allow that optimization](https://github.com/JuliaLang/julia/pull/37621/files#diff-e01d0e7f3b294b2f2e3ca556c2b8c9ebR617-R621), so that the platform augmentation hook is consulted literally every time you call `@artifact_str()`.  It's a (very small) performance penalty, but it allows your environment to change and the `Platform` to not get stuck with a stale representation of the environment.

> With this, how much do I need to worry about the state of the machine I build the sysimage on? What would a good interface for create_app be when it comes to artifact handling?

I would very much like to push people to use this mechanism sparingly; when I designed this functionality I had in mind very "slow-moving" things, like driver versions. I suppose there is the option that we could use this to implement "variants", e.g. I have a `BLAS` key to my `Arpack_jll` artifacts, and some of them are `MKL` and some of them are `OpenBLAS`.  But I'm not totally sold on that, and characteristics of your environment are probably the "fastest-changing" thing that I think should be used with this mechanism; it's probably not a good idea to use this for something that can change over the course of a computation, for instance.

For `create_app()`, I think you would probably need to expose a way to capture the full range of `Platform` objects that could possibly arise on the target machine.  So, for instance, if you don't know precisely what CUDA version is going to be used on the target machine, but you know it will be either CUDA v10 or v11 (but not v9 or v12) then we need a way to call `Pkg.Operations.download_artifact(;platform=p)` for multiple `p`.

staticfloat:
uh-oh, a test is passing that should be failing.  Don't merge this yet.

src/Operations.jl
```diff
@@ -643,25 +643,56 @@ function download_artifacts(ctx::Context, pkgs::Vector{PackageSpec}; platform::A
...
 end
 
+"""
+    load_pkg_hook(hook_path::String)
+
+Given a path to a hook file (which should end in a function definition) include
+it into an anonymous module and return that function.
+"""
+function load_pkg_hook(hook_path::String)
+    m = Module(:__anon__)
```
fredrikekre, line 653: Doesn't this need to be `gensym` or something in order not to cause clashes?
```suggestion
    m = Module(gensym())
```

src/Operations.jl
```diff
@@ -643,25 +643,56 @@ function download_artifacts(ctx::Context, pkgs::Vector{PackageSpec}; platform::A
...
+    f = Pkg.API.activate(dirname(dirname(hook_path))) do
+        Base.include(m, hook_path)::Function
+    end
```
fredrikekre, line 656: You need to performe the `LOAD_PATH` dance here.


What is the point of running with the package as the active project? Presumably you can't load the package anyway since we are still installing it? And this project might not even be valid (missing or inconsistent manifest, uninstalled dependencies etc) so you can't even expect to load other packages either.

src/Operations.jl
```diff
@@ -643,25 +643,56 @@ function download_artifacts(ctx::Context, pkgs::Vector{PackageSpec}; platform::A
...
+    # Beacuse this function will most likely be used in an old worldage,
```
fredrikekre, line 657:
```suggestion
    # Because this function will most likely be used in an old worldage,
```

src/Operations.jl
staticfloat: What is the `LOAD_PATH` dance?

The point is that perhaps someone's hook requires some fundamental piece of infrastructure; for instance, perhaps the routines that can tell us which CUDA artifact to download are in some `CUDAUtils` package that `CUDAnative` relies upon; it makes sense to me that the hook should be run with respect to that package's environment, even if it should be vanishingly rare for it to need that.

In what situations would there be invalid manifests or uninstalled dependencies?

fredrikekre: > What is the LOAD_PATH dance?

```
old_path = copy(LOAD_PATH)
empty!(LOAD_PATH)
push!(LOAD_PATH, "@")
try
   #...
finally
    empty!(LOAD_PATH)
    append!(LOAD_PATH, old_path)
end
```

> In what situations would there be invalid manifests or uninstalled dependencies?

Basically all the time? If the package doesn't commit the manifest it will error since it is missing, if it does come with a manifest it is unlikely to contain the same versions as Pkg decided upon and downloaded in the actual project.

> The point is that perhaps someone's hook requires some fundamental piece of infrastructure; for instance, perhaps the routines that can tell us which CUDA artifact to download are in some CUDAUtils package that CUDAnative relies upon; it makes sense to me that the hook should be run with respect to that package's environment, even if it should be vanishingly rare for it to need that.

Setting this up correctly is essentially what Pkg.build does. It creates a new project where you can load your own dependencies and starts a new Julia process with that env. 

I think that (i) the code you put in the hook file is completely self contained (not even using stdlibs) or (ii) we reuse Pkg.build, but the process prints out a program/config to `stdout` or something that can be read by the hook machinery in Artifacts.

staticfloat: Yeah, I think you're right, let's just require it to be completely self-contained.  That's more in the spirit of a hook anyway.

fredrikekre: Another option would be to just mark all artifacts as lazy and just download the correct one in Pkg.build

staticfloat:
Okay, I fixed the tests and eliminated something that was just a little too magical.  This will fail until https://github.com/JuliaLang/julia/pull/37592 is merged.

src/Operations.jl
staticfloat: Heh, that is exactly what we're trying to avoid.  ;)  That's what is possible now, but it's a better experience for the ecosystem as a whole if we can know beforehand what will be downloaded.

fredrikekre: Right, but with this hook you can't "statically" decide that anyway since you need to run code, so it already is essentially `deps/build.jl` in disguise. Anyway, just a suggestion.

staticfloat: You're right that it's not purely static, but once you have all package source available, it is a stateless operation to determine what artifacts need to be downloaded, which I think is as close to static as we can get without pushing more intelligence into base Julia itself.

src/Operations.jl
```diff
@@ -643,25 +643,55 @@ function download_artifacts(ctx::Context, pkgs::Vector{PackageSpec}; platform::A
...
 
     for path in pkg_roots
         # Check to see if this package has an (Julia)Artifacts.toml
         for f in artifact_names
             artifacts_toml = joinpath(path, f)
             if isfile(artifacts_toml)
-                push!(artifacts_tomls, artifacts_toml)
+                hook_path = joinpath(path, ".pkg", "platform_augmentation_hook.jl")
+                if isfile(hook_path)
+                    try
```
KristofferC, line 673: Consistently with the build script now throwing an exception if it fails, this should most likely also do so. If the script is so broken that it throws you don't want a tiny log message that is easy to miss when e.g. deploying it.

Personally, I think almost every try catch in the Artifact loading system should be removed since they make it easy for bugs to slip through, and then you run with a broken configuration. But I don't want to push that too hard.

src/Operations.jl
```diff
@@ -643,25 +643,55 @@ function download_artifacts(ctx::Context, pkgs::Vector{PackageSpec}; platform::A
     return download_artifacts(ctx, pkg_roots; platform=platform, verbose=verbose)
 end
 
+"""
+    load_pkg_hook(hook_path::String)
+
+Given a path to a hook file (which should end in a function definition) include
+it into an anonymous module and return that function.
+"""
+function load_pkg_hook(hook_path::String)
```
KristofferC, line 652: I think this should be executed in a separate process to hide the side effects of executing the file (like if it loads dependencies) in the same way as `build.jl` does it.

src/Operations.jl
staticfloat: Sure, that's fair enough.

src/Operations.jl
staticfloat: I'm conflicted on this; on the one hand, we have a natural serialization of platforms already defined, so communication is not a problem.  On the other, spinning up a subprocess to do this will add ~100ms * number of packages with hooks.  That's just on the cusp of being what I would consider to be expensive.

I've walked back on my stance that the hooks should be able to load things from their environment; they're now executed in whatever environment the `Pkg.add()` code itself is running in, so I think package authors will not be able to load things to effect things that much.  Even if they try to load a package, it would usually fail (because it may not have been installed yet, as this is happening mid-`Pkg.add()`) and if it succeeds, it will have come from the calling environment so it is probably the version you'd eventually load anyway.

src/Operations.jl
KristofferC: The `LOAD_PATH` dance doesn't really work because anything you have loaded in the current process will be used no matter what. That's why I suggested a separate process.

src/Operations.jl
KristofferC: > On the other, spinning up a subprocess to do this will add ~100ms * number of packages with hooks. That's just on the cusp of being what I would consider to be expensive.

True, but there shouldn't be many packages doing this, it only happens at `install` time (where we have to download the package), etc. So will that really be significant?

staticfloat: Not only at `install` time, but also things like `instantiate`, which will otherwise be much faster.  It will need to happen at `instantiate` time because we must check to see if each needed Artifact is actually installed or not.

staticfloat: I just realized one reason the serialization idea won't work; if a Pkg hook needs to add a custom comparison strategy (e.g. CUDA doesn't just match `cuda+11` to `cuda+10`, it is a version cap, just like `os_version` or `libstdcxx_version`, so it needs to call `set_compare_strategy!()` on the platform that is passed through it).

Right now, we have a small library of alternate comparison strategies, but in general the comparison strategies are a `Dict{String,Function}`, and we don't want to serialize the `Function`'s as well, I don't think.  The comparison functions should be completely self-contained, and I'm fine with erroring if they're not, but I don't actually know if serializing functions across Julia sessions is supported.

docs/src/artifacts.md
```diff
@@ -227,3 +227,50 @@ Note that while that hash was previously overridden, it is no longer, and theref
 Most methods that are affected by overrides have the ability to ignore overrides by setting `honor_overrides=false` as a keyword argument within them.
 For UUID/name based overrides to work, `Artifacts.toml` files must be loaded with the knowledge of the UUID of the loading package.
 This is deduced automatically by the `artifacts""` string macro, however if you are for some reason manually using the `Pkg.Artifacts` API within your package and you wish to honor overrides, you must provide the package UUID to API calls like `artifact_meta()` and `ensure_artifact_installed()` via the `pkg_uuid` keyword argument.
+
+## Extending Platform Selection
+
+New in Julia 1.6, `Platform` objects can have extended attributes applied to them, allowing artifacts to be tagged with things such as CUDA driver version compatibility, microarchitectural compatibility, julia version compatibility and more!
+Note that this API is experimental and may change in the future.
```
StefanKarpinski, line 234: I would move this to the first sentence or maybe a `!!! warning` or something.

fredrikekre:
Since this is just a _"I know better than BinaryPlatforms what artifact I need"_ thing (IIUC), why can't the file just return the TOML key instead of a function? That should make it simpler to do this in a separate process too instead of the `eval`ing since we just have to read a string from stdout?

staticfloat:
You're assuming there's only one artifact; the platform object _is_ the key.  You can't get the artifact hash unless you know both the name and the platform.

fredrikekre:
> the platform object is the key

Yes, and this PR is about modifying the key, right? So why can't the function just return the key in stringified form instead?

staticfloat:
Mostly an implementation detail; comparison strategies aren’t serialized.

src/Operations.jl
KristofferC: I still think it is wrong to include a bunch of files into the running julia process when running `Pkg.add`. Who knows what state the process now is in? What packages are loaded, what methods have been defined etc. Especially if you are running Pkg more as an API and not just as a user sitting in front of your terminal.

StefanKarpinski:
Can you elaborate on that, @staticfloat? I don't really follow. Seems easier to just generate a key value.

staticfloat:
Okay, I had to alter my way of thinking about this, but I now agree with you three that we can turn this into an interface where we run a separate _program_ that returns the hashes that need to be installed.  Example:

Package `EdgeCase` contains within it four artifacts: one that is a "normal" platform-specific artifact, one that is a platform-unspecific artifact, one that uses a special platform tag that Julia doesn't know about, and one lazy artifact.

Pkg will execute `.pkg/select_artifacts.jl`, which will walk over the artifacts defined in `Artifacts.toml`, select the ones that need to be installed, and print the hashes out to stdout.  In the example case, it would print out the hashes of the normal, platform-unspecific and special artifacts, but not for the lazy artifact.  Being unable to find an artifact would result in an error.  This will require a small patch to the `Artifacts` stdlib to give us an easy verb for "select all hashes that match this platform", but I think it should satisfy all parties.  Also, we'll probably want to encourage users to define this hook such that they can re-use their platform augmentation in their own package, so that they can invoke `@artifact_str` properly.

The only downside is the extra time required to spin up Julia, but perhaps that's just the price we'll have to pay for this functionality.  :)

We now have the opportunity to execute this hook in whatever environment we want; I kind of feel that we should just always generate a clean environment and run it in that; so it only has access to Base/stdlibs.  What do you guys think?

staticfloat:
Alright!  I've had the chance to revisit this and I think it's ready for review again.

In summary, the interface is now to run an external Julia process (with a temporary project) that invokes `.pkg/artifact_selection.jl` whose responsibility is to output a TOML representation of the artifacts that should be installed.  The format of the output matches [the new `select_downloadable_artifacts()` method](https://github.com/JuliaLang/julia/pull/38233), so most people should just be augmenting their platform object, then calling `select_downloadable_artifacts()`, then using `TOML.print()`.

For an example of how to use this, see the `test/test_packages/PlatformAugmentationHook` test package.  Most packages should mimic the structure (including the two separate files in `.pkg` so that the package itself can easily import the platform augmentation function for use with `@artifact_str` as well), and best of all this makes it easier for Pkg to get a list of all resources that should be downloaded without needing to actually download them.  This allows us to decouple resource discovery from resource acquisition, because the output of `select_downloadable_artifacts()` can directly be passed to `ensure_artifact_installed()`.

staticfloat:
This will require https://github.com/JuliaLang/julia/pull/38233 before it can be merged

KristofferC:
Just so I understand things:

- The reason for this is for e.g. CUDA to be able to install a specific artifact based on some arbitrary local property of the machine?
- If that is the case, why can't the artifact be made lazy but installed by CUDA in the build script?

docs/src/artifacts.md
```diff
@@ -227,3 +227,65 @@ Note that while that hash was previously overridden, it is no longer, and theref
...
+```
+
+```julia
+# .pkg/select_artifacts_hook.jl
+using Artifacts, TOML
+
+# Load in our platform augmentation function
+include("platform_augmentation.jl")
+
+# Select all artifacts that mattch our augmented platform and print the TOML dict
```
StefanKarpinski, line 273:
```suggestion
# Select all artifacts that match our augmented platform and print the TOML dict
```

staticfloat:
> The reason for this is for e.g. CUDA to be able to install a specific artifact based on some arbitrary local property of the machine?

Correct.

> If that is the case, why can't the artifact be made lazy but installed by CUDA in the build script? The build script can run arbitrary code.

It's true that in the near-term the only true benefit of this is that the artifacts get downloaded at `Pkg.add()` time as opposed to `using CUDA` time.  But in the long-term this could be the mechanism by which things like `CUDA` can become compatible with things like `PackageCompiler`.

From a certain perspective, this is very similar to a `deps/build.jl` script, except that its scope is drastically reduced:

* It runs in a bare environment; it doesn't require any dependencies.
* It doesn't modify any state, just outputs the artifacts that should be loaded.

This allows us to further disentangle resource selection and installation.

KristofferC:
> It's true that in the near-term the only true benefit of this is that the artifacts get downloaded at Pkg.add() time as opposed to using CUDA time.

And CUDA could just add a `build` file to have them be downloaded during `Pkg.add` time?



> But in the long-term this could be the mechanism by which things like CUDA can become compatible with things like PackageCompiler.

In what way would this be done? IIUC, CUDA is already compatible with PackageCompiler, the issue right now is that PackageCompiler bundles all lazy artifacts. If that would be changed, it would still be compatible, it would just download the artifacts lazily when the app is run. Do you envision that this script is run as an installation step on the machine the app is run on? Is there an advantage to that instead of running it at `__init__` with a message that things are being set up for first time usage?

maleadt:
> Is there an advantage to that instead of running it at `__init__` with a message that things are being set up for first time usage?

Wasn't that invalid, as I recall from an earlier compiler call?

Anyway, teaching the system about the CUDA platform has other advantages. We need this not only for CUDA.jl, but for any package that builds against the CUDA runtime (Torch.jl, CUTENSOR.jl, CUDNN.jl, SCS.jl). Also, some of these do not support the latest CUDA, so the artifact selected by Pkg for CUDA_jll.jl would need to be bounded by that. I'm not sure how that would work with a simple `build.jl` (but I'm also not certain that the current tools already allow this kind of selection)?

KristofferC:
> Anyway, teaching the system about the CUDA platform has other advantages. We need this not only for CUDA.jl, but for any package that builds against the CUDA runtime (Torch.jl, CUTENSOR.jl, CUDNN.jl, SCS.jl). 

Could that be solved by having CUDA use it's scratch space to dump whatever information it wants and have the dependent packages read that?

staticfloat:
> In what way would this be done? IIUC, CUDA is already compatible with PackageCompiler, the issue right now is that PackageCompiler bundles all lazy artifacts. If that would be changed, it would still be compatible, it would just download the artifacts lazily when the app is run.

We want to avoid using lazy artifacts for this, because we want to be able to construct a fully-functional application that does not require internet access afterwards.  We could download all possible artifacts, but that gets onerous when there are 15 different 100MB artifacts and you only need one of them.

As for using `deps/build.jl` files, I want to get to the point where we can say that if you are using `deps/build.jl` files, you're doing something wrong.  It would be much more preferable to be able to know what resources are needed without running code that traditionally mutates the local environment.  The resource selection and state mutation are separated using this PR (`select_artifacts_hook.jl` selects, then `Pkg.add`/`Pkg.instantiate` mutates), whereas with `build.jl` files it's merged (you don't know what `build.jl` wants and it's difficult to figure out what it added).

> Do you envision that this script is run as an installation step on the machine the app is run on?

No, it's run on the machine that you're building the application on, whose architecture and environment matches the one you want to run it on.  Cross-building will require a little more communication between `Pkg` (doing the installation) and this hook.  That was easy when the hook code was run in the same process as `Pkg`, but it's still possible through either environment variables or command-line arguments to the hook, to influence `Platform` construction.  I haven't decided on how to do this yet as it's still a little far off.

>  Is there an advantage to that instead of running it at `__init__` with a message that things are being set up for first time usage?

In addition to the concerns above about being able to determine what resources are needed for a package in an idempotent fashion, Jameson doesn't like things like that being run within `__init__()`.  You can dodge those concerns by just running it lazily but not in `__init__()`, so I'd say the main issue is that I think it is well worth the trouble to have the ability to declare what resources a package needs without actually installing it.

KristofferC:
Would this have to run every time one does `Pkg.instantiate`? We don't re-run e.g. `Pkg.build` wehen instantiating an already installed package.

> It would be much more preferable to be able to know what resources are needed without running code that traditionally mutates the local environment. 

CUDA etc can chose not to  do any mutation. At the end of the day, it is still an arbitrary code that is running. So what is the advantage of this vs e.g. putting a note in `Pkg.build` with something like:

>  "modifying the package directory is discouraged `Pkg.build` should be used to obtain resources (like artifacts) that are to be used by the package

One problem I see here is that this assumes that artifacts are the only thing that will be used in this manner (initialized at `Pkg.add` time). But maybe packages want to initialize something else, like a scratch space for example. Where will they put the code for that? Either in `build` or piggy-back on this new thing. That's why I think it makes sense to put all arbitrary code running into one step because otherwise we have to add another step for each new "thing" in the future.

staticfloat:
> CUDA etc can chose not to do any mutation. At the end of the day, it is still an arbitrary code that is running. So what is the advantage of this vs e.g. putting a note in Pkg.build

The advantage is both technical and cultural: the technical side is that a `deps/build.jl` script does both resource selection and resource installation.  A `.pkg/artifacts_selection_hook.jl` script does only resource selection, and resource installation is a separate step.  This makes it possible to ascertain what resources _should_ be downloaded and unpacked _before_ you actually try to download and unpack them.

The cultural side is that while we _could_ put such a disclaimer on `Pkg.build`, in practice users are already modifying their own packages and whatnot.  It is much cleaner and simpler if we instead say "never use `deps/build.jl`" and provide a new mechanism that is purposefully scoped, not to mention named in such a way as to guide users toward a better behavior.

> One problem I see here is that this assumes that artifacts are the only thing that will be used in this manner (initialized at Pkg.add time). But maybe packages want to initialize something else, like a scratch space for example.

If there is a common need for something that is not artifacts but can be done beforehand, we can add a separate hook for that kind of resource acquisition.  The fact that there might be other needs that we'll need to meet in the future shouldn't stop us from providing for needs that we know of right now, IMO.

It's kind of like saying that we don't need the `Artifacts` module because people can download files to directories on their own, and in the future users may need to modify the files they download.  That is true, but it is beneficial to users to provide to them an API that meets a specific need, which is streamlined downloading of read-only files that get stored in a consistent place.  This PR meets a specific need, and is scoped to only fulfill that need, so that it can't be abused like `deps/build.jl` is.

StefanKarpinski:
I'm not too worried about there being an unending series of hooks like this because there are only a limited set of different times when things could possibly run. This adds a hook that can affect what is installed before a package is fully installed. Are there an unlimited number of other hooks that might run at this point? I can't see how that could be the case. Artifacts, by design, are the only thing that we install in a platform-specific way and I don't see what else it would ever make sense to run at this point. Everything else that someone might want to run would have to happen after the package is fully installed, at which point we're back in `deps/build.jl` territory, no?

While this kind of selection could be done by `deps/build.jl` (because it can do anything), the advantage of this is that you can run it at any point without having actually installed anything, and on a different system than the target system even. ecause the hook is, by definition, only allowed to use Base+stdlib functions. @KristofferC, what is your alternative plan for being able to determine what artifacts a package needs when PackageCompiling it? Just ship all possible artifacts? With all the potential microarchitecture variations, that seems highly impractical since the resulting project bundles will be incredibly large. With `deps/build.jl` I don't see any alternative since the choice of which lazy artifacts are downloaded during the build phase is completely opaque and can only be determined by actually installing the package on the target machine.

That said, @staticfloat, can you clarify how these hooks would be used to target different architectures? Is there a way to feed target system profiles into these hooks so that we can get different possible configurations out?

fredrikekre:
`Pkg.instantiate` and `Pkg.add` has a `platform` kwarg, so that could be sent to `Pkg.build` too to build a package with a different target system in mind.

What is the advantage of knowing at `Pkg.add` time what artifacts you need?

StefanKarpinski:
The advantage is that you can eliminate the build step entirely for many cases. Build is, as far as I understand it, pretty much incompatible with being able to package up a complete app that can just be run on a different system. What is the plan for that? Is the plan something like this?:

1. Build everything with a different target platform
2. Record (somehow?) which lazy artifacts the build script happened to install
3. Bundle up those artifacts and the potentially modified content of packages in the app

If that's the plan, it seems like there are several big issues:

1. I'm skeptical that build scripts pay any attention to the platform argument. Maybe I'm wrong.
2. I'm also skeptical that build scripts don't do platform or environment-specific things that are completely wrong on other systems.
3. Do we really believe that the arbitrary package modifications that `deps/build.jl` can make are a good thing to bundle up and ship to another machine?

If that's the plan for building apps for other platforms, I have to say, I'm deeply unconvinced that it will work well. The first two points here mean that most build scripts won't work for this, so we have to go badger people to change their build scripts to work for cross-compilation (so to speak). As to the last point, instead of shipping a bundle of immutable, content-addressed packages and artifacts which we can perfectly identify and reproduce, who the fuck knows what we're shipping? It's whatever was left after running some arbitrary code that may have done arbitrary things based on the global state of the system that it was run on. Not only is it allowed to download whatever artifacts it wants, heck, it can download things that aren't even artifacts, and it can change the source of the package in any way it wants! 😜🤣

The point of this design is that you can run a script that tells you what artifacts you need in a controlled, managed way: you run the hook for the platform you want to target (the current one, by default), it does a bit of isolated computation and outputs information about what artifacts are needed, _which you can record_ so that you reproduce the results in the future; the hook is not allowed to download anything or modify the content of anything. No source code modification, no modification of any artifacts, no downloading of arbitrary things, none of that. It is the smallest bit of code that can be allowed into the system. The counteragument here seems to be "but `deps/build.jl` can already do that". Well, no kidding, because `deps/build.jl` can do literally _anything_, which is why it's such a problem.

fredrikekre:
> The counteragument here seems to be "but deps/build.jl can already do that". Well, no kidding, because deps/build.jl can do literally anything, which is why it's such a problem.

But couldn't packages just not do that then? They could be written such that the only thing they do is download the artifact they need. The solution here can also run arbitrary code so I guess I just don't understand the need for 2 solutions for this.

Since the solution here also requires Pkg to download and unpack the package before it can examine the hook it seems that it would happen almost at the same "time" anyway, so the solution here does not even allow for collecting resources to download in paralelle IIUC?

StefanKarpinski:
This isn't about a performance improvement. It's about two things:

1. allow packages to depend on artifacts specialized beyond the platform tuple in an extensible way
2. not require a mechanism that forces packages to be mutable in order to do so.

There's a huge difference between `deps/build.jl` and this hook:

* `deps/build.jl` already exists and it has always been allowed to do anything at all to build a package, including modifying the contents of the package directory.
* `.pkg/artifacts_selection_hook.jl` is new and while it can run arbitrary code to produce its output, it cannot modify the package source at all; the only effect it can have is to indicate which artifacts to install.

Any package that has a `deps/build.jl` cannot safely be PackageCompiled. Also, since `deps/build.jl` can modify the package in any way, we lose reproducibility. A package that only has `.pkg/artifacts_selection_hook.jl` is guaranteed to be PackageCompileable and it's perfectly reproducible because the only thing the hook can do is choose a subset of pre-defined artifacts.

StefanKarpinski:
> But couldn't packages just not do that then? They could be written such that the only thing they do is download the artifact they need. The solution here can also run arbitrary code so I guess I just don't understand the need for 2 solutions for this.

How do you know which packages are which? We have to assume that any package with `deps/build.jl` needs to be mutable and that the build script might mutate it. Any package that has this hook and no `deps/build.jl` is safe: we know that the package can be installed fully immutable and unmodified. If you install all possible artifacts then it will definitely work. If you want to know which precise subset of artifacts are needed, you can run the hook and the _only_ thing the hook can do is select a set of artifacts.

fredrikekre:
You can already achieve both 1 and 2 with `deps/build.jl` though?

> it cannot modify the package source at all; the only effect it can have is to indicate which artifacts to install.

What do you mean? It absolutely can be used to create stuff in the package itself, just something like
```julia
open(joinpath(@__DIR__, "file"), "w") do io
    print(io, "something needed for the package")
end
```

fredrikekre:
From the example in the docs here:
```
# .pkg/select_artifacts_hook.jl
using Artifacts, TOML

# Load in our platform augmentation function
include("platform_augmentation.jl")

# Select all artifacts that match our augmented platform and print the TOML dict
artifacts_toml = joinpath(dirname(@__DIR__), "Artifacts.toml")
platform = augment_platform!(HostPlatform())
artifacts = select_downloadable_artifacts(artifacts_toml; platform)
TOML.print(stdout, artifacts)
```

wouldn't it be better if Pkg did the last 2 steps and that this script only prints out the platform? I guess that wouldn't allow this script to include other artifacts that is not expressable as a platform though.

staticfloat:
> wouldn't it be better if Pkg did the last 2 steps and that this script only prints out the platform?

The issue is things like this: https://github.com/JuliaLang/julia/blob/master/base/binaryplatforms.jl#L319-L324

In the case of `CUDA`, this is easy; either an artifact is `-cuda10` or `-cuda11`, and only one will work, so exact tag matching works.  It gets more complicated if we were to do something like this for, say, microarchitectural optimization levels.

Let's say there are two different artifacts in the TOML file, one that has optimization levels `x86_64`, and `avx512`.  The other one has `x86_64` and `avx`.  The computer itself is a Haswell processor so it supports avx and avx2 but not avx512.  If you ask our platform code what the `march` tag for this host is, it will inspect the system and return `march=avx2`.  If you then blindly return a platform string that looks like `x86_64-linux-gnu-march+avx2`, that accurately describes the system, but it then will not match any of the platform strings in the `Artifacts.toml`, because by default all tags are set to compare via exact equality.  This is changed by setting an arbitrary function as a comparator for certain tags for a given Platform object, which naturally is a difficult thing to serialize (conventionally only host platforms have these special values set, so that artifact platforms can be easily serialized).

It's also not possible to cram more intelligence into the platform selection hook and just select a triplet that would perfect-equality-match the platforms in the `Artifacts.toml`; in the above example there's no single `march` tag that matches both artifacts (except the base `x86_64` level).  We'd need to emit an exact platform string for each artifact that needs to be installed, and at that point... why not just emit the artifacts themselves in their natural serialization state?  And that is precisely what this PR does.

StefanKarpinski:
I buy it. Thank you for laying out the problem and all your thinking, @staticfloat. Hopefully this is helpful for @fredrikekre and @KristofferC to understand where this feature is coming from.

staticfloat:
Windows is failing with an unrelated error; I tested on a recent checkout of `master` and all tests pass.

@KristofferC @fredrikekre have we reached agreement here, or is there more to discuss?

staticfloat:
Bump

docs/src/artifacts.md
```diff
@@ -227,3 +227,65 @@ Note that while that hash was previously overridden, it is no longer, and theref
...
+
+## Extending Platform Selection
+
+!!! compat "Julia 1.6"
+    Pkg's extended platform selection requires at least Julia 1.6, and is considered experimental.
+
+New in Julia 1.6, `Platform` objects can have extended attributes applied to them, allowing artifacts to be tagged with things such as CUDA driver version compatibility, microarchitectural compatibility, julia version compatibility and more!
+Note that this feature is considered experimental and may change in the future.
+If you as a package developer find yourself needing this feature, please get in contact with us so it can evolve for the benefit of the whole ecosystem.
+In order to support artifact selection at `Pkg.add()` time, `Pkg` will run the specially-named file `<project_root>/.pkg/select_artifacts_hook.jl` and which should output a `TOML`-serialized dictionary representing the artifacts that this package needs.
```
KristofferC, line 239: Maybe we can remove the `hook` in the file name? Or what do you think? `select_artifacts.jl` feels kinda clean.

KristofferC, line 239: Maybe change "output a ...." to "print a `TOML`-serialized dictionary to `stdout` ..."

src/Operations.jl
```diff
@@ -597,22 +598,36 @@ end
...
         # Check to see if this package has an (Julia)Artifacts.toml
         for f in artifact_names
             artifacts_toml = joinpath(path, f)
             if isfile(artifacts_toml)
-                push!(artifacts_tomls, artifacts_toml)
+                hook_path = joinpath(path, ".pkg", "select_artifacts_hook.jl")
+
+                if isfile(hook_path)
+                    mktempdir() do project
+                        meta_toml = String(read(`$(Base.julia_cmd()) --project=$(project) $(hook_path)`))
```
KristofferC, line 612: Should probably turn off the startup file here?

src/Operations.jl
```diff
@@ -597,22 +598,36 @@ end
...
+                        push!(artifacts_tomls, (artifacts_toml, TOML.parse(meta_toml)))
```
KristofferC, line 613: In the future, it would be good if we had a typed `struct` that we parse the Artifact.toml into, instead of passing around the raw dictionary from `TOML.parse`.

KristofferC, line 613:
```suggestion
                        push!(artifacts_tomls, (artifacts_toml, Base.parsed_toml(meta_toml)))
```

if that is ok.

KristofferC:
Some small comments, but looking mostly good.

src/Operations.jl
KristofferC: I don't understand the temporary `project` temp. If we include a file there with the default `LOAD_PATH`, you will be able to load stdlibs that you don't depend on.

docs/src/artifacts.md
```diff
@@ -227,3 +227,65 @@ Note that while that hash was previously overridden, it is no longer, and theref
...
+The format of the dictionary should match that returned from `Artifacts.select_downloadable_artifacts()`, and indeed most packages should simply call that function with an augmented `Platform` object.
+An example hook definition might look like the following, split across two files:
+
+```julia
+# .pkg/platform_augmentation.jl
+using Libdl, Base.BinaryPlatforms
```
KristofferC, line 245: What packages are you allowed to load in this file? Why can you load `Libdl`?

docs/src/artifacts.md
```diff
@@ -227,3 +227,65 @@ Note that while that hash was previously overridden, it is no longer, and theref
...
+    end
+
+    # Return possibly-altered `Platform` object
+    return p
+end
+```
+
+```julia
+# .pkg/select_artifacts_hook.jl
+using Artifacts, TOML
```
KristofferC, line 268: Same question about what packages you can load.

docs/src/artifacts.md
fredrikekre: Couldn't we also "reuse" the `deps` folder instead of a new `.pkg` folder?

KristofferC: I think it makes sense to look at this without thinking about `build`. I kinda like `.pkg`, it feels sort of similar to what many other projects do for aux files in a repo.

docs/src/artifacts.md
staticfloat: I assumed that it would be fine to load any standard libraries, but perhaps we should build an explicit list.  I think that `Artifacts`, `Libdl` and `TOML` are the bare minimum.

KristofferC: That doesn't seem 100% viable to me. A package can only load its dependencies, stdlibs included. Alternatively, we can say that Pkg is the package that runs this file and then Pkg needs to take on a dependency on everything that is allowed to be loaded.

staticfloat: It feels weird that it's the `Pkg` environment.  I'd rather ship a `Project.toml` within `Pkg` that has the explicit list embedded within it.

KristofferC: Yeah, thinking back on it, I agree it is weird to use the Pkg one.

src/Operations.jl
staticfloat: Should be fine; that file is only ever read once, since it's a temporary file and immediately deleted.

staticfloat: Actually, I lied.  The `meta_toml` is the TOML content itself, so there's really no benefit to using `Base.parsed_toml` here.

staticfloat:
Green CI!  What a welcome surprise.

staticfloat:
Updated to sort artifact dependencies and invoke `select_artifacts.jl` within the parent project.

src/Operations.jl
```diff
@@ -596,22 +602,37 @@ end
...
         for f in artifact_names
             artifacts_toml = joinpath(path, f)
             if isfile(artifacts_toml)
-                push!(artifacts_tomls, artifacts_toml)
+                selector_path = joinpath(path, ".pkg", "select_artifacts.jl")
+
+                if isfile(selector_path)
+                    meta_toml = String(read(```
+                        $(Base.julia_cmd()) --color=no --history-file=no -O0 --startup-file=no
+                                            --project=$(path) $(selector_path)
```
fredrikekre, line 617: This will not work unless you also instantiate this project (which we shouldn't).

fredrikekre (changes requested):

staticfloat:
Posting this here so that slack doesn't lose it:

How can I test the workflow that `Pkg.add("foo")` should install/instantiate dependencies of `foo` before running the `.pkg/select_artifacts.jl` hook?  I would have thought that we'd have a test for that for `Pkg.build()` but it does not appear that we do.

src/Operations.jl
staticfloat: I'm not sure why this won't work; in all my tests (such as the one I've added to `test/artifacts.jl`) it works just fine.  Can you help me to write a test case that fails?

staticfloat: ......huh, apparently it works locally but not on CI.  Great! :D

staticfloat:
So I think I'm going to need some help dragging this across the finish line.  I've been doing my best to mess with `sandbox()` to get it to provide the environment needed to run the hook in the proper environment, but it seems to me that invoking `sandbox()` during an `instantiate` will always infinitely recurse.

staticfloat:
Okay so Fredrik and I went through this together today, and in the end we decided that allowing for arbitrary environments gets us too close to undecidable installation regimes; if a package that uses a `select_artifacts.jl` hook wants to use another package that also uses this hook, we need an iterative installation procedure where we can't know what artifacts we need to download until other packages have downloaded their artifacts.  This makes me very uncomfortable, as I'd like to stay as far away from iterative installation procedures as possible; I want to take a single pass over the entire graph and walk away with knowledge of all resources that must be installed.

To that end, we decided to cut back and simply run the hook in a fixed environment that includes all stdlibs and nothing else.  It's possible we will in fact need something else in the future, but as it stands, we don't have any need for anything other than `Base`, `Libdl` and maybe `Sockets`, so we're going to call this good for now.

staticfloat:
This is passing tests and is, in my mind, feature complete.  If I can get a green light from someone else, I say let's finally put this to rest.

src/Operations.jl
```diff
@@ -586,12 +587,48 @@ function install_git(
     end
 end
 
+function temp_base_stdlibs_env(f::Function)
+    mktempdir() do dir
+        project_file = joinpath(dir, "Project.toml")
+        open(project_file, "w") do io
+            TOML.print(Dict(
```
fredrikekre, line 594:
```suggestion
            TOML.print(io, Dict(
```

src/Operations.jl
```diff
@@ -586,12 +587,48 @@ function install_git(
...
+    # Check to see if this package has an (Julia)Artifacts.toml
+    artifacts_tomls = Tuple{String,Dict}[]
+    for f in artifact_names
+        artifacts_toml = joinpath(pkg_root, f)
+        if isfile(artifacts_toml)
+            selector_path = joinpath(pkg_root, ".pkg", "select_artifacts.jl")
+
+            # If there is a dynamic artifact selector, run that in an appropriate sandbox to select artifacts
+            if isfile(selector_path)
+                temp_base_stdlibs_env() do project
```
fredrikekre, line 612: Instead of this, I think it is better to set `JULIA_LOAD_PATH` in the new process to `"@stdlib"`.

src/Operations.jl
```diff
@@ -586,12 +587,48 @@ function install_git(
...
+                    meta_toml = String(read(`$(gen_build_code(selector_path)) --project=$(project) $(triplet(platform))`))
```
fredrikekre, line 613:
```suggestion
                    meta_toml = String(read(`$(gen_build_code(selector_path)) $(triplet(platform))`))
```

src/Operations.jl
fredrikekre: (That is effectively what was used in the tests here anyway since your TOML.print went to stdout)

staticfloat: That..... makes sense

src/Operations.jl
```diff
@@ -586,12 +587,36 @@ function install_git(
...
+    # Check to see if this package has an (Julia)Artifacts.toml
+    artifacts_tomls = Tuple{String,Dict}[]
+    for f in artifact_names
+        artifacts_toml = joinpath(pkg_root, f)
+        if isfile(artifacts_toml)
+            selector_path = joinpath(pkg_root, ".pkg", "select_artifacts.jl")
+
+            # If there is a dynamic artifact selector, run that in an appropriate sandbox to select artifacts
+            if isfile(selector_path)
+                select_cmd = Cmd(`$(gen_build_code(selector_path)) --project=$(project) $(triplet(platform))`)
```
fredrikekre, line 600: This doesn't actually do anything with this load path, and code should not be using this env. `project` also seems undefined?
```suggestion
                select_cmd = Cmd(`$(gen_build_code(selector_path)) $(triplet(platform))`)
```

KristofferC (approved):

staticfloat:
Woohoo!! Thanks guys!

docs/src/artifacts.md
```diff
@@ -227,3 +227,77 @@ Note that while that hash was previously overridden, it is no longer, and theref
 Most methods that are affected by overrides have the ability to ignore overrides by setting `honor_overrides=false` as a keyword argument within them.
 For UUID/name based overrides to work, `Artifacts.toml` files must be loaded with the knowledge of the UUID of the loading package.
 This is deduced automatically by the `artifacts""` string macro, however if you are for some reason manually using the `Pkg.Artifacts` API within your package and you wish to honor overrides, you must provide the package UUID to API calls like `artifact_meta()` and `ensure_artifact_installed()` via the `pkg_uuid` keyword argument.
+
+## Extending Platform Selection
+
+!!! compat "Julia 1.6"
```
vchuravy, line 233: The docs say 1.6, but I don't think this was ever backported?

PR #2031, opened 2020-09-21
Add Julia version-independent resolution capability

staticfloat:
In the BinaryBuilder world, we use `Pkg.add()` to install dependent JLLs
in the build prefix.  By relying on the Pkg resolver, we implicitly
limit ourselves to the package versions that are compatible with the
currently-running Julia.  This prevents us from, for example, using
Julia 1.6 in BB to generate packages that link against the binaries used
by Julia 1.5, or from constructing "impossible" deps trees (from Julia's
perspective, that is) where we want to build something for Julia 1.6
that was compiled against GMP 6.1.2; although the ABIs are compatible,
Julia's resolver must be coerced to allow it.

This PR adds a `julia_version` keyword argument to all the methods from
`add()` down to the resolver.  `julia_version` defaults to `VERSION`,
but if it is given another version number, it resolves with respect to
that Julia version.  As the final word in any resolutionary argument, if
`julia_version` is set to `nothing`, then Julia compatibility bounds are
ignored completely, and only the `PackageSpec` bounds on other packaegs
are considered at all.

Usage of this faculty is not recommended for typical Julia package
installation.


KristofferC:
Makes a lot of sense to me to have this. The only question is if all operations should accept this. In that case, it might be simpler to just make it a field in the Context struct so that automatically all functions accept that as a keyword argument.

staticfloat:
Hmmm, yeah, I could see that.  It would be good to ensure that it's consistent where it needs to be.

KristofferC:
There's another thing that Pkg uses the current running Julia version for and that is to get the list of all the stdlibs and their dependencies. That also has an effect on the end result of the resolver. So if you want to truly get the same result as when running with the given `julia_version` then that has to be dealt with as well.

staticfloat:
> There's another thing that Pkg uses the current running Julia version for and that is to get the list of all the stdlibs and their dependencies

In my case I don't care about stdlibs, but assuming I did, how would I do that?  We don't really have a list of stdlibs by version.  :(

I've updated this PR to store `julia_version` within the `Context`; if you have a good idea on how to do the stdlib checking, I'm all ears, otherwise IMO we should just merge this and address that once we have a usecase for it.  Perhaps by collecting a dictionary of UUIDs per Julia version.

src/Types.jl
```diff
@@ -328,7 +328,9 @@ Base.@kwdef mutable struct Context
     use_only_tarballs_for_downloads::Bool = false
     num_concurrent_downloads::Int = 8
     graph_verbose::Bool = false
-    currently_running_target::Bool = false
```
staticfloat, line 331: I didn't want this to be confused with the `julia_version` since they sound almost similar; and since this flag doesn't seem to be used anymore (since we have dropped support for the Pkg 2 feature that used this) it seemed fine to remove it.

src/Resolve/graphtype.jl
```diff
@@ -241,21 +241,26 @@ mutable struct Graph
...
-        versions[uuid_julia] = Set([VERSION])
-        deps[uuid_julia] = DepsValDict(VERSION => valtype(DepsValDict)())
-        compat[uuid_julia] = CompatValDict(VERSION => valtype(CompatValDict)())
+        if julia_version !== nothing
+            fixed[uuid_julia] = Fixed(julia_version)
+            versions[uuid_julia] = Set([julia_version])
+            deps[uuid_julia] = DepsValDict(julia_version => valtype(DepsValDict)())
+            compat[uuid_julia] = CompatValDict(julia_version => valtype(CompatValDict)())
+        else
+            versions[uuid_julia] = Set([])
```
KristofferC, line 260: I don't really understand how this works. If we say there are no versions of Julia available, why isn't the resolver complaining about that when it resolves packages that have a dependency on Julia?

src/Resolve/graphtype.jl
```diff
@@ -241,21 +241,26 @@ mutable struct Graph
...
+        end
 
-        extra_uuids = union(keys(reqs), union(keys(fixed), map(fx->keys(fx.requires), values(fixed))...))
+        extra_uuids = union(collect(keys(reqs)), union(collect(keys(fixed)), map(fx->keys(fx.requires), values(fixed))...))
```
KristofferC, line 263: Out of curiosity, what made this needed?

KristofferC (approved):
Can deal with the stdlib stuff later if needed

src/Resolve/graphtype.jl
staticfloat: If `fixed` is an empty array, then some codepath of `union()` ends up calling `copy(keys(fixed))`, and it complains that there is no `copy(::KeySet{...})` method.  You can recreate the fun yourself with the following snippet:


```
julia> using Pkg
       fixed = Dict{Base.UUID,Pkg.Resolve.Fixed}()
       union(keys(fixed))
ERROR: MethodError: no method matching copy(::Base.KeySet{Base.UUID,Dict{Base.UUID,Pkg.Resolve.Fixed}})
Closest candidates are:
  copy(::Pkg.Resolve.FieldValue) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Pkg/src/Resolve/fieldvalues.jl:63
  copy(::Random._GLOBAL_RNG) at /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.5/Random/src/RNGs.jl:325
  copy(::Expr) at expr.jl:35
  ...
Stacktrace:
 [1] union(::Base.KeySet{Base.UUID,Dict{Base.UUID,Pkg.Resolve.Fixed}}) at ./abstractset.jl:49
 [2] top-level scope at REPL[12]:3
```

This is happening because the splat has zero elements.  This never happened before, because the `julia` entry ensured that at least one fixed element always existed.

src/Resolve/graphtype.jl
staticfloat: TBH I don't know.  I wouldn't have done this at all, except that it broke when I didn't do this, so I initialized it to the empty set and it works.

src/Operations.jl
```diff
@@ -328,9 +328,11 @@ end
 # all versioned packges should have a `tree_hash`
 function resolve_versions!(ctx::Context, pkgs::Vector{PackageSpec})
     # compatibility
-    v = intersect(VERSION, project_compatibility(ctx, "julia"))
-    if isempty(v)
-        @warn "julia version requirement for project not satisfied" _module=nothing _file=nothing
+    if ctx.julia_version !== nothing
```
KristofferC, line 331: I think this means that the julia compatibility of the current project is ignored when running with `julia_version=nothing`. I think this needs to be passed to the resolver so that the julia compatibility is respected.

staticfloat: Can you explain why you think that is a bad idea?  The whole reasoning behind `julia_version = nothing` is for it to ignore all Julia version constraints during resolution.  This is what allows us to, for instance, resolve versions of JLLs that obey the compat constraints between stdlib JLLs, but ignoring the Julia version constraints.

KristofferC: Okay I understand. I thought it would treat julia as a normal package (with a non-fixed version) and the resolver would give back the most recent julia version that satisfies all the julia constraints of the packages that end up in the environment. But for that, the set of available julia versions has to be available which it currently is not. If https://github.com/JuliaRegistries/General/blob/master/J/julia/Versions.toml is updated then this could be done.

staticfloat: Hmmm, that’s interesting. In what situation would you want that?

KristofferC: It's been discussed in the context of juliaup in the situation where you, for example, have a project file (but no manifest) and want to launch the most recent julia that also satisfies your resolved project. You would then "bootstrap" this by running some julia version that resolves the environment with the julia version "floating" and then after, find out what julia version was actually resolved to and launch that.

staticfloat: Maybe we set julia_version to “missing”, heh.

PR #2033, opened 2020-09-23
Add opt-in auto precompilation after Pkg.add, instantiate, update

ianshmean:
Runs `Pkg.precompile()` after `Pkg.add()`, `instantiate()`, `update()`

Enabled by setting `ENV["JULIA_PKG_PRECOMPILE_AUTO"]=1` etc.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -186,6 +186,7 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PR
     end
 
     Operations.add(ctx, pkgs, new_git; preserve=preserve, platform=platform)
+    _do_auto_precompile() && Pkg.precompile()
     return
 end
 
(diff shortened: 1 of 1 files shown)


staticfloat:
It's going to be a string; you want something like `if parse(Int, get(ENV, "JULIA_PKG_PRECOMPILE_AUTO", "0")) == 1`

ianshmean:
What would be best? I didn't mean to pick a strategy.

staticfloat (approved):

tkf:
Setting environment variables is not the cleanest way to configure the behavior of a package https://github.com/JuliaLang/julia/issues/34726. Why not use (run-time) Preferences.jl?

KristofferC:
This is just a way for people to be able to try it out to collect feedback. It is not a permanent setting.

PR #2034, opened 2020-09-23
follow up #2018:

aviatesk:
UUIDs are not in `keys` but `values` of `Project.deps::Dict{String,UUID}`

/cc @ianshmean

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -895,27 +895,28 @@ end
 precompile() = precompile(Context())
 function precompile(ctx::Context)
     printpkgstyle(ctx, :Precompiling, "project...")
-    
+
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
     parallel_limiter = Base.Semaphore(num_tasks)
-    
+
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -895,27 +895,28 @@ end
...
+
     if ctx.env.pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", ctx.env.pkg.name * ".jl") )
         push!(pkgids, Base.PkgId(ctx.env.pkg.uuid, ctx.env.pkg.name))
-        push!(pkg_dep_uuid_lists, collect(keys(ctx.env.project.deps)))
-    end    
-    
+        push!(pkg_dep_uuid_lists, collect(values(ctx.env.project.deps)))
+    end
+
+    filter!.(!is_stdlib, pkg_dep_uuid_lists)
```
aviatesk, line 911: the actual diffs are these 4 lines

ianshmean:
Thanks. Already fixed by https://github.com/JuliaLang/Pkg.jl/pull/2021 which does things a bit more properly

aviatesk:
aha, I just missed that. Please keep focus on that ;)

PR #2037, opened 2020-09-23
Clarify activate command signatures

danielmatz:
The signatures in the help for the activate command were confusing. It seemed to imply that the `activate --shared` and `activate --temp` each took an optional path, when in fact `activate --shared` requires a path and `activate --temp` does not accept a path. This clarifies the command signatures for these two cases.

### DIFF
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -249,7 +249,8 @@ PSA[:name => "activate",
     :description => "set the primary environment the package manager manipulates",
     :help => md"""
     activate
-    activate [--shared|--temp] [path]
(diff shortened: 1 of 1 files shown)


fredrikekre (approved):

StefanKarpinski:
I restarted the doc build; the macOS Travis failures are unrelated (due to PCRE upgrade).

00vareladavid:
Is the best way to restart tests still to close then open the PR?

00vareladavid:
I'm merging because macOS tests are all green, not sure why this is still pointing to Travis tests (seems we do not use it at all anymore)

PR #2039, opened 2020-09-24
Launch opt-in auto precompilation from appropriate gen funcs

ianshmean:
I wasn't familiar with the generated Pkg.API functions when doing #2033 

I think this is actually the correct launch point.

Works as expected for:
```
pkg> add Foo
pkg> pin Foo
pkg> free Foo
pkg> update
pkg> update Foo
pkg> build
pkg> build Foo
pkg> instantiate
```

Activated with `ENV["JULIA_PKG_PRECOMPILE_AUTO"]=1`

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -64,7 +64,10 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
     @eval begin
         $f(pkg::Union{AbstractString, PackageSpec}; kwargs...) = $f([pkg]; kwargs...)
         $f(pkgs::Vector{<:AbstractString}; kwargs...)          = $f([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
-        $f(pkgs::Vector{PackageSpec}; kwargs...)               = $f(Context(), pkgs; kwargs...)
+        function $f(pkgs::Vector{PackageSpec}; kwargs...)
+            $f(Context(), pkgs; kwargs...)
+            ($(f in (:add, :up, :pin, :free, :build)) && _do_auto_precompile()) && Pkg.precompile()
+        end
         $f(ctx::Context; kwargs...) = $f(ctx, PackageSpec[]; kwargs...)
         function $f(; name::Union{Nothing,AbstractString}=nothing, uuid::Union{Nothing,String,UUID}=nothing,
                       version::Union{VersionNumber, String, VersionSpec, Nothing}=nothing,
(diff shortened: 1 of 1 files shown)


fredrikekre:
But this misses the case when you call the functions with `ctx` directly?

KristofferC:
True, like from the REPL?

ianshmean:
In my testing it works for the direct equivalents of above `Pkg.add("Foo")`, `Pkg.update()` etc. 

What's an example of the case you mean @fredrikekre ?

Has anyone else given this a go? Interested to see what others think

KristofferC:
Okay, I don't think the REPL mode passes the `Context` (https://github.com/JuliaLang/Pkg.jl/blob/5bf9e6a4b28fe0e8abe9a6325528b1e7be62d56b/src/REPLMode/REPLMode.jl#L400). And if you pass a context yourself you are doing pretty advanced stuff so perhaps not needed there.

fredrikekre:
> And if you pass a context yourself you are doing pretty advanced stuff so perhaps not needed there.

Maybe not. In any case, isn't it better to do it locally in each function as before since you there know if you should precompile or not. Now we do it blindly always.

ianshmean:
Where they were launched from before caused unexpected invocations. Like during a `] add Foo` the one in `up` was called, as well as the one in `add`

Plus, the no-op is pretty fast now. But I guess it is always printing the `precompiling project...` message.. I did wonder if we could detect the no-op and not print the message in that case.

ianshmean:
Or maybe it's reassuring to know precomp has been covered each time?

fredrikekre:
> Like during a ] add Foo the one in up was called

Why? Are we calling the toplevel Pkg.up from Pkg.add internals?

> Or maybe it's reassuring to know precomp has been covered each time?

If you have this option enabled you already know that in some sense, and it looks a bit strange if e.g. `instantate` or `add` doesn't do anything but we still precompile:

```
(Pkg) pkg> instantiate
Precompiling project...

(Pkg) pkg> add Example
  Resolving package versions...
No Changes to `~/dev/Pkg/Project.toml`
No Changes to `~/dev/Pkg/Manifest.toml`
Precompiling project...
```

ianshmean:
I think `precompile` is the best judge of whether it needs to do anything, so I think the thing is to only print that message if any packages get precomped.

ianshmean:
Made that change here https://github.com/JuliaLang/Pkg.jl/pull/2043

Issue #2041, opened 2020-09-24
Things auto precompile over and over with no changes to manifest


KristofferC:
```
(@v1.6) pkg> add StatsBase#master
    Cloning git-repo `https://github.com/JuliaStats/StatsBase.jl.git`
   Updating git-repo `https://github.com/JuliaStats/StatsBase.jl.git`
  Resolving package versions...
Updating `~/.julia/environments/v1.6/Project.toml`
  [2913bbd2] ~ StatsBase v0.33.1 ⇒ v0.33.1 `https://github.com/JuliaStats/StatsBase.jl.git#master`
Updating `~/.julia/environments/v1.6/Manifest.toml`
  [2913bbd2] ~ StatsBase v0.33.1 ⇒ v0.33.1 `https://github.com/JuliaStats/StatsBase.jl.git#master`
Precompiling project...
[ Info: Precompiling StatsBase [2913bbd2-ae8a-5f71-8c99-4fb6c76f3a91]
[ Info: Precompiling ImageCore [a09fc81d-aa75-5fe9-8630-4744c3626534]
[ Info: Precompiling ColorVectorSpace [c3611d14-8923-5661-9e6a-0046d554d3a4]
[ Info: Precompiling Distributions [31c24e10-a181-5473-b8eb-7969acd0382f]
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
[ Info: Precompiling FreeTypeAbstraction [663a7486-cb36-511b-a19d-713bb74d65c9]
[ Info: Precompiling KernelDensity [5ab0869b-81aa-558d-bb23-cbf5423bbe9b]
┌ Warning: Precompilation failed for indirect dependency ImageCore [a09fc81d-aa75-5fe9-8630-4744c3626534]
└ @ Pkg.API ~/JuliaPkgs/Pkg.jl/src/API.jl:978
[ Info: Precompiling PNGFiles [f57f5aa1-a3ce-4bc8-8ab9-96f992907883]
┌ Warning: Precompilation failed for indirect dependency PNGFiles [f57f5aa1-a3ce-4bc8-8ab9-96f992907883]
└ @ Pkg.API ~/JuliaPkgs/Pkg.jl/src/API.jl:978
[ Info: Precompiling ImageIO [82e4d734-157c-48bb-816b-45c225c6df19]
[ Info: Precompiling AbstractPlotting [537997a7-5e4e-5d89-9595-2241ea00577e]
[...]

PR #2043, opened 2020-09-24
Precomp Tweaks: Don't print precomp message when no-op + Suspend precomp on pkgs that have errored, until they change

ianshmean:
Don't print precomp message when precompile is a no-op.

Makes for a cleaner experience when precompilation is auto-enabled

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -65,8 +65,9 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
         $f(pkg::Union{AbstractString, PackageSpec}; kwargs...) = $f([pkg]; kwargs...)
         $f(pkgs::Vector{<:AbstractString}; kwargs...)          = $f([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
         function $f(pkgs::Vector{PackageSpec}; kwargs...)
-            $f(Context(), pkgs; kwargs...)
-            ($(f in (:add, :up, :pin, :free, :build)) && _do_auto_precompile()) && Pkg.precompile(true)
+            ret = $f(Context(), pkgs; kwargs...)
+            $(f in (:add, :up, :pin, :free, :build)) && _auto_precompile()
+            return ret
         end
         $f(ctx::Context; kwargs...) = $f(ctx, PackageSpec[]; kwargs...)
         function $f(; name::Union{Nothing,AbstractString}=nothing, uuid::Union{Nothing,String,UUID}=nothing,
@@ -903,12 +904,19 @@ function _is_stale(paths::Vector{String}, sourcepath::String, toml_c::Base.TOMLC
     return true
 end
 
-_do_auto_precompile() = parse(Int, get(ENV, "JULIA_PKG_PRECOMPILE_AUTO", "0")) == 1
+function _auto_precompile()
+    if parse(Int, get(ENV, "JULIA_PKG_PRECOMPILE_AUTO", "0")) == 1
+        Pkg.precompile(internal_call=true)
+    end
+end
 
-precompile(auto::Bool=false) = precompile(Context(), auto)
-function precompile(ctx::Context, auto::Bool=false)    
+precompile(;internal_call::Bool=false) = precompile(Context(), internal_call=internal_call)
+function precompile(ctx::Context; internal_call::Bool=false)    
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
     parallel_limiter = Base.Semaphore(num_tasks)
+    
+    # when manually called, unsuspend all packages that were suspended due to precomp errors
+    !internal_call && Operations.precomp_unsuspend!() 
     
     direct_deps = [
         Base.PkgId(uuid, name) 
@@ -938,7 +946,6 @@ function precompile(ctx::Context, auto::Bool=false)
         was_recompiled[pkgid] = false
     end
     
-    msg_printed = false
     errored = false
     toml_c = Base.TOMLCache()
     @sync for (pkg, deps) in depsmap
@@ -958,8 +965,7 @@ function precompile(ctx::Context, auto::Bool=false)
             
             # skip stale checking and force compilation if any dep was recompiled in this session
             any_dep_recompiled = any(map(dep->was_recompiled[dep], deps))
-            #              only observe suspension state if auto-invoked
-            if !errored && (!auto || !Operations.precomp_suspended(pkg)) && (any_dep_recompiled || _is_stale(paths, sourcepath, toml_c))
+            if !errored && !Operations.precomp_suspended(pkg) && (any_dep_recompiled || _is_stale(paths, sourcepath, toml_c))
                 
                 Base.acquire(parallel_limiter)
                 if errored # catch things queued before error occurred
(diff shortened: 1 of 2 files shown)


ianshmean:
Now includes a candidate fix for #2041, where packages that have errored during precomp were repeatedly being attempted by the auto precomp. 

Now packages that fail precomp are excluded from precomp until they change in the Manifest 

cc. @KristofferC

ianshmean:
Updated now so that manually calling `pkg> precompile` will still attempt full precompilation. 
The suspensions of packages that errored during precomp are only observed during auto precomp.

```
(dev) pkg> st
Status `~/Documents/dev/Project.toml` (empty project)

(dev) pkg> add Plots
   Updating git-repo `https://github.com/JuliaRegistries/General`
  Resolving package versions...
Updating `~/Documents/dev/Project.toml`
  [91a5bcdd] + Plots v1.6.6
Updating `~/Documents/dev/Manifest.toml`
  [79e6a3ab] + Adapt v2.2.0
  [6e34b625] + Bzip2_jll v1.0.6+4
...
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
Precompiling project...
[ Info: Precompiling HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
┌ Warning: Precompilation failed for indirect dependency HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
└ @ Pkg.API ~/Documents/src/API.jl:984
[ Info: Precompiling GR [28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71]

(dev) pkg> up
   Updating git-repo `https://github.com/JuliaRegistries/General`
No Changes to `~/Documents/dev/Project.toml`
No Changes to `~/Documents/dev/Manifest.toml`

(dev) pkg> precompile
Precompiling project...
[ Info: Precompiling HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
┌ Warning: Precompilation failed for indirect dependency HTTP [cd3eb016-35fb-5094-929b-558a96fad6f3]
└ @ Pkg.API ~/Documents/src/API.jl:984
[ Info: Precompiling GR [28b8d3ca-fb5f-59d9-8090-bfdbd6d07a71]

(dev) pkg>
```

src/API.jl
```diff
@@ -66,7 +66,7 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
         $f(pkgs::Vector{<:AbstractString}; kwargs...)          = $f([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
         function $f(pkgs::Vector{PackageSpec}; kwargs...)
             $f(Context(), pkgs; kwargs...)
-            ($(f in (:add, :up, :pin, :free, :build)) && _do_auto_precompile()) && Pkg.precompile()
+            ($(f in (:add, :up, :pin, :free, :build)) && _do_auto_precompile()) && Pkg.precompile(true)
```
KristofferC, line 69: By the way, I think this should still return the value of `f` (which I guess is `nothing`)?

src/API.jl
```diff
@@ -905,10 +905,8 @@ end
 
 _do_auto_precompile() = parse(Int, get(ENV, "JULIA_PKG_PRECOMPILE_AUTO", "0")) == 1
 
-precompile() = precompile(Context())
-function precompile(ctx::Context)
-    printpkgstyle(ctx, :Precompiling, "project...")
-    
+precompile(auto::Bool=false) = precompile(Context(), auto)
```
fredrikekre, line 908: Maybe make this a kwarg? And maybe a more "scary internal name" like `internal_clall` or something?

src/API.jl
```diff
@@ -940,6 +938,7 @@ function precompile(ctx::Context)
         was_recompiled[pkgid] = false
     end
     
+    msg_printed = false
```
fredrikekre, line 941: There is no race condition to worry about here if multiple tasks modify this value?

src/API.jl
```diff
@@ -1036,7 +1042,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     Operations.download_artifacts(ctx, [dirname(ctx.env.manifest_file)]; platform=platform, verbose=verbose)
     # check if all source code and artifacts are downloaded to exit early
     if Operations.is_instantiated(ctx) 
-        _do_auto_precompile() && Pkg.precompile()
+        _do_auto_precompile() && Pkg.precompile(true)
```
fredrikekre, line 1045: The call to `precompile` could be moved in to the `_do_auto_precompile` function (I know this isn't from this PR but just something I noted).

src/Operations.jl
```diff
@@ -18,6 +18,12 @@ import ...Pkg: pkg_server
 #########
 # Utils #
 #########
+
+const pkgs_precompile_suspended = Base.PkgId[]
```
fredrikekre, line 22: Doesn't this need to be `empty!`d somewhere?

src/API.jl
ianshmean: At this point the tasks are all async, so on the same thread. So I don’t believe so

src/Operations.jl
ianshmean: Specific packages are filtered out in the manifest update method. I'm not sure what action would warrant fully emptying it. Maybe at the start of a manual precomp invoke, which could serve as the means to ignore the suspension

src/API.jl
ianshmean: Fixed

src/API.jl
ianshmean: Done

src/API.jl
ianshmean: I realized I can just check for `!any(values(was_recompiled))` so I removed `msg_printed`. Should be just as ok, with less code

src/API.jl
ianshmean: Done

src/Operations.jl
ianshmean: This empties `pkgs_precompile_suspended` as a means to retry precomp during manual calls

fredrikekre: Perhaps I am misunderstanding what this is used for, but if a package fails to precompile, why should I have to restart Julia in order to precompile it again (perhaps at a different version)?

src/Operations.jl
```diff
@@ -129,6 +136,7 @@ function update_manifest!(ctx::Context, pkgs::Vector{PackageSpec}, deps_map)
             entry.deps = deps_map[pkg.uuid]
         end
         ctx.env.manifest[pkg.uuid] = entry
+        precomp_unsuspend!(pkg)
```
ianshmean, line 139: @fredrikekre this is where it's unsuspended. If any change is made to that package in the manifest

src/Operations.jl
ianshmean: See line 139. The un-suspension happens when the package changes in the manifest

fredrikekre: Thanks. Wouldn't it be easier to just empty it everytime we enter `precompile`?

ianshmean: That's what I just changed it to, but only in manual enters. 

We always enter if auto is enabled, the suspension is observed on a per-pkg basis, so we only want to empty! for the manual call.

fredrikekre: But consider if `Example@1.2.0` fails precompilation, and I add `Example@1.2.1`. Doesn't this mechanism prevent that from precompiling?

fredrikekre: Nevermind, you do empty it completely now.

ianshmean: > Doesn't this mechanism prevent that from precompiling?

It's not the emptying that makes this work, it's that if you did change that package, it will be updated in the manifest, which will unsuspend that specific package.

KristofferC:
Good to go?

ianshmean:
Yep

PR #2049, opened 2020-09-29
Improve inference in string-related operations

timholy:


timholy:
Error looks maybe like a different variant of #2035?

ianshmean:
Re. the error, there was a change in `Base.compilecache` args. Fixed by #2048

src/REPLMode/completions.jl
```diff
@@ -30,13 +30,14 @@ function complete_local_dir(s, i1, i2)
...
+end
 
+function complete_expanded_local_dir(s, i1, i2)
     cmp = REPL.REPLCompletions.complete_path(s, i2)
     cmp2 = cmp[2]
     completions = [REPL.REPLCompletions.completion_text(p) for p in cmp[1]]
-    completions = let s=s
-        filter!(x -> isdir(s[1:prevind(s, first(cmp2)-i1+1)]*x), completions)
-    end
+    completions = filter!(x -> isdir(s[1:prevind(s, first(cmp2)-i1+1)]*x), completions)
```
KristofferC, line 40: Looks like `expanded_user` also needs to be passed down.

Issue #2057, opened 2020-09-30
Precompile deadlock


timholy:
While it's entirely possible I'm running a broken branch (I have #2049, plus https://github.com/JuliaLang/julia/pull/37754), I'm getting deadlocks presumably due to the wonderful new parallel precompilation:

```julia
(@v1.6) pkg> precompile
PrecompilingPrecompilingPrecompilingPrecompilingPrecompiling project...
┌ Warning: Precompilation failed for indirect dependency ExprTools [e2ba6199-217a-4e67-a87a-7c52f15ade04]
└ @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:987
 project...┌ Warning: Precompilation failed for indirect dependency Mocking [78c3b35d-d492-501b-9361-3d52fe80e533]
└ @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:987

┌ Warning: Precompilation failed for indirect dependency ISVD [0de75c66-c0d4-11e8-3de9-d1632eed9267]
└ @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:987
┌ Warning: Precompilation failed for indirect dependency Opus_jll [91d4177d-7536-5919-b921-800302f37372]
└ @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:987
 project...┌ Warning: Precompilation failed for indirect dependency StableRNGs [860ef19b-820b-49d6-a774-d7a799459cd3]
└ @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:987
┌ Warning: Precompilation failed for indirect dependency PCRE_jll [2f80f16e-611a-54ab-bc61-aa92de5b98fc]
└ @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:987

┌ Warning: Precompilation failed for indirect dependency Xorg_libpthread_stubs_jll [14d82f49-176c-5ed1-bb49-ad3f5cbd8c74]
└ @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:987
[...]

PR #2058, opened 2020-09-30
Precompilation: Try adding failsafe against deadlocks

ianshmean:
As discussed in #2057 it'd be good if deadlocks weren't possible.

This does two things:
- Assert that all deps of the entries in a manifest are themselves entries in the Manifest (if not, that would've deadlocked)
- As a more general unknown deadlock avoidance, allow each task to skip waiting for a dep if the precomp tasks have been idle for >~250ms

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -950,6 +950,7 @@ function precompile(ctx::Context; internal_call::Bool=false)
     errored = false
     toml_c = Base.TOMLCache()
     @sync for (pkg, deps) in depsmap
+        @assert all(map(dep->dep in keys(depsmap), deps)) "$pkg has dependencies that are missing from the manifest of the current environment"
         paths = Base.find_all_in_cache_path(pkg)
         sourcepath = Base.locate_package(pkg, toml_c)
         sourcepath === nothing && continue
@@ -961,14 +962,10 @@ function precompile(ctx::Context; internal_call::Bool=false)
         
         @async begin
             for dep in deps # wait for deps to finish
-                not_doing_anything_elapsed = 0
-                while was_processed[dep].set == false && not_doing_anything_elapsed < 0.25 # failsafe against deadlocks
+                idle_elapsed = 0
+                while was_processed[dep].set == false && idle_elapsed < 0.25 # failsafe against deadlocks
                     sleep(0.01)
-                    if parallel_limiter.curr_cnt == 0
-                        not_doing_anything_elapsed += 0.01
-                    else
-                        not_doing_anything_elapsed = 0
-                    end
+                    idle_elapsed = parallel_limiter.curr_cnt == 0 ? idle_elapsed + 0.01 : 0
                 end
             end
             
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -939,29 +939,34 @@ function precompile(ctx::Context; internal_call::Bool=false)
         ]
     end    
     
-    was_processed = Dict{Base.PkgId,Base.Event}()
+    was_processed = Dict{Base.PkgId,Bool}()
```
ianshmean, line 942: With this PR, `was_processed` Events are just being used for their underlying Bool, so I just switched it to a Bool

ianshmean:
I tried deleting an `ColorTypes` from the env Manifest that another package depended on, and it was caught by `Context()`, so the `@assert`  isn't needed

```
julia> Pkg.precompile()
ERROR: `ColorVectorSpace=c3611d14-8923-5661-9e6a-0046d554d3a4` depends on `ColorTypes`, but no such entry exists in the manifest.
Stacktrace:
  [1] pkgerror(::String, ::Vararg{String, N} where N)
    @ Pkg.Types ~/Documents/GitHub/Pkg.jl/src/Types.jl:52
  [2] normalize_deps(name::String, uuid::Base.UUID, deps::Vector{String}, manifest::Dict{String, Vector{Pkg.Types.Stage1}})
    @ Pkg.Types ~/Documents/GitHub/Pkg.jl/src/manifest.jl:95
  [3] validate_manifest(stage1::Dict{String, Vector{Pkg.Types.Stage1}})
    @ Pkg.Types ~/Documents/GitHub/Pkg.jl/src/manifest.jl:109
  [4] Dict{Base.UUID, Pkg.Types.PackageEntry}(raw::Dict{String, Any})
    @ Pkg.Types ~/Documents/GitHub/Pkg.jl/src/manifest.jl:158
  [5] read_manifest(f_or_io::String)
    @ Pkg.Types ~/Documents/GitHub/Pkg.jl/src/manifest.jl:170
  [6] Pkg.Types.EnvCache(env::Nothing)
    @ Pkg.Types ~/Documents/GitHub/Pkg.jl/src/Types.jl:294
  [7] EnvCache
    @ ~/Documents/GitHub/Pkg.jl/src/Types.jl:274 [inlined]
  [8] Pkg.Types.Context()
    @ Pkg.Types ./util.jl:442
  [9] #precompile#195
    @ ~/Documents/GitHub/Pkg.jl/src/API.jl:913 [inlined]
 [10] precompile
    @ ~/Documents/GitHub/Pkg.jl/src/API.jl:913 [inlined]
```

I'm also going off the approach in this PR.. it does away with a fair amount of elegance, and increases the no-op about 10-20%..
I'm going to mark this draft and try to figure out how the #2057 deadlock is happening..

ianshmean:
Not needed anymore. `precompile` guards against circular deps, and we haven't seen any other deadlocks being reported

PR #1807, opened 2020-05-04
More invalidation-related fixes

timholy:
This is the last piece of https://github.com/JuliaLang/julia/pull/35714

PR #2062, opened 2020-10-01
PlatformEngines: stop disabling logging during symlink test

### LINKED ISSUE #2061: Pkg should not mutate the logger
Calling `Pkg.PlatformEngines.probe_platform_engines!` (probably `probe_symlink_creation` more specifically) results in the Logging subsystem getting mutated, causing issues with subsequent attempts to log messages. This is the cause of https://github.com/JuliaLang/julia/issues/37731

StefanKarpinski:
Closes #2061. Should be followed up by changing `symlink` not to
warn but only throw an error (whose error message can include a
warning to hint that symlinks don't work on Windows).


KristofferC:
So I think https://github.com/JuliaLang/Pkg.jl/pull/1807 by maybe storing something into the precompiled state that wasn't allowed.

StefanKarpinski:
This PR fixes it in any case 😁

fredrikekre:
Looks like Pkg's dependency on Logging can be deleted after this.

Project.toml
```diff
@@ -1,5 +1,5 @@
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b780"
```
fredrikekre, line 2: This should be reverted.

StefanKarpinski: Hah. Funny that our CI doesn't catch that.

KristofferC: Woow.

PR #2063, opened 2020-10-01
Precompilation: detect and suspend precomp for circular deps

ianshmean:
Circular deps could deadlock the parallel `Pkg.precompile` process, which may be what was happening in #2057 

This adds a lightweight check to detect and suspend any circular deps within the full dep tree (not just first level, as in the case below)
Warnings are only printed only during  explicit `pkg> precompile` calls, not when `precompile` is automatically called.

These new checks added ~0.006 seconds in my case, for an environment with 333 dependencies in the manifest, which was insignificant even in the no-op case.

```
(@v1.6) pkg> precompile
┌ Warning: Circular dependency detected. Precompilation skipped for CircularDep2 [206312f7-dc9c-4d80-acd8-43a486edfa59]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:959
┌ Warning: Circular dependency detected. Precompilation skipped for CircularDep1 [5f60ae3d-344a-4690-ba7b-6488d0e2fe2e]
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:959
Precompiling project...
[ Info: Precompiling Roots [f2b01f46-fcfa-551c-844a-d8ac1e96c665]
[ Info: Precompiling Opus_jll [91d4177d-7536-5919-b921-800302f37372]
[ Info: Precompiling CEnum [fa961155-64e5-5f13-b03f-caf6b980ea82]
[ Info: Precompiling IniFile [83e8ac13-25f8-5344-8a64-a9f2b223428f]
...
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -937,13 +937,27 @@ function precompile(ctx::Context; internal_call::Bool=false)
             Base.PkgId(last(x), first(x)) 
             for x in ctx.env.project.deps if !Base.in_sysimage(Base.PkgId(last(x), first(x)))
         ]
-    end    
+    end
     
     was_processed = Dict{Base.PkgId,Base.Event}()
     was_recompiled = Dict{Base.PkgId,Bool}()
     for pkgid in keys(depsmap)
         was_processed[pkgid] = Base.Event()
         was_recompiled[pkgid] = false
+    end
+    
+    # guarding against circular deps
+    function in_deps(pkg, deps, dmap)
+        length(deps) == 0 && return false
+        pkg in deps && return true
+        return any(map(dep->in_deps(pkg, dmap[dep], dmap), deps))
+    end
+    for (pkg, deps) in depsmap
+        if in_deps(pkg, deps, depsmap)
+            notify(was_processed[pkg])
+            Operations.precomp_suspend!(pkg)
+            !internal_call && @warn "Circular dependency detected. Precompilation skipped for $pkg"
+        end
     end
     
     print_lock = stdout isa Base.LibuvStream ? stdout.lock : ReentrantLock()
--- a/src/API.jl
+++ b/src/API.jl
@@ -950,7 +950,7 @@ function precompile(ctx::Context; internal_call::Bool=false)
     function in_deps(pkg, deps, dmap)
         isempty(deps) && return false
         pkg in deps && return true
-        return any(map(dep->in_deps(pkg, dmap[dep], dmap), deps))
+        return any(dep->in_deps(pkg, dmap[dep], dmap), deps)
     end
     for (pkg, deps) in depsmap
         if in_deps(pkg, deps, depsmap)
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -946,6 +946,20 @@ function precompile(ctx::Context; internal_call::Bool=false)
         was_recompiled[pkgid] = false
     end
     
+    # guarding against circular deps
+    function in_deps(pkg, deps, dmap)
+        length(deps) == 0 && return false
```
oxinabox, line 951:
```suggestion
        isempty(deps)  && return false
```

tkf:
Can we have an MWE for #2057 as a regression test?

src/API.jl
```diff
@@ -946,6 +946,20 @@ function precompile(ctx::Context; internal_call::Bool=false)
...
+        pkg in deps && return true
+        return any(map(dep->in_deps(pkg, dmap[dep], dmap), deps))
```
oxinabox, line 953:
```suggestion
        return any(dep->in_deps(pkg, dmap[dep], dmap)
```

src/API.jl
```diff
@@ -946,6 +946,20 @@ function precompile(ctx::Context; internal_call::Bool=false)
...
+    end
+    for (pkg, deps) in depsmap
+        if in_deps(pkg, deps, depsmap)
+            notify(was_processed[pkg])
+            Operations.precomp_suspend!(pkg)
+            !internal_call && @warn "Circular dependency detected. Precompilation skipped for $pkg"
```
oxinabox, line 959: Should pkg be interpolated or should it be a log message argument?
In general I am never clear on that.
It seems unlikely someone will be using va structured log consumer with Pkg, but. Idk, may a Pkg graphical front-enf might be made

src/API.jl
ianshmean: This suggestion doesn't seem right, and I'm not sure how to fix it.. 🤔

ianshmean:
> Can we have an MWE for #2057 as a regression test?

I'm only hypothesizing that circular deps are the culpret in #2057, as I can't figure out another way it would happen.
@timholy do you still have the same environment set up? Could you try this PR?

I've added precompile tests for circular deps, and for a normal environment

src/API.jl
fredrikekre:
```suggestion
        return any(dep->in_deps(pkg, dmap[dep], dmap), deps)
```

oxinabox: sorry yeah, i was on my phone

ianshmean: Ah, of course. Thanks

timholy:
I'm kicking myself for not saving my Project and Manifest. I tried recreating it on a vanilla master and couldn't get back to the state that caused it. Sorry!

src/API.jl
ianshmean: I don't think I'm the person to answer this. Should it hold up merge?

ianshmean:
No worries, who'd want to keep a broken environment around anyway..

I think given the minimal impact of this circular dep check we should proceed with this and watch out for issues.

We could leave #2057 open for now

ianshmean:
CI has got confused. It's actually all green

PR #2052, opened 2020-09-29
[Artifacts] add equality operator for `BinaryPlatforms` shims

staticfloat:
This first fixes the BinaryPlatforms shims (e.g. the `Linux` and `MacOS` objects) by adding a better `==` operator, which solves https://github.com/JuliaLang/Pkg.jl/issues/2050, then removes their usage in the tests to use the new `Base.BinaryPlatforms` objects.

PR #2065, opened 2020-10-02
Travis CI: test that the Pkg UUID has not been changed

### LINKED ISSUE #2064: CI should probably catch changing Pkg's UUID
Since this is a standard dev trick and has been accidentally included in multiple PRs (including mine), we should probably check that it hasn't changed in CI.

DilumAluthge:
Fixes #2064 

This pull request adds a test to make sure that the UUID in the Pkg.jl `Project.toml` file has not changed. The test only runs on Travis CI.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,6 +31,7 @@ stages:
 
 script:
   - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
+  - julia test/pkg-uuid.jl
   - julia --project -e 'using UUIDs; write("Project.toml", replace(read("Project.toml", String), r"uuid = .*?\n" =>"uuid = \"$(uuid4())\"\n"));'
   - julia --project --check-bounds=yes -e 'import Pkg; Pkg.build(); Pkg.test(; coverage=true)'
 
--- a/test/pkg-uuid.jl
+++ b/test/pkg-uuid.jl
@@ -1,0 +1,10 @@
+using Pkg
+using Test
+
+@testset "Pkg UUID" begin 
+    project_filename = joinpath(dirname(@__DIR__), "Project.toml")
+    project = Pkg.TOML.parsefile(project_filename)
+    uuid = project["uuid"]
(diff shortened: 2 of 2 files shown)


KristofferC:
Tbh, this feels excessive to me. Like, are we going to test that not all tests are commented out because some people do that when they test locally (https://github.com/JuliaLang/Pkg.jl/pull/2052#issuecomment-700879182)?

fredrikekre:
Agree that it is a bit strange, but CI was green for once so I pressed the button :shrug:

StefanKarpinski:
It's different because you _have_ to change or comment out the UUID in the project file to develop Pkg, which makes it really likely to accidentally change the UUID in a PR and then forget about it. Also, unlike merging commented out tests, which don't actually break anything, even if changing the UUID doesn't break Pkg's own tests, it does break Julia if you try to test Julia with Pkg master. We could build and test Julia here, but this seems cheaper.

DilumAluthge:
> Like, are we going to test that not all tests are commented out because some people do that when they test locally ([#2052 (comment)](https://github.com/JuliaLang/Pkg.jl/pull/2052#issuecomment-700879182))?

I think that the easiest way to catch errors like that is to install the CodeCov GitHub app on this repository. Then, for each PR, CodeCov will report a status indicating whether coverage has increased, decreased, or remained the same.

Then, if someone comments out a bunch of tests, CodeCov will report a failing (red ❌) status on that PR because coverage will significantly decrease on that PR.

DilumAluthge:
> It's different because you _have_ to change or comment out the UUID in the project file to develop Pkg, which makes it really likely to accidentally change the UUID in a PR and forget about it. Also, unlike merging commented out tests, which don't actually break anything, even if changing the UUID doesn't break Pkg's own tests, it does break Julia if you try to test Julia with Pkg master. We could build and test Julia here, but this seems cheaper.

Testing Julia would definitely be expensive.

But would it be crazy to do the "build Julia with the PR branch of Pkg" part? A regular build of Julia (use BB, use default options for everything) might not be insane.

PR #2067, opened 2020-10-02
Use Travis for Windows CI, and stop using AppVeyor

### LINKED ISSUE #2066: Proposal: move all CI to GitHub Actions CI
Currently the Linux and Mac tests run on Travis, and the Windows tests run on AppVeyor.

Would it more convenient to have all of CI on a single provider?

Options include:
1. ~~Put everything in Travis CI~~ not an option, as Travis on Windows is too slow
2. Put everything in GitHub Actions CI

DilumAluthge:
Fixes #2066


DilumAluthge:
It looks like deleting the `appveyor.yml` file is not sufficient to disable AppVeyor.

If this PR is merged, then someone with the appropriate access will need to log in to the AppVeyor dashboard and disable AppVeyor on this repository.

DilumAluthge:
Hmmm. Looking at old AppVeyor runs, it looks like the AppVeyor Windows tests would usually take more than 50 minutes to run. Travis has a 50 minute limit. However, GH Actions has a six hour limit, so that might be a better option.

GH Actions jobs also tend to start up much quicker than either Travis or AppVeyor jobs.

KristofferC:
What's the issue with AV?

DilumAluthge:
I think the benefits for having a single CI provider are that we would only have to maintain a single CI config and build script.

Currently we have to maintain separate configs for Travis and AppVeyor, which results in duplicated information that we need to keep in sync.

Also, IIUC, AppVeyor is a bottleneck - the jobs take a while to start up, and the limit on concurrent jobs is low.

I don't think Travis is the solution; I think GH Actions CI will probably work better.

DilumAluthge:
Closing because Travis on Windows is too slow.

PR #2069, opened 2020-10-02
some performance improvements to uncompressing registry data

KristofferC:
Takes the benchmark of

```jl
using TOML
using Pkg
using UUIDs
using Pkg.Operations: load_package_data, VersionSpec, load_versions

const ctx = Pkg.Types.Context()
function bench_load_package()
    reg = joinpath(homedir(), ".julia/registries/General/")
    d = TOML.parsefile(joinpath(reg, "Registry.toml"))
    for (uuid, pkg) in d["packages"]
        pkgpath = pkg["path"] #
        path = joinpath(reg, pkgpath)
        version_info = load_versions(ctx, path)
        versions = sort!(collect(keys(version_info)))
        deps_data = load_package_data(UUID, joinpath(path, "Deps.toml"), versions)
        compat_data = load_package_data(VersionSpec, joinpath(path, "Compat.toml"), versions)
    end
end
```

from 407 ms, 339 Mib, to 330 ms, 248 Mib. For `Pkg.add("DifferentialEquations")` the impact is around 10-13% improvement.,

### DIFF
--- a/src/versions.jl
+++ b/src/versions.jl
@@ -132,12 +132,16 @@ VersionRange(t::Integer...)                  = VersionRange(VersionBound(t...))
 VersionRange(v::VersionNumber)               = VersionRange(VersionBound(v))
 VersionRange(lo::VersionNumber, hi::VersionNumber) = VersionRange(VersionBound(lo), VersionBound(hi))
 
-
+# The vast majority of VersionRanges are in practice equal to "1"
+const VersionRange_1 = VersionRange(VersionBound("1"), VersionBound("1"))
 function VersionRange(s::AbstractString)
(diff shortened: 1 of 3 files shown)


src/versions.jl
```diff
@@ -251,12 +255,12 @@ end
 Base.intersect(a::VersionNumber, B::VersionSpec) = a in B ? VersionSpec(a) : empty_versionspec
 Base.intersect(A::VersionSpec, b::VersionNumber) = intersect(b, A)
 
-Base.union(A::VersionSpec, B::VersionSpec) = union!(copy(A), B)
-function Base.union!(A::VersionSpec, B::VersionSpec)
```
KristofferC, line 255: The point of this is to consider a `VersionSpec` immutable so that we can use multiple instances of the same `VersionSpec` and not have to create tens of thousands of `VersionSpec("1")` just in case it get mutated somewhere.

KristofferC:
I'll merge this because it is an improvement but some of this gets rewritten in #2072

PR #2070, opened 2020-10-02
Document that dev --local path is not allowed.

tpapp:
See [discussion](https://discourse.julialang.org/t/pkg-dev-local-some-path-doesnt-make-a-local-clone-in-dev/47316/).

Clarification on whether is just not implemented, or not allowed for some deeper reason, is welcome; then I will update the PR.

### DIFF
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -156,12 +156,16 @@ PSA[:name => "develop",
     :description => "clone the full package repo locally for development",
     :help => md"""
     [dev|develop] [--shared|--local] pkg[=uuid] ...
+    [dev|develop] [--shared] path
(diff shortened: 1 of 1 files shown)


fredrikekre:
The same applies to `--global` so it should be
```
    [dev|develop] [--shared|--local] pkg[=uuid] ...
    [dev|develop] path
```

And from discourse:

> Is there a rationale for this, or just not implemented yet?

I think that it should not be supported. When you give dev a path it is because you want Pkg to point at that folder on your harddrive and this is a folder that you manage yourself. I don't think it would be appropriate for Pkg to copy that folder with `--local` for example since it can be in any state.

00vareladavid:
Merged in #2467

Issue #2044, opened 2020-09-25
Stop using the raw manifest `Dict' inside Pkg operations


KristofferC:
Looking at invalidations, a lot of them come from indexing into the manifest dict which is non-typed, getting inference to infer `Any` and things are bad. I think it is better if we just statically type the whole manifest, read everything into a `Manifest` struct, and then just use that one. The raw dictionary need not be stored anywhere after being read.

PR #2072, opened 2020-10-02
more structured Registry handling

KristofferC:
Some background: Pkg is a little bit of spaghetti now and it is hard to know where things for a certain area is done. For example (relevant for this PR) where is the code that handles the registry? Well, it's a bit all over the place, depending on what the function needs, it might even go and parse some files in the registry and get what it wants. I (mostly) blame myself for this but in my defense when I did most of the stuff on Pkg there was little time until 1.0 and a lot of features to implement to get to feature parity with the old package manager. Now when things are calmed down I think it would be good to re-evaluate a bit on how things are structured.

As a start, I want to look exactly at the way we access the information in the registries. As I said, right now, there are callees that parse registry files all over the place, the same TOML file thus gets parsed multiple times, etc. There is an attempt of a caching system in the `EnvCache` struct but it doesn't work properly. Also, https://github.com/JuliaLang/Pkg.jl/issues/2044 is relevant since every time you go and parse a TOML file you lose type information. In addition, we might want to have access to the registry information when we e.g. do `status` so that we can show what packages are being held back. But there is no way to conveniently do that right now.

To start improving this, this PR starts implementing a proper "interface" to query the registry. It is lazy and constructs things as you access them as to not parse and uncompress a bunch of unnecessary data. Right now, only the resolver part has been moved to use the new interface (and there is still some wip there) but I want to start moving all lookups that Pkg does about registry information to it. Also, the information about the registry will be stored in e.g. the `EnvCache` so that information about the registry is available throughout the whole Pkg operation and not only when it is needed.

I'll probably force push to this a lot.

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -18,6 +18,7 @@ end
 
 # setup for doctesting
 DocMeta.setdocmeta!(Pkg.BinaryPlatforms, :DocTestSetup, :(using Base.BinaryPlatforms); recursive=true)
+DocMeta.setdocmeta!(Pkg.LazilyInitializedFields, :DocTestSetup, :(using Pkg.LazilyInitializedFields))
 
 # Run doctests first and disable them in makedocs
 Documenter.doctest(joinpath(@__DIR__, "src"), [Pkg])
--- a/ext/LazilyInitializedFields/LazilyInitializedFields.jl
+++ b/ext/LazilyInitializedFields/LazilyInitializedFields.jl
@@ -1,0 +1,328 @@
+"""
+A package for handling lazily initialized fields.
+
+### Exports:
+* macros: `@lazy`, `@init!`, `@uninit!`, `@isinit`.
+* functions: `init!` `uninit!`, `isinit`.
+* objects: `uninit`.
+* exceptions: `NonLazyFieldException`, `UninitializedFieldException`, `AlreadyInitializedException`
+
+### Example usage:
+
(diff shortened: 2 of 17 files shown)


DilumAluthge:
> Also, the information about the registry will be stored in e.g. the `EnvCache` so that information about the registry is available throughout the whole Pkg operation and not only when it is needed.

Just out of curiosity, is the cache thread safe?

DilumAluthge:
Would this PR be a good opportunity to also implement the "parse the registry without untarring the tarball" feature? Or would it be better for that feature to go in a different PR?

KristofferC:
> Would this PR be a good opportunity to also implement the "parse the registry without untarring the tarball" feature?

That's a separate thing.

DilumAluthge:
> > Would this PR be a good opportunity to also implement the "parse the registry without untarring the tarball" feature?
> 
> That's a separate thing.

Sounds good.

KristofferC:
This should be in a decent state to merge now (let's leave it out from 1.6 though). There is still some work left to integrate some of the rest of the registry handling in it but it can be done in a piecewise manner.

KristofferC:
I want to merge this soon because locally https://github.com/JuliaLang/Pkg.jl/pull/2103 is starting to mix with the changes here. This is not in perfect condition because I want to move some of the registry update handling etc over to this stuff but I think it is a large enough PR already.

PR #2076, opened 2020-10-06
Print a warning when using the REPL interface in non-interactive mode.

fredrikekre:


src/REPLMode/REPLMode.jl
```diff
@@ -373,6 +374,10 @@ end
 # Execution #
 #############
 function do_cmd(repl::REPL.AbstractREPL, input::String; do_rethrow=false)
+    if !isinteractive() && !TEST_MODE[] && !PRINTED_REPL_WARNING[]
+        @warn "The Pkg REPL interface is intended for interactive use, use with caution from scripts."
```
KristofferC, line 378: Maybe the `log_once` option to `@warn` could be used?

PR #2078, opened 2020-10-06
PlatformEngines: replace ad hoc stuff with Downloads & Tar stdlibs

### LINKED ISSUE #1830: Use libCURL instead of platform engines
_This is very speculative, so mainly for brainstorming_

Currently HTTTP downloads within Pkg are done using the `PlatformEngines` subsystem, which shell out to locally installed commandline tools to perform the download. This generally means the `curl`/`wget`/`fetch` executables on unix and `PowerShell` on windows. This code is reasonably complex and difficult to maintain (I think)

In order to make things streamlined, one thought is to replace this with HTTP.jl. However, that is a large library, and splitting it will be some effort. 

Another idea is to use libCurl. 

* libCURL is *widely* used and *very* stable. Every car, phone and TV has multiple instances. 
* It's very efficient, in CPU and memory -- it can do multiple downloads simultaneously in async mode. 
* There is now a well maintained libCURL_jll, and the recent jll work on base julia makes it easy to pull in. One important feature here is that the jll correctly builds against the mbedTLS jll
* libCURL has http2 support (and ftp support and a dozen other protocols ...)

A few concerns, not insurmountable:

* libCURL has a low level C API. A higher level Julia API [needs to be written](https://curl.haxx.se/libcurl/c/libcurl-tutorial.html), ideally using the [multi-api](https://curl.haxx.se/libcurl/c/libcurl-multi.html). There are julia examples using the [easy api](https://github.com/JuliaWeb/LibCURL.jl#example-fetch-a-url)
* proxies (potentially authenticated)  need to be handled 
* root certificates needs to be handled

cc: @StefanKarpinski @staticfloat

StefanKarpinski:
Fixes #1830

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -8,6 +8,7 @@ version = "1.5.0"
 [deps]
 Artifacts = "56f22d72-fd6d-98f1-02f0-08ddc0907c33"
 Dates = "ade2ca70-3891-5945-98fb-dc099432e06a"
+Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 LibGit2 = "76f85450-5226-5b5a-8eaa-529ad045b433"
 Libdl = "8f399da3-3557-5675-b5ff-fb832c97cbdb"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,5 @@
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+uuid = "b4ba1df2-8887-4706-8585-f0048dd7ac84"
 keywords = ["package management"]
 license = "MIT"
 desc = "The next-generation Julia package manager."
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,5 @@
 name = "Pkg"
-uuid = "b4ba1df2-8887-4706-8585-f0048dd7ac84"
+uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 keywords = ["package management"]
 license = "MIT"
 desc = "The next-generation Julia package manager."
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -7,9 +7,7 @@ using SHA
 import ...Pkg: Pkg, TOML, pkg_server, depots1
 using Base.BinaryPlatforms
 
-export probe_platform_engines!, parse_7z_list, parse_tar_list, verify,
-       download_verify, unpack, package, download_verify_unpack,
-       list_tarball_files, list_tarball_symlinks
+export probe_platform_engines!, verify, unpack, package, download_verify_unpack
 
 # In this file, we setup the `gen_download_cmd()`, `gen_unpack_cmd()` and
 # `gen_package_cmd()` functions by providing methods to probe the environment
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -682,13 +682,7 @@ function download(
     for header in get_metadata_headers(url)
         push!(headers, header)
     end
-    # TODO: once we get a new nightly, delete all but last
-    yield()
-    if applicable(Downloads.download, url, dest, headers)
-        Downloads.download(url, dest, headers)
-    else
-        Downloads.download(url, dest, headers=headers)
-    end
+    Downloads.download(url, dest, headers=headers)
 end
 
 """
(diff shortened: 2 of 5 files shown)


src/PlatformEngines.jl
```diff
@@ -802,18 +682,7 @@ function download(
...
-    end
-    try
-        run(download_cmd, (devnull, verbose ? stdout : devnull, verbose ? stderr : devnull))
-    catch e
-        if isa(e, InterruptException)
-            rethrow()
-        end
-        error("Could not download $(url) to $(dest):\n$(e)")
-    end
+    Downloads.download(url, dest, headers=headers)
```
KristofferC, line 685:
```suggestion
    Downloads.download(url, dest, headers)
```

KristofferC:
I tried this with the tweak I commented and then it starts to download but I think some concurrent download tasks get stuck. I did an `instantiate` in my global env and it got really slow and then it just hung completely.

StefanKarpinski:
Yeah, there's some kind of task deadlock issue with a lot of downloads. Definitely needs debugging.

StefanKarpinski:
Inserting a `yield` before calling `Downloads.download` "fixes" the issue.

KristofferC:
> Inserting a yield before calling Downloads.download "fixes" the issue.

I tried this on the latest nightly and it still hangs for me. It might be related but the throughput seems quite a lot slower as well (but maybe this is because tasks are hanging):

PR:
Screenshot-20201008170442-1119x132

1.5:

Screenshot-20201008170631-532x114

StefanKarpinski:
So this mostly works but the throughput regression needs to be investigated and may be the cause of the timeouts on various CI runs (I restarted a couple of them).

KristofferC:
You don't get the hang when running instantiate on a big environment? I'll try again later.

KristofferC:
I made a benchmark script:

```
using Pkg
using Pkg.TOML
import Downloads
function gather_urls(env="v1.6")
    d = TOML.parsefile(joinpath(homedir(), ".julia/environments/$env/Manifest.toml"))
    regpath = joinpath(homedir(), ".julia/registries/General")
    reg = TOML.parsefile(joinpath(regpath, "Registry.toml"))
    urls = String[]
    for (pkg, infos) in d
        for info in infos
            uuid = info["uuid"]
            tree = get(info, "git-tree-sha1", nothing)
            tree === nothing && continue
            pkg = get(reg["packages"], uuid, nothing)
            pkg === nothing && continue
            path = pkg["path"]
            repo = TOML.parsefile(joinpath(regpath, path, "Package.toml"))["repo"]
            push!(urls, string(Pkg.pkg_server(), "/package/$uuid/$tree"))
        end
    end
    return urls
end
urls = gather_urls()
function download_urls(urls; num_concurrent=8, path=mktempdir(), old_way=true)
    Pkg.PlatformEngines.probe_platform_engines!()
    broke = false
    @sync for i in 1:num_concurrent
        @async while true
            broke && break
            isempty(urls) && break
            url = popfirst!(urls)
            @info "downloading $url"
            file = joinpath(path, basename(url))
            try
                if old_way
                    Pkg.PlatformEngines.download(url, file; verbose=false)
                else
                    Downloads.download(url, file)
                end
            catch e
                if e isa InterruptException
                    broke = true
                end
                @show e
                @warn "url failed to download"
            end
        end
    end
end
@time download_urls(urls; old_way=false)
```

and running this on my mac I get faster speed with `old_way=false` than `old_way=true`, that is faster with `Downloads`. So I am not sure why I got the bad throughput in Pkg, I'll have to test this on my linux machine tomorrow where I made the post above.

KristofferC:
I also get the hang in Pkg on my mac, which runs the benchmark script fine. So it doesn't seem to be related to only running the downloads..

KristofferC:
Tried it again with latest Downloads.jl and it downloads fast(!) but unfortunately there still seem to be some tasks getting stuck.

However, on the benchmark above it works fine and it is more than twice as fast as the old way.

StefanKarpinski:
Awesome to hear that. I had tried it hoping the the hangs to go away, but alas, not yet. So the new speed is great, but I still need to investigate the hanging. I've already got a branch of downloads that prints a bunch of instrumented data in CSV format so I can see which downloads are getting stuck. Will continue with that tomorrow.

Project.toml
```diff
@@ -1,5 +1,5 @@
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+uuid = "b4ba1df2-8887-4706-8585-f0048dd7ac84"
```
KristofferC, line 2:
```suggestion
uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```

KristofferC:
bild

:tada:

PR #2082, opened 2020-10-07
Travis CI: if the tests fail, automatically retry them one time

DilumAluthge:
It is very common for one of the CI jobs to fail because of a network error. For example, a clone from GitHub fails. The solution is to restart the single failed job.

See e.g. this run: https://travis-ci.com/github/JuliaLang/Pkg.jl/builds/188489284

In that run, the first three jobs passed, but the fourth failed due to a sporadic network error.

It's annoying to have to always restart those jobs.

This pull request adds a bounded retry loop. We try at most two times total. We sleep 60 seconds between tries - if GitHub was rate limiting our downloads for some reason, this sleep may give time for the rate limit to reset.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,11 +29,7 @@ stages:
   - name: test
   - name: docs
 
-script:
-  - if [[ -a .git/shallow ]]; then git fetch --unshallow; fi
-  - julia test/pkg-uuid.jl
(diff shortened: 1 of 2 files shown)


DilumAluthge:
🤦 

I completely forgot - we've already discussed this. The consensus was that we should identify the specific Pkg tests that are being flaky, and do some combination of:
1. Make those individual tests less flaky
2 Put only those individual tests inside bounded retry loops

Issue #2085, opened 2020-10-07
Install the Codecov GitHub app on this repo?


DilumAluthge:
Would it be a good idea to install the Codecov GitHub app on the Pkg.jl repo? This would result in Codecov posting a status on each pull request.

This might make it easier to detect e.g. if someone commented out all of the tests and forgot to un-comment them before committing and pushing. Such a scenario would result in an ❌ (failing) Codecov status on the pull request because the coverage would have significantly dropped.

We already submit coverage to Codecov:
https://github.com/JuliaLang/Pkg.jl/blob/f0cacf64d1a5fa767098cab43ba5c1c2c5f21e37/.travis.yml#L49-L51

So this would just be a matter of installing the Codecov GitHub app.

DilumAluthge:
See also: https://github.com/JuliaLang/Pkg.jl/pull/2086

fredrikekre:
I can install the app if we don't do #2086 :)

fredrikekre:
It looks like it is already enabled on this repo...

DilumAluthge:
That's strange... it should be posting statuses.

I wonder if Codecov is getting confused because we submit to coverage from Travis but not AppVeyor. So it detects the AppVeyor builds (by polling the GitHub API), but it never gets coverage from AppVeyor, so it keeps waiting.

For completeness we should submit coverage from all platforms anyway.

DilumAluthge:
Even when we submit coverage from AppVeyor, the CodeCov status doesn't show up: https://github.com/JuliaLang/Pkg.jl/pull/2088

DilumAluthge:
It'll probably be easiest to just do https://github.com/JuliaLang/Pkg.jl/pull/2086

DilumAluthge:
@fredrikekre Can you try uninstalling and reinstalling the Codecov app? Codecov should be posting statuses on both PRs and branches. I'm not sure why it's not doing that.

fredrikekre:
I nuked the install and re-authorized it.

DilumAluthge:
@fredrikekre I think you've cured it! See this PR, which has the Codecov statuses: https://github.com/JuliaLang/Pkg.jl/pull/2089

fredrikekre:
Perhaps it was confused about the travis org -> com move?

DilumAluthge:
> Perhaps it was confused about the travis org -> com move?

Probably. I've heard of other people having various parts of their workflow broken by the migration.

PR #2086, opened 2020-10-07
Allow Codecov to post comments on pull requests

DilumAluthge:
The comment will include information on whether coverage has dropped, and if so, by how much. If there is a significant drop in coverage, this may indicate a problem, e.g. maybe a bunch of tests have been commented out.

See also: https://github.com/JuliaLang/Pkg.jl/issues/2085

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,1 +1,0 @@
-comment: false


KristofferC:
I must say that 99.9% of the time these comments are just noise so personally I don't think this is very useful.

fredrikekre:
> I must say that 99.9% of the time these comments are just noise so personally I don't think this is very useful.

Agree (https://github.com/JuliaLang/Pkg.jl/issues/2085#issuecomment-704799069).

DilumAluthge:
In that case, let's keep working on trying to get the PR status to work.

PR #2091, opened 2020-10-07
Precompilation: Prettier printing, opt-out auto

### LINKED ISSUE #2090: Ctrl-C should probably interrupt the automatic precompilation
Right now, it just marks the precompiled package as failed.

### LINKED ISSUE #2093: Enable precompiling after Pkg operations by default
It would be good to enable this by default so it gets properly tested in the 1.6 alpha and RCs.

@tkf you mentioned you wanted to use preferences for this, how would that look in practice?

cc @ianshmean

ianshmean:
An attempt to make the firehose more manageable https://www.dropbox.com/s/r3x7p3dk46n0o5b/Julia%20Precomp%20-%20Prettier%20printing.mov?dl=0

Some things to note:

1. The code could definitely be tidier. This is more of a UX demo.

2. No throwing. This doesn't throw errors, not even for direct deps (unlike master), but it shows when errors occur. It's proven really challenging to make the errors and stack traces throw cleanly from within the parallel process, and I thought that perhaps the pkg repl isn't the best place to show the full stack traces anyway. So the failing deps are just highlighted, and the user can go and load the code to see the errors in full.

3. Erroring packages are suspended during auto-precompilation. (This isn't new to this PR, but the video illustrates this nicely). I threw an error into `ImageCore`. Note how the errors that happen in the first `precompile` mean that those failing packages are skipped in subsequent auto precomp sessions. Then to force full precompilation, `precompile` is called.

4. Need a basic mode for IJulia, CI etc. We'd need to disable this for dumb terminals and just do a linear print like it used to do.

5. Timing tweaking. The successful packages only stay for a moment after succeeding. Increasing that might make it less busy, but would limit how much could be on screen at one time

Unrelated. Note the long pause at 1:53 immediately after `pkg> add` looks a bit like invalidation-related lag? Anyone else see that?

____

Edit: Given auto is now enabled in this PR, Ctrl-C interrupts are handled, and some other fixes are included
Closes #2093 
Closes #2102 
Fixes #2090


KristofferC:
Coolest shit ever!

timholy:
That's really fun. The biggest problem I foresee is that people are then going to ask for `rm_precompile` just so they can watch it precompile again.

What's the difference between the white-bold and the rest? I expected the ones underneath to be dependencies but it seems they are not.

> Note the long pause at 1:53 immediately after pkg> add looks a bit like invalidation-related lag? Anyone else see that?

I have seen something like that occasionally and wondered the same thing. It's a bit unexpected because precompilation doesn't load new code in the parent session, right?

ianshmean:
> What's the difference between the white-bold and the rest?

White bold = direct dependencies from the Project.toml
The rest = indirect dependencies, but not specifically of the packages above/near them.. the ordering is first-come first-served so rather unsorted

ianshmean:
Perhaps the indent of the gray indirect deps is inaccurately making it look tree-like. I'll remove that.

> precompilation doesn't load new code in the parent session, right?

As far as I know, yeah. Could it just be because I'm dev-ing Pkg, so it's not in the sysimage?

KristofferC:
You can run with `--trace-compile=pkgprec.jl` and then run `tail` on that. Maybe it will shown if new things are compiled. But I'm not sure how that acts with invalidations.

haampie:
Using the 128 core machine once more:

https://asciinema.org/a/6I9AN6uBTIHFoX4d7uCfnqh1q

ianshmean:
@haampie still a bit of a firehose, but does seem more comprehendible 

On the lag, I just built this branch into the julia sysimage and it disappeared 👍🏻

KristofferC:
> The code could definitely be tidier. This is more of a UX demo.

It's pretty isolated code though so tidiness isn't too important. As long as it works well :P


> No throwing. This doesn't throw errors, not even for direct deps (unlike master), but it shows when errors occur. 

Perhaps it can throw in the very end if something failed? Just so if someone runs `Pkg.precompile` in their CI script it will abort if a package fails to compile.

> Need a basic mode for IJulia, CI etc. We'd need to disable this for dumb terminals and just do a linear print like it used to do.

Maybe check if `stdout` / `stderr` (whatever is used) is a `Base.TTY`?

src/API.jl
```diff
@@ -959,9 +962,81 @@ function precompile(ctx::Context; internal_call::Bool=false)
...
+                term_size = Base.displaysize(stdout)::Tuple{Int,Int}
+                num_deps_show = term_size[1] - 2
+                pkg_queue_show = if !finished && length(pkg_queue) > num_deps_show 
+                    last(pkg_queue, num_deps_show)
+                else
+                    pkg_queue
+                end
+                str = ""
+                if i > 1
+                    str *= "\e[$(last_length)A\e[1G\e[0J"
```
KristofferC, line 996: Can these ANSI codes be assigned to variables with names instead?

src/API.jl
```diff
@@ -959,9 +962,81 @@ function precompile(ctx::Context; internal_call::Bool=false)
...
+                end
+                for dep in pkg_queue_show
+                    finished && was_recompiled[dep] && continue
+                    name = dep in direct_deps ? "  $(dep.name)" : "  \e[38;5;244m$(dep.name)\e[0m"
+                    if dep in failed_deps
+                        str *= string(name, " \e[38;5;160m✗\e[0m\n")
```
KristofferC, line 1002: Could we use one of the standard 16 colors here? Could perhaps use `Base.error_color` for the red?

fredrikekre:
> Perhaps the indent of the gray indirect deps is inaccurately making it look tree-like. I'll remove that.

Perhaps instead indent everything? Then it looks more like it is under the `Precompiling project...` "header".

ianshmean:
Yeah, I agree. I did actually end up doing that. It's already committed.

ianshmean:
@KristofferC 
> Perhaps it can throw in the very end if something failed? Just so if someone runs Pkg.precompile in their CI script it will abort if a package fails to compile.

But if multiple packages fail, and it throws at the end, we'd have to figure out how to show a stack trace for all packages that failed which could be a huge mess, otherwise the CI session wouldn't show the actual error and would be a bit opaque to the user. Instead, this approach reserves throwing proper package load errors to linear load time, which would be a lot more manageable. I can't think of a CI use case that would include `Pkg.precompile()`  but not package loading.

This approach to not throw does change the current behavior though, so perhaps that's an unacceptable breaking change for a 1.x release?

KristofferC:
Couldn't we just throw an error in the end showing what packages failed? I don't understand the part about stack traces.

ianshmean:
We could do that, but then it would be opaque on CI why the package failed. The stack trace wouldn't show the root reason why.

Happy to go with your judgement though 👍

ianshmean:
This now throws at the end if any direct deps fail, with a list
image

Also, more basic printing is done when stdout isn't a `Base.TTY`
image

ianshmean:
It seemed that there are a few others in favor of enabling auto-precompilation by default, so I've gone ahead and done that here

I've also added a proposal for control to the API, while retaining the ENV var control (the API function sets the ENV var)
```
(@v1.6) pkg> precompile_auto false

(@v1.6) pkg> precompile_auto true

julia> Pkg.precompile_auto(false)

julia> ENV["JULIA_PKG_PRECOMPILE_AUTO"] = "0"
```

src/API.jl
```diff
@@ -1093,6 +1092,14 @@ function precompile(ctx::Context; internal_call::Bool=false)
     finished = true
     notify(first_started) # in case of no-op
     wait(t_print)
+    failed_direct = filter(in(direct_deps), failed_deps)
+    if !isempty(failed_direct)
+        failed_list = ""
+        for d in failed_direct
+            failed_list *= "  $d\n"
+        end
+        throw(ErrorException("The following direct dependencies failed to precompile:\n$(failed_list)"))
```
KristofferC, line 1121:
```suggestion
        pkgerror("The following direct dependencies failed to precompile:\n$(failed_list)"))
```

(the difference is that it doesn't show a backtrace when you call it from the Pkg REPL)

ianshmean: perfect, thanks

src/API.jl
```diff
@@ -66,7 +67,7 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
         $f(pkgs::Vector{<:AbstractString}; kwargs...)          = $f([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
         function $f(pkgs::Vector{PackageSpec}; kwargs...)
             ret = $f(Context(), pkgs; kwargs...)
-            $(f in (:add, :up, :pin, :free, :build)) && _auto_precompile()
```
KristofferC, line 69: Move the `Context()` creation out of the `f` call and also pass it into `auto_precompile`?

src/API.jl
```diff
@@ -904,25 +905,34 @@ function _is_stale(paths::Vector{String}, sourcepath::String, toml_c::Base.TOMLC
...
+precompile(; kwargs...) = precompile(Context(); kwargs...)
+function precompile(ctx::Context; internal_call::Bool=false, io::IO=stdout)    
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
     parallel_limiter = Base.Semaphore(num_tasks)
-    
+    fancy_print = (io isa Base.TTY) && (get(ENV, "CI", nothing) != "true")
+
     # when manually called, unsuspend all packages that were suspended due to precomp errors
     !internal_call && Operations.precomp_unsuspend!() 
+    action_help = internal_call ? " (tip: control auto with `pkg> precompile_auto true/false`)" : ""
```
KristofferC, line 929: Imo, it would be better to only tell about the env variable. With this, you would have to do it for every session and that seems annoying enough that you would want to learn how to do it the persistent way.

src/API.jl
```diff
@@ -904,25 +905,34 @@ function _is_stale(paths::Vector{String}, sourcepath::String, toml_c::Base.TOMLC
...
-    if parse(Int, get(ENV, "JULIA_PKG_PRECOMPILE_AUTO", "0")) == 1
+    if parse(Int, get(ENV, "JULIA_PKG_PRECOMPILE_AUTO", "1")) == 1
         Pkg.precompile(internal_call=true)
     end
 end
 
-precompile(;internal_call::Bool=false) = precompile(Context(), internal_call=internal_call)
-function precompile(ctx::Context; internal_call::Bool=false)    
+precompile(; kwargs...) = precompile(Context(); kwargs...)
+function precompile(ctx::Context; internal_call::Bool=false, io::IO=stdout)
```
fredrikekre, line 922:
```suggestion
function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
```

src/API.jl
ianshmean: Updated
image

ianshmean:
I'm not sure about the underscore in `(@v1.6) pkg> precompile_auto false` as it feels like it breaks pkg REPL style. Would `(@v1.6) pkg> precompileauto false` be better? 

`(@v1.6) pkg> precompile auto false` would be need a bit of a hack to work from what I can tell

Otherwise, I'm happy with this now

ianshmean:
I've gone with `autoprecompile`. Simpler & tidier

KristofferC:
Imo we don't need really a REPL option to turn this off. After all, we want this to be on, it is just in the rare case that people should opt-out and then they can just set the env variable?

fredrikekre:
I agree. If this is on by default that means it is good enough that most users should not have to disable it.

ianshmean:
Sounds good to me!

KristofferC (approved):

ianshmean:
I think this is finally ready to go

KristofferC:
Did you see https://github.com/JuliaLang/julia/pull/37906 by the way? It removes the need to use `TOMLCache`. What order do you prefer, merge that first, wait until the nightly update and fix up things here, or merge this, bump Pkg on master, merge that PR and then fix Pkg again?

ianshmean:
Oh cool. Happy to hold off until that's merged & nightlied

src/API.jl
```diff
@@ -904,25 +906,27 @@ function _is_stale(paths::Vector{String}, sourcepath::String, toml_c::Base.TOMLC
...
+precompile(; kwargs...) = precompile(Context(); kwargs...)
+function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
     parallel_limiter = Base.Semaphore(num_tasks)
-    
+    fancy_print = (io isa Base.TTY) && (get(ENV, "CI", nothing) != "true")
+
     # when manually called, unsuspend all packages that were suspended due to precomp errors
     !internal_call && Operations.precomp_unsuspend!() 
+    action_help = internal_call ? " (tip: to disable auto set ENV[\"JULIA_PKG_PRECOMPILE_AUTO\"]=0)" : ""
```
fredrikekre, line 923: I don't think we need this, will get pretty annoying the 100th time you see it.

KristofferC: I think it is fine to have it for now, just until it has gotten a bit more usage. Then we remove it after it's been tested properly in the wild.

ianshmean: I think you're right. I've removed it and added info about the automatic mode to the `Pkg.precompile` docs, including this tip on how to disable it

ianshmean: Ok. I'll add it back.

KristofferC: Sorry for the back and forth :D

KristofferC:
I merged the `TOMLCache` PR so if you want, you could remove all references to `TOMLCache` from here and then we wait until e.g. tomorrow when nightlies are updated to rerun CI here and merge.

ianshmean:
I've added interrupt handling for `Ctrl-C`. Whether the interrupt happens during the print loop or precomp tasks it will exit simply after a pause while the top level tasks stop.
It will end pretty un-gracefully if `Ctrl-C` is button mashed.. but that's always going to be hard to handle.

Fixes #2090

ianshmean:
I merged #2102 in here as there were mutually exclusive bugs preventing CI success

DilumAluthge:
@ianshmean Can you add "closes #2101" somewhere in the first post of this PR?

ianshmean:
@DilumAluthge I think you meant #2102, which I added up top

DilumAluthge:
> @DilumAluthge I think you meant #2102, which I added up top

Yep you are correct.

DilumAluthge:
Looks like Travis is passing now. @KristofferC is this good to merge once AppVeyor passes?

DilumAluthge:
I'm not sure why AppVeyor is reporting a failure. If I look at the AppVeyor test log:
```
[00:59:49]     Testing Pkg tests passed 
[00:59:49] C:\julia\bin\julia --project=test/coverage -e "using Pkg; Pkg.instantiate(); using Coverage; Codecov.submit(Codecov.process_folder())"
[00:59:50]    Updating registry at `C:\Users\appveyor\.julia\registries\General`
```

So it looks like the Pkg tests passed, but then something went wrong while AppVeyor was trying to upload to Codecov.

DilumAluthge:
@KristofferC or @fredrikekre Can you restart the AppVeyor job?

PR #2094, opened 2020-10-08
Add verbose option to activate and precompile

### LINKED ISSUE #2036: [FR] Quiet environment setup
I wonder if it would make sense to have an option for quiet environment setup in projects. Currently, whenever we activate, instantiate, precompile an environment in the beginning of a script, we get information messages sent to stdout. It is fine when you only have one worker process, but when the script is being executed on a computer cluster with say 100 processes, it is a lot as each worker prints the same information.

I am considering using Supressor.jl to deal with this verbose printing, but would like to ask here if it makes sense to make this a feature of the package manager, something like an option `Pkg.setup(verbose=false)`.

juliohm:
Fix #2036 

This PR adds a simple `verbose` option to `activate` and `precompile`. Could you please review and double check it is ok? It would be nice to have this in Julia v1.6.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -910,8 +910,7 @@ function _auto_precompile()
     end
 end
 
-precompile(;internal_call::Bool=false) = precompile(Context(), internal_call=internal_call)
-function precompile(ctx::Context; internal_call::Bool=false)    
+function precompile(ctx::Context=Context(); internal_call::Bool=false, verbose::Bool=true)
     num_tasks = parse(Int, get(ENV, "JULIA_NUM_PRECOMPILE_TASKS", string(Sys.CPU_THREADS + 1)))
     parallel_limiter = Base.Semaphore(num_tasks)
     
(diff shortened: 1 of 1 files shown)


fredrikekre:
I think those should just accept an `io` kwarg like other functions. `verbose` is usually used to give some more/extra information together with the default output.

juliohm:
@fredrikekre I followed the `verbose` option in `instantiate`. Could you quickly work on this io suggestion? I have little experience with Julia's IO objects, and besides that, I find it is difficult to properly run the test suite without the master branch of Julia. I only have the latest stable release with me.

DilumAluthge:
It looks like the AppVeyor failure is a download failure, which is most likely the usual sporadic network failure.

@KristofferC or @ianshmean can you manually restart the AppVeyor job?

DilumAluthge:
Sorry, posted on the wrong thread.

fredrikekre:
It is still missing for `activate` I think?

juliohm:
Both `precompile` and `activate` miss options for a quiet setup. I understood from your comment that this PR wouldn't make it, and that instead people should add an `io` option and redirect to `devnull` somehow.

PR #2102, opened 2020-10-11
fix test after `bytes` field in `SHA1` changed into a tuple

KristofferC:


ianshmean:
I believe this also needs https://github.com/JuliaLang/Pkg.jl/pull/2091/commits/e68d43c7f95b763865b1655b959001ba2bb2a99e which I've added to #2091 along with the above commit to try to get CI passing

PR #2115, opened 2020-10-13
fix invalidation sources

KristofferC:


src/API.jl
```diff
@@ -972,11 +975,13 @@ function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
     failed_deps = Base.PkgId[]
     skipped_deps = Base.PkgId[]
 
-    print_lock = stdout isa Base.LibuvStream ? stdout.lock : ReentrantLock()
+    print_lock= (stdout isa Base.LibuvStream ? stdout.lock : ReentrantLock())::ReentrantLock
```
timholy, line 978:
```suggestion
    print_lock = (stdout isa Base.LibuvStream ? stdout.lock : ReentrantLock())::ReentrantLock
```

src/API.jl
```diff
@@ -972,11 +975,13 @@ function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
...
     first_started = Base.Event()
-    finished = false
-    should_exit = !fancy_print # exit print loop immediately if not fancy printing
-    interrupted = false
+    # We make these Refs so they do not have to be boxed
```
timholy, line 980: Probably prettier if you change it to
```julia
finished::Bool = false
should_exit::Bool = !fancy_print
interrupted::Bool = false
show_report::Bool = true
```

src/PlatformEngines.jl
```diff
@@ -598,7 +598,7 @@ function get_server_dir(url::AbstractString, server=pkg_server())
         @warn "malformed Pkg server value" server
         return
     end
-    joinpath(depots1(), "servers", m.captures[1]::AbstractString)
+    joinpath(depots1(), "servers", String(m.captures[1]))
```
timholy, line 601: This change probably isn't necessary:
```julia
julia> fieldtype(RegexMatch, :captures)
Vector{Union{Nothing, SubString{String}}} (alias for Array{Union{Nothing, SubString{String}}, 1})
```
and so the `::AbstractString` should be enough for Julia to know that it's a `SubString{String}`. If that's not true, then the best might be `m::RegexMatch`?

timholy (commented):
Looks mostly fine, but I think you can ditch a lot of these using a different 15276 strategy.

src/PlatformEngines.jl
KristofferC: The issue is that if it becomes a `SubString` then `joinpath` will not specialize on all arguments in `joinpath(::String, ::String, ::SubString{String}` but instead consider the last argument as a `Vararg{AbstractString, String}`.

timholy: Got it, carry on!

src/API.jl
KristofferC: Yes, would be prettier but doesn't remove the boxes from what I can see. I've found type asserts to be quite ineffective when it comes to boxes.

timholy: It may not remove the boxes but with this change they don't matter. But either way is fine.

KristofferC: Alright, I changed it to the type asserts.

timholy (approved):

PR #2123, opened 2020-10-15
Wait for and terminate async download tasks to avoid a hang.

maleadt:


KristofferC:
Works for me locally, let's merge it to get CI to run on Stefans branch.

PR #2128, opened 2020-10-16
add a progress bar to precompile process

KristofferC:
https://asciinema.org/a/a26toW3opklSEtaRAeXzpcW8W

cc @ianshmean

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -992,6 +992,7 @@ function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
     ansi_movecol1 = "\e[1G"
     ansi_cleartoend = "\e[0J"
     show_report::Bool = true
+    n_done::Int = 0
 
     t_print = @async begin # fancy print loop
         try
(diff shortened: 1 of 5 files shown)


KristofferC:
I don't really understand the printing code here. It looks like there are almost two copies of the printing code. For example:

https://github.com/JuliaLang/Pkg.jl/blob/0314b791bd81ae36fee31f89e34044f16e7e8783/src/API.jl#L1052

https://github.com/JuliaLang/Pkg.jl/blob/0314b791bd81ae36fee31f89e34044f16e7e8783/src/API.jl#L1104

or

https://github.com/JuliaLang/Pkg.jl/blob/0314b791bd81ae36fee31f89e34044f16e7e8783/src/API.jl#L1094

https://github.com/JuliaLang/Pkg.jl/blob/0314b791bd81ae36fee31f89e34044f16e7e8783/src/API.jl#L1026

when are both of these used?

ianshmean:
The first two are needed because either loop could be interrupted and trigger the shutdown. 

The second two are dependent on `fancy_print`. i.e. whether the terminal can handle ansi cursor movements. In CI and IJulia etc it does a basic linear print

ianshmean:
This is great, btw!
I thought for a second that total of the progress bar should be the number of packages that actually need precompiling, but the process doesn't know that in advance.

With this approach, adding a new package to a mostly already precompiled project would show a progress bar that jumps to nearly complete instantly, so that might not be the best UX.. but it is truthful.. After all it is `Precompiling project...` and it jumping a bunch instantly wouldn't be inaccurate in that case

KristofferC:
Yeah, I was a bit unsure about that as well. I guess we could run through everything and count how much is up to date first, before doing any precompiling?

ianshmean:
I think it's good as-is. It reinforces the idea that it's the project being precompiled, and that it's checking everything in there

KristofferC:
Personally, I think it would be better to only show progress for what is actually going to happen. But maybe that is hard with the current code?

ianshmean:
Done, over in #2137 which merges into here

PR #2129, opened 2020-10-16
Precompile: Fix and test per-project precomp suspension state retention

ianshmean:
The project-specific retention of auto-precompile suspended packages wasn't quite right, and was missing tests

As for burden, the addition of `recall_suspended_packages` to the first Pkg API call adds very little on my system: `0.000206 seconds (92 allocations: 6.750 KiB)` on my machine

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1094,7 +1094,7 @@ function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
             # skip stale checking and force compilation if any dep was recompiled in this session
             any_dep_recompiled = any(map(dep->was_recompiled[dep], deps))
             is_stale = true
-            if (any_dep_recompiled || (!suspended && is_stale = _is_stale(paths, sourcepath)))
+            if (any_dep_recompiled || (!suspended && (is_stale = _is_stale(paths, sourcepath))))
                 Base.acquire(parallel_limiter)
                 is_direct_dep = pkg in direct_deps
                 iob = IOBuffer()
--- a/src/API.jl
+++ b/src/API.jl
@@ -1090,6 +1090,7 @@ function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
                     Base.acquire(parallel_limiter)
                     is_direct_dep = pkg in direct_deps
                     iob = IOBuffer()
+                    name = is_direct_dep ? pkg.name : string(color_string(pkg.name, :light_black))
                     try
                         !fancy_print && lock(print_lock) do
                             isempty(pkg_queue) && printpkgstyle(io, :Precompiling, "project...$action_help")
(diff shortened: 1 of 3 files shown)


src/API.jl
```diff
@@ -60,12 +60,18 @@ function require_not_empty(pkgs, f::Symbol)
...
+
 # Provide some convenience calls
 for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
     @eval begin
         $f(pkg::Union{AbstractString, PackageSpec}; kwargs...) = $f([pkg]; kwargs...)
         $f(pkgs::Vector{<:AbstractString}; kwargs...)          = $f([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
         function $f(pkgs::Vector{PackageSpec}; kwargs...)
+            if first_pkg_action[]
+                first_pkg_action[] = false
+                Operations.recall_suspended_packages()
```
KristofferC, line 73: I don't get why this is done here and not in e.g. `_auto_precompile` or `precompile`.

ianshmean: It needs to be recalled before any pkg actions that could change the manifest and unsuspend a suspension.
Say a user does this:
- `pkg> add Foo` which installs v0.1.0, which fails during auto-precomp and is suspended
- Restarts julia
- `pkg> up` which updates Foo to v0.1.1, which should fix the package

If we recall state for the first time after the `up` during the auto-precomp, it will restore the old list that says `Foo` should be ignored.

Basically it needs to be recalled at the activation of the environment, but `Pkg.activate` isn't used to activate the first environment.. that seems to be done in Base, where we can't run `recall_suspended_packages()`

So I tack it onto the first (and only first) time the key Pkg actions are run, and subsequently any time `Pkg.activate` is done

ianshmean:
This is much better now. The suspension pays attention to a much more specific `PackageSpec` that includes the `tree_hash`, so we don't need to explicitly free a package if it's updated.. it will just have a different `PackageSpec`, so won't be blocked.

The only thing is that the project-specific lists of suspended packages will just grow with failures, unless the user manually calls `pkg> precompile` which wipes the list.
That would likely make the de/serialization slow down, but not sure how much, maybe it's insignificant.

I'm not aware of another time we could wipe the list. Perhaps we could put a tip somewhere about `pkg> precompile` trying harder

src/API.jl
ianshmean: OK, this is fixed now. See below

KristofferC:
> The only thing is that the project-specific lists of suspended packages will just grow with failures, unless the user manually calls pkg> precompile which wipes the list.

Why would the list contain anything more than the packages that are currently known to fail to precompile and are also in that project? Couldn't every precompile (including auto) clear out packages that are no longer in the project?

ianshmean:
> Why would the list contain anything more than the packages that are currently known to fail to precompile and are also in that project? Couldn't every precompile (including auto) clear out packages that are no longer in the project?

This hadn't occurred to me. It's now implemented 👍🏻

PR #2134, opened 2020-10-17
send Pkg server value itself as a header to the Pkg server

StefanKarpinski:
This is useful because there may be redirection going on and it may not be clear how someone got to a particular pkg server. Setting this header is not a risk even though it captures the content of the `JULIA_PKG_SERVER` environment variable, because if the content of the variable happens to be something sensitive instead of the address of a package server, then the request will never reach the package server in the first place, let alone the header.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -639,6 +639,7 @@ function get_metadata_headers(url::AbstractString)
     server_dir = get_server_dir(url, server)
     server_dir === nothing && return headers
     push!(headers, "Julia-Pkg-Protocol" => "1.0")
+    push!(headers, "Julia-Pkg-Server" => server)
     push!(headers, "Julia-Version" => string(VERSION))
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
I don't get why the Travis run appears as cancelled and the AppVeyor failure doesn't seem like it can possibly be related since all this PR does is set an extra header on download requests to Pkg servers.

DilumAluthge:
AppVeyor tests actually passed (see https://ci.appveyor.com/project/JuliaLang/pkg-jl/builds/35809291#L1099). There was some kind of issue when uploading code coverage to CodeCov; probably a network issue, I've seen it when uploading to CodeCov from GH Actions as well.

Travis tests actually passed as well. It's just the docs build that got "cancelled", which was probably either a Travis error or a timeout of some kind.

KristofferC (approved):

PR #2135, opened 2020-10-17
get rid of the mode field in PackageSpec

KristofferC:
This is never used in Pkg individually, but always as something applied to all packages. Pretty pointless to store that in every PackageSpec.


00vareladavid:
These changes are in #2454

PR #2138, opened 2020-10-18
Add the public `Pkg.active_pkg_server` function, and (only if on CI) print the active Pkg server once

DilumAluthge:
It's pretty common for people to share Pkg logs when they are trying to figure out whether or not their problem is a Pkg server problem or something else. Sometimes people share logs from their local machines, and other times people share logs from one or more CI providers.

It can be difficult to figure out which Pkg server is actually being used. If the user is on their local machine, we can try to guess based on the user's geographic location. If this is a log from CI, then we have to figure out whether the particular CI job was using the CI-specific Pkg server or one of the regular Pkg servers.

This pull request tries to fix this problem. One time per Julia session, if you are running on CI, Pkg will print the active Pkg server.

Example output looks like this:
```julia
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.6.0-DEV.1275 (2020-10-18)
 _/ |\__'_|_|_|\__'_|  |  Commit 44b55d1584 (0 days old master)
|__/                   |

julia> import Pkg

julia> Pkg.update()
        Info The active Pkg server is: https://us-east2.pkg.julialang.org/
No Changes to `~/Downloads/Pkg.jl/Project.toml`
No Changes to `~/Downloads/Pkg.jl/Manifest.toml`

julia> Pkg.update()
No Changes to `~/Downloads/Pkg.jl/Project.toml`
No Changes to `~/Downloads/Pkg.jl/Manifest.toml`

julia> Pkg.update()
No Changes to `~/Downloads/Pkg.jl/Project.toml`
No Changes to `~/Downloads/Pkg.jl/Manifest.toml`
```

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -20,6 +20,8 @@ using ..Pkg.Versions
 import Base: SHA1
 using SHA
 
+import Downloads
+
 export UUID, SHA1, VersionRange, VersionSpec,
     PackageSpec, EnvCache, Context, PackageInfo, ProjectInfo, GitRepo, Context!, Manifest, Project, err_rep,
     PkgError, pkgerror, has_name, has_uuid, is_stdlib, stdlibs, write_env, write_env_usage, parse_toml, find_registered!,
@@ -1077,11 +1079,35 @@ function remove_registries(io::IO, regs::Vector{RegistrySpec})
     return nothing
 end
 
+function active_pkg_server()
+    server = pkg_server()
+    if server !== nothing
+        request = Downloads.Request(devnull, server, Vector{Pair{String,String}}(undef, 0))
+        response = Downloads.request(request)
+        active_server = response.url
+        return active_server
+    end
+    return nothing
+end
+
+function print_active_pkg_server(ctx::Context)
+    active_server = active_pkg_server()
+    if active_server !== nothing
+        printpkgstyle(ctx, :Info, "The active Pkg server is: " * active_server)
+    end
+    return nothing
+end
+
+function is_ci()
+    return lowercase(get(ENV, "CI", "")) == "true"
+end
+
 # entry point for `registry up`
 function update_registries(io::IO, regs::Vector{RegistrySpec} = collect_registries(depots1());
                            force::Bool=false)
     Pkg.OFFLINE_MODE[] && return
     !force && UPDATED_REGISTRY_THIS_SESSION[] && return
+    UPDATED_REGISTRY_THIS_SESSION[] || (is_ci() && print_active_pkg_server(ctx))
     errors = Tuple{String, String}[]
     registry_urls = nothing
     for reg in unique(r -> r.uuid, find_installed_registries(io, regs); seen=Set{Union{UUID,Nothing}}())
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -543,6 +543,19 @@ Below is a comparison between the REPL version and the API version:
 """
 const RegistrySpec = Types.RegistrySpec
 
+"""
+active_pkg_server()
+
+Get the active Pkg server.
+
+## Example
+
+```julia
+julia> Pkg.active_pkg_server()
+"https://us-east.pkg.julialang.org/"
+```
+"""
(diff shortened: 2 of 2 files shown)


DilumAluthge:
The current implementation uses command-line `curl` if it is available. Ideally, we would instead use `Downloads` (or another stdlib), but I'm not sure how to do that.

DilumAluthge:
@StefanKarpinski Can you review this?

fredrikekre:
I think this will feel spammy pretty fast.

DilumAluthge:
> Perhaps it can be printed only on CI? And perhaps in combination with the `versioninfo()` call that is usually already there? E.g. here https://github.com/julia-actions/setup-julia/blob/c5d801f77bd83a046b600a4b42e942cc70e00fb1/src/setup-julia.ts#L80 and similarly for other providers.

Sure. Instead of patching each CI provider manually (idk how to even do it for Travis or AppVeyor), we could just put `if get(ENV, "CI", "") == "true"` in this PR.

DilumAluthge:
> I think this will feel spammy pretty fast and it seems pretty unlikely that a user happens to get an error on first use and hence include this in a report/discourse post.

How about we make public (document and export) the `active_pkg_server` function? That way, when users do report having issues, we can say, "Can you post the output of `Pkg.active_pkg_server()`?"

StefanKarpinski:
Isn't that already available as `Pkg.pkg_server()`?

src/Types.jl
```diff
@@ -1098,11 +1100,31 @@ function remove_registries(ctx::Context, regs::Vector{RegistrySpec})
     return nothing
 end
 
+function active_pkg_server()
+    server = pkg_server()
+    if server !== nothing
+        request = Downloads.Request(devnull, server, Vector{Pair{String,String}}(undef, 0))
```
StefanKarpinski, line 1106: This is not a stable API and I'm [about to break it](https://github.com/JuliaLang/Downloads.jl/pull/46), so we definitely shouldn't merge before that.

StefanKarpinski:
Oh, I see, you want to find out which server the user is redirected to. I was just thinking about doing some fancy client-side server selection based on pinging the servers so that if one server is down we pick another one.

DilumAluthge:
> Isn't that already available as `Pkg.pkg_server()`?

`Pkg.pkg_server()` will return e.g. `pkg.julialang.org`

`active_pkg_server()` returns e.g. `us-east.pkg.julialang.org` or `us-east-ci.pkg.julialang.org` or whatever, which is useful for figuring out whether an issue is related to Pkg server downtime or something else.

It's not always intuitive- I recently saw someone on Slack who is in South America but who gets routed to `us-east` I think `Pkg.active_server()` is probably easier than playing the "where do you live" game.

DilumAluthge:
> Oh, I see, you want to find out which server the user is redirected to. I was just thinking about doing some fancy client-side server selection based on pinging the servers so that if one server is down we pick another one.

That also seems like a good idea to do.

src/Types.jl
DilumAluthge: I can switch to using a different method. Is there a public API in Downloads that will expose the same info?

DilumAluthge:
> > Oh, I see, you want to find out which server the user is redirected to. I was just thinking about doing some fancy client-side server selection based on pinging the servers so that if one server is down we pick another one.
> 
> That also seems like a good idea to do.

We already have the concept of Pkg server siblings. So we could pull the list of siblings, compare it against the active server (the one the user is redirected to), and if the active server is down, pick a different sibling.

DilumAluthge:
Alright, here's what I've done:
1. We will only automatically print the active Pkg server if we are on CI
2. I've added a docstring for `active_pkg_server` and made it available as `Pkg.active_pkg_server`.

So on CI, it will automatically print. It won't automatically print otherwise, but users can easily access it with `Pkg.active_pkg_server()`.

DilumAluthge:
It might be easier to put this in the `setup-Julia` GitHub Action.

PR #2143, opened 2020-10-20
Precomp: Better error handling, hide cursor during print

ianshmean:
- Improves specificity of error handling ~~with a new `Base.PrecompileFailedException` type thrown by `Base.compilecache`. Dependent on https://github.com/JuliaLang/julia/pull/38109~~

- Hides cursor during pretty printing

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1060,18 +1060,18 @@ function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
                 i += 1
                 wait(t)
             end
-            println(io, ansi_enablecursor)
         catch err
             interrupted = true
             show_report = false
             if err isa InterruptException
                 lock(print_lock) do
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -1057,16 +1060,17 @@ function precompile(ctx::Context; internal_call::Bool=false, io::IO=stderr)
                 i += 1
                 wait(t)
             end
+            println(io, ansi_enablecursor)
```
KristofferC, line 1063: Couldn't this be done in a `finally`?

ianshmean: Indeed. Thanks

KristofferC:
Could we update this for now not to depend on JuliaLang/julia#38109?

ianshmean:
Done

ianshmean:
Just a CI glitch on one of the MacOS machines

PR #2147, opened 2020-10-21
PlatformEngines: use Tar.jl for tarball operations

StefanKarpinski:


StefanKarpinski:
Hard to tell if the AppVeyor failures are because I broke something or just network flakiness.

StefanKarpinski:
I don't believe that the AppVeyor CI failure is caused by this change, I think it's caused by https://github.com/JuliaLang/julia/pull/38070.

KristofferC:
What was the Windows problem in the end?

src/PlatformEngines.jl
```diff
@@ -888,19 +365,10 @@ end
 Compress `src_dir` into a tarball located at `tarball_path`.
 """
 function package(src_dir::AbstractString, tarball_path::AbstractString)
```
KristofferC, line 367: This is also test only, right?

StefanKarpinski: Not sure, I thought the Artifacts code used it too, but I could be wrong.

StefanKarpinski:
It was these two things:

https://github.com/JuliaLang/Pkg.jl/pull/2147/files#diff-dffc9e68c799ed23b244d9d42f8ae9c56136907a1e036638272b7d07448c10f1

https://github.com/JuliaLang/Pkg.jl/pull/2147/files#diff-9a6a909fb1a0128f652bc4ca79a4df77d5dbb6bafeac08d5cc2e6e9be9bc07fb

yuyichao:
This broke system 7z build!

KristofferC:
https://github.com/JuliaLang/Pkg.jl/issues/2181

PR #2152, opened 2020-10-22
fix get_auth_header: headers are now string pairs

StefanKarpinski:
Missing change from https://github.com/JuliaLang/Pkg.jl/pull/2078


KristofferC:
Ref https://github.com/JuliaLang/Pkg.jl/pull/2148

StefanKarpinski:
Hah! My CI luck was better than yours 😝

PR #2160, opened 2020-10-23
Don't error if a package does not have a `version`

### LINKED ISSUE #2154: Can't add/update packages that have no version in Project.toml
After #2133, you can't add a package that has no version. If you already had one installed, updating will now fail with the same error.

To reproduce, create a package with `generate`, remove the `version` in its `Project.toml`, init a Git repo, and commit. Then run `add` on that path.

```jl
(@v1.6) pkg> add ./Foo
    Updating git-repo `/home/degraafc/scratch/Foo`
    Updating registry at `~/.local/share/julia/registries/General`
   Resolving package versions...
ERROR: project file for Foo is missing a `version` entry
```

Found this out via Nanosoldier which has `BaseBenchmarks` installed, who has no version:

```jl
(@v1.6) pkg> st
Status `~/.julia/environments/v1.6/Project.toml`
  [d7f09723] BaseBenchmarks v0.0.0+ `~/.julia/dev/BaseBenchmarks`
  [6e4b80f9] BenchmarkTools v0.5.0
  [34da2185] Compat v3.21.0
  [682c06a0] JSON v0.21.1

(@v1.6) pkg> up
ERROR: project file for BaseBenchmarks is missing a `version` entry
```

DilumAluthge:
Fixes #2154 
Closes #2161 

@KristofferC Would this be sufficient to fix #2154?

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -263,7 +263,8 @@ function collect_project!(ctx::Context, pkg::PackageSpec, path::String,
     if project.version !== nothing
         pkg.version = project.version
     else
-        pkgerror("project file for $(pkg.name) is missing a `version` entry")
+        @warn("project file for $(pkg.name) is missing a `version` entry")
+        pkg.version = VersionNumber(0)
     end
     return
 end


KristofferC:
Yep, this should work. Not sure about the warning but let's keep it for now.

PR #2162, opened 2020-10-23
Use `Base.Filesystem.prepare_for_deletion()` instead of defining our own

staticfloat:
Note that this function is only available in Julia 1.6+


fredrikekre:
Should the definition be backported to 1.5 (https://github.com/JuliaLang/Pkg.jl/pull/2150)

staticfloat:
Yeah, I think that's a good idea.

KristofferC (approved):

PR #2163, opened 2020-10-24
change the way compat is uncompressed

KristofferC:
The registry has two files `Deps.toml` and `Compat.toml`. This is (I presume!) to make it easier for a human to read the `Compat.toml` file since one can read package names and mostly ignore UUIDs. However, this has also "leaked" into the implementation in Pkg where two dictionaries are created and passed along, and then finally in the Resolver the mapping is done to achieve the finally desired `UUID => VersionRange` compat. This changes it so that the uncompression creates this final data structure at once instead of having the resolver do that work.


StefanKarpinski:
Yeah, that's much simpler.

PR #2167, opened 2020-10-25
Only do fancy ansi printing in terminals that support it, specifically not on CI

ianshmean:
I've noticed ansi cursor codes being printed as text in CI, and it may be happening in IJulia etc.

This introduces a general setting for whether ansi overprinting and more advance cursor movements should be used, based on the heuristic:
```
(io isa Base.TTY) && (get(ENV, "CI", nothing) != "true")
```

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -34,6 +34,8 @@ const UPDATED_REGISTRY_THIS_SESSION = Ref(false)
 const OFFLINE_MODE = Ref(false)
 const DEFAULT_IO = Ref{Union{Nothing,IO}}(nothing)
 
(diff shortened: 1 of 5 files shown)


src/Pkg.jl
```diff
@@ -549,6 +550,7 @@ function __init__()
         end
     end
     OFFLINE_MODE[] = get(ENV, "JULIA_PKG_OFFLINE", nothing) == "true"
+    FANCY_PRINT[] = (stdout isa Base.TTY) && (get(ENV, "CI", nothing) != "true")
```
KristofferC, line 553: Imo this should be a function on `io` and the `ctx.io` or e.g. `stderr` should be passed to it.

ianshmean: Good point. Changed

PR #2169, opened 2020-10-25
remove unused Artifact struct

KristofferC:
At least I think so... CI will tell.


KristofferC:
Or is this for some backwards compat? cc @staticfloat

staticfloat:
I don’t think this is used anymore; please feel free to remove with my blessing.

PR #2170, opened 2020-10-28
PlatformEngines: use new download progress API

StefanKarpinski:
Let's see if the nightlies are new enough for this...


StefanKarpinski:
You're going to need this on 1.6 if it's not already there. Otherwise progress bars will break since I've already removed the now-legacy progress API from the Downloads version that master uses.

KristofferC:
It's already on the release-1.6 branch

DilumAluthge:
Kristoffer: fastest backporter in the West

PR #2171, opened 2020-10-28
use the TOML invalidation mechanism in Base in more places in Pkg

KristofferC:
Edit: This won't work right now since we mutate the dictionary that we cache... Ugh.


KristofferC:
But for now, we can just make a copy of the cached values. That's a lot faster than hitting disc anyway.

PR #2175, opened 2020-10-29
Windows: prefer git cloning registries (close #2014)

### LINKED ISSUE #2014: Pkg 1.5.2: clone git registries by default on Windows
See https://discourse.julialang.org/t/status-of-pkg-speed-improvements-outside-us-in-v1-5/46395/17. In 1.5.2, we should clone registries by default on Windows since unpacking tarballs with lots of small files is so slow there.

StefanKarpinski:
Since untarring a tarball like the registry is so slow on Windows,
we prefer to git clone registries instead. Git is slow on Windows
as well, but git updates are faster since git only updates the
files that need to be modified, which is much faster than unpacking
a whole new registry.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -974,7 +974,8 @@ function clone_or_cp_registries(ctx::Context, regs::Vector{RegistrySpec}, depot:
         # clone to tmpdir first
         mktempdir() do tmp
             url, registry_urls = pkg_server_registry_url(reg.uuid, registry_urls)
-            if url !== nothing
+            # on Windows we prefer git cloning because untarring is so slow
+            if !Sys.iswindows() && url !== nothing
                 # download from Pkg server
                 try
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Can we backport this to 1.6?

StefanKarpinski:
Can and should. To 1.5 also, if we're going to do any more releases off that version (which we should, imo).

KristofferC:
I wonder if not https://github.com/JuliaLang/Pkg.jl/pull/2186 caused a lot of the slowness..., not untarring. Probably a combination of both.

StefanKarpinski:
Probably both.

j-fu:
Found this one: https://github.com/microsoft/WinDev/issues/27 - there seems to be a  possible communication channel to the windows devs.

PR #2176, opened 2020-10-30
Add the `force_latest_compat` keyword argument to `Pkg.test`

DilumAluthge:
This pull request adds the `force_latest_compat` keyword argument to the `Pkg.test` function.

When `force_latest_compat` is `true`, then, for each `[compat]` entry in the active project, only the latest compatible version will be allowed.

The default value is `force_latest_compat = false`.

### Motivation

Suppose that you have a bot (e.g. CompatHelper or Dependabot) that opens pull requests to widen your `[compat]` entries whenever one of your dependencies makes a new breaking release. 

And suppose that originally your package's `Project.toml` file looks like this:
```toml
[compat]
SomePackage = "1"
```

The bot opens a PR to change your package's `Project.toml` file to this:
```toml
[compat]
SomePackage = "1, 2"
```

Now, your CI runs your test suite with this new `Project.toml` file. The problem is that, during the test suite, `Pkg.test` might choose to run your package's tests with version 1 of SomePackage.jl. Therefore, even if your CI tests pass, you don't know whether or not your package is compatible with version 2 of SomePackage.jl.

The solution is to modify your CI test script. Instead of using this as your test script (in Travis CI, AppVeyor CI, GitHub Actions CI, etc.):
```julia
Pkg.test(coverage=true)
```

You would have this as your CI test script instead:
```julia
Pkg.test(coverage=true, force_latest_compat=true)
```

This will guarantee that your tests will run with version 2 of SomePackage.jl.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -10,8 +10,10 @@ import REPL
 using REPL.TerminalMenus
 using ..Types, ..Resolve, ..PlatformEngines, ..GitTools, ..MiniProgressBars
 import ..depots, ..depots1, ..devdir, ..set_readonly, ..Types.PackageEntry
-import ..Artifacts: ensure_artifact_installed, artifact_names, extract_all_hashes,
-                    artifact_exists, select_downloadable_artifacts
+import ..Types.latest_compat, ..Types.force_latest_compat
+import ..Artifacts: ensure_artifact_installed, ensure_all_artifacts_installed,
+                    artifact_names, extract_all_hashes, artifact_exists,
+                    select_downloadable_artifacts
 using Base.BinaryPlatforms
 import ...Pkg
 import ...Pkg: pkg_server, Registry, pathrepr, can_fancyprint, printpkgstyle, DEFAULT_IO
@@ -1423,7 +1425,7 @@ end
 
 # ctx + pkg used to compute parent dep graph
 function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::String,
-                 sandbox_path::String, sandbox_project_override)
+                 sandbox_path::String, sandbox_project_override; force_latest_compat::Bool=false)
     active_manifest = manifestfile_path(dirname(ctx.env.project_file))
     sandbox_project = projectfile_path(sandbox_path)
 
@@ -1436,6 +1438,10 @@ function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::S
             Types.write_project(sandbox_project_override, tmp_project)
         elseif isfile(sandbox_project)
             cp(sandbox_project, tmp_project)
+            chmod(tmp_project, 0o600)
+        end
+        if force_latest_compat
+            Types.force_latest_compat(tmp_project)
             chmod(tmp_project, 0o600)
         end
         # create merged manifest
@@ -1566,7 +1572,7 @@ testdir(source_path::String) = joinpath(source_path, "test")
 testfile(source_path::String) = joinpath(testdir(source_path), "runtests.jl")
 function test(ctx::Context, pkgs::Vector{PackageSpec};
               coverage=false, julia_args::Cmd=``, test_args::Cmd=``,
-              test_fn=nothing)
+              test_fn=nothing, force_latest_compat::Bool=false)
     Pkg.instantiate(ctx; allow_autoprecomp = false) # do precomp later within sandbox
 
     # load manifest data
@@ -1602,7 +1608,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
             gen_target_project(ctx.env, ctx.registries, pkg, source_path, "test")
         # now we sandbox
         printpkgstyle(ctx.io, :Testing, pkg.name)
-        sandbox(ctx, pkg, source_path, testdir(source_path), test_project_override) do
+        sandbox(ctx, pkg, source_path, testdir(source_path), test_project_override; force_latest_compat=force_latest_compat) do
             test_fn !== nothing && test_fn()
             sandbox_ctx = Context()
             status(sandbox_ctx.env; mode=PKGMODE_COMBINED, io=sandbox_ctx.io)
--- a/src/force-latest-compat.jl
+++ b/src/force-latest-compat.jl
@@ -1,0 +1,54 @@
+using TOML
+
+function latest_compat(range::VersionRange)
+    lower = range.lower
+    upper = range.upper
+    t_lower = Int.(lower.t[1:lower.n])
+    t_upper = Int.(upper.t[1:upper.n])
+    # `t_lower` is a tuple of integers
+    # `t_upper` is a tuple of integers
+    # `t_lower` and `t_upper` do not necessarily have the same length
+    if VersionNumber(t_lower) > VersionNumber(t_upper)
+        return t_lower
+    else
+        return t_upper
+    end
+end
+
+function latest_compat(s::VersionSpec)
+    ts = latest_compat.(s.ranges)
+    # `ts` is a vector of tuples of integers
+    # the tuples in `ts` may have different lengths
+    vers = VersionNumber.(ts)
+    i = argmax(vers)
+    t = ts[i]
+    new_compat_entry = join(string.(t), ".")
+    return new_compat_entry
+end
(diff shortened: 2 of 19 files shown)


KristofferC:
If this is decided to be a good idea, I think an option to `Pkg.test` makes more sense instead of actual rewriting of the project file.

DilumAluthge:
Sure, that would be fine. We could have an optional keyword argument `force_latest_compat`. Where in the `Pkg.test` code would I implement this?

I definitely think that we need this feature somewhere. Otherwise, you don't actually know if your CompatHelper/Dependabot PRs are passing CI.

DilumAluthge:
@KristofferC Alright, it is now a keyword argument to `Pkg.test`. Example usage:
```julia
Pkg.test(coverage = true, force_latest_compat = true)
```

DilumAluthge:
Bump. @KristofferC take a look at the new implementation and let me know what you think.

DilumAluthge:
Bump.

src/force-latest-compat.jl
```diff
@@ -0,0 +1,47 @@
+using TOML
+
+@inline function latest_compat(range::VersionRange)
```
KristofferC, line 3: Random question but why is everything here inlined?

DilumAluthge: Not sure; I'm sure it made sense in my head when I wrote it.

Fixed now.

src/Operations.jl
```diff
@@ -1405,6 +1406,10 @@ function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::S
             cp(sandbox_project, tmp_project)
             chmod(tmp_project, 0o600)
         end
+        if force_latest_compat
+            Types.force_latest_compat(tmp_project)
```
KristofferC, line 1410: Why reparse the project and resave it here, can't you  tweak the compat before it is written?

src/force-latest-compat.jl
```diff
@@ -0,0 +1,47 @@
...
+    if l_ver > u_ver
+        return l_t, l_ver
+    else
+        return u_t, u_ver
+    end
+end
+
+function latest_compat(s::VersionSpec)
+    x = latest_compat.(s.ranges)
+    return join(string.(x[argmax(getindex.(x, 2))][1]), ".")::String
```
KristofferC, line 19: This is quite hard to understand what it does.

DilumAluthge: Agreed.

I'll break it up into multiple lines to make it clearer what's going on.

src/Operations.jl
DilumAluthge: I'm not sure I follow. There are two possible places that the project could come from:
1. `Types.write_project(sandbox_project_override, tmp_project)` - line 1404
2. `cp(sandbox_project, tmp_project)` - line 1406

For case 1, I'd have to modify the `Types.write_project` function and do the compat-tweaking there. And then for case 2, I'd have to reparse and resave the project anyway, right?

So it seemed simpler to just have the compat-tweaking in a single place.

Maybe I'm misunderstanding what you're suggesting?

src/force-latest-compat.jl
DilumAluthge: Alright, take a look now.

christopher-dG:
Just to put what's been said in Slack somewhere permanent:

It might be a good idea to make forcing latest versions of dependencies the default `Pkg.test` behaviour in CI environments for PRs created by CompatHelper or Dependabot. For example, in GitHub Actions the environment variable `GITHUB_REF` will be `refs/heads/compathelper/new_version/...`.

Personally, I don't see this getting much buy-in if it needs to be done manually.

DilumAluthge:
> Just to put what's been said in Slack somewhere permanent:
> 
> It might be a good idea to make forcing latest versions of dependencies the default `Pkg.test` behaviour in CI environments for PRs created by CompatHelper or Dependabot. For example, in GitHub Actions the environment variable `GITHUB_REF` will be `refs/heads/compathelper/new_version/...`.
> 
> Personally, I don't see this getting much buy-in if it needs to be done manually.

I agree.

Do you think that should go in this PR, or a follow-up PR? I was thinking that we first get this PR merged, and then I'll make a follow-up PR to add the automatic CI detection.

The alternative approach would just to be to set `force_latest_compat = true` by default and call it a day.

christopher-dG:
If we're going to make latest compat default all the time, it's basically the same as just not having the option to at all. Not a good idea IMO. It doesn't really matter to me in which PR stuff happens, just as long as it doesn't get delayed a whole minor Julia version. If it's easier to do it after this is merged, then sure.

DilumAluthge:
How about this: I'll put the CI changes in a new PR targeting this branch. So Kristoffer and others can review the two PRs separately. But if they do want to review everything at once, they can merge that PR into this one.

I'll get the CI changes PR up later today.

DilumAluthge:
Okay, the "set `force_latest_compat=true` on Dependabot/CompatHelper" pull request is up here: #2234 

@KristofferC Do you want to review these two pull requests (#2176 and #2234) separately? Or do you want me to merge them into a single branch, and then you can review it as a single pull request?

DilumAluthge:
Based on the discussion in #2234, I have closed #2234, and the "auto-detect Dependabot/CompatHelper" functionality has been moved into https://github.com/julia-actions/julia-runtest/pull/20.

src/force-latest-compat.jl
DilumAluthge: I think this is resolved now. The new version of the function splits things up into multiple lines, and should be much more readable.

Let me know if this is not resolved.

quinnj:
Bump; this seems really useful for CompatHelper

DilumAluthge:
Any additional thoughts or review comments on this PR? If not, I'm inclined to wait a little bit, and then merge this PR.

src/force-latest-compat.jl
```diff
@@ -0,0 +1,54 @@
...
+
+function force_latest_compat(compat_old::AbstractDict{<:AbstractString, <:Any})
+    compat_new = Dict{String,Any}()
+    for (pkg_name, old_compat_entry) in pairs(compat_old)
+        compat_new[pkg_name] = latest_compat(old_compat_entry)
+    end
+    return compat_new
+end
+
+function force_latest_compat(filename::AbstractString)
```
KristofferC, line 42: This function looks kinda pointless. We already have functions to read and write projects. What you want to do is just change the compat of an existing project. So the only thing that should be modified is https://github.com/JuliaLang/Pkg.jl/blob/8b0045daf928e55e06a4701d959caba2487bd63f/src/Types.jl#L210.

src/Operations.jl
```diff
@@ -1438,6 +1440,10 @@ function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::S
             cp(sandbox_project, tmp_project)
             chmod(tmp_project, 0o600)
         end
+        if force_latest_compat
```
KristofferC, line 1443: Modify the `sandbox_project_override.env.project.compat` before it is written (at line 1437) and the correct compat should just be printed automatically. No need to write to disk, chmod, read it back in, modify compat, write it back out to disk, chmod again.

src/force-latest-compat.jl
DilumAluthge: Sure, we can get rid of the `force_latest_compat(filename::AbstractString)` method. However, we would just end up moving this code (the code to load the project from the `Project.toml` file) from this location to line 1443 of `src/Operations.jl`.

If you look at lines 1437 through 1443 of `src/Operations.jl`: the temp project that is going to be used for the test sandbox is in the form of a `Project.toml` file that has been written to disk at the filename `tmp_project`. This temp project does not exist in memory yet; it is only a project file on disk. So we will need to read it in from the file.

src/Operations.jl
DilumAluthge: That works if `sandbox_project_override !== nothing`. But what if we are in the other branch, i.e. `isfile(sandbox_project)`?

DilumAluthge:
See #2439 for an alternative to this PR.

DilumAluthge:
Closing in favor of #2439

PR #2179, opened 2020-10-30
remove a bunch of ctx usage, part 1

KristofferC:
The `Context` struct has become something kinda ridiculous where it is passed into almost any function and then one just extract the tiny thing inside it for using it in the function. This begins the tedious process of reversing this. My main goal with 1.7 is to decouple and make the "flow" of information through Pkg more clear and to use less mutation, in order to more easily enable new features to be made. This `Context` is a road block for that since it ties everything together so let's try reduce it's usage as much as possible. Perhaps at some point it can go away completely.


src/API.jl
```diff
@@ -1423,7 +1423,7 @@ function activate(path::AbstractString; shared::Bool=false, temp::Bool=false)
     p = Base.active_project()
     if p !== nothing
         n = ispath(p) ? "" : "new "
-        printpkgstyle(Context(), :Activating, "$(n)environment at $(pathrepr(p))")
+        printpkgstyle(stdout, :Activating, "$(n)environment at $(pathrepr(p))")
```
KristofferC, line 1426: Should probably use the default IO instead to make it not print in tests.

KristofferC:
Windows timed out but got almost all the way through. Linux and mac passed so I think this is fine.

PR #2183, opened 2020-10-31
move all registry handling under the Registry module

KristofferC:
The current registry handling is spread out a bit all over the place. We have `Registry.jl` that does a bit off stuff, then calls into registry function in `Types.jl`, and there used to be registry parsing in `Operations.jl`.

With https://github.com/JuliaLang/Pkg.jl/pull/2179 and https://github.com/JuliaLang/Pkg.jl/pull/2072 we can now move pretty much all registry handling under a separate `Registry` module that does not have any ties to `Pkg.Types` (the exception being `PkgError` but we could move that later). It just uses some utility functions in Pkg, download (git and normal).

In https://github.com/JuliaLang/Pkg.jl/pull/2072 I chose the struct name `Registry` but that collides with the `Registry` module that is already documented. So I changed the name of it to `RegistryInstance` here.

This PR also moves a few helper functions from `Types.jl` into `utils.jl` and move some struct definitions from `Types.jl` to where they are used.

The next step is to do something similar for Environments, then https://github.com/JuliaLang/Pkg.jl/pull/2103 should be easier to finish.


KristofferC:
I'm tired of rebasing this so let's merge :P.

PR #2186, opened 2020-11-01
remove duplicated tests and move registry into place instead of copying

KristofferC:
Makes a significant impact on CI time (and probably on user time) on especially Windows.


KristofferC:
28 min -> 17 min CI time on apple with Pkg server, 37 min -> 21 without... lol.

PR #2187, opened 2020-11-01
CI: make registry available in temp_pkg_dir

KristofferC:
Takes off another chunk on mac builder CI time. Let's see how it does on Windows.


KristofferC:
Windows 30 min to 20 min.

PR #2192, opened 2020-11-03
fix artifacts API reference

### LINKED ISSUE #2131: missing docstrings in artifacts API reference
Documenter complains about some docstrings missing [here](https://julialang.github.io/Pkg.jl/dev/api/#Artifacts-Reference). It seems like `@artifact_str` at least does actually have a docstring, so it looks like there's something else that's broken.

simeonschaub:
This fixed the issue for me locally.
fixes #2131

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -2,6 +2,7 @@
 
 using Documenter, DocumenterLaTeX
 using Pkg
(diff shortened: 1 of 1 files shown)


KristofferC:
Maybe we can just remove the ones that come from the Artifacts stdlib now?

PR #2198, opened 2020-11-10
Allow `7z` selection to pick something up off of the PATH

staticfloat:
This uses `Sys.which()` to search the `PATH` for an available `7z`,
preferring the one shipped by default with normal Julia distributions.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -16,11 +16,22 @@ const exe7z = Ref("")
 function __init__()
     exe7z[] = ""
     name = "7z"
-    Sys.iswindows() && (name = "$name.exe")
-    for dir in (joinpath("..", "libexec"), ".")
-        path = normpath(Sys.BINDIR, dir, name)
-        isfile(path) || continue
-        exe7z[] = path
+    if Sys.iswindows()
+        name = "$name.exe"
+    end
+
+    paths = [
+        joinpath(Sys.BINDIR, Base.LIBEXECDIR),
+        Sys.BINDIR,
+    ]
+    PATH = get(ENV, "PATH", nothing)
+    if PATH !== nothing
+        push!(paths, PATH)
+    end
+    pathsep = @static Sys.iswindows() ? ";" : ":"
+
+    withenv("PATH" => join(paths, pathsep)) do
+        exe7z[] = something(Sys.which(name), "")
     end
     isempty(exe7z[]) && error("7z binary not found")
 end


src/PlatformEngines.jl
```diff
@@ -16,11 +16,22 @@ const exe7z = Ref("")
 function __init__()
     exe7z[] = ""
     name = "7z"
-    Sys.iswindows() && (name = "$name.exe")
-    for dir in (joinpath("..", "libexec"), ".")
-        path = normpath(Sys.BINDIR, dir, name)
-        isfile(path) || continue
-        exe7z[] = path
+    if Sys.iswindows()
```
KristofferC, line 19: Tangential question, but what is the reason for doing this in `init` instead of lazily at first call?

staticfloat: Probably to avoid the pain point that we had with `probe_platform_engines!()`; when other tooling wanted to use the stuff in PlatformEngines but didn't want to call the same functions that would lazily-initialize things, we'd only notice that something would break when that function would be called after not calling a `Pkg.add()` operation first, or similar.

I think in this case would be simple enough to add a `get_7z_exe()` function or somesuch though.

KristofferC: Yeah, I didn't mean to have a dedicated function to initialize. Just initialize when it is called the first time. I can PR that, np.

StefanKarpinski: My suggestion would be to have an `exe7z()` function that does this logic the first time it's called and then saves the path in a global and returns it from there on. Then we can just call `exe7z()` wherever we use this.

KristofferC: https://github.com/JuliaLang/Pkg.jl/pull/2204 maybe. Made towards this branch. cc @staticfloat

src/PlatformEngines.jl
```diff
@@ -16,11 +16,22 @@ const exe7z = Ref("")
...
+        name = "$name.exe"
+    end
+
+    paths = [
+        joinpath(Sys.BINDIR, Base.LIBEXECDIR),
+        Sys.BINDIR,
+    ]
+    PATH = get(ENV, "PATH", nothing)
+    if PATH !== nothing
+        push!(paths, PATH)
```
StefanKarpinski, line 29: Why are you pushing the value of `PATH` onto `paths` when it's a list of directories to look in?

StefanKarpinski (commented):
This seems unnecessarily convoluted and fragile. We should really avoid changing the environment if we can. Also, what happens if `Sys.BINDIR` has `pathsep` in it? This patch seems simpler and more reliable:
```diff
diff --git a/src/PlatformEngines.jl b/src/PlatformEngines.jl
index fd25ae6f..8f266258 100644
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -21,6 +21,10 @@ function __init__()
         isfile(path) || continue
         exe7z[] = path
     end
+    if isempty(exe7z[])
+        path = Sys.which(name)
+        path !== nothing && (exe7z[] = path)
+    end
     isempty(exe7z[]) && error("7z binary not found")
 end
 __init__()
```

StefanKarpinski:
I made my own version of this PR with @KristofferC's lazy amendment included.

Issue #2201, opened 2020-11-11
MethodError: no method matching registry_info(::Base.UUID)


fredrikekre:
Probably related to the recent registry handling work.

```
julia> Pkg.add("Example")
    Updating registry at `~/.julia/registries/General`
    Updating registry at `~/.julia/registries/Staging`
    Updating git-repo `https://github.com/fredrikekre/Staging`
   Resolving package versions...
ERROR: MethodError: no method matching registry_info(::Base.UUID)
Closest candidates are:
  registry_info(::Pkg.RegistryHandling.PkgEntry) at /home/fredrik/dev/Pkg/src/RegistryHandling.jl:129
Stacktrace:
  [1] load_urls(registries::Vector{Pkg.RegistryHandling.Registry}, pkgs::Vector{Pkg.Types.PackageSpec})
    @ Pkg.Operations ~/dev/Pkg/src/Operations.jl:465
  [2] #download_source#39
    @ ~/dev/Pkg/src/Operations.jl:643 [inlined]
  [3] download_source
    @ ~/dev/Pkg/src/Operations.jl:642 [inlined]
  [4] add(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}, new_git::Vector{Base.UUID}; preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform)
    @ Pkg.Operations ~/dev/Pkg/src/Operations.jl:1156
  [5] add(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}; preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API ~/dev/Pkg/src/API.jl:199
  [6] add
    @ ~/dev/Pkg/src/API.jl:150 [inlined]
  [7] add(pkgs::Vector{Pkg.Types.PackageSpec}; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API ~/dev/Pkg/src/API.jl:75
  [8] add
    @ ~/dev/Pkg/src/API.jl:73 [inlined]
[...]

PR #2203, opened 2020-11-11
fix default value in registry_info

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2201

Could you try this @fredrikekre

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -460,7 +460,7 @@ function load_urls(registries::Vector{Registry}, pkgs::Vector{PackageSpec})
         uuid = pkg.uuid
         urls[uuid] = Set{String}()
         for reg in registries
-            reg_pkg = get(reg, pkg.uuid, uuid)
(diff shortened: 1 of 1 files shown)


src/Operations.jl
```diff
@@ -460,7 +460,7 @@ function load_urls(registries::Vector{Registry}, pkgs::Vector{PackageSpec})
         uuid = pkg.uuid
         urls[uuid] = Set{String}()
         for reg in registries
-            reg_pkg = get(reg, pkg.uuid, uuid)
+            reg_pkg = get(reg, pkg.uuid, nothing)
```
fredrikekre, line 463:
```suggestion
            reg_pkg = get(reg, uuid, nothing)
```

fredrikekre (commented):
Do we not have tests with multiple registries?

KristofferC:
> Do we not have tests with multiple registries?

Apparently not where we need to download the package form a registry which is not the first one.

PR #2210, opened 2020-11-13
Allow `7z` selection to pick something up off of the PATH

StefanKarpinski:
Replacement for https://github.com/JuliaLang/Pkg.jl/pull/2198 and https://github.com/JuliaLang/Pkg.jl/pull/2204.


StefanKarpinski:
Errors seem unrelated.

PR #2212, opened 2020-11-16
normalize potentially invalid chars in get_server_dir

pfitzseb:
`get_server_dir` breaks on Windows if the package server includes a port, because 
```
julia> joinpath("/asd", "foo:20")
"foo:20"
```

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -50,7 +50,9 @@ function get_server_dir(url::AbstractString, server=pkg_server())
         @warn "malformed Pkg server value" server
         return
     end
-    joinpath(depots1(), "servers", String(m.captures[1]))
+    # normalize a few characters
+    domain = replace(m.captures[1], r"[\\\/\:\*\?\"\<\>\|]" => "-")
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
I guess the question is whether we should differentiate between servers based on port numbers. The better choice here might be to just strip off the port number and map all servers to the same host name regardless of port number. If, in the future, we were to support something like `ssh://user@host:port/path` via libcurl, we might also want to strip the user name part out, although maybe not...

pfitzseb:
Hm, we probably shouldn't (although I doubt this is going to come up in practice much anyways).
8e899e0 strips usernames (`@` is invalid in normal URLs, right?) and ports from the dirname.

StefanKarpinski (approved):

StefanKarpinski:
The user info is actually a standardized part of URL syntax: https://en.wikipedia.org/wiki/URL#Syntax. Not common for HTTP requests since there are other, better authentication mechanisms, but still valid.

pfitzseb:
Sorry, I meant host instead of URL. Tests should pass now too, I think.

PR #2216, opened 2020-11-18
download artifacts from urls first if given

Yujie-W:
Downloading customized data artifacts (Julia 1.5.3) is extremely slow because Pkg.Artifacts always tries to fetch data from Pkg server first.
It should be more effcient to download the data this way:
- if there is no url, download the artifact from Pkg server
- if urls are given, then download the artifact from given urls first
- if downloading from gievn urls fails, then try Pkg server

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -396,32 +396,40 @@ function ensure_artifact_installed(name::String, meta::Dict, artifacts_toml::Str
     hash = SHA1(meta["git-tree-sha1"])
 
     if !artifact_exists(hash)
-        # first try downloading from Pkg server
-        # TODO: only do this if Pkg server knows about this package
-        if (server = pkg_server()) !== nothing
-            url = "$server/artifact/$hash"
-            download_success = with_show_download_info(name, quiet_download) do
-                download_artifact(hash, url; verbose=verbose, quiet_download=quiet_download)
-            end
-            download_success && return artifact_path(hash)
-        end
-
         # If this artifact does not exist on-disk already, ensure it has download
-        # information, then download it!
+        # if no customized url is provided, try to download it from Pkg server
+        # if customized urls are given, try to download it from given url, then the Pkg server
         if !haskey(meta, "download")
-            error("Cannot automatically install '$(name)'; no download section in '$(artifacts_toml)'")
-        end
-
-        # Attempt to download from all sources
-        for entry in meta["download"]
-            url = entry["url"]
-            tarball_hash = entry["sha256"]
-            download_success = with_show_download_info(name, quiet_download) do
-                download_artifact(hash, url, tarball_hash; verbose=verbose, quiet_download=quiet_download)
+            # first try downloading from Pkg server
+            # TODO: only do this if Pkg server knows about this package
+            if (server = pkg_server()) !== nothing
+                url = "$server/artifact/$hash"
+                download_success = with_show_download_info(name, quiet_download) do
+                    download_artifact(hash, url; verbose=verbose, quiet_download=quiet_download)
+                end
+                download_success && return artifact_path(hash)
             end
-            download_success && return artifact_path(hash)
+            error("Cannot automatically install '$(name)'; no download section in '$(artifacts_toml)'")
+        else
+            # Attempt to download from all sources
+            for entry in meta["download"]
+                url = entry["url"]
+                tarball_hash = entry["sha256"]
+                download_success = with_show_download_info(name, quiet_download) do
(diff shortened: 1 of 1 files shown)


fredrikekre:
> Downloading customized data artifacts (Julia 1.5.3) is extremely slow because Pkg.Artifacts always tries to fetch data from Pkg server first.

Compared to downloading data, an instant 404 should not be "extremely slow", are you sure this is the problem?

Yujie-W:
Example below is what I get when using something like artifact"ARTIFACT_NAME". It often takes about 120 time out before downloading data from given url. I have multiple datasets to download and test, and it takes about 30 minutes to run the test.

When I force the code to predownload the datasets, e.g., https://github.com/CliMA/GriddingMachine.jl/blob/main/src/Utils/download.jl#L20, it takes only 7 minutes to finish the tests. So I think it is the Pkg server that troubled.
```
Wed, 18 Nov 2020 18:07:23 GMT 25l25hDownloading artifact: GPP_VPM_v20_1X_8D
Wed, 18 Nov 2020 18:07:24 GMT #=#=#                                                                         
... ...
Wed, 18 Nov 2020 18:09:24 GMT -=O=-                         #     #     #     #                             
Wed, 18 Nov 2020 18:09:24 GMT curl: (22) The requested URL returned error: 504 
Wed, 18 Nov 2020 18:09:24 GMT 25l25hDownloading artifact: GPP_VPM_v20_1X_8D
Wed, 18 Nov 2020 18:09:24 GMT                                                                            0.1%
Wed, 18 Nov 2020 18:09:25 GMT                                                                            0.9%
... ...
Wed, 18 Nov 2020 18:09:27 GMT #######################################################################   99.1%
Wed, 18 Nov 2020 18:09:27 GMT ######################################################################## 100.0%
```

Yujie-W:
Here the github action record that when I did not force a pre-download from given urls: https://github.com/CliMA/GriddingMachine.jl/runs/1419791006?check_suite_focus=true
This is the github action record when I forced the download from given url directly: https://github.com/CliMA/GriddingMachine.jl/runs/1420168639?check_suite_focus=true

KristofferC:
> It often takes about 120 time out before downloading data from given url.

Maybe that's a problem with the Pkg server? 

Tangentially, It would also be good to be able to ask the server if it has a certain artifact available without having to reveal the artifact tree hash itself.

StefanKarpinski:
cc @staticfloat. Maybe this is hitting some slow path on the Pkg server? If the pkg server in question knows about a package then its artifacts absolutely should be requested via the package server. However, we currently don't check that and just try getting all artifacts from the package server, regardless of whether it knows about the package or not. That ought to be fixed.

staticfloat:
I don't see any 404's anywhere; It's only 504's, which means the Pkg server crashed.  It's a known issue, we're working on it.

> Tangentially, It would also be good to be able to ask the server if it has a certain artifact available without having to reveal the artifact tree hash itself.

I'm not sure how we would do that; the only identifier we have for artifacts is their treehash (the Pkg server doesn't know which artifacts belong to which packages).

StefanKarpinski:
If a package server knows about a package hash, then it should also know about the artifacts associated with that package hash (i.e. listed in the `Artifacts.toml` file in that package tree). Even if we ask a server about a hash that it doesn't know about, it doesn't reveal any useful information since the hash is not reversible.

staticfloat:
> If a package server knows about a package hash, then it should also know about the artifacts associated with that package hash (i.e. listed in the Artifacts.toml file in that package tree).

Sure, we can do that.  But you'd still need to identify the artifact by hash.  Otherwise we need to come up with a way to request "do you have the Artifact named `LibFoo` for platform `x86_64-linux-gnu-cxx11-cuda10` that belongs to package `LibFoo_jll v1.3.0+6`", which seems worse than just transmitting the treehash; I don't see what reason there is for not wanting to transmit the hash.  I suppose you could instead say "do you have all of the artifacts for "LibFoo_jll v1.3.0+6", but that doesn't seem very useful as the answer will basically always be "no", since some platforms will be very unpopular and very unlikely to be in the cahce, like armv6l-linux.

StefanKarpinski:
I'm not suggesting that: I'm saying that if we have downloaded a package version from a pkg server, then we should download the artifacts associated with that package version from the pkg server as well and we should do so by tree hash. If we remember downloading a package version from the pkg server, then we just go ahead and download the artifacts from it as well. If we don't remember, we can do a HEAD request to ask if it knows about the package version and if that's a 200 OK, then we request all the artifacts from the same server. If we've downloaded a registry tree from the pkg server and a package version is in that registry, then we can also assume that the pkg server knows about that package version and just download the artifacts without checking.

PR #2220, opened 2020-11-19
pkg client: add a session header

StefanKarpinski:


KristofferC:
What is this used for? Are there any docs for the package server / protocol for it?

StefanKarpinski:
Not used for anything. It would allow server-side log analysis to tell which requests are from the same Julia client session without being able to tie it back to the client in any way since this ID is never stored anywhere.

KristofferC:
So what would it be used for?

StefanKarpinski:
> It would allow server-side log analysis to tell which requests are from the same Julia client session without being able to tie it back to the client in any way since this ID is never stored anywhere.

PR #2222, opened 2020-11-22
If the Pkg server does not know about the current registry hash, then do not update the registry

### LINKED ISSUE #2215: prevent package downgrade when pkg server is out-of-date
I was having some conectivity issues with the registry today. When it finally worked, it downgraded a few packages.
Most likely the mirror I'm connecting to is out-of-date. I would suggest to fail instead of downgrade if the local registry copy is ahead of the mirror's copy

```
(@v1.5) pkg> up
No Changes to `~/.julia/environments/v1.5/Project.toml`
Updating `~/.julia/environments/v1.5/Manifest.toml`
  [69de0a69] ↓ Parsers v1.0.12 ⇒ v1.0.11
```

DilumAluthge:
Fixes #2215 

Before we update a registry, we do a HEAD request to see if the Pkg server knows about the current registry hash. If the Pkg server does know about the current registry hash, then we proceed to update the registry as normal. If the Pkg server does NOT know about the current registry hash, then we do not update the registry; instead, we keep the current copy of the registry.

Example output:
```julia
julia> Pkg.update()
┌ Warning: Some registries failed to update:
│     — `~/.julia/registries/General` — the local copy of the registry may be ahead of the Pkg server
└ @ Pkg.Types ~/Downloads/Pkg.jl/src/Types.jl:1208
```

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -20,6 +20,8 @@ using ..Pkg.Versions
 import Base: SHA1
 using SHA
 
+import Downloads
+
 export UUID, SHA1, VersionRange, VersionSpec,
     PackageSpec, EnvCache, Context, PackageInfo, ProjectInfo, GitRepo, Context!, Manifest, Project, err_rep,
     PkgError, pkgerror, has_name, has_uuid, is_stdlib, stdlibs, write_env, write_env_usage, parse_toml, find_registered!,
@@ -1077,6 +1079,24 @@ function remove_registries(io::IO, regs::Vector{RegistrySpec})
     return nothing
 end
 
+get_status_code(response::Downloads.Response) = response.status
+function get_status_code(url::AbstractString, timeout_seconds::Real)
+    task = @async Downloads.request(url; method = "HEAD")
+    while timeout_seconds > 0
+        if istaskdone(task)
+            response = fetch(task)
+            return get_status_code(response)
+        end
+        sleep(1)
+        timeout_seconds -= 1
+    end
+    try
+        Base.throwto(task, InterruptException())
+    catch
+    end
+    return 408
+end
+
 # entry point for `registry up`
 function update_registries(io::IO, regs::Vector{RegistrySpec} = collect_registries(depots1());
                            force::Bool=false)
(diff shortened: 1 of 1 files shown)


src/Types.jl
```diff
@@ -1077,6 +1079,24 @@ function remove_registries(io::IO, regs::Vector{RegistrySpec})
     return nothing
 end
 
+get_status_code(response::Downloads.Response) = response.status
+function get_status_code(url::AbstractString, timeout_seconds::Real)
```
KristofferC, line 1083: This function raises many warning flags to me. The `istaskdone` in a loop with `sleep` and the `throwto` to a task and an empty `catch` section. I'm not super well versed in the task system but `throwto` I've only heard bad stuff about. Why would the request take so long to return?

DilumAluthge: Yeah I definitely would like a better way to do this.

If the registry hash is not known to the Pkg server, the request never returns - it seems to hang indefinitely. So we need some kind of timeout.

E.g. try out the URLs in my comment here: https://github.com/JuliaLang/Pkg.jl/issues/2215#issuecomment-731721470

The hashes that are known to the Pkg server return immediately with 200. But the hashes that the Pkg server does not know about will hang indefinitely.

DilumAluthge: I opened a related feature request here: https://github.com/JuliaLang/Downloads.jl/issues/70

DilumAluthge:
Based on the feedback from @fredrikekre and @KristofferC, it looks like this PR probably isn't the right approach.

I'll close this PR, but we can continue the discussion in https://github.com/JuliaLang/Pkg.jl/issues/2215 to figure out what the best solution will be.

DilumAluthge:
`Downloads.request` now has a `timeout` keyword argument, so we could get rid of all the task system stuff, if we want.

StefanKarpinski:
I'd have to do the version bump dance in Base before you can use it here, but yeah, it's a feature now.

DilumAluthge:
> I'd have to do the version bump dance in Base before you can use it here, but yeah, it's a feature now.

https://github.com/JuliaLang/julia/pull/38569

DilumAluthge:
Once that's done, here's what it might look like:

```julia
import Downloads

function url_exists_200(url::AbstractString; timeout::Real = 15)
    try
        response = Downloads.request(url; method = "HEAD", timeout = timeout)
        return response.status == 200
    catch 
        c isa Downloads.RequestError || rethrow()
        return false
    end
end
```

StefanKarpinski:
The `request` function has a `throw = false` option that causes it to return a `RequestError` instead of throwing it. So you could write this instead:
```jl
response = Downloads.request(url; method = "HEAD", throw = false, timeout = timeout)
return response isa Downloads.RequestError && response.status == 200
```

DilumAluthge:
> The `request` function has a `throw = false` option that causes it to return a `RequestError` instead of throwing it. So you could write this instead:
> 
> ```julia
> response = Downloads.request(url; method = "HEAD", throw = false, timeout = timeout)
> return response isa Downloads.RequestError && response.status == 200
> ```

Wow that's much nicer.

DilumAluthge:
I think you need a `not` in there.

```julia
import Downloads

function url_exists_200(url::AbstractString; timeout::Real = 15)
    response = Downloads.request(url; method = "HEAD", throw = false, timeout = timeout)
return !(response isa Downloads.RequestError) && response.status == 200
end
```

StefanKarpinski:
Ah yeah, you can just check `response isa Response` which is what I was thinking of.

PR #2231, opened 2020-11-23
Clarify `Pkg.add` docstring

mcabbott:
This wasn't clear to me, and didn't include mention of what version of Julia it applies to.

I'm still not completely sure what the noun "tier" is meant to refer to here:
* One possibility is that it means the mode in which the resolver operates. I think there are there 4 such modes.
* Another is that it refers to a subset of packages, to be treated differently. Such as those already installed, vs. those being installed by this command. Then there are exactly 2 tiers, always?

So I picked one, but would be happy to re-write if this is wrong. 

It's possible that the story about how test deps are added should be referenced / explained a bit in the `Pkg.test` docstring, too. But I haven't done that here.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -102,11 +102,11 @@ Add a package to the current project. This package will be available by using th
 `import` and `using` keywords in the Julia REPL, and if the current project is
 a package, also inside that package.
 
-## Resolution Tiers
-`Pkg` resolves the set of packages in your environment using an algorithm that operates in one of four
-tiers, or modes. Which of these is used can be controlled by the `preserve` keyword argument,
-which should take one of the values shown below (starting with the strictest). The command will fail if
-there is no solution preserving what you request.
+## Tiered Resolution
+`Pkg.add` resolves the set of packages in your environment using an algorithm that treats previously installed 
+packages in one of four ways. Which of these is used is chosen automatically by default, and can be controlled
+with the `preserve` keyword argument, which takes the values shown below.
+The command will fail if there is no solution preserving what you request.
 
 | Value             | Description                                                                         |
 |:------------------|:------------------------------------------------------------------------------------|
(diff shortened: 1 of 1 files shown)


src/Pkg.jl
```diff
@@ -103,24 +103,32 @@ Add a package to the current project. This package will be available by using th
...
-The following table describes the argument values for `preserve` (in order of strictness):
+`Pkg` resolves the set of packages in your environment using an algorithm that operates in one of four
+tiers, or modes. Which of these is used can be controlled by the `preserve` keyword argument,
+which should take one of the following values (in order of decreasing strictness):
 
 | Value             | Description                                                                         |
 |:------------------|:------------------------------------------------------------------------------------|
 | `PRESERVE_ALL`    | Preserve the state of all existing dependencies (including recursive dependencies)  |
-| `PRESERVE_DIRECT` | Preserve the state of all existing direct dependencies                              |
+| `PRESERVE_DIRECT` | Preserve the state of all existing direct dependencies. Error if this is impossible.|
```
fredrikekre, line 113: Doesn't this apply to all of them?

mcabbott: It does, the others just lacked space. Maybe it should be elsewhere.

DilumAluthge:
> One possibility is that it means the mode in which the resolver operates. I think there are there 4 such modes.

This is the right one, but there are five tiers ("modes").

src/Pkg.jl
KristofferC: Isn't this kind of implied?

mcabbott:
> This is the right one, but there are five tiers ("modes").

This is part of what seemed confusing about the previous version. Does "tier" refer to the algorithm ultimately used for solving, or to the possible values of the keyword?

src/Pkg.jl
```diff
@@ -103,24 +103,32 @@ Add a package to the current project. This package will be available by using th
...
 | `PRESERVE_SEMVER` | Preserve semver-compatible versions of direct dependencies                          |
 | `PRESERVE_NONE`   | Do not attempt to preserve any version information                                  |
-| `PRESERVE_TIERED` | Use the tier which will preserve the most version information (this is the default) |
+| `PRESERVE_TIERED` | Select from the above tiers the one which preserves the most version information, while still alowing a solution. (This is the default.) |
```
KristofferC: "alowing" -> "allowing"

Adding the "while still allowing" feels redundant.

mcabbott: Indeed, will fix.

mcabbott: And perhaps I mis-read you. But I think "while allowing a solution" is important, as otherwise this sentence implies that it would always choose the first. It preserves the most, conditional on working.

IanButterworth:
I think this can be closed. The versions notes refer to versions beyond the current LTS (1.6) and this section has since been updated.

Please feel free to reopen if I'm wrong

PR #2234, opened 2020-11-23
Set `force_latest_compat=true` by default on Dependabot/CompatHelper PRs, and set `force_latest_compat=false` by default otherwise

DilumAluthge:
This pull request is made against #2176 

You can view the correct diff (#2234 against #2176) at this link: https://github.com/DilumAluthge/Pkg.jl/compare/dpa/force-latest-compat...dpa/autodetect-dependabot

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -10,7 +10,7 @@ import REPL
 using REPL.TerminalMenus
 using ..Types, ..Resolve, ..PlatformEngines, ..GitTools, ..MiniProgressBars
 import ..depots, ..depots1, ..devdir, ..set_readonly, ..Types.PackageEntry
-import ..Types.latest_compat, ..Types.force_latest_compat
+import ..Types.latest_compat, ..Types.force_latest_compat, ..Types.decide_force_latest_compat
 import ..Artifacts: ensure_all_artifacts_installed, artifact_names, extract_all_hashes, artifact_exists
 using Base.BinaryPlatforms
 import ...Pkg
@@ -1391,7 +1391,7 @@ end
 
 # ctx + pkg used to compute parent dep graph
 function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::String,
-                 sandbox_path::String, sandbox_project_override; force_latest_compat::Bool=false)
+                 sandbox_path::String, sandbox_project_override; force_latest_compat::Union{Bool, Symbol}=:autodetect)
     active_manifest = manifestfile_path(dirname(ctx.env.project_file))
     sandbox_project = projectfile_path(sandbox_path)
 
@@ -1406,7 +1406,7 @@ function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::S
             cp(sandbox_project, tmp_project)
             chmod(tmp_project, 0o600)
         end
-        if force_latest_compat
+        if Types.decide_force_latest_compat(force_latest_compat)
             Types.force_latest_compat(tmp_project)
             chmod(tmp_project, 0o600)
         end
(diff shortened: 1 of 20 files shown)


DilumAluthge:
@christopher-dG Can you review this?

christopher-dG:
I think you set the target branch to master.

DilumAluthge:
> I think you set the target branch to master.

Yeah. GitHub won't let me set the target (base) branch to `dpa/force-latest-compat` because the branch is in a fork.

DilumAluthge:
You can view the correct diff (#2234 against #2176) at this link: https://github.com/DilumAluthge/Pkg.jl/compare/dpa/force-latest-compat...dpa/autodetect-dependabot

src/Operations.jl
```diff
@@ -1390,7 +1391,7 @@ end
 
 # ctx + pkg used to compute parent dep graph
 function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::String,
-                 sandbox_path::String, sandbox_project_override)
+                 sandbox_path::String, sandbox_project_override; force_latest_compat::Union{Bool, Symbol}=:autodetect)
```
christopher-dG, line 1394: Maybe it would be cleaner to just do `::Bool=Types.decide_force_latest_compat()`?

DilumAluthge: Do you mean `::Bool=Types.is_dependabot_job()`

christopher-dG: Oh yeah, sorry!

KristofferC:
I must say, I am quite uncomfortable coding stuff like this into Pkg. There must be some better option than this. Like having this logic in the CI file and pass the argument based on that.

fredrikekre:
I agree (had a similar comment already written in a review that I failed to post...).

christopher-dG:
The problem with that is it's 1 fix vs N fixes where N is the number of Julia packages. As great as it would be if everyone instantly updated their CI configs when it was needed, that just doesn't happen. Which is fine, since it's their responsibility, but it's still going to annoy and confuse loads of people.

KristofferC:
CompatBot could try update the CI file?

DilumAluthge:
> CompatBot could try update the CI file?

It would have to be Dependabot. @christopher-dG Is that feasible?

DilumAluthge:
Actually, here's one option: if users use the `julia-runtest` action, we could implement this "autodetect" logic there.

This would automatically cover most people using GitHub Actions.

Travis and AppVeyor people will need to fix their scripts manually.

christopher-dG:
I think encoding this in CI scripts (the ones that go in people's repos) is a non-starter. Doing it once per CI provider seems okay, and it's not tied to Julia releases then.

DilumAluthge:
@christopher-dG If it's okay with you, I'll close this PR, and I'll work on adding this "autodetect" functionality into the `julia-runtest` GitHub Action.

DilumAluthge:
I've moved the "auto-detect Dependabot/CompatHelper" functionality into https://github.com/julia-actions/julia-runtest/pull/20, so I have closed this PR.

SaschaMann:
> Travis and AppVeyor people will need to fix their scripts manually.

You could potentially add it to Travis here: https://github.com/travis-ci/travis-build/blob/master/lib/travis/build/script/julia.rb

christopher-dG:
Pretty much everyone using AppVeyor is doing so via [AppVeyor.jl](https://github.com/JuliaCI/Appveyor.jl). GHA+Travis+AppVeyor probably covers >95% of open source package CI.

DilumAluthge:
See https://github.com/JuliaRegistries/CompatHelper.jl/issues/298 to track the deployment of the `force_latest_compatible_version` feature across the most commonly used CI platforms.

PR #2197, opened 2020-11-10
Don't do autoprecomp after Pkg.dev

ianshmean:
Fixes #2195 

I also tweaked the message for `?` precomp errors, which are usually because the package is loaded and thus not in a precompilable state, but that's not the only reason.. so the advice is to try restarting julia which will usually fix it.

PR #2236, opened 2020-11-23
Revert "Don't do autoprecomp after Pkg.dev"

DilumAluthge:
Reverts JuliaLang/Pkg.jl#2197

I just want to make sure that CI is green on this PR, to confirm that we're not missing something.


DilumAluthge:
Hmmm okay in that case, something else has broken Pkg CI on master.

KristofferC:
The failure now seems to be something with Artifacts.

PR #2237, opened 2020-11-23
CI: Don't use the Pkg server when building the docs

DilumAluthge:


fredrikekre:
Why?

StefanKarpinski:
Because it's been flaky? Or is it because it's not sufficiently up-to-date?

DilumAluthge:
> Because it's been flaky? Or is it because it's not sufficiently up-to-date?

The former. Docs builds are intermittently failing because of Pkg server flakiness.

StefanKarpinski:
Until we can fix the reliability of the pkg servers this is fine. Once we do that, we should revert this.

fredrikekre:
What is the failure mode? It is not like we need the most up-to-date registry here so this would only be used when the server is completely dead, and even then, don't we fall back to the old methods?

DilumAluthge:
Here's an example: https://github.com/JuliaLang/Pkg.jl/runs/1444642614?check_suite_focus=true

For some reason it didn't fall back, it just errored out.

Issue #2221, opened 2020-11-20
idea: make Pkg.precompile happen when exiting Pkg REPL


vtjnash:
I'm not sure I prefer that the mode is sticky, but since it is, it'd be really nice to be able to queue a sequence of operations, and have them all applied when I switch back to the main REPL. (ignore the version number in the prompt, I'm on master, but I like the name v1.1)

For example:
```
julia> ]

(@v1.1) pkg> up
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
    Updating git-repo `https://github.com/chriselrod/VectorizationBase.jl.git`
   Installed FastGaussQuadrature ─ v0.4.4
   Installed StatsFuns ─────────── v0.9.6
   Installed DualNumbers ───────── v0.6.3
   Installed DiffRules ─────────── v1.0.2
   Installed Coverage ──────────── v1.2.0
   Installed NaNMath ───────────── v0.3.5
   Installed LoweredCodeUtils ──── v1.2.6
   Installed Requires ──────────── v1.1.1
   Installed DoubleFloats ──────── v1.1.15
   Installed Revise ────────────── v3.1.9
   Installed TimeZones ─────────── v1.5.1
   Installed DSP ───────────────── v0.6.9
   Installed Documenter ────────── v0.25.4
Updating `~/.julia/environments/v1.1/Project.toml`
  [a2441757] ↑ Coverage v1.1.1 ⇒ v1.2.0
  [e30172f5] ↑ Documenter v0.25.3 ⇒ v0.25.4
  [fa6b7ba4] ↑ DualNumbers v0.6.2 ⇒ v0.6.3
  [295af30f] ↑ Revise v3.1.7 ⇒ v3.1.9
Updating `~/.julia/environments/v1.1/Manifest.toml`
  [082447d4] + ChainRules v0.6.5
  [d360d2e6] ↓ ChainRulesCore v0.9.17 ⇒ v0.8.1
  [a2441757] ↑ Coverage v1.1.1 ⇒ v1.2.0
  [717857b8] ↑ DSP v0.6.8 ⇒ v0.6.9
  [b552c78f] ↑ DiffRules v1.0.1 ⇒ v1.0.2
  [e30172f5] ↑ Documenter v0.25.3 ⇒ v0.25.4
  [497a8b3b] ↑ DoubleFloats v1.1.13 ⇒ v1.1.15
[...]

PR #2241, opened 2020-11-25
Add pc as short name for Pkg.precompile

IanButterworth:
Adds `pkg> pc` as short form of `pkg> precompile`. 

Should be nicer for those who don't want auto precomp after each Pkg change, but want to manually invoke precompile routinely, for instance just before leaving Pkg REPL mode. i.e. https://github.com/JuliaLang/Pkg.jl/issues/2221

```
(jl_dyzBJC) pkg> pc
Precompiling project...
  Progress [========================================>]  9/9
9 dependencies successfully precompiled in 8 seconds
```

### DIFF
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -303,10 +303,11 @@ Create a project called `pkgname` in the current folder.
 """,
 ],
 PSA[:name => "precompile",
+    :short_name => "pc",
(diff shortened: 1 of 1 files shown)


fredrikekre:
Not a fan; `pkg> pr<TAB>` is just one more keystroke and as far as I know `pc` is not a known acronym for precompile?

PR #2243, opened 2020-11-25
CI: Re-enable testing on Pkg server

DilumAluthge:
Closes #2237 

This pull request re-enables testing both with and without Pkg server.

This applies to:
- Linux (GitHub Actions)
- macOS (GitHub Actions)
- Windows (AppVeyor)


DilumAluthge:
@KristofferC What do you think?

PR #2245, opened 2020-11-28
WIP: An example for a system-wide installation of packages

mauro3:
After struggling a lot with getting a JupyterHub install going and some chat with @fredrikekre over at:
https://github.com/mauro3/JupyterHubWithJulia/issues/1
I decided to do this documentation PR as long as my memory and motivations are fresh, even though Fredrik had some reservations:

> I agree it would be good with some more "official" documentation, but this seems highly system dependent so not sure it would fit in the regular manual.

If this is useful, let me know how I can improve it.  In particular, I have no idea where to put it, so I put it as a new section.  Also, please review whether the proposed way is the "best" way to do this.


IanButterworth:
It does seem like we're missing a general section on depots 👍

IanButterworth:
I'm planning to do a pass on this with a more broad scope about depot stuff, including all the modern considerations.

docs/src/depots.md
```diff
@@ -0,0 +1,115 @@
...
+Since Julia v1.10, multiple processes using the same depot coordinate via pidfile locks
+to ensure only one process precompiles a package while others wait. However, due to
+the caching of native code in pkgimages since v1.9, you may need to set the `JULIA_CPU_TARGET`
+environment variable appropriately to ensure cache compatibility across different
+worker nodes with varying CPU capabilities.
+
+For more details, see the [FAQ section on distributed computing](https://docs.julialang.org/en/v1/manual/faq/#Computing-cluster)
+and the [environment variables documentation](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_CPU_TARGET).
+
+## Example: System-wide package installation on a *nix OS
```
IanButterworth: @KristofferC what do you think about this section from the original PR? Meaningfully useful/too specific/outdated? (I've not ussed IJulia etc. for ages)

docs/src/depots.md
```diff
@@ -0,0 +1,115 @@
+# **15.** Depots
+
+The packages installed for a particular environment, defined in the
+files `Project.toml` and `Manifest.toml` within the directory
+structure, are not actually installed within that directory but into a
+"depot". The location of the depots are set by the variable
+`DEPOT_PATH`, which has by default the following entries:
+
+1. `~/.julia` where `~` is the user home as appropriate on the system
+2. an architecture-specific shared system directory, e.g. `/usr/local/share/julia`
```
KristofferC:
```
julia> DEPOT_PATH
3-element Vector{String}:
 "/Users/kc/.julia"
 "/Users/kc/.julia/juliaup/julia-1.12-nightly/local/share/julia"
 "/Users/kc/.julia/juliaup/julia-1.12-nightly/share/julia"
 ```
 
 Doesn't look shared, or?

File: docs/src/depots.md [hunk 3]
```diff
@@ -0,0 +1,115 @@
...
+
+For more details, see the [FAQ section on distributed computing](https://docs.julialang.org/en/v1/manual/faq/#Computing-cluster)
+and the [environment variables documentation](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_CPU_TARGET).
+
+## Example: System-wide package installation on a *nix OS
+
+Usually, most of the packages will be installed by the user into their
+own depot.  However, sometimes a system-wide installation of some
+packages are desired or necessary.  One example where it is necessary
+to install a package system-wide is in a
```
KristofferC: I don't really understand this, I don't need to install IJulia system wide?

IanButterworth: Yeah I have no idea. That's original to the PR. If it's not needed then I think we drop it. I've not used IJulia for ages but I don't hear people talking about needing to do this kind of thing.

mauro3: I dug into this when setting up a jupyter hub instance with a Julia kernel installed. I'm pretty sure that this needs a system wide installation of IJulia as each JHub user is setup as a Linux user on that system, at least in the setup I did: https://github.com/mauro3/JupyterHubWithJulia. But I could be wrong. Albeit, if system-wide install has no use then the question would be why it exists in the first place... In that sense, an example would maybe be good, as apparently not that many use cases exist.

When trying to find out how to do this JHub install there was no docs about depots and such, thus this PR was born. Feel free to use what is useful and ditch the rest (or start from scratch).

x-ref: https://github.com/JuliaLang/Pkg.jl/issues/2246

IanButterworth: Ok. I've tidied things up, removed duplication against the Base docs, and removed the main section you added @mauro3. If it does prove to be needed, I think we need to understand why and that can be a follow-on.

Because I've already made the changes I'll just go ahead on this PR.

Here's the stuff I removed, in case it's lost from the git history.

`````
## Example: System-wide package installation on a *nix OS

Usually, most of the packages will be installed by the user into their
own depot.  However, sometimes a system-wide installation of some
packages are desired or necessary.  One example where it is necessary
to install a package system-wide is in a
[JupyterHub](https://github.com/jupyterhub/jupyterhub), where the
[IJulia.jl](https://github.com/jupyterhub/jupyterhub) package needs to
be available as it provides the kernel which JupyterHub needs to run
to provide a Julia notebook.

Here the depot `DEPOT_PATH[2]` is chosen as the install depot, the
corresponding environment is placed in
`DEPOT_PATH[2]/environments/v#.#`, with `v#.#` denoting the version of
Julia.

First remove the standard, user-depot, such that installs are not
occurring there and set a new environment at the same location:

```julia-repl
julia> deleteat!(DEPOT_PATH, [1])

julia> env_path = "$(DEPOT_PATH[1])/environments/v$(VERSION.major).$(VERSION.minor)"

julia> using Pkg; Pkg.activate(env_path);

julia> Pkg.add("YourFavoritePackage")
```

Notes:

- This setup mirrors the layout used in the "standard" depot
  `./julia`. However, another setup can be chosen with different depot
  and environment paths.
- Depending on the write permissions on the system, the above commands may
  require administrator privileges.
- The precompile cache will be available to system users if they have
  read access to the depot.

To make this environment available to the system users, it has to be
added to the `LOAD_PATH`:

```bash
JULIA_LOAD_PATH="@":"@v#.#":/usr/local/share/julia/environments/v1.13:"@stdlib" julia
```

Note that the `DEPOT_PATH` is already set, as we installed into
`DEPOT_PATH[2]`. The path `/usr/local/share/julia/environments/v1.13` should be
adjusted according to your Julia version and chosen depot location.
The environment variable `JULIA_LOAD_PATH` will need to
be set for all users. Once that is done, all users will have access to the system-wide
installed package(s).

Regarding the motivating example: don't forget to copy the
Julia-kernel created by IJulia to a place where the JupyterHub install
can pick it up.
`````

docs/src/depots.md
IanButterworth: This was from the Base.DEPOT_PATH docstring, which I now reference rather than copy.

docs/src/depots.md
IanButterworth: See discussion below

PR #2248, opened 2020-11-28
Make a <return> keypress on an empty line return to julia REPL via precomp

IanButterworth:
A proposal to provide an optional exit back to julia repl via precompilation, as per https://github.com/JuliaLang/Pkg.jl/issues/2221

(This is with `ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0`)

With this, on an empty line:
`<backspace>` remains the same and returns to julia repl promptly
`<return>` no longer just prints another empty prompt, it _returns_ to julia repl via `Pkg.precompile()`

```
pkg> add Foo
...
pkg> dev Bar
...
pkg> up
...
pkg> <return>
Precompiling project...
...
julia> 
```

That way, the precomp only takes place once the project is stable.

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -547,10 +547,16 @@ function create_mode(repl::REPL.AbstractREPL, main::LineEdit.Prompt)
         ok || return REPL.transition(s, :abort)
         input = String(take!(buf))
         REPL.reset(repl)
-        do_cmd(repl, input)
+        if isempty(input)
+            Pkg.precompile()
+        else
+            do_cmd(repl, input)
+        end
         REPL.prepare_next(repl)
         REPL.reset_state(s)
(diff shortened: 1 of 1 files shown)


KristofferC:
I personally think we should slightly hold off on too many configuration things for the auto precompile stuff until we have a bit more feedback on the current system. Environment variables, explicit commands, special REPL actions etc is starting to be a bit much, perhaps.

ianshmean:
Yeah ok 👍

StefanKarpinski:
If we do decide to make this configurable, the right way is probably to have a single variable that can take different values for when to precompile, e.g. `JULIA_PKG_PRECOMPILE_TIME` with valid values of `install`, `repl-exit` or `load`. If there are other options later on we can change the default and add other options.

IanButterworth:
Closing as probably not necessary

Issue #2247, opened 2020-11-28
Relative repo paths in manifest


GunnarFarneback:
This is a simplified reproducer showing two issues with relative paths in `Manifest.toml`.
Start with
```
mkdir /tmp/testdir
cd /tmp/testdir
mkdir empty_depot
git clone https://github.com/JuliaLang/Example.jl.git
cd Example.jl/
mkdir test/reference_environment
julia --project=test/reference_environment
    using Pkg
    Pkg.add(PackageSpec(name="Example", path=".", rev="3c440fe40c6e37c3ad76542f2c9f010cac984dac"))
    exit()
```
The resulting `test/reference_environment/Manifest.toml` becomes
```
[[Example]]
git-tree-sha1 = "4da5b9200d71b81738241f5213c811acc074aa6c"
repo-rev = "3c440fe40c6e37c3ad76542f2c9f010cac984dac"
repo-url = "../.."
uuid = "7876af07-990d-54b4-ab0e-23690620f79a"
version = "0.5.3"
```
The `repo-url` is the relative path from the project, which is fine.

**Issue 1**: If this is done on WIndows, the `repo-url` gets a backslash instead of a forward slash, which is not fine when you try to use the Manifest on another platform, e.g. for CI.

Next emulate instantiating this Manifest in a clean CI (it's assumed `Example.jl` is cloned in the CI, so we continue where we are but with an empty depot):
```
JULIA_DEPOT_PATH=../empty_depot julia --project=test/reference_environment
    using Pkg
    Pkg.instantiate()
```
**Issue 2**: Despite recording a relative path from the project in the manifest, `instantiate` tries to use it relative to the current working directory:
```
ERROR: git repository not found at `../..`
[...]
caused by: GitError(Code:ENOTFOUND, Class:Repository, could not find repository from '/tmp')
[...]
```
Interestingly, with the incorrect manifest recorded on Windows, it is interpreted as a path relative to the project:
```
ERROR: Did not find path `/tmp/testdir/Example.jl/test/reference_environment/..\..` for `Example [7876af07]`
[...]
```

GunnarFarneback:
Issue 2 is addressed in #2249.

Issue 1 requires switching the slash direction somewhere. Possible candidates are
1. In [`relative_project_path`](https://github.com/JuliaLang/Pkg.jl/blob/6f6d60017b389ae87a037f0eb925e3e13a2c73f1/src/Types.jl#L408).
[...]

PR #2249, opened 2020-11-29
Use repo_source when cloning.

GunnarFarneback:
This solves part 2 of #2247 and looks like an obvious correction. Possibly `repo_source` was inadvertently changed back to `pkg.repo.source` in #1419.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1330,11 +1330,11 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
             pkgerror("Did not find path `$(repo_source)` for $(err_rep(pkg))")
         end
         repo_path = Types.add_repo_cache_path(repo_source)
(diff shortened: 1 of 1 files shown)


KristofferC:
Thanks, yes, that seems likely.

KristofferC:
Bonus points for adding a test at some point ;)

GunnarFarneback:
I think it's supposed to be tested here: https://github.com/JuliaLang/Pkg.jl/blob/5b57bacb41e010b03053ac9089a2fc7262c6e963/test/new.jl#L674-L699
I don't know why it passed.

GunnarFarneback:
Possibly because the test cleans out `packages` rather than `clones`.

GunnarFarneback:
Confirmed, if you remove also the clone, the test would fail without this PR and pass with it. I don't understand what is meant by the comment: `# We can use the loaded depot here, it will help us avoid the original clone.`

KristofferC:
> I don't understand what is meant by the comment: # We can use the loaded depot here, it will help us avoid the original clone.

Me neither.

Issue #2146, opened 2020-10-20
Packages outside sysimage without versions in Project file are not compatible with Pkg


vtjnash:
This reproducibly fails for me, when Pkg is not baked into the system image:
```julia
julia> Pkg.update()
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
    Updating git-repo `https://github.com/chriselrod/VectorizationBase.jl.git`
No Changes to `~/.julia/environments/v1.1/Project.toml`
No Changes to `~/.julia/environments/v1.1/Manifest.toml`
ERROR: MethodError: Cannot `convert` an object of type 
  Nothing to an object of type 
  Union{Pkg.Types.UpgradeLevel, VersionNumber, Pkg.Types.VersionSpec}
Closest candidates are:
  convert(::Type{T}, ::T) where T at essentials.jl:205
Stacktrace:
  [1] Pkg.Types.PackageSpec(name::String, uuid::Base.UUID, version::Nothing, tree_hash::Nothing, repo::Pkg.Types.GitRepo, path::Nothing, pinned::Bool, mode::Pkg.Types.PackageMode)
    @ Pkg.Types /data/vtjnash/julia/usr/share/julia/stdlib/v1.6/Pkg/src/Types.jl:91
  [2] Pkg.Types.PackageSpec(; name::String, uuid::Base.UUID, version::Nothing, tree_hash::Nothing, repo::Pkg.Types.GitRepo, path::Nothing, pinned::Bool, mode::Pkg.Types.PackageMode)
    @ Pkg.Types ./util.jl:444
  [3] precompile(ctx::Pkg.Types.Context; internal_call::Bool, io::Base.TTY)
    @ Pkg.API /data/vtjnash/julia/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:959
  [4] _auto_precompile(ctx::Pkg.Types.Context; kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API /data/vtjnash/julia/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:912
  [5] _auto_precompile
    @ /data/vtjnash/julia/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:911 [inlined]
  [6] #up#38
[...]

PR #2251, opened 2020-12-01
Work around precompiling packages with no `version` entry

staticfloat:
While experimenting with non-precompiled stdlibs, I found that a
`Project.toml` that has no `version` entry set will cause `precompile()`
to fail in a nasty way.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/2146


vtjnash:
Does this fix https://github.com/JuliaLang/Pkg.jl/issues/2146?

staticfloat:
Probably; that error looks exactly like the one I worked around.

PR #2253, opened 2020-12-01
Use Sys.isexecutable() to calculate git hashes

staticfloat:
Since our libuv now understands file permissions better on windows, we can finally use `Sys.isexecutable()` on Windows.  :)

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -174,8 +174,7 @@ function gitmode(path::AbstractString)
         return mode_symlink
     elseif isdir(path)
         return mode_dir
-    # We cannot use `Sys.isexecutable()` because on Windows, that simply calls `isfile()`
-    elseif !iszero(filemode(path) & 0o100)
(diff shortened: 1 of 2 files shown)


KristofferC:
One of the tests stopped being broken so I pushed a commit that tweaked that. However, another test seems to have started to break:

``` 
git tree hash computation: Test Failed at C:\projects\pkg-jl\test\new.jl:2315
  Expression: "0a890bd10328d68f6d85efd2535e3a4c588ee8e6" == tree_hash(dir)
   Evaluated: "0a890bd10328d68f6d85efd2535e3a4c588ee8e6" == "952cfce0fb589c02736482fa75f9f9bb492242f8"
```

staticfloat:
This is actually a good thing!  The test was assuming an initial permissions that wasn't true on Windows, but because we were permissions blind before, the test still passed.  I've changed it to now explicitly set the permissions of the file, so we should be good to go on that test.

PR #2259, opened 2020-12-02
fix paths to stdlibs being put into test manifest again

KristofferC:


KristofferC:
Hm, not that easy apparently...

PR #2261, opened 2020-12-03
Precomp: Missing fix from #2251 (Work around precompiling packages with no `version` entry)

ianshmean:
#2251 missed one place where a missing version would break precomp.

Also, the errors weren't throwing due to coming from async tasks, so they now print via `@error`


ianshmean:
I added two more tweaks:
- Ensured one extra loop of the print loop after the last async task finishes, as occasionally the last task wasn't scrubbed
- Added a bit of semi-random jitter into the precomp spinner animation phases. On larger font terminals it could look a bit odd that they're all synchronised, given that they're individual tasks that started at different times.

KristofferC:
Error looks like the resolve test that was added to test for specific julia_versions.

PR #2265, opened 2020-12-03
Canonicalize repo url when writing manifest.

GunnarFarneback:
This solves part 1 of #2247, adds a test for part 1, and modifies a test so that it captures part 2 and would have failed without #2249.


fredrikekre (approved):

GunnarFarneback:
This is ready to go?

PR #2266, opened 2020-12-03
allow changing stdlib dir

Roger-luo:
Tho in most cases we can assume `Pkg` ships with a complete installation of Julia, but when one generates a standalone app with PackageCompiler, this might not be true, and not all the files in stdlibs will be shipped with such kind of app. But there are cases that packages might depend on `Pkg`:

- they might be an enhancement tool on julia's package/project, e.g PkgTemplates
- the user may use Julia's Pkg system as a hot plugin loading system since Julia is dynamic and can actually load packages in runtime, and update methods table by requirements.

This PR make `stdlib_dir` available to change in `Context`, so this won't prevent developers to ship their own standalone app with certain stdlib files (or alternative stdlib files located other places).

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -1,5 +1,5 @@
 name = "Pkg"
-uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
+uuid = "54cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 keywords = ["package management"]
 license = "MIT"
 desc = "The next-generation Julia package manager."
--- a/src/API.jl
+++ b/src/API.jl
@@ -1376,12 +1376,12 @@ function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=
 end
 
 
-function activate(;temp=false,shared=false)
+function activate(;temp=false,shared=false, stdlib_dir::String=Types.default_stdlib_dir())
     shared && pkgerror("Must give a name for a shared environment")
     temp && return activate(mktempdir())
     Base.ACTIVE_PROJECT[] = nothing
     p = Base.active_project()
-    p === nothing || printpkgstyle(DEFAULT_IO[], :Activating, "environment at $(pathrepr(p))")
+    p === nothing || printpkgstyle(DEFAULT_IO[], :Activating, "environment at $(pathrepr(p, stdlib_dir))")
     add_snapshot_to_undo()
     return nothing
 end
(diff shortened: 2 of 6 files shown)


KristofferC:
What's the current behavior? Does it error? Perhaps Pkg should read and store the stdlib projects during compilation time.

Roger-luo:
yeah, the simplest case is wrapping `PkgTemplates` in the app. Since plugins like `Tests` in `PkgTemplates` calls `Pkg.update`, and `Pkg.update` will use `stdlib` to resolve and update the `Manifest.toml`. But since these `stdlib` åctually locates in a different directory, `Pkg.update` gets `nothing` then it errors.

Roger-luo:
> Perhaps Pkg should read and store the stdlib projects during compilation time of it.

what does compilation mean here? the PackageCompiler compilation stage? e.g let PackageCompiler modify the `stdlib` path?

Roger-luo:
Ah I think I get what @KristofferC mean now, do you mean `Pkg` should just read the `Project.toml` of stdlibs during compilation of julia's system image and store that as a constant instead of read it in runtime? Would this affect the ability of developing stdlibs without building julia etc. tho?

KristofferC:
> Ah I think I get what @KristofferC mean now, do you mean Pkg should just read the Project.toml of stdlibs during compilation of julia's system image and store that as a constant instead of read it in runtime?

Yes, any reading of stdlib files during runtime seems vulnerable to the issue here when Pkg is run in a way where these files are not accessible.

> Would this affect the ability of developing stdlibs without building julia etc. tho?

I dunno, probably not? You would have to describe the exact workflow you mean for developing stdlibs.

Roger-luo:
OK, so I go through the implementation related to this in `Pkg`, and tried to store stdlib's `Project` in compilation, here is what I found:

> I dunno, probably not? You would have to describe the exact workflow you mean for developing stdlibs.

I was not sure, but I find the answer should be no since all stdlibs can be developed using a different UUID so there won't be anything to do with stored stdlib info.

for stdlibs in most cases, there are no version specifications except for some fake JLL packages. However, if we load stdlib's `Project` object to store all the information of stdlibs, currently this won't work since the fake `LibUnwind_jll` package has an invalid `Project.toml`. But I guess that was because `JLLWrappers` is not a stdlib yet, and current implementation of Pkg only loads stdlib's `Project.toml` when it's being resolved, which won't error. I guess we need to find a way to workaround this, or somehow add the valid `Project.toml` to `LibUnwind_jll`.

KristofferC:
>  currently this won't work since the fake LibUnwind_jll package has an invalid Project.toml.

That should just be fixed there then (https://github.com/JuliaLang/julia/issues/38743).

PR #2268, opened 2020-12-04
use PackageSpec less after the resolve step, part 1

KristofferC:
`PackageSpec` are inputs from the user. After the resolve step, there should be no need to use these anymore. However, for some reason, we keep passing this vector of `PackageSpec` along into different operations even though we have a perfectly contained environment struct containing the full manifest with all the information we need. This is a start to remove usage of these after the resolve step.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -981,7 +981,7 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
     foreach(pkg -> pkg.mode = PKGMODE_MANIFEST, pkgs)
     manifest_resolve!(ctx.env.manifest, pkgs)
     ensure_resolved(ctx.env.manifest, pkgs)
-    Operations.build(ctx, pkgs, verbose)
+    Operations.build(ctx, Set{UUID}(pkg.uuid for pkg in pkgs), verbose)
 end
 
 function _is_stale(paths::Vector{String}, sourcepath::String)
@@ -1465,15 +1465,11 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     end
 
     # Install all packages
-    new_apply = Operations.download_source(ctx, pkgs)
-    # Install artifacts for deps and artifacts for current package, if applicable
-    art_pkgs = copy(pkgs)
-    if ctx.env.pkg !== nothing
-        push!(art_pkgs, ctx.env.pkg)
-    end
-    Operations.download_artifacts(ctx.env, art_pkgs; platform, verbose, io=ctx.io)
+    new_apply = Operations.download_source(ctx)
+    # Install all artifacts
+    Operations.download_artifacts(ctx.env; platform=platform, verbose=verbose)
     # Run build scripts
-    Operations.build_versions(ctx, union(UUID[pkg.uuid for pkg in new_apply], new_git); verbose)
+    Operations.build_versions(ctx, union(new_apply, new_git); verbose=verbose)
 
     allow_autoprecomp && Pkg._auto_precompile(ctx)
 end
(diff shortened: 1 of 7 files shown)


src/API.jl
```diff
@@ -1465,15 +1465,11 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     end
 
     # Install all packages
-    new_apply = Operations.download_source(ctx, pkgs)
-    # Install artifacts for deps and artifacts for current package, if applicable
```
KristofferC, line 1469: Current package is now handled inside `download_artifacts`.

test/artifacts.jl
```diff
@@ -464,17 +460,13 @@ end
         generate_flooblegrank_artifacts(ap_path)
 
         Pkg.activate(ap_path)
-        add_this_pkg()
-        @test !isdir(artifact_path(engaged_hash))
-        @test !isdir(artifact_path(disengaged_hash))
```
staticfloat, line 469: I don't understand what operation is installing the flooblecrank artifacts here.

test/artifacts.jl
```diff
@@ -464,17 +460,13 @@ end
         generate_flooblegrank_artifacts(ap_path)
```
staticfloat, line 460: Hah, just realized that I mis-spelled `flooblecrank` in the function name.  😂

staticfloat (commented):
Overall looks right to me

test/artifacts.jl
KristofferC: After we add a package we also make sure that everything is instantiated. That has been there for a long time. But there was a bug where artifacts for the current active package did not get installed, but that is fixed now. So when we do `add_this_pkg`, it instantiates and that installs the flubidooki artifact of the current package.

KristofferC: The bug fix is basically https://github.com/JuliaLang/Pkg.jl/pull/2268/files#diff-6dfb4e40705b22a427451362ea83b4e3a763028e3dcaa3b9b1846aa1ec84370fR596-R597.

staticfloat: That makes sense to me, but I don't see where the `add_this_pkg()` is; you've removed it.

src/Operations.jl
```diff
@@ -587,18 +582,25 @@ function collect_artifacts(pkg_root::String; platform::AbstractPlatform=HostPlat
...
-        pkg_root === nothing && continue
+        pkg_root === nothing || push!(pkg_roots, pkg_root)
+    end
+    envpkg = env.pkg
+    envpkg === nothing || push!(pkg_roots, envpkg.path)
+    for pkg_root in pkg_roots
         for (artifacts_toml, artifacts) in collect_artifacts(pkg_root; platform)
             # For each Artifacts.toml, install each artifact we've collected from it
             for name in keys(artifacts)
+                @show name, artifacts[name], artifacts_toml
```
staticfloat, line 602: Just to make sure you don't accidentally leave the `@show` in here

test/artifacts.jl
KristofferC: It is on (the new) line 469:

https://github.com/JuliaLang/Pkg.jl/pull/2268/files#diff-bf3b8e7b503ee397f9d6ba823767c04bf212d23b4669c50d9753e6897d9c1ecbR469

staticfloat: But that is below these lines?  So wouldn't these `@test` lines still pass?  The intent here was to verify that these two artifacts are not present, so we can assert that they were added within those `withenv()` blocks, and not accidentally added by something else earlier in the test recipe.

KristofferC: I see, yes, that makes sense. I should be able to add these back indeed.

staticfloat: Great!  Then we're totally on the same page.  :)

src/Operations.jl
KristofferC: Jeez, today has really been a spectacle in this type of stuff for me. :sweat:

PR #2271, opened 2020-12-05
Revert "Use Sys.isexecutable() to calculate git hashes"

KristofferC:
Reverts JuliaLang/Pkg.jl#2253

This caused hashes to not give the correct result anymore (even on non-Windows systems) which of course causes a bunch of errors (https://github.com/jheinen/GR.jl/issues/360). I added some debug prints to `gitmode` when adding `GR_jll` (which fails on nightly) and it showed that for example for the file

```
"~/.julia/artifacts/jl_LIRHTe/Applications/GKSTerm.app/Contents/Resources/English.lproj/InfoPlist.strings"
```

`Sys.isexecutable` and `!iszero(filemode(path) & 0o100)` does not give the same result causing the hash to change.

The permission of the file is `555` octal.

cc @staticfloat

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -175,7 +175,8 @@ function gitmode(path::AbstractString)
         return mode_symlink
     elseif isdir(path)
         return mode_dir
-    elseif Sys.isexecutable(path)
+    # We cannot use `Sys.isexecutable()` because on Windows, that simply calls `isfile()`
+    elseif !iszero(filemode(path) & 0o100)
         return mode_executable
     else
         return mode_normal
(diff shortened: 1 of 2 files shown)


KristofferC:
Fixes GR_jll but still:

```
┌ Warning: tarball content does not match git-tree-sha1 68/282
└ @ Pkg.Operations ~/julia/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:523
```

... Maybe something weird in Tar.jl as well?

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/2273

PR #2273, opened 2020-12-05
Only use `Sys.isexecutable()` on Windows

staticfloat:
Supercedes #2271

@KristofferC I think this is a better solution; it will keep windows improved (and hopefully allow us to no longer have hash mismatches when users create artifacts on Windows then use them on Linux) while keeping the behavior we have had on Linux/MacOS.  My initial guess right now is that `Sys.isexecutable()` (which uses `access()` internally) has corner cases on MacOS that we don't know about yet, so we should just stay away from using that for now and keep using what we've been using so far.

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -171,11 +171,21 @@ Base.string(mode::GitMode) = string(UInt32(mode); base=8)
 Base.print(io::IO, mode::GitMode) = print(io, string(mode))
 
 function gitmode(path::AbstractString)
+    # Windows doens't deal with executable permissions in quite the same way,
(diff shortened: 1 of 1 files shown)


KristofferC:
Alright, makes sense!

StefanKarpinski:
Had to fix Tar.jl to match: https://github.com/JuliaIO/Tar.jl/pull/87. Will be nice when we only use Tar.tree_hash for this and there's only one place to worry about this.

StefanKarpinski:
This seems fine: the only CI issue is code coverage.

Issue #2185, opened 2020-11-01
`PlatformEngines.package` creates directories with wrong permissions if symlinked


giordano:
Reproducer:
```julia
using Pkg.PlatformEngines

probe_platform_engines!()

mktempdir() do dir
    mkpath(joinpath(dir, "lib", "icu", "67.1"))
    symlink("67.1", joinpath(dir, "lib", "icu", "current"))
    package(dir, joinpath(@__DIR__, "test.tar.gz"))
end
```
The generated tarball has the directory `lib/icu/67.1` with permissions set to `40555` instead of `40755`.  This happens only if there is the symlink `lib/icu/current` pointing to it.

```julia
julia> versioninfo()
Julia Version 1.6.0-DEV.1334
Commit 6edf6d9e15 (2020-10-26 17:30 UTC)
```

giordano:
I cannot reproduce the issue on
```julia
julia> versioninfo()
Julia Version 1.6.0-DEV.1397
Commit 8d55e04b69 (2020-10-30 21:02 UTC)
```
Maybe the switch to `Tar.jl` fixed it?  However it would be good to understand what was happening, a test to prevent this issue from coming back would be great, too.

StefanKarpinski:
Seems likely that `7z` had a bug where it applied permissions to the target of a symlink instead of the symlink itself. Since `7z` is primarily developed for Windows, we've generally found that it's pretty unreliable for anything on UNIX beyond handling pure data de/compression, which was one of the motivations for switching to Tar.jl.

PR #2274, opened 2020-12-05
Add tests for permissions of files in archives created by `package`

giordano:
In particular, make sure directories pointed by symlinks don't get borked
read-only permissions, as it happened in Yggdrasil for some time.

This addresses the issue reported in #2185.  I'm not sure this test makes sense
on all operating systems, it can also be restricted to Unix- or Linux-only if
necessary.

### DIFF
--- a/test/platformengines.jl
+++ b/test/platformengines.jl
@@ -1,7 +1,7 @@
 module PlatformEngineTests
 import ..Pkg # ensure we are using the correct Pkg
 
-using Test, Pkg.PlatformEngines, Pkg.BinaryPlatforms, SHA
+using Test, Pkg.PlatformEngines, Pkg.BinaryPlatforms, SHA, Tar
 
 
 @testset "Packaging" begin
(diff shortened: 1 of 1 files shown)


test/platformengines.jl
```diff
@@ -43,6 +43,25 @@ using Test, Pkg.PlatformEngines, Pkg.BinaryPlatforms, SHA
         end
     end
 
+    @testset "Issue #2185" begin
+        probe_platform_engines!()
```
KristofferC, line 47:
```suggestion
```

Can be deleted, it doesn't do anything

giordano: Oh, wow.  Thanks!

KristofferC:
I'm not sure what part of Pkg this is testing? I'm thinking if this should be in Tar.jl?

StefanKarpinski:
Now that we're always using [Tar.jl](https://github.com/JuliaIO/Tar.jl) for unpacking, it does seem better to test this there.

giordano:
But we were experiencing the issue _before_ the switch to Tar.jl, and the problem was with the generated tarballs, not with their unpacking

StefanKarpinski:
Pkg doesn't control the generation of tarballs though, so why test here?

giordano:
Because it used to, when we had the problem, reason why the reproducer in #2185 had the now useless `probe_platform_engines!()`.

StefanKarpinski:
Can you elaborate? I don't get it.

giordano:
There isn't really much to explain, until a month ago `PlatformEngines.package` was this: https://github.com/JuliaLang/Pkg.jl/blob/4a08f33fd42bed62aee7ee16feaddeef2841245a/src/PlatformEngines.jl#L886-L906 no calls to other julia packages. We've been experienced issues with tarballs generated with wrong permissions since September (see https://github.com/JuliaPackaging/BinaryBuilder.jl/pull/911 and https://github.com/JuliaPackaging/Yggdrasil/issues/1969).

My fault is having opened this PR with 2+ months of delay

StefanKarpinski:
So the regressions we're protecting against with this test are:

1. switching to another method of packing tarballs that reintroduces this problem
2. `Tar.create` introducing the same behavior that `z7` exhibited when unpacking them

The first problem doesn't seem like something that could really happen by accident, whereas the second could (and probably should) be tested in the Tar.jl tests instead. I think that's the point @KristofferC was getting at too.

giordano:
I'm mainly concerned with 1, that's what we've seen.  I agree it's unlikely to see it again in the future, but better safe than sorry?

StefanKarpinski:
🤷🏻‍♂️

PR #2275, opened 2020-12-05
Prioritize package name before local directory for REPL add and develop

GunnarFarneback:
This PR revives #1636, i.e. interpret `pkg> add Example` as a package name also when you have a directory `Example` in your current directory and require `pkg> add ./Example` if you want the local directory. It differs somewhat though:

* Still accepts the name as a local directory if it's not a valid package name.
* Is more chatty in the ambiguous case.
* Uses `normpath` to normalize the local directory name after parsing. This replaces `./Example` with just `Example`.

I'm open for discussion whether these changes are desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,0 +1,4 @@
+Pkg v1.7 Release Notes
+======================
+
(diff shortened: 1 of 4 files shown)


KristofferC:
I think we should add back a `NEWS.md` file and then we could put a blurb about this in it.

KristofferC:
I added a changelog file, will merge when CI is green.

PR #2284, opened 2020-12-09
wip: --outdated flag to status

KristofferC:
bild

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,9 @@
+Pkg v1.8 Release Notes
+======================
+
+- New `compat::Bool` kwarg to `Pkg.status` (`--compat` in the REPL mode) to show
+  information about packages not at the latest version.
+
 Pkg v1.7 Release Notes
 ======================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,7 @@
 Pkg v1.8 Release Notes
 ======================
 
-- New `compat::Bool` kwarg to `Pkg.status` (`--compat` in the REPL mode) to show
+- New `outdated::Bool` kwarg to `Pkg.status` (`--outdated` or `-o` in the REPL mode) to show
   information about packages not at the latest version.
 
 Pkg v1.7 Release Notes
--- a/src/API.jl
+++ b/src/API.jl
@@ -1520,9 +1520,8 @@ end
 
 @deprecate status(mode::PackageMode) status(mode=mode)
 
-function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=PKGMODE_PROJECT,
-                io::IO=stdout, kwargs...)
-    Operations.status(ctx.env, pkgs, mode=mode, git_diff=diff, io=io)
+function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=PKGMODE_PROJECT, compat::Bool=false, io::IO=stdout, kwargs...)
+    Operations.status(ctx.env, ctx.registries, pkgs; mode, git_diff=diff, io, compat)
     return nothing
 end
 
@@ -1662,7 +1661,7 @@ function redo_undo(ctx, mode::Symbol, direction::Int)
     snapshot = state.entries[state.idx]
     ctx.env.manifest, ctx.env.project = snapshot.manifest, snapshot.project
     write_env(ctx.env; update_undo=false)
-    Operations.show_update(ctx.env; io=ctx.io)
+    Operations.show_update(ctx.env, ctx.registries; io=ctx.io)
 end
 
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -1520,8 +1520,8 @@ end
 
 @deprecate status(mode::PackageMode) status(mode=mode)
 
-function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=PKGMODE_PROJECT, compat::Bool=false, io::IO=stdout, kwargs...)
-    Operations.status(ctx.env, ctx.registries, pkgs; mode, git_diff=diff, io, compat)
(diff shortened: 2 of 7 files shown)


StefanKarpinski:
What about showing a limited form of this by default? Like `HTTP v0.8.19 (< v0.9.1)`?

brenhinkeller:
I'll put in a word for displaying the full names of each package causing a hold back (as in the screenshot) -- this would hugely decrease the barrier of entry for folks that aren't experts in the nuances of `Pkg` to see what is holding back what, and maybe even go make some PRs to help fix the holdouts.

StefanKarpinski:
That seems like way too much information to put in the default status output.

vtjnash:
Something like this for `Pkg.update` seems excellent, though I think the current state of the PR output is confusing. My suggestion would be to mark them more kindly (without an `x`), and in the message, tell the user how to get the more detailed output.
```
...
[abcdefg] Example v1.1 ([yellow]v2.9 available[/yellow])
...
Info Some packages are not at the highest released version. Call `Pkg.upgradable()` for additional information.
```

KristofferC:
> My suggestion would be to mark them more kindly (without an x), and in the message, tell the user how to get the more detailed output.

This is with a `-c` to status so the output here is pretty much what `Pkg.upgradable()` would show.

I am not sure having `([yellow]v2.9 available[/yellow])` in the default status output is a good idea.

KristofferC:
I've gotten around to showing something by default in the normal status update. I'll work on how it would look and update here.

KristofferC:
I updated this. Still need to fix tests etc. The way it works now is that if you give the `--compat` flag it filters out things to only show packages that are not on the latest version:

```
(Pkg) pkg> st
Project Pkg v1.8.0
Status `~/JuliaPkgs/Pkg.jl/Project.toml`
 [a8cc5b0e] Crayons v3.0.0
 [28b8d3ca] GR v0.57.5
 [5c1252a2] GeometryBasics v0.3.13
...

(Pkg) pkg> st -c
Project Pkg v1.8.0
Status `~/JuliaPkgs/Pkg.jl/Project.toml`
 [28b8d3ca] GR v0.57.5 (<0.58.1): Plots
 [5c1252a2] GeometryBasics v0.3.13 (<0.4.1): Plots
 [91a5bcdd] Plots v1.15.3 (<1.20.0)
```

Also, Crayons here is not the latest version but it is the latest version according to our own specific compact (I added `Crayons = "4"` to the Pkg Project.toml file). If you want to ignore that you can do:

```
(Pkg) pkg> st -c --ignore-project-compat
Project Pkg v1.8.0
Status `~/JuliaPkgs/Pkg.jl/Project.toml` 
 [a8cc5b0e] Crayons v3.0.0 (<4.0.4) # <-------------- now visible
 [28b8d3ca] GR v0.57.5 (<0.58.1): Plots
 [5c1252a2] GeometryBasics v0.3.13 (<0.4.1): Plots
 [91a5bcdd] Plots v1.15.3 (<1.20.0)
```

I think it would be good to get this in before deciding on what exactly to show in the default `status` output because then people can play with it and based on that experience the default status can be tweaked.

IanButterworth:
I gave this a bit of a go.

 > if you give the --compat flag it filters out things to only show packages that are not on the latest version

To me the word "--compat" doesn't lend itself to that behavior. Perhaps "--heldback"? of if staying with `--compat` perhaps show all, not just those being held back.

Also, perhaps it should indicate when the package is held back by an explicit `[compat]` entry, and use the same language as the resolver to tie the concepts together.

KristofferC:
Thanks for the comment. 

Regarding showing all packages, and just have a note for those that are upper bounded, I tried that first but it was kind of noisy and for big environments, you were mostly scrolling around trying to find the upper-bounded packages.
How about the following:

```
(Pkg) pkg> st -c
Project Pkg v1.8.0
Status `~/JuliaPkgs/Pkg.jl/Project.toml`
 [a8cc5b0e] Crayons v3.0.0 (<4.0.4) [compat]
 [28b8d3ca] GR v0.57.5 (<0.58.1): Plots
```
This shows that Crayons is upper bounded to 3.0.0 due to project compat. GR is upper bounded to 0.57.5 due to Plots etc.

```
(Pkg) pkg> add Crayons@2
...

(Pkg) pkg> st -c
Project Pkg v1.8.0
Status `~/JuliaPkgs/Pkg.jl/Project.toml`
 [a8cc5b0e] Crayons v2.0.0 (<3.0.0, <4.0.4)
 [28b8d3ca] GR v0.57.5 (<0.58.1): Plots
 [5c1252a2] GeometryBasics v0.3.13 (<0.4.1): Plots
 [91a5bcdd] Plots v1.15.3 (<1.20.0)
```

Here we see that Crayons is upper bounded to 2.0.0 and the max version from compat is 3.0.0 (shown in green) and the max total version is 4.0.4 (shown in yellow).

StefanKarpinski:
Brainstorming option names: `--capped`, `--conflicts`, `--old`?

IanButterworth:
What about `--latest` and group by whether they're on latest or not?

```
On latest release
...
Not on latest release
...
```

KristofferC:
What I liked with `compat` was that it kind of shows the effect that the `compat` section of the packages has on your versions. If there was no `compat`, everything would be on the latest version, with `compat` you have packages that might be on lower version and this option gives you the packages that have added `compat` as to make that happen.

IanButterworth:
Yeah ok. I guess I was thinking about it from the perspective of #2682 but that issue seems to need a separate solution that's more about _latest_.

KristofferC:
Another thing, you could also be restricted by Julia compat. I don't think this is taken into account right now in this code. Gonna have to test it.

martinholters:
`--outdated`?

StefanKarpinski:
That does seem like a good name for this option 👍🏻

KristofferC:
Yep, let's go with that

PR #2292, opened 2020-12-13
Force mv in `create_artifact`

charleskawczynski:
Over at [ClimateMachine.jl](https://github.com/CliMA/ClimateMachine.jl), we're seeing race conditions when using `create_artifact` on multiple processors (e.g., [here](https://buildkite.com/clima/climatemachine-ci/builds/572#d76710eb-c466-4c30-ad3d-4636e5b33bd6/208-218)). This is likely not the best fix but, based on [this thread](https://github.com/JuliaLang/julia/pull/36638#discussion_r453820564), it seems like it may work. Looping in @kpamnany. I'm hoping that this will partially fix [ClimateMachine's 1678](https://github.com/CliMA/ClimateMachine.jl/issues/1678).

Doing this the "right way" seems ultimately dependent on resolving [cross-platform file locking](https://github.com/JuliaLang/julia/issues/7176). [Pidfile.jl](https://github.com/vtjnash/Pidfile.jl) seems like a promising solution, but I'm not sure what the status is on [moving this into Base](https://github.com/JuliaLang/julia/issues/7176#issuecomment-411102022).

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -55,7 +55,7 @@ function create_artifact(f::Function)
         new_path = artifact_path(artifact_hash; honor_overrides=false)
         if !isdir(new_path)
             # Move this generated directory to its final destination, set it to read-only
-            mv(temp_dir, new_path)
+            mv(temp_dir, new_path; force=true)
             chmod(new_path, filemode(dirname(new_path)))
             set_readonly(new_path)
         end


kpamnany:
Ah, you may have missed the [subsequent comment](https://github.com/JuliaLang/julia/pull/36638#discussion_r455879928). So, not sure this will actually help.

charleskawczynski:
Ah, crud I see. You're right. This won't fix it. What if we also passed `force` through from `mv` to `rename` in `julia/base/file.jl`? (right now `force` is not passed through). Even if that did work, I guess this is not really moving things in the right direction, though ☹️.

kpamnany:
There are many, many things depending on particular behavior of the file I/O primitives; can't really change things in Base easily.

We need to MPI I/O all our stuff, which will mean significant PRs to packages we use or specialized forks. Neither option is appealing. 🤷‍♂️ 

Simplest solution for now is:
```julia
If mpirank == 0
    ...
end
MPI.Barrier(MPI.COMM_WORLD)
```
Which is bad for performance, but should avoid distributed races.

PR #2296, opened 2020-12-19
Throw an error if the Julia version requirement is not satisfied

DilumAluthge:
In all other cases, if the `[compat]` entry is not satisfiable, we throw an exception.

Therefore, for consistency, I think it makes sense that if the `[compat]` entry for `julia` is not satisfiable, we should throw an exception.

---

See also: https://discourse.julialang.org/t/resolve-mismatch-between-running-julia-and-the-compatibility-section/52064


KristofferC:
It makes sense from a correctness point of view but at the same time, the first thing you are going to do after this error is to go and change that compat straight away so I wonder if the warning here is arguably more convenient. Not sure.

DilumAluthge:
Let's keep it as a warning for now.

PR #2297, opened 2020-12-20
Fix typo

rikhuijzer:
This seems like a typo


rikhuijzer:
@KristofferC

KristofferC:
Thanks, to be honest, I'm not even sure why we need this method heh.

PR #2301, opened 2020-12-25
Use occursin instead of findfirst(...) === nothing

rikhuijzer:


StefanKarpinski:
This PR seems good. Was there a reason you closed it, @rikhuijzer?

rikhuijzer:
Yes, I boldly opened this PR without testing locally. Then, CI failed and I realized that you cannot always replace `findfirst(...) === nothing` with `occursin`, so I assumed that this PR is just a braking change and hoped that no-one would ever look at it again.

KristofferC:
Yeah, `occursin` is just for looking for occurances inside a string.

PR #2307, opened 2020-12-29
Add REPL hook to prompt to install missing packages, if available in registry

IanButterworth:
Paired with https://github.com/JuliaLang/julia/pull/39026

Adds hooks into REPL to prompt to install packages if missing, and available in registry

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -561,6 +561,7 @@ function __init__()
             end
         end
     end
+    push!(empty!(REPL.install_packages_hooks), REPLMode.try_prompt_pkg_add)
     OFFLINE_MODE[] = get(ENV, "JULIA_PKG_OFFLINE", nothing) == "true"
     return nothing
 end
@@ -671,6 +672,7 @@ const precompile_script = """
     Pkg.add("TestPkg")
     Pkg.develop(Pkg.PackageSpec(path="TestPkg.jl"))
     Pkg.add(Pkg.PackageSpec(path="TestPkg.jl/"))
+    Pkg.REPLMode.try_prompt_pkg_add(Symbol[:notapackage])
     ] add Te\t\t$CTRL_C
     ] st
     $CTRL_C
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -650,4 +650,45 @@ end
 
 const help = gen_help()
 
+function try_prompt_pkg_add(pkgs::Vector{Symbol})
+    ctx = Context()
+    available_uuids = [Types.registered_uuids(ctx.registries, String(pkg)) for pkg in pkgs] # vector of vectors
+    available_pkgs = pkgs[isempty.(available_uuids) .== false]
+    isempty(available_pkgs) && return false
+    resp = try
+        plural1 = length(pkgs) == 1 ? "" : "s"
+        plural2 = length(available_pkgs) == 1 ? "a package" : "packages"
+        plural3 = length(available_pkgs) == 1 ? "is" : "are"
+        plural4 = length(available_pkgs) == 1 ? "" : "s"
+        missing_pkg_list = length(pkgs) == 1 ? String(pkgs[1]) : "[$(join(pkgs, ", "))]"
+        available_pkg_list = length(available_pkgs) == 1 ? String(available_pkgs[1]) : "[$(join(available_pkgs, ", "))]"
+        printstyled(ctx.io, " │ "; color=:green)
+        println(ctx.io, "Package$(plural1) $(missing_pkg_list) not found,",
+                            " but $(plural2) named $(available_pkg_list) $(plural3) available from a registry.")
+        printstyled(ctx.io, " │ "; color=:green)
+        println(ctx.io, "Install package$(plural4)?")
+        printstyled(ctx.io, " │ "; color=:green)
+        print(ctx.io, "  ")
+        printstyled(ctx.io, REPLMode.promptf(); color=:blue)
+        println(ctx.io, "add ", join(available_pkgs, ' '))
+        printstyled(ctx.io, " └ "; color=:green)
+        Base.prompt(stdin, ctx.io, "(y/n)", default = "n")
+    catch err
+        if err isa InterruptException
+            println(ctx.io)
+            return false
+        end
+        rethrow()
+    end
+    if lowercase(resp) in ["y", "yes"]
+        Pkg.add(string.(available_pkgs))
+        if length(available_pkgs) < length(pkgs)
+            return false # declare that some pkgs couldn't be installed
+        else
+            return true
+        end
+    end
+    return false
+end
+
 end #module
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -667,4 +667,9 @@ end
     @inferred Pkg.REPLMode.CompoundSpecs(Pair{String,Vector{Pkg.REPLMode.CommandDeclaration}}[])
 end
 
+@testset "REPL missing package install hook" begin
+    @test Pkg.REPLMode.try_prompt_pkg_add(Symbol[:notapackage]) == false
+    # cannot test installation of findable packages given requires user input to stdin on the prompt
+end
+
 end # module


src/Operations.jl
```diff
@@ -743,6 +743,28 @@ function download_source(ctx::Context, pkgs::Vector{PackageSpec},
...
+    !Types.isresolved(pkg) && return false
+    resp = try
+        Base.prompt(" \033[32m\033[1m└\033[22m\033[0m Package $(repr(String(mod))) not found, but $(repr(String(mod))) is available from a registry.
+   Active project: $(Base.active_project())
+   Add $(repr(String(mod))) to the active project? (y/n)", default = "n")
+    catch err
+        if err isa InterruptException
+            println()
+            return false
+        end
```
oxinabox, line 759: Were you intending to swallow all exceptions silently here?

```suggestion
        end
        rethrow()
```

ianshmean: Nope, thanks!

IanButterworth:
The base hook and handler support is now available given JuliaLang/julia#39026 has been merged. 

The formatting in this PR now looks like 

Screen Shot 2021-04-28 at 10 50 55 AM

fredrikekre:
`CSV` vs `"CSV"` looks a bit strange.

IanButterworth:
My thinking is that..

On the load side: Julia currently prints the load error message like `ERROR: ArgumentError: Package CSV not found in current path:`.
So on the load side it's unquoted

On the registry side: Julia/Pkg can't guarantee that the package in the registry is the same package the user wants by name alone, so the `"CSV"` hints that the user should check that it's the one they want

I've been going back and forth on this though..

fredrikekre:
I don't really see how it hints to that. Perhaps the message can be tweaked somehow instead.

IanButterworth:
How about 
```
julia> using CSV, ImageIO, DoesNotExist
 │ Packages [CSV, ImageIO, DoesNotExist] not found, but packages named [CSV, ImageIO] are available from a registry.
 │ Project `~/.julia/environments/v1.7/Project.toml`
 └ Add [CSV, ImageIO] to the active project and load? (y/n) [n]: 
 ```

**Comment by fredrikekre** [comment]
Yea that looks good. Perhaps printing the project path is not needed either? Or instead print
```
julia> using CSV, ImageIO, DoesNotExist
 │ Packages [CSV, ImageIO, DoesNotExist] not found, but packages named [CSV, ImageIO] are available from a registry.
 └ Add [CSV, ImageIO] to the active project (`~/.julia/environments/v1.7/Project.toml`) and retry? (y/n) [n]: 
```

Perhaps "retry" instead of "load"?

File: src/Pkg.jl [hunk 2]
```diff
@@ -671,6 +672,7 @@ const precompile_script = """
     Pkg.add("TestPkg")
     Pkg.develop(Pkg.PackageSpec(path="TestPkg.jl"))
     Pkg.add(Pkg.PackageSpec(path="TestPkg.jl/"))
+    Pkg.try_prompt_pkg_add(Symbol[:notapackage])
```
fredrikekre, line 675: Won't this block on `prompt`?

IanButterworth: No because `:notapackage` won't be found. But you did make me realize it's missing a module reference
```
julia> Pkg.Operations.try_prompt_pkg_add(Symbol[:notapackage])
false
```

IanButterworth:
Ok. Updated to this.
I don't love adding the project to the prompt line. It makes the prompt line much longer which makes it a bit less visually obvious that it's waiting on user input

Screen Shot 2021-04-28 at 1 33 36 PM

fredrikekre:
Perhaps it doesn't have to be included at all? I guess in general you either don't care about environments and it will mostly confuse you, or you know about environments, and then you know which one you run in .

src/Pkg.jl
fredrikekre: Ok. Is the lag noticable without this?

src/Operations.jl
```diff
@@ -739,6 +739,43 @@ function download_source(ctx::Context; readonly=true)
...
+    available_uuids = [Types.registered_uuids(ctx.registries, String(pkg)) for pkg in pkgs] # vector of vectors
+    available_pkgs = pkgs[isempty.(available_uuids) .== false]
+    isempty(available_pkgs) && return false
+    resp = try
+        plural1 = length(pkgs) == 1 ? "" : "s"
+        plural2 = length(available_pkgs) == 1 ? "a package" : "packages"
+        plural3 = length(available_pkgs) == 1 ? "is" : "are"
+        missing_mod_list = length(pkgs) == 1 ? String(pkgs[1]) : "[$(join(pkgs, ", "))]"
+        available_mod_list = length(available_pkgs) == 1 ? String(available_pkgs[1]) : "[$(join(available_pkgs, ", "))]"
+        println(ctx.io, " \033[32m\033[1m│\033[22m\033[0m Package$(plural1) $(missing_mod_list) not found,",
```
fredrikekre, line 753: This should respect `--color` I guess?

src/Pkg.jl
IanButterworth: I think so

IanButterworth:
In the discussions that lead to the base PR a few people mentioned that showing the active project was important for not confusing the user, given they're not in Pkg mode, and hence don't have the env prefix visible on this action. 

If I was to vote I'd go for the original style with a 2nd line devoted to the project

fredrikekre:
> In the discussions that lead to the base PR a few people mentioned that showing the active project was important for not confusing the user, given they're not in Pkg mode, and hence don't have the env prefix visible on this action.

I see no discussion about this at all in https://github.com/JuliaLang/julia/pull/39026. Honestly I don't quite understand why it would be important. The fact that you typed `using CSV` seems to suggest that you expected `CSV` to already be installed in the current environmen (or the stack).

IanButterworth:
It's from your notes in https://github.com/JuliaLang/julia/pull/26469#issuecomment-523150333 plus slack discussions.

> The fact that you typed using CSV seems to suggest that you expected CSV to already be installed in the current environmen (or the stack).

True, but I see another use model for this.. people wanting to install packages without switching to `pkg>` mode. It's actually quite a slick way of doing it, because all you type us `using Foo` then `y [return]` and the package is installed, precompiled quickly and loaded.

I've added the single line project print back. Not to be forceful, but to present it alongside the other changes. Happy to revert if the consensus is something else

Current form:
image

src/Operations.jl
IanButterworth: Indeed. Thanks. Fixed

IanButterworth:
Two more options

Screen Shot 2021-04-30 at 9 23 02 AM

crstnbr:
I was just about to suggest the same: i.e. why not show it in a similar style as in the Pkg REPL mode. So +1 for this solution from me.

IanButterworth:
It does feel minimal but informative. I've pushed the last one of those two

IanButterworth:
@fredrikekre what do you think about that last example? Given this isn't being backported, perhaps the formatting could be trialled and adjusted based on use up to 1.7?

IanButterworth:
The discussion is still open here but I’m thinking it’s better to merge sooner rather than too close to the 1.7 feature freeze, to see how it goes. I think I’ll merge but happily work with any formatting/info feedback. I’ll merge and update Pkg on master in an hour or so if no one has objection

fredrikekre:
How about making it look kinda like the Pkg REPL mode syntax?
```
(@v1) pkg> add CSV ImageIO
```

IanButterworth:
Yeah there's a really nice elegance to that, but I think the triage discussion was that the default action when the user hits return should be not to install.

So it would have to be something like

```
julia> using CSV, ImageIO, DoesNotExist
 │ Packages [CSV, ImageIO, DoesNotExist] not found, but packages named [CSV, ImageIO] are available from a registry.
 └ (@v1) pkg> add CSV ImageIO <- Run and retry? (y/n) [n]: 
```

IanButterworth:
Or perhaps just
```
julia> using CSV, ImageIO, DoesNotExist
 │ Packages [CSV, ImageIO, DoesNotExist] not found, but packages named [CSV, ImageIO] are available from a registry.
 └ (@v1) pkg> add CSV ImageIO ? (y/n) [n]: 
```

fredrikekre:
Yea, I meant something like the latter.

IanButterworth:
With offline help from @fredrikekre we arrived at this, which seems quite nice

Screen Shot 2021-05-02 at 2 34 13 PM
Screen Shot 2021-05-02 at 2 32 33 PM

PR #2309, opened 2020-12-31
Align all status printing

ianshmean:
Alignment during printing of status info is inconsistent with the rest of pkg indenting and stands out most when the status prints are embedded within other actions. Notice the two `Updating`'s here

Master
Screen Shot 2021-01-04 at 4 17 00 PM

This PR
Screen Shot 2021-01-04 at 4 19 09 PM

This PR turns alignment on for all status prints.
That includes for basic `pkg> status` reports, which stood out less given they were the only text printed, which I understand the reasoning for, but it seems better to be consistent?

Master
Screen Shot 2020-12-31 at 12 41 56 AM

This PR
Screen Shot 2020-12-31 at 12 41 25 AM

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1023,7 +1023,7 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
     prune_manifest(ctx.env)
     # update project & manifest
     write_env(ctx.env)
-    show_update(ctx, ignore_indent=false)
+    show_update(ctx)
 end
 
 update_package_add(ctx::Context, pkg::PackageSpec, ::Nothing, is_dep::Bool) = pkg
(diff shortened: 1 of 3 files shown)


StefanKarpinski:
Seems visually nicer to me.

ianshmean:
@KristofferC @fredrikekre Thoughts on this? It's minor, but thought I'd check

src/Operations.jl
```diff
@@ -1023,7 +1023,7 @@ function rm(ctx::Context, pkgs::Vector{PackageSpec})
     prune_manifest(ctx.env)
     # update project & manifest
     write_env(ctx.env)
-    show_update(ctx)
+    show_update(ctx, ignore_indent=false)
```
fredrikekre, line 1026: What is actually ignoring indent? It seems like these should not be necessary since false is the default?

fredrikekre (commented):
For a sole `status` command I think the current behaviour looks slighly better, but this makes it more consistent and looks better in combination with the rest.

src/Operations.jl
ianshmean: True. Removed. Thanks

ianshmean:
I feel the same. Shall I merge?


I don't plan on getting in the habit of merging.. but this seems so minor I don't want to waste people's time with it

fredrikekre:
Go for it!

PR #2310, opened 2020-12-31
Make a <return> on an empty line return to julia REPL

mcabbott:
This lets you type `] add This, That` and press return *twice*,  then immediately paste in the code you wanted to try, and leave that window alone until it's finished downloading, precompiling, loading, etc. 

For some reason I frequently want to do this, but there seems to be no way at present -- the backspace key to exit the `pkg>` mode isn't cached. I don't know how to change the backspace; the change here is simply half of #2248.

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -549,7 +549,9 @@ function create_mode(repl::REPL.AbstractREPL, main::LineEdit.Prompt)
         do_cmd(repl, input)
         REPL.prepare_next(repl)
         REPL.reset_state(s)
-        s.current_mode.sticky || REPL.transition(s, main)
+        if isempty(input) || !s.current_mode.sticky
+            REPL.transition(s, main)
+        end
     end
 
     mk = REPL.mode_keymap(main)


KristofferC:
You could also just use `Pkg.add(["This", "That"])`. Personally, it seems slightly odd/surprising to special case the Pkg REPL like this, why shouldn't that be the case for other REPL modes like shell mode?

mcabbott:
Yes that's one work-around, but more to type. And you have to think of it before starting the command (and realising you're on the slow network today).

Maybe this should be marked RFC, to see if it sounds weird or useful to others. I guess it's a way of making modes slightly less sticky, and perhaps shell mode could follow?

KristofferC:
I think a better solution would be to extend the prompt pasting  functionality of the REPL (the one that allows you to directly paste `julia> 1+1` into the REPL) to also handle different modes. Then you can just copy-paste the whole thing in one chunk.

mcabbott:
That would be neat, but solves related but different problem: It works if you have a REPL printout to paste, but won't work if you manually install some packages and then wish to immediately run code, without waiting for installation / registry update etc to finish.

Maybe I should make an issue for the issue, instead of for the proposed solution? It just seemed simpler to show than tell.

PR #2312, opened 2021-01-02
document activate --temp

### LINKED ISSUE #1944: document --temp
Could not find any documentation online for the `--temp` flag for `activate`. It also does not show for `?Pkg.activate`, only for `?activate` (from `pkg` mode).

```
julia> using Pkg

help?> Pkg.activate
  Pkg.activate([s::String]; shared::Bool=false)

  Activate the environment at s. The active environment is the environment that is modified by executing package commands. The logic for what path is activated is as follows:

    •    If shared is true, the first existing environment named s from the depots in the depot stack will be activated. If no such environment exists, create and activate that environment in the first depot.

    •    If s is an existing path, then activate the environment at that path.

    •    If s is a package in the current project and s is tracking a path, then activate the environment at the tracked path.

    •    Otherwise, s is interpreted as a non-existing path, which is then activated.

  If no argument is given to activate, then activate the home project. The home project is specified by either the --project command line option to the julia executable, or the JULIA_PROJECT environment variable.

  Examples
  ≡≡≡≡≡≡≡≡≡≡

  Pkg.activate()
  Pkg.activate("local/path")
  Pkg.activate("MyDependency")
```

```
(@v1.5) pkg> ?activate
  activate
  activate [--shared|--temp] [path]

  Activate the environment at the given path, or the home project environment if no path is specified. The active environment is the environment that is modified by executing package commands. When the option --shared is given,
  path will be assumed to be a directory name and searched for in the environments folders of the depots in the depot stack. In case no such environment exists in any of the depots, it will be placed in the first depot of the
  stack. Use the temp option to create temporary environments. This should be useful for experimenting with packages.
```

oxinabox:
closes #1944

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -391,7 +391,7 @@ const status = API.status
 
 
 """
-    Pkg.activate([s::String]; shared::Bool=false)
+    Pkg.activate([s::String]; temp::Bool=false, shared::Bool=false)
 
 Activate the environment at `s`. The active environment is the environment
 that is modified by executing package commands.
@@ -400,6 +400,9 @@ The logic for what path is activated is as follows:
   * If `shared` is `true`, the first existing environment named `s` from the depots
     in the depot stack will be activated. If no such environment exists,
     create and activate that environment in the first depot.
+  * If `temp` is `true` this will create and activate a temporary enviroment which will
+    be deleted when the julia process is exitted. The `s` argument must not be passed
+    with the `temp` option set to `true`.
   * If `s` is an existing path, then activate the environment at that path.
   * If `s` is a package in the current project and `s` is tracking a path, then
     activate the environment at the tracked path.
@@ -414,6 +417,7 @@ the julia executable, or the `JULIA_PROJECT` environment variable.
 Pkg.activate()
 Pkg.activate("local/path")
 Pkg.activate("MyDependency")
+Pkg.activate(; temp=true)
 ```
 """
 const activate = API.activate
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -391,7 +391,8 @@ const status = API.status
 
 
 """
-    Pkg.activate([s::String]; temp::Bool=false, shared::Bool=false)
+    Pkg.activate([s::String]; shared::Bool=false)
+    Pkg.activate(; temp::Bool=false)
 
 Activate the environment at `s`. The active environment is the environment
 that is modified by executing package commands.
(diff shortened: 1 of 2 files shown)


src/Pkg.jl
```diff
@@ -400,6 +400,9 @@ The logic for what path is activated is as follows:
   * If `shared` is `true`, the first existing environment named `s` from the depots
     in the depot stack will be activated. If no such environment exists,
     create and activate that environment in the first depot.
+  * If `temp` is `true` this will create and activate a temporary enviroment which will
+    be deleted when the julia process is exitted. The `s` argument must not be passed
```
oxinabox, line 404:
```suggestion
    be deleted when the julia process is exited. The `s` argument must not be passed
```

src/Pkg.jl
```diff
@@ -391,7 +391,7 @@ const status = API.status
 
 
 """
-    Pkg.activate([s::String]; shared::Bool=false)
+    Pkg.activate([s::String]; temp::Bool=false, shared::Bool=false)
```
fredrikekre, line 394:
```suggestion
    Pkg.activate([s::String]; shared::Bool=false)
    Pkg.activate(; temp::Bool=false)
```

src/Pkg.jl
```diff
@@ -400,6 +400,9 @@ The logic for what path is activated is as follows:
   * If `shared` is `true`, the first existing environment named `s` from the depots
     in the depot stack will be activated. If no such environment exists,
     create and activate that environment in the first depot.
+  * If `temp` is `true` this will create and activate a temporary enviroment which will
+    be deleted when the julia process is exited. The `s` argument must not be passed
+    with the `temp` option set to `true`.
```
fredrikekre, line 405:
```suggestion
    be deleted when the julia process is exited.
```

src/REPLMode/command_declarations.jl
```diff
@@ -256,7 +256,8 @@ The active environment is the environment that is modified by executing package
 When the option `--shared` is given, `path` will be assumed to be a directory name and searched for in the
 `environments` folders of the depots in the depot stack. In case no such environment exists in any of the depots,
 it will be placed in the first depot of the stack.
-Use the `temp` option to create temporary environments. This should be useful for experimenting with packages.
+Use the `--temp` option to create temporary environments which are removed when the julia
+process is exited. This is useful for experimenting with packages.
```
fredrikekre, line 260: Don't think we need to motivate this?
```suggestion
process is exited.
```

src/Pkg.jl
oxinabox: Oh that is much better

src/REPLMode/command_declarations.jl
oxinabox: I don't think so either, but i wasn't sure i wanted to fully delete the motivation that was alread there

PR #2319, opened 2021-01-07
[master] Teach`Pkg.add()` to write Manifests with old Julia versions

staticfloat:
This extends the `julia_version` context field to effect `julia_version`
within `Pkg.add()` such that if you pass an older `julia_version` (or,
in the extreme case, `nothing`) you can force Julia to write out a
Manifest with new stdlibs (such as `NetworkOptions`) with versions,
without treating them as standard libraries.  If you are running a newer
Julia that has these standard libraries baked into the system image, you
cannot (yet) actually load these new packages, but in the future this
may be possible.  In a more practical sense, this will allow
BinaryBuilder to collect all packages (and their attendant
`Artifacts.toml` files) for a particular Julia version, or completely
ignoring Julia version by using `julia_version=nothing` and getting the
latest possible version of all stdlibs.

Version targeting `release-1.6`: https://github.com/JuliaLang/Pkg.jl/pull/2318


KristofferC:
LGTM from a cursory review.

src/Types.jl
```diff
@@ -363,6 +364,36 @@ function stdlibs()
...
+    # will be treated like normal packages.
+    if julia_version === nothing
+        return false
+    end
+
+    # If we are given an actual version, find the entry in `STDLIBS_BY_VERSION`
+    # that corresponds to the requested version, and use that.
+    last_stdlibs = Dict{UUID,String}()
+    for (version, stdlibs) in STDLIBS_BY_VERSION
+        if julia_version < version
```
fredrikekre, line 384: I think this might be wrong
```
julia> VERSION
v"1.6.0-beta1.2"

julia> VERSION < v"1.6"
true
```

PR #2322, opened 2021-01-08
Move artifact to the expected location

### LINKED ISSUE #2317: Artifact hashes wrong on MacOS when depot resides on a FAT32 drive
In testing performance related to slow disk access, I moved the depot onto a FAT32 formatted USB pen drive and noticed that the artifact hashes are repeatably different.

This is probably highly unlikely to happen in the wild, but I thought I'd report.

```
julia> versioninfo()
Julia Version 1.7.0-DEV.185
Commit 770e228c61 (2020-12-29 00:32 UTC)
Platform Info:
  OS: macOS (x86_64-apple-darwin19.5.0)
  CPU: Intel(R) Core(TM) i7-8850H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-11.0.0 (ORCJIT, skylake)
Environment:
  JULIA_NUM_THREADS = 6
  JULIA_DEPOT_PATH = /Volumes/pendrive/.julia

(@v1.7) pkg> add libpng_jll
   Resolving package versions...
   Installed libpng_jll ─ v1.6.37+6
 Downloading artifact: libpng
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   fb2695d9f2ba598ca2213b0768aac906ebcd1000
│   Calculated git-tree-sha1: 0e2015d6296e3f1119adcc39a621077dfe5ce50f
  Downloaded artifact: libpng

(@v1.7) pkg> st
Status `/Volumes/pendrive/.julia/environments/v1.7/Project.toml`
→ [b53b4c65] libpng_jll v1.6.37+6
Info packages marked with → not downloaded, use `instantiate` to download

(@v1.7) pkg> instantiate
 Downloading artifact: libpng
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   fb2695d9f2ba598ca2213b0768aac906ebcd1000
│   Calculated git-tree-sha1: 0e2015d6296e3f1119adcc39a621077dfe5ce50f
  Downloaded artifact: libpng
 Downloading artifact: libpng
┌ Error: Tree Hash Mismatch!
│   Expected git-tree-sha1:   fb2695d9f2ba598ca2213b0768aac906ebcd1000
│   Calculated git-tree-sha1: 0e2015d6296e3f1119adcc39a621077dfe5ce50f
  Downloaded artifact: libpng

julia> using libpng_jll
ERROR: InitError: Artifact "libpng" was not installed correctly. Try `using Pkg; Pkg.instantiate()` to re-install all missing resources.
Stacktrace:
```

fredrikekre:
Move artifact to the expected location when tree hash computation goes wrong, fixes #2317.

~Probably need to update some tests.~ Nope, but added one.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -349,10 +349,22 @@ function download_artifact(
             msg  = "Tree Hash Mismatch!\n"
             msg *= "  Expected git-tree-sha1:   $(bytes2hex(tree_hash.bytes))\n"
             msg *= "  Calculated git-tree-sha1: $(bytes2hex(calc_hash.bytes))"
+            # Since tree hash calculation is still broken on some systems, e.g. Pkg.jl#1860,
+            # and Pkg.jl#2317 so we allow setting JULIA_PKG_IGNORE_HASHES=1 to ignore the
+            # error and move the artifact to the expected location and return true
+            ignore_hash = get(ENV, "JULIA_PKG_IGNORE_HASHES", nothing) == "1"
+            if ignore_hash
+                msg *= "\n\$JULIA_PKG_IGNORE_HASHES is set to 1: ignoring error and moving artifact to the expected location"
+            end
             @error(msg)
-            # Tree hash calculation is still broken on some systems, e.g. Pkg.jl#1860,
-            # so we return true here and only raise the warning on the lines above.
-            # return false
+            if ignore_hash
+                # Move it to the location we expected
+                src = artifact_path(calc_hash; honor_overrides=false)
+                dst = artifact_path(tree_hash; honor_overrides=false)
+                mv(src, dst; force=true)
+                return true
+            end
+            return false
         end
     end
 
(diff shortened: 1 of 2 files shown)


staticfloat:
Are we okay with this?  I'm not sure we should be installing something with an incorrect hash by default.  I would rather we only do this if the user has exported `JULIA_PKG_IGNORE_HASHES=1` or something, because this is the only check that prevents a nightmare scenario like e.g. a compromised Pkg server sending out viruses to everyone.  Disabling this check completely breaks the chain of trust, so I think we should treat it like `-k` on `curl`.  Useful in rare situations, but definitely not something you want on by default.

fredrikekre:
Sure, having an option seems fine.

fredrikekre:
Should probably revert #1885 too then

staticfloat (approved):

PR #2330, opened 2021-01-11
Change refs/* to refs/heads/* to speed up repo cloning w/ many branches

### LINKED ISSUE #2329: Pkg.add on repo w/ many branches can be extremely slow
Trying to run:
```julia
Pkg.add(url="https://github.com/apache/arrow", subdir="julia/Arrow")
```
takes an extremely long time (though I'm told some people were patient enough to have it actually finish). It was suggested that it's probably due to the `apache/arrow` repo having a ton of branches and that changing [this line](https://github.com/JuliaLang/Pkg.jl/blob/0af9d427a6ee3e71092779a32e6cc348f0f86dce/src/Types.jl#L409) to:
```julia
const refspecs = ["+refs/heads/*:refs/remotes/cache/*"]
```
might solve the problem.

quinnj:
Fixes #2329. Fix suggested by @staticfloat and discussed with @KristofferC. I've tested this only on the command in the #2329 issue and it makes the `Pkg.add` much much faster (like 5-10 seconds vs. several minutes).


quinnj:
Hmmmm, unfortunately I dont' know anything about this code, so not sure what this test failure means:
```julia
update: package state changes: Test Failed at /home/runner/work/Pkg.jl/Pkg.jl/test/new.jl:1461
  Expression: pkg.version == v"0.3.0"
   Evaluated: v"0.2.0" == v"0.3.0"
Stacktrace:
  [1] (::Main.PkgTests.NewTests.var"#430#451")(pkg::Pkg.API.PackageInfo)
    @ Main.PkgTests.NewTests ~/work/Pkg.jl/Pkg.jl/test/new.jl:1461
  [2] dependencies
    @ ~/work/Pkg.jl/Pkg.jl/src/API.jl:90 [inlined]
  [3] (::Main.PkgTests.NewTests.var"#427#448")(tempdir::String)
    @ Main.PkgTests.NewTests ~/work/Pkg.jl/Pkg.jl/test/new.jl:1459
  [4] mktempdir(fn::Main.PkgTests.NewTests.var"#427#448", parent::String; prefix::String)
    @ Base.Filesystem ./file.jl:732
  [5] mktempdir
    @ ./file.jl:730 [inlined]
  [6] #426
    @ ~/work/Pkg.jl/Pkg.jl/test/new.jl:1425 [inlined]
  [7] (::Main.PkgTests.Utils.var"#4#7"{Bool, Main.PkgTests.NewTests.var"#426#447"})()
    @ Main.PkgTests.Utils ~/work/Pkg.jl/Pkg.jl/test/utils.jl:53
  [8] withenv(::Main.PkgTests.Utils.var"#4#7"{Bool, Main.PkgTests.NewTests.var"#426#447"}, ::Pair{String, Nothing}, ::Vararg{Pair{String, Nothing}})
    @ Base ./env.jl:161
  [9] isolate(fn::Main.PkgTests.NewTests.var"#426#447"; loaded_depot::Bool)
    @ Main.PkgTests.Utils ~/work/Pkg.jl/Pkg.jl/test/utils.jl:44
 [10] macro expansion
    @ ~/work/Pkg.jl/Pkg.jl/test/new.jl:1425 [inlined]
 [11] macro expansion
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.7/Test/src/Test.jl:1152 [inlined]
 [12] top-level scope
    @ ~/work/Pkg.jl/Pkg.jl/test/new.jl:1384
```

staticfloat:
In that test, we first create a git repo with a package in it:

https://github.com/JuliaLang/Pkg.jl/blob/master/test/new.jl#L1426

We ensure it is of the expected version, then we bump it and commit that new version:

https://github.com/JuliaLang/Pkg.jl/blob/master/test/new.jl#L1438-L1443

Then we `Pkg.update()` and see if the changes have taken effect:

https://github.com/JuliaLang/Pkg.jl/blob/master/test/new.jl#L1459-L1465

This makes me think that the `refs/*` somehow catches an update that `refs/heads/*` doesn't.  Perhaps the way that tracking is setup within a git repository on file remotes is different than on HTTPS remotes?

DilumAluthge:
I think that in addition to `refs/heads/*`, you'll also need `refs/remotes/*` or something like that?

quinnj:
I'm not sure I'll have the time/skills to do much more on this soon, so if anyone would be willing to pick it up in the meantime, I'd appreciate it.

KristofferC:
Claude suggested the last commit as a fix and it seems to work. If I do `pkg> add https://github.com/apache/arrow` (which will fail after all the cloning and updating is done etc but it works fine as a benchmark) it is much faster here than with the default.

IanButterworth:
Great. Backport?

KristofferC:
I think we can do even better than this by specifying the exact branch we want to use etc but this is an incremental improvement so should hopefully be fine.

MilesCranmer:
I think this is causing some CI failures in Julia packages that try to check out code from PRs, e.g., https://github.com/MilesCranmer/AirspeedVelocity.jl/issues/113. Is there a way we can manually disable this?

Edit: #4422

KristofferC:
I mean you can push to the https://github.com/JuliaLang/Pkg.jl/pull/2330/files#diff-39d5f9e47d567ec5ed9f74c56cd46d75ad2d1b2752e4853eb5bb0cffa5234498R702 variable but that is obviously not ideal.

MilesCranmer:
Will do that for now. Can the change be reverted? It seems it fundamentally alters the behavior of `Pkg.add`. I understand there's this speedup, but that seems to only be because it removes some functionality, not that it's more efficient.

I think it would be better to have a special keyword for this rather than changing the standard behavior, perhaps `heads=false` by default, and `heads=true` to enable this reduced refspec

KristofferC:
Can you elaborate the exact situation that causes this to happen? It doesn't be wide spread so what is the unique thing about this situation?

KristofferC:
Imo let's continue the discussion in https://github.com/JuliaLang/Pkg.jl/issues/4422 instead of in a closed PR.

Issue #2302, opened 2020-12-26
AssertionError in is_package_downloaded


timholy:
```julia
(AbstractPlotting) pkg> up
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `git@github.com:JuliaRegistries/General.git`
    Updating registry at `~/.julia/registries/HolyLabRegistry`
    Updating git-repo `git@github.com:HolyLab/HolyLabRegistry.git`
No Changes to `~/.julia/dev/AbstractPlotting/Project.toml`
No Changes to `~/.julia/dev/AbstractPlotting/Manifest.toml`
ERROR: AssertionError: sourcepath !== nothing
Stacktrace:
  [1] is_package_downloaded(ctx::Pkg.Types.Context, pkg::Pkg.Types.PackageSpec)
    @ Pkg.Operations ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:1757
  [2] #12
    @ ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:118 [inlined]
  [3] _all
    @ ./reduce.jl:923 [inlined]
  [4] #all#696
    @ ./reducedim.jl:886 [inlined]
  [5] all
    @ ./reducedim.jl:886 [inlined]
  [6] is_instantiated(ctx::Pkg.Types.Context)
    @ Pkg.Operations ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:118
  [7] instantiate(ctx::Pkg.Types.Context; manifest::Nothing, update_registry::Bool, verbose::Bool, platform::Base.BinaryPlatforms.Platform, allow_autoprecomp::Bool, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:1301
  [8] precompile(ctx::Pkg.Types.Context; internal_call::Bool, kwargs::Base.Iterators.Pairs{Union{}, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API ~/src/julia-master/usr/share/julia/stdlib/v1.6/Pkg/src/API.jl:917
  [9] _auto_precompile(ctx::Pkg.Types.Context)
[...]

PR #2338, opened 2021-01-14
Fix `is_instantiated()`

staticfloat:
Previously, `is_instantiated()` would only check for the presence of
dependencies, which caused artifacts belonging to the top-level project
itself to be missed in its check.  This rectifies that oversight.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -110,7 +110,11 @@ end
 function is_instantiated(env::EnvCache)::Bool
     # Load everything
     pkgs = load_all_deps(env)
-    # Make sure all paths exist
+    # If the top-level project is a package, ensure it is instantiated as well
+    if env.pkg !== nothing
+        push!(pkgs, Types.PackageSpec(name=env.pkg.name, uuid=env.pkg.uuid, version=env.pkg.version, path=dirname(env.project_file)))
+    end
+    # Make sure all paths/artifacts exist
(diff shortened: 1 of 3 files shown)


src/Operations.jl
```diff
@@ -110,7 +110,11 @@ end
 function is_instantiated(env::EnvCache)::Bool
     # Load everything
     pkgs = load_all_deps(env)
-    # Make sure all paths exist
+    # If the top-level project is a package, ensure it is instantiated as well
+    if env.pkg !== nothing
+        push!(pkgs, Types.PackageSpec(name=env.pkg.name, uuid=env.pkg.uuid, version=env.pkg.version, path=dirname(env.project_file)))
```
fredrikekre, line 115: This might also have fixed https://github.com/JuliaLang/Pkg.jl/issues/2302

timholy: Yes!

fredrikekre: Not quite, because with the my testcase from https://github.com/JuliaLang/Pkg.jl/issues/2302#issuecomment-763594149 still fails since, in that case, the package is already in the `pkgs` vector. SO this code here should first search for a package with the same uuid in `pkgs` and modify that, and if it does not exist, then `push!`.

timholy: Unless I screwed something up, for me that testcase worked.

PR #2345, opened 2021-01-18
Catch errors during `recursive_dir_size()`

staticfloat:
While trying to clean up artifacts on Yggdrasil's CI bots, we found that some malformed artifacts (that had things like extremely-deeply-nested directories) would cause `ENAMETOOLONG` in a `stat()` within `walkdir()` itself.  Let's catch these kinds of issues and report them to the user properly.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -839,15 +839,19 @@ function gc(ctx::Context=Context(); collect_delay::Period=Day(7), verbose=false,
 
     function recursive_dir_size(path)
         size = 0
-        for (root, dirs, files) in walkdir(path)
-            for file in files
-                path = joinpath(root, file)
-                try
-                    size += lstat(path).size
-                catch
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -839,15 +839,19 @@ function gc(ctx::Context=Context(); collect_delay::Period=Day(7), verbose=false,
...
+                    path = joinpath(root, file)
+                    try
+                        size += lstat(path).size
+                    catch ex
+                        @warn("Failed to calculate size of $path", ex)
+                    end
                 end
             end
+        catch ex
+            @warn("Failed to calculate size of $path", ex)
```
KristofferC, line 854: Perhaps

``` 
@error "foo" exception = ex
```

gives better output? I think that is the idiomatic way to attach an exception to a log message.

KristofferC (approved):

Issue #2348, opened 2021-01-19
Performance of Pkg.update


DhruvaSambrani:
Running `] update` is extremely slow on Windows, and gets stuck at 
```
Updating registry at `C:?Users/username/.julia/registries/General` 
```
for well above 3 minutes. The rest of the steps are fine

Running `] update` on julia on WSL on the same device updates packages instantaneously.
Issue persists even if RealTime Scanning is stopped on Windows McAfee

StefanKarpinski:
What version of Julia are you using?

DhruvaSambrani:
V1.5.3 on both, WSL and Windows

ianshmean:
I profiled `Pkg.update()` on a fresh depot in Windows on 1.5.3. It took 3m42s. Here's the profile report
https://gist.github.com/ianshmean/e31a31d7287c9a5b843d2dac7b56b2b2

It seems like a majority of time is spent during `cp`. Master uses `mv`.

I tried the same for nightly, which took 2m20s
https://gist.github.com/ianshmean/ccfe5ff232b87f468183507f651d1ebe

I also tried a 2nd time without wiping the depot and hit the error below, which highlighted that while installing a fresh registry no longer uses `cp` updating an existing one still does
https://github.com/JuliaLang/Pkg.jl/blob/fe759b2449b3ace291b6a5e43bd7ecff1767ff91/src/Registry/Registry.jl#L305

Could that be a `mv`?

```
julia> import Pkg; Pkg.update()
    Updating registry at `C:\Users\ian\.julia\registries\General`
ERROR: IOError: stat("C:\\Users\\ian\\.julia\\registries\\General"): permission denied (EACCES)
Stacktrace:
  [1] uv_error
    @ .\libuv.jl:97 [inlined]
  [2] stat(path::String)
    @ Base.Filesystem .\stat.jl:69
  [3] ispath
    @ .\stat.jl:311 [inlined]
  [4] checkfor_mv_cp_cptree(src::String, dst::String, txt::String; force::Bool)
    @ Base.Filesystem .\file.jl:301
  [5] cptree(src::String, dst::String; force::Bool, follow_symlinks::Bool)
    @ Base.Filesystem .\file.jl:324
  [6] cp(src::String, dst::String; force::Bool, follow_symlinks::Bool)
    @ Base.Filesystem .\file.jl:356
[...]

Issue #1667, opened 2020-02-12
 SystemError: opening file "C:\\Users\\justi\\.julia\\registries\\General\\Registry.toml": No such file or directory


justintong97:
If you have a question please search or post to our Discourse site: https://discourse.julialang.org.
We use the GitHub issue tracker for bug reports and feature requests only.

If you're submitting a bug report, be sure to include as much relevant information as
possible, including a minimal reproducible example and the output of `versioninfo()`.
If you're experiencing a problem with a particular package, open an issue on that
package's repository instead.
When i type Pkg.add("IJulia")
it returns: SystemError: opening file "C:\\Users\\justi\\.julia\\registries\\General\\Registry.toml": No such file or directory

Thanks for contributing to the Julia project!

fredrikekre:
Can you try to delete `C:\Users\justi\.julia\registries\General` and try again?

justintong97:
it works， thanks for solving my problem

获取 Outlook for iOS<https://aka.ms/o0ukef>
________________________________
发件人: Fredrik Ekre <notifications@github.com>
发送时间: Wednesday, February 12, 2020 1:11:49 AM
收件人: JuliaLang/Pkg.jl <Pkg.jl@noreply.github.com>
抄送: Tong, Jiacheng <jiacheng-tong@uiowa.edu>; Author <author@noreply.github.com>
主题: [External] Re: [JuliaLang/Pkg.jl] SystemError: opening file "C:\\Users\\justi\\.julia\\registries\\General\\Registry.toml": No such file or directory (#1667)


Can you try to delete C:\Users\justi\.julia\registries\General and try again?

―

00vareladavid:
Perhaps we can detect this ourselves and re-download General if this is the case?

AndyAbok:
> Can you try to delete `C:\Users\justi\.julia\registries\General` and try again?

when you say delete how do you mean facing the same issue over here when tried to add the general after removing it since DataFrames was having issues with versioning

julia> Pkg.Registry.add("General")
    Cloning registry from "https://github.com/JuliaRegistries/General.git"
ERROR: SystemError: opening file "C:\\Users\\andrewa\\.julia\\registries\\General\\Registry.toml": No such file or directory
Stacktrace:
 [1] systemerror(::String, ::Int32; extrainfo::Nothing) at .\error.jl:168
 [2] #systemerror#50 at .\error.jl:167 [inlined]
[...]

PR #2349, opened 2021-01-20
Registry: Change update from `cp` to `mv`

ianshmean:
- ~~Don't leave an incomplete registry if mv fails or is interrupted
An attempt to reduce the chances of https://github.com/JuliaLang/Pkg.jl/issues/1667 happening.
Might be a fix, although the move from `cp` to `mv` may have made the likelihood much smaller already?~~

- Use `mv` rather than `cp` when updating the registry. 
`cp` seems A LOT slower on windows than `mv` https://github.com/JuliaLang/Pkg.jl/issues/2348#issuecomment-763346086

### DIFF
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -186,7 +186,12 @@ function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=d
                         "`$(Base.contractuser(joinpath(depot, "registries", registry.name*"-2")))`."))
                 end
             else
-                mv(tmp, regpath)
+                try
+                    mv(tmp, regpath)
+                catch e
+                    isdir(regpath) && rm(regpath, recursive=true, force=true) # don't leave an incomplete registry
+                    rethrow()
+                end
                 printpkgstyle(io, :Added, "registry `$(registry.name)` to `$(Base.contractuser(regpath))`")
             end
         end
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -293,16 +293,17 @@ function update(regs::Vector{RegistrySpec} = RegistrySpec[]; io::IO=DEFAULT_IO[]
                 if url !== nothing && (new_hash = pkg_server_url_hash(url)) != old_hash
                     let new_hash = new_hash
                         # TODO: update faster by using a diff, if available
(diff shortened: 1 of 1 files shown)


KristofferC:
Shouldn't the change from cp to mv have fixed this? Mv of a folder doesn't move things one by one, or?

ianshmean:
Yeah true.. Perhaps that issue needs someone to chip in and say it's fixed in 1.6. There's a lot of speculation in there

KristofferC:
I think we can drop everything from here except the `cp` -> `mv` change.

src/Registry/Registry.jl
```diff
@@ -293,17 +298,16 @@ function update(regs::Vector{RegistrySpec} = RegistrySpec[]; io::IO=DEFAULT_IO[]
...
-                            try
-                                download_verify_unpack(url, nothing, tmp, ignore_existence = true)
-                            catch err
-                                @error "could not download $url" exception=err
-                            end
-                            tree_info_file = joinpath(tmp, ".tree_info.toml")
-                            hash = pkg_server_url_hash(url)
-                            write(tree_info_file, "git-tree-sha1 = " * repr(string(new_hash)))
-                            cp(tmp, reg.path, force=true)
+                        tmp = mktempdir(cleanup=false) # no cleanup since we mv it
```
StefanKarpinski, line 301: Cleanup shouldn't do anything if the temp path is already gone, so it should be safe to leave cleanup on in case it somehow doesn't get cleaned up otherwise.

ianshmean: Ok, makes sense. Updated

ianshmean:
> I think we can drop everything from here except the cp -> mv change.

Done. 

5 commits for a 2 char change.
i tend to think it's better to leave the history of commits like this so the discussion can be reviewed, and given it'll be squashed anyway. Is that generally ok practice?

ianshmean:
@KristofferC I actually don't believe this needs backporting to 1.6. It appears to `mv` already

https://github.com/JuliaLang/Pkg.jl/blob/cb090a1517f2d398842841375a47330598331a2c/src/Types.jl#L1045

KristofferC:
Oh, ok.

PR #2352, opened 2021-01-21
Add test for #2302

timholy:
Closes #2302

PR #2354, opened 2021-01-23
[release-1.6] more backport

ianshmean:
- Add test for #2302 (https://github.com/JuliaLang/Pkg.jl/pull/2352)


ianshmean:
I've re-run tests several times and keep hitting `failed to get tarball from ["https://github.com/JuliaData/Parsers.jl.git"]` for ubuntu without the pkg server.
Which happens during https://github.com/JuliaLang/Pkg.jl/blob/1f12cac2e35261bf93063383bafe8f2072731469/test/pkg.jl#L408

fredrikekre:
https://github.com/JuliaLang/Pkg.jl/pull/2354/commits/8f6ee65bbabcc7f17b3552e22c18e1da947322c4 shouldn't include the new registry file I guess.

KristofferC:
oops

PR #2355, opened 2021-01-23
Fix download prints respecting IO, and show when downloads fail

### LINKED ISSUE #2328: Artifact downloading does not respect io
```
$ JULIA_DEPOT_PATH=$(mktemp -d) julia --project=. -e 'using Pkg; Pkg.add("SQLite"; io=devnull)'
  Downloaded artifact: SQLite
```

IanButterworth:
- Fixes #2328. All download activity printing is passed to the given io (which can be `devnull` to suppress printing)

- Now reports when an artifact download fails. 
Previously Pkg was reporting "Downloaded.." even if an artifact download failed, which looks a bit odd.
If the first download attempt fails for an artifact, I've enabled printing of the domain of both the failure and the subsequent success/failures.
There's an existing todo to only try the pkg server if it knows about the artifact, which would make this all a bit tidier for artifacts that aren't actually served there.
If the first download attempts just work, the printing remains minimal.

## Before
- Download IO isn't suppressed by `io=devnull`
- Download failures are reported as successful downloads, and no detail is provided

Screen Shot 2021-01-23 at 1 34 38 AM

## This PR

Screen Shot 2021-01-23 at 3 08 35 AM

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -410,14 +410,16 @@ function ensure_artifact_installed(name::String, meta::Dict, artifacts_toml::Str
     hash = SHA1(meta["git-tree-sha1"])
 
     if !artifact_exists(hash)
+        first_attempt = true
         # first try downloading from Pkg server
         # TODO: only do this if Pkg server knows about this package
         if (server = pkg_server()) !== nothing
             url = "$server/artifact/$hash"
-            download_success = with_show_download_info(io, name, quiet_download) do
+            download_success = with_show_download_info(io, name, url, quiet_download, first_attempt) do
                 download_artifact(hash, url; verbose=verbose, quiet_download=quiet_download, io=io)
             end
             download_success && return artifact_path(hash)
+            first_attempt = false
         end
 
         # If this artifact does not exist on-disk already, ensure it has download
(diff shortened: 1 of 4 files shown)


ianshmean:
Thoughts on this? It's more correct, but are we ok showing failures like this when the artifact ultimately gets downloaded?

Perhaps the non-fatal download failure could be warning colored rather than red.

KristofferC:
Maybe we can add the URL to the error message when it fails?

StefanKarpinski:
I think it's better for now that we show these so that we can get feedback on non-fatal issues people are having. There's a more fine-grained distinction to be made which is between downloads that we expect may fail so we probably don't want to warn about them failing and downloads that we expect should succeed and if they don't, even if we have another way of getting them, we would want to print a warning so that the matter can be investigated. I don't think the current architecture of Pkg makes this something we can reasonably distinguish, however.

fredrikekre:
It is a bit inconsistent with how registry failures are printed (see your screenshot).

ianshmean:
> It is a bit inconsistent with how registry failures are printed

Yeah I thought about that. The issue is that the artifact download works differently to the package tarball download process. Artifacts do discrete download actions walking through the different services, while for package tarballs a list of urls is populated, then download is tried on each. 

It seems a bit easier with the package tarball approach to fail/log gracefully, because you know if there are more downloads to try at each download step.

That being said, I'm happy to try conforming the reporting style of both

IanButterworth:
The IO handling was fixed by https://github.com/JuliaLang/Pkg.jl/pull/2391
and some debug output was added for artifact downloads in https://github.com/JuliaLang/Pkg.jl/pull/2544

There may be some ideas to salvage here but I don't think it needs to stay open

Issue #2358, opened 2021-01-23
Detecting a precompile failure


omus:
As part of a Docker build we're using `Pkg.API.precompile()`. We've noticed that if a failure occurs with this function no exception or return is provided to indicate that a failure has taken place. This is problematic as it's hard to provide Docker with a non-zero return codes to indicate the `RUN` instruction has failed. With newer versions of Docker the flag `--progress=auto` is the default which automatically hides the output generated during successful `RUN` commands making pre-compilation failures practically invisible.

Could we update `Pkg.API.precompile()` with an option to throw and exception or return success/failure?

ianshmean:
When you call `precompile` directly, it will throw for any errors from direct dependencies. i.e. deps in the Project.toml

Indirect dependencies may fail but won't throw, unless they are actually called by direct dependencies during their own precompilation.

During autoprecomp, direct deps will report failure in the log but won't throw

omus:
Interesting. In this particular case the failure occurred during the `__init__` of an indirect dependency.

ianshmean:
If your direct deps don't fail, then maybe that indirect dep isn't actually used my your direct deps in your config? Being in a Project.toml doesn't mean the dep with definitely be used. 

Or if it's guarded behind Requires.jl you might want to add the erroring dep to your Project.toml

omus:
The private package is definitely being used as an indirect dep. I think what I want is an option to to `precompile` which will throw an exception when any failure occurs. That way Julia will produce a non-zero return code causing Docker to report the failure.

omus:
As a work around I'm validating that the `.ji` file exists for my package:
```julia
using Pkg
Pkg.API.precompile()

# Validate precompile succeeded
pkg_uuid = Base.UUID(Pkg.TOML.parsefile(joinpath(PKG_PATH, "Project.toml"))["uuid"])
pkg_id = Base.PkgId(pkg_uuid, PKG_NAME)
isfile(Base.compilecache_path(pkg_id)) || error("precompile failed")
```
Note: I can't use the module to create the `PkgId` as that involve loading the package which triggers another precompile

ianshmean:
How does this look? https://github.com/JuliaLang/Pkg.jl/pull/2362

PR #2362, opened 2021-01-25
Precompile: add `strict` kwarg to control whether indirect deps throw

ianshmean:
By default we prevent failures when precompiling indirect dependencies to throw, with the expectation that if they are used by direct dependencies the failure will be caught then and throw.

However, there seems to be some use cases where there is a need to force failures from indirect dependencies.
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2358

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -980,7 +980,7 @@ function make_pkgspec(man, uuid)
 end
 
 precompile(; kwargs...) = precompile(Context(); kwargs...)
-function precompile(ctx::Context; internal_call::Bool=false, throw_indirect_deps::Bool=false, kwargs...)
+function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false, kwargs...)
     Context!(ctx; kwargs...)
     instantiate(ctx; allow_autoprecomp=false, kwargs...)
     time_start = time_ns()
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -1278,7 +1278,7 @@ function precompile(ctx::Context; internal_call::Bool=false, kwargs...)
             err_str = ""
             n_direct_errs = 0
             for (dep, err) in failed_deps
-                if dep in direct_deps
+                if fail_indirect_deps || (dep in direct_deps)
                     err_str *= "\n" * "$dep" * "\n\n" * err * (n_direct_errs > 0 ? "\n" : "")
                     n_direct_errs += 1
```
omus, line 1283: Looks like the `pkgerror` later needs to be revised as it assumes direct dependencies

ianshmean:
@omus I just realized `throw_indirect_deps` would be a better name than `fail_indirect_deps`

omus:
> @omus I just realized `throw_indirect_deps` would be a better name than `fail_indirect_deps`

I'd be okay with a more general name indicating no failure are allowed

ianshmean:
But this is nice because it's specific to behavior that is explained in the docstring

ianshmean:
@KristofferC /  @fredrikekre can I get a 2nd opinion on this?

src/Pkg.jl
```diff
@@ -130,12 +130,13 @@ See also [`PackageSpec`](@ref).
 const add = API.add
 
 """
-    Pkg.precompile()
+    Pkg.precompile(; throw_indirect_deps::Bool=false)
```
fredrikekre, line 133: Perhaps a simpler keyword like
```suggestion
    Pkg.precompile(; strict::Bool=false)
```
would be nicer?

ianshmean: Ok, yeah. That's I think what @omus was suggesting too. I'll go with it

omus:
Thanks!

PR #2357, opened 2021-01-23
Debugging CI download failure

ianshmean:
There's a repeatable error during tarball download on ubuntu without the package server, and the error message is currently a little unhelpful. 
This adds the actual tarball urls being tried

PR #2364, opened 2021-01-26
only use git on Windows for default pkg server

StefanKarpinski:
Using git for registries on Windows was definitely a big improvement for the vast majority of Windows users, but for Windows users behind firewalls that can't access github.com, it's a problem. This changes the heuristic to using git when on Windows if the pkg server isn't explicitly set. If someone explicitly sets their pkg server then we get the registry from there instead of trying to git clone it. This can be overridden by explicitly git cloning the registry once, after which it will always be git updated. But a very small number of people should need to do that.

### DIFF
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -136,6 +136,9 @@ function populate_known_registries_with_urls!(registries::Vector{RegistrySpec})
     end
 end
 
+registry_use_pkg_server() =
+    !Sys.iswindows() || haskey(ENV, "JULIA_PKG_SERVER")
+
 function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=depots1())
     populate_known_registries_with_urls!(regs)
     registry_urls = nothing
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
What's up with the CI dumpster fire? I'm guessing some of it is https://github.com/JuliaLang/Pkg.jl/pull/2357. What's the rest?

fredrikekre:
Something returning `nothing`
```
# First argument is wrong
MethodError: no method matching download_verify_unpack(::Nothing, ::Nothing, ::String; ignore_existence=true)
Closest candidates are:
  download_verify_unpack(::AbstractString, ::Union{Nothing, AbstractString}, ::AbstractString; tarball_path, ignore_existence, force, verbose, quiet_download) at /Users/runner/work/Pkg.jl/Pkg.jl/src/PlatformEngines.jl:428
  ```

**Comment by StefanKarpinski** [comment]
I'm afraid I don't have time for screwing around with whatever's broken on master, so I'm going to make this PR directly to 1.6 where hopefully it will pass CI. Someone else can forward port it.

File: src/Types.jl [hunk 1]
```diff
@@ -997,8 +1008,7 @@ function clone_or_cp_registries(ctx::Context, regs::Vector{RegistrySpec}, depot:
         # clone to tmpdir first
         mktempdir() do tmp
             url, registry_urls = pkg_server_registry_url(reg.uuid, registry_urls)
-            # on Windows we prefer git cloning because untarring is so slow
-            if !Sys.iswindows() && url !== nothing
+            if registry_try_git(url)
```
fredrikekre, line 1011: I believe the CI errors you saw where from the logic here
```suggestion
            if registry_try_git(url) && url !== nothing
```

StefanKarpinski:
This seems good to me and the CI failures are the Ubuntu GitHub downloads issue.

DilumAluthge:
If I understand correctly, @vchuravy is working on another patch release for Julia 1.5. In that case, should we backport this PR to 1.5?

PR #2369, opened 2021-01-29
Allow custom `package` function in `archive_artifact`

omus:
Compressing large artifacts can be [quite time consuming](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/106#issuecomment-769173686) when using 7zip's non-threaded gzip support. In BinaryBuilderBase we're [experimenting with using `pigz`](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/107) as an alternative way of performing threaded gzip compression. Having support for a custom `package` function would avoid having to duplicate the functionality provided in `archive_artifact`.

Possibly `pigz` or another compression format should be used here in the future.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -127,16 +127,22 @@ function verify_artifact(hash::SHA1; honor_overrides::Bool=false)
 end
 
 """
-    archive_artifact(hash::SHA1, tarball_path::String; honor_overrides::Bool=false)
+    archive_artifact(hash::SHA1, tarball_path::String;
+                     honor_overrides::Bool=false,
+                     package::Function=package)
 
 Archive an artifact into a tarball stored at `tarball_path`, returns the SHA256 of the
 resultant tarball as a hexidecimal string. Throws an error if the artifact does not
-exist.  If the artifact is overridden, throws an error unless `honor_overrides` is set.
+exist.  If the artifact is overridden, throws an error unless `honor_overrides` is set.  A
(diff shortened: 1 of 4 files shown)


KristofferC:
> Having support for a custom package function would avoid having to duplicate the functionality provided in archive_artifact.

Personally, I almost think "duplication" is better here since otherwise you are taking a dependency on Pkg internals.

omus:
> Personally, I almost think "duplication" is better here since otherwise you are taking a dependency on Pkg internals.

Reasonable. I think a more ideal solution is to allow users to specify an archive format to `archive_artifact`. I believe BB will want to move to `7z` in the future.

giordano:
> otherwise you are taking a dependency on Pkg internals.

`Pkg.archive_artifact` is documented and exported, is it really "internal"?  Besides, BinaryBuilder already relies _a lot_ on Pkg internals :grimacing:

KristofferC:
Well, personally I think that way way way too much of `Artifacts` is exported and considered public. Why would the package manager expose an `archive_artifact`?? Oh well, if it is already public... :/

KristofferC:
Also, do you guys really want to wait for a whole Julia release to be able to do updates to this function?

omus:
Related: #2375

omus:
BinaryBuilderBase has gone the duplication route for now: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/113

PR #2370, opened 2021-01-29
Explicitly set User-Agent for downloads.

fredrikekre:
It is possible defaults should be reused for multiplexed connections and that this is a bug in Downloads/libcurl but it seems like a good idea to explicitly set it here anyway.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -247,6 +247,7 @@ function download(
     for header in get_metadata_headers(url)
         push!(headers, header)
     end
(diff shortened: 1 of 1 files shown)


KristofferC:
Ref https://github.com/JuliaLang/Pkg.jl/pull/2357#issuecomment-769700263.

StefanKarpinski:
If this is a problem for Pkg then it would be a problem for other people trying to download things as well, which suggests that it should be fixed in Downloads rather than here.

fredrikekre:
https://github.com/JuliaLang/Downloads.jl/pull/88

PR #2371, opened 2021-01-29
show the exception when download fails in artifact downloading

KristofferC:
also, verify the tree hash on windows which I think (correct me if I am wrong @staticfloat) should be ok now.

Might have a bit of overlap with https://github.com/JuliaLang/Pkg.jl/pull/2355.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -313,6 +313,11 @@ function download_artifact(
     calc_hash = try
         create_artifact() do dir
             download_verify_unpack(tarball_url, tarball_hash, dir, ignore_existence=true, verbose=verbose, quiet_download=quiet_download)
+            f = joinpath(dir, "lib", "libc_simple.dll")
+            if isfile(f)
+                @info("immediate post-extraction")
+                run(`icacls $(f)`)
+            end
         end
     catch e
         if isa(e, InterruptException)
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -303,70 +303,47 @@ function download_artifact(
         return true
     end
 
-    if Sys.iswindows()
-        # The destination directory we're hoping to fill:
-        dest_dir = artifact_path(tree_hash; honor_overrides=false)
-        mkpath(dest_dir)
-
-        # On Windows, we have some issues around stat() and chmod() that make properly
-        # determining the git tree hash problematic; for this reason, we use the "unsafe"
-        # artifact unpacking method, which does not properly verify unpacked git tree
-        # hash.  This will be fixed in a future Julia release which will properly interrogate
-        # the filesystem ACLs for executable permissions, which git tree hashes care about.
-        try
-            download_verify_unpack(tarball_url, tarball_hash, dest_dir, ignore_existence=true,
-                                   verbose=verbose, quiet_download=quiet_download)
-        catch e
-            # Clean that destination directory out if something went wrong
-            rm(dest_dir; force=true, recursive=true)
-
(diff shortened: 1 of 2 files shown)


staticfloat:
Yeah, this looks good to me.

staticfloat (approved):

KristofferC:
Hmm...

```
┌ Error: failed to download tarball from "https://pkg.julialang.org/artifact/69a468bd51751f4ed7eda31c240e775df06d6ee6"
│   exception = HTTP/1.1 404 Not Found while requesting https://pkg.julialang.org/artifact/69a468bd51751f4ed7eda31c240e775df06d6ee6
└ @ Pkg.Artifacts C:\projects\pkg-jl\src\Artifacts.jl:321
ERROR: LoadError: could not load library "C:\Users\appveyor\AppData\Local\Temp\1\jl_pWD6Uf\artifacts\444cecb70ff39e8961dd33e230e151775d959f37\bin\libc_simple.dll"
Access is denied. 
```

on Win CI

KristofferC:
Could it be something in `create_artifact`? I think that does something with the permissions.

StefanKarpinski:
The trouble with this is that sometimes we try the package server for things that we really ought to know it doesn't have and this will make that pretty noisy. Perhaps that's better, but to do this really correctly is more subtle.

KristofferC:
Hm, yeah that's true. Perhaps only show if it is not a 404?

StefanKarpinski:
That's a pretty good approximation of what we want, yeah. That will miss cases where the pkg server should have something but for some reason doesn't, but it's still better in 99% of cases.

staticfloat:
That's weird..... I tried to instrument the download part and it didn't print.  We aren't caching artifacts somewhere are we?

staticfloat:
I've been trying to debug this; the fundamental error is that the owner of the file (`appveyor`) isn't getting the permissions it's supposed to.  While trying to reproduce locally however, I ran into another error:

```
julia> mktempdir() do dir
       mkdir(joinpath(dir, "foo"))
       symlink("foo", joinpath(dir, "link"))
       stat(joinpath(dir, "link"))
       end
ERROR: IOError: stat("C:\\Users\\Administrator\\AppData\\Local\\Temp\\jl_7nHqNZ\\link"): permission denied (EACCES)
Stacktrace:
 [1] uv_error
   @ .\libuv.jl:97 [inlined]
 [2] stat(path::String)
   @ Base.Filesystem .\stat.jl:69
 [3] (::var"#5#6")(dir::String)
   @ Main .\REPL[3]:4
 [4] mktempdir(fn::var"#5#6", parent::String; prefix::String)
   @ Base.Filesystem .\file.jl:729
 [5] mktempdir(fn::Function, parent::String)
   @ Base.Filesystem .\file.jl:727
 [6] top-level scope
   @ REPL[3]:1
```

This is breaking one of the earlier tests in Artifacts.  Basically, relative symlinks to directories don't seem to be working on windows.  I have no idea why this isn't causing problems on Appveyor.  I'm cooked for today, but I think we're going to have to fix this sooner than later.

DilumAluthge:
I rebased this on master, so we can look at the Windows test results on both GHA and AppVeyor. cc: @ianshmean

IanButterworth:
The conflict with master is a little messy here for me to understand, perhaps @staticfloat or @KristofferC, would you mind resolving conflicts?

KristofferC:
I think this has been subsumed by https://github.com/JuliaLang/Pkg.jl/pull/2964

PR #2372, opened 2021-01-30
Enable windows on github actions CI

ianshmean:


DilumAluthge:
Can you also delete the `appveyor.yml` file in this PR? If we get GHA working on Windows, no need for AV.

ianshmean:
Yeah, absolutely, I'll do that if I get there. I wanted to show equivalence on the same commit before deleting it.

Also, I think appveyor still launches after the yml is deleted. We may need to disable it on the AV side when we come to it

DilumAluthge:
Yeah that's easy enough to do.

ianshmean:
Looks like we're now hitting the same issue as being explored in https://github.com/JuliaLang/Pkg.jl/pull/2371#issuecomment-770142417

https://github.com/JuliaLang/Pkg.jl/pull/2372/checks?check_run_id=1796099273#step:5:868

cc. @staticfloat

ianshmean:
@DilumAluthge given the work in #2371 perhaps we should merge this as-is, and retain appveyor so that #2371 can be rebased to get the new windows CI?

AV could then be disabled when windows works on both?

DilumAluthge:
SGTM.

Issue #2361, opened 2021-01-24
Pkg.precompile sometimes doesn't avoid precompilation during load time


IanButterworth:
There is an edge case in Pkg's precompilation for which we currently don't have a fix. I'm making this issue to make it easier to track given it has been discussed in a few places

One might see
```
(jl_dDFSry) pkg> precompile

julia> @time (using Plots; display(plot(1:0.1:10, sin.(1:0.1:10))))
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
```

This has improved since julia 1.5 (https://github.com/JuliaLang/Pkg.jl/issues/1578), but is still known to happen as a result of this kind of sequence:
```
pkg> add Plots, GR
... # add and fully precompile packages

julia> using Plots

pkg> pin GR@0.52.0 # roll back GR
... # precomp will ? for Plots given it's loaded
```

restart julia

```
pkg> precompile

julia> using Plots
[ Info: Precompiling Plots [91a5bcdd-55d7-5caf-9e0b-520d859cae80]
```

The problem is that `Base.stale_cachefile` can only check whether dependencies of a module have been updated if they are loaded in `Main`.
`Pkg.precompile` works around this, given the modules aren't actually loaded into `Main`, by keeping track itself of whether dependencies are updated.
But if `Pkg.precompile` isn't able to precompile a package due to it being loaded in `Main`, and julia is restarted, the new `precompile` attempt has no idea that dependencies have been recompiled.

In writing this up it occurs to me that the fix could be to add a `force_precomp` list of packages that's saved to the scratch space, to remember packages that still need precomp (i.e. packages with `?`), like the way `pkgs_precompile_suspended` works. 

Previous discussion around this:
- https://github.com/JuliaLang/Pkg.jl/pull/2018#issuecomment-691766768
- https://github.com/JuliaLang/Pkg.jl/issues/1578#issuecomment-694474501

PR #2373, opened 2021-01-30
Precompile: Remember packages that need precompilation to catch stale dependents

IanButterworth:
This adds an environment-specific memory between sessions for packages that need precompilation but couldn't be updated due to being loaded. i.e. those marked as `?`
As described in #2361, the cache invalidation heuristics in base only work for dependents if the dependencies that have changed are loaded, which isn't the case during the `Pkg.precompile` checks, so this memory is needed.

If they succeed or error, they're removed from this new memory, except for if a `?` error occurs again

Fixes https://github.com/JuliaLang/Pkg.jl/issues/2361

## Current behavior
```
pkg> add Plots, GR
... # add and fully precompile packages

julia> using Plots

pkg> pin GR@0.52.0 # roll back GR
... # precomp will ? for Plots given it's loaded
```
restart julia
```
pkg> precompile

julia> using Plots

```

## This PR
```
(@v1.7) pkg> add Plots, GR
...
111 dependencies successfully precompiled in 68 seconds

julia> using Plots

(@v1.7) pkg> pin GR@0.52.0 # roll back GR
...
Precompiling project...
  Progress [========================================>]  2/2
  ? Plots
1 dependency successfully precompiled in 8 seconds (60 already precompiled)
1 dependency failed but may be precompilable after restarting julia

```
restart julia
```
(@v1.7) pkg> precompile
  Progress [========================================>]  1/1
1 dependency successfully precompiled in 29 seconds (61 already precompiled)

julia> using Plots
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -994,8 +994,8 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
     io = ctx.io
     fancyprint = can_fancyprint(io)
 
-    # when manually called, unsuspend all packages that were suspended due to precomp errors
-    internal_call ? recall_suspended_packages() : precomp_unsuspend!()
+    recall_precompile_state() # recall suspended and force-queued packages
+    !internal_call && precomp_unsuspend!() # when manually called, unsuspend all packages that were suspended due to precomp errors
 
     direct_deps = [
         Base.PkgId(uuid, name)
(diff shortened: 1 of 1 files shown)


IanButterworth:
GHA Windows failures are #2378

IanButterworth:
@KristofferC LMK if I should just merge, I tested this locally and I'd be happy to monitor for issues on nightly.

IanButterworth:
I wanted to see if there were any issues with this in the wild before backporting it to 1.6, given it's more of a nice-to-have and 1.6.0 doesn't need holding up further.

I've not seen any issues on nightly, so I've added a 1.6 backport label. If other changes are going in before 1.6.0 is released, this _could_ go in, otherwise wait for 1.6.1

PR #2375, opened 2021-02-02
Use `xz` as package compression format

omus:
Threading is supported on p7zip with the `xz` format (unlink `gzip`) and the compression ratios are better than `gzip` and comparable to `7z`. It appears that Pkg has has support for handling `xz` since artifacts were introduced so this should be a backwards compatible change.

See these compression [benchmarks](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/issues/106#issuecomment-769173686) for runtime and compression size differences.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -384,7 +384,7 @@ Compress `src_dir` into a tarball located at `tarball_path`.
 """
 function package(src_dir::AbstractString, tarball_path::AbstractString)
     rm(tarball_path, force=true)
-    cmd = `$(exe7z()) a -si -tgzip -mx9 $tarball_path`
+    cmd = `$(exe7z()) a -si -txz -mx9 $tarball_path`
     open(pipeline(cmd, stdout=devnull), write=true) do io
         Tar.create(src_dir, io)
     end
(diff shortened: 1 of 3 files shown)


giordano:
On a further thought, I don't know if it's safe to _always_ use xz compression.  In Julia v1.6 we ship 7z for all platforms, but I believe this wasn't the case before?

omus:
Taking a look at Pkg 1.3.0 it does appears that [`xz` support is present](https://github.com/JuliaLang/Pkg.jl/blob/v1.3.0/src/PlatformEngines.jl#L318) (but who knows how well tested).

I agree handling this in BinaryBuilder is probably a better approach to start with.

giordano:
> Taking a look at Pkg 1.3.0 it does appears that xz support is present (but who knows how well tested).

But it relies on `xz` being present in the system.  We can't guarantee that, it wasn't a requirement, we can't change it retroactively.  In BinaryBuilder we can be smart and enable it for JLL that we know will require Julia v1.6

omus:
> But it relies on `xz` being present in the system. We can't guarantee that, it wasn't a requirement, we can't change it retroactively. In BinaryBuilder we can be smart and enable it for JLL that we know will require Julia v1.6

Very good point. Looking more closely at that release `tar` is being used for listing archive contents which would require `xz` to be present on the system. I believe 7zip always includes `xz` support which is why I didn't think there would be an issue.

I'll make some changes to https://github.com/JuliaLang/Pkg.jl/pull/2369 to make life easier and abandon this default for now. Maybe this should be revisited when 1.6 is the LTS.

PR #2377, opened 2021-02-06
If the `JULIA_PKG_SERVER` environment variable is not set, use Git for cloning registries

DilumAluthge:
See also: #2175, #2364 

Before this PR:
| OS | `JULIA_PKG_SERVER` environment variable | Registry is obtained via | 
| --- | --- | --- | 
| Windows | not set | Git |
| Windows | set to `""` | Git |
| Windows | set to `"pkg.julialang.org"` (or any other valid Pkg server value) | Pkg server |
| Not Windows | not set | **Pkg server** |
| Not Windows | set to `""` | Git |
| Not Windows | set to `"pkg.julialang.org"` (or any other valid Pkg server value) | Pkg server |

After this PR:
| OS | `JULIA_PKG_SERVER` environment variable | Registry is obtained via | 
| --- | --- | --- | 
| Windows | not set | Git |
| Windows | set to `""` | Git |
| Windows | set to `"pkg.julialang.org"` (or any other valid Pkg server value) | Pkg server |
| Not Windows | not set | **Git** |
| Not Windows | set to `""` | Git |
| Not Windows | set to `"pkg.julialang.org"` (or any other valid Pkg server value) | Pkg server |

### DIFF
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -136,7 +136,13 @@ function populate_known_registries_with_urls!(registries::Vector{RegistrySpec})
     end
 end
 
-registry_use_pkg_server() = haskey(ENV, "JULIA_PKG_SERVER")
+function registry_use_pkg_server(url)
+    if url === nothing
+        return false
+    else
+        return haskey(ENV, "JULIA_PKG_SERVER")
+    end
+end
 
 function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=depots1())
     populate_known_registries_with_urls!(regs)
@@ -148,7 +154,7 @@ function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=d
         # clone to tmpdir first
         mktempdir() do tmp
             url, registry_urls = pkg_server_registry_url(reg.uuid, registry_urls)
-            if url !== nothing && registry_use_pkg_server()
+            if registry_use_pkg_server(url)
                 # download from Pkg server
                 try
                     download_verify_unpack(url, nothing, tmp, ignore_existence = true)
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -136,8 +136,7 @@ function populate_known_registries_with_urls!(registries::Vector{RegistrySpec})
     end
 end
 
-registry_use_pkg_server() =
-    !Sys.iswindows() || haskey(ENV, "JULIA_PKG_SERVER")
+registry_use_pkg_server() = haskey(ENV, "JULIA_PKG_SERVER")
 
 function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=depots1())
     populate_known_registries_with_urls!(regs)
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -144,6 +144,20 @@ function registry_use_pkg_server(url)
     end
 end
 
+function check_registry_state(reg, url)
+    reg_currently_uses_pkg_server = reg.tree_info !== nothing
+    reg_should_use_pkg_server = registry_use_pkg_server(url)
+    if reg_currently_uses_pkg_server && !reg_should_use_pkg_server
+        msg = string(
+            "Your registry may be outdated. We recommend that you run the ",
+            "following command: ",
+            "using Pkg; Pkg.Registry.rm(\"$(reg.name)\"); Pkg.Registry.add(\"$(reg.name)\")",
+        )
+        @warn(msg)
+    end
+    return nothing
+end
+
 function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=depots1())
     populate_known_registries_with_urls!(regs)
     registry_urls = nothing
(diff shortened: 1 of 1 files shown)


DilumAluthge:
If https://github.com/JuliaLang/Pkg.jl/pull/2175 and https://github.com/JuliaLang/Pkg.jl/pull/2364 are backported to 1.5, then this PR should also be backported to 1.5.

ExpandingMan:
In my opinion unless the sync issue is fully resolved in time for the 1.6 RC, this should be merged.  I suspect this will wreak far more chaos for people less plugged into the community than it has for us.

DilumAluthge:
My apologies for the long list of requested reviewers. I do think that this change needs to have a lot of eyes on it.

Feel free to ping anyone else that should review this.

DilumAluthge:
The problem with this PR is that it doesn't fix the underlying issue (Pkg servers being out-of-date). It's just a band-aid.

The best solution is to make the Pkg servers (and the storage servers) more reliable so that the delay between a PR being merged to the General registry and the new version being available from the Pkg servers is very short (ideally no longer than 10-15 minutes).

KristofferC:
Without diff support for the registry, I do agree that downloading the registry from PkgServer doesn't make much sense. As an example, I think e.g. Cargo uses git for their registry even though they have the equivalent of our PkgServers.

KristofferC (approved):

KristofferC:
The problem with this, as it is right now, is that everyone who has a Pkg Server registry will not get any updates now until they remove that one.

DilumAluthge:
> The problem with this, as it is right now, is that everyone who has a Pkg Server registry will not get any updates now until they remove that one.

Yeah, that's a big problem as well.

I guess in the Discourse announcement for when Julia 1.6 is released, we could add a section like this:

Once you have installed Julia 1.6, we recommend that you run this command one time:
```julia
julia> using Pkg; Pkg.Registry.rm("General"); Pkg.Registry.add("General")
```

You only need to run this command once.

IanButterworth:
Can a warning be added if the registry is in a non-updatable state, and give that advice?

DilumAluthge:
> Can a warning be added if the registry is in a non-updatable state, and give that advice?

This is tricky. How do we know if the registry is in a non-updateable state?

IanButterworth:
Perhaps something like this, inserted here https://github.com/JuliaLang/Pkg.jl/blob/8e802b8dd34766904a85aca10b4482b6de8d2e8c/src/Registry/Registry.jl#L291-L292

```
            if reg.tree_info !== nothing
                if registry_use_pkg_server() == false
                    pkgerror("""The local registry at $regpath was downloaded via a pkg server,
                    but `JULIA_PKG_SERVER` must now be explicitly set to specify the server.
                    To fix the local registry, either:
                    - set the pkg server with `ENV["JULIA_PKG_SERVER"]="pkg.julialang.org"` etc.
                    - remove and re-add the existing registry to revert to a git-cloned registry
                      `pkg> registry rm General` `pkg> registry add General`""")
                end
                printpkgstyle(io, :Updating, "registry at " * regpath)
```

DilumAluthge:
Alright, the third commit on this PR adds a warning if the registry is currently being obtained from the Pkg server but it should actually be obtained from Git.

The warning will not be shown if the user has set the `JULIA_PKG_SERVER` environment variable.

Here's what the warning looks like:

```julia
julia> using Pkg

julia> Pkg.update()
┌ Warning: Your registry may be outdated. We recommend that you run the following command: using Pkg; Pkg.Registry.rm("General"); Pkg.Registry.add("General")
└ @ Pkg.Registry ~/Downloads/Pkg.jl/src/Registry/Registry.jl:156
  No Changes to `~/Downloads/Pkg.jl/Project.toml`
  No Changes to `~/Downloads/Pkg.jl/Manifest.toml`

julia> using Pkg; Pkg.Registry.rm("General"); Pkg.Registry.add("General")
    Removing registry `General` from ~/.julia/registries/General

julia> Pkg.update()
  No Changes to `~/Downloads/Pkg.jl/Project.toml`
  No Changes to `~/Downloads/Pkg.jl/Manifest.toml`
```

DilumAluthge:
Are we going to do a second RC? If so, this PR can make it into 1.6.0. If not, it'll have to be 1.6.1.

DilumAluthge:
@StefanKarpinski @staticfloat @fredrikekre Any thoughts on this?

staticfloat (approved):

DilumAluthge:
I'll merge this tomorrow if there are no objections.

IanButterworth (approved):

fredrikekre:
Whats up with all the test failures?

IanButterworth:
It's #2378. Relative symlinks are broken on Windows but we only picked it up since adding GitHub Actions windows. 

Windows passes on Appveyor though, perhaps because it runs with elevated permissions and might fall back to a different link method.

fredrikekre:
Perhaps those tests could be removed then for now?

IanButterworth:
The upstream fix has been merged. I was waiting for nightly to re-run the CI here

The run here should be on nightly, and succeed: https://github.com/JuliaLang/Pkg.jl/pull/2373

KristofferC:
FWIW, In my opinion, we shouldn't put this in until 1.6.1. We really need to cut anything non-crucial for 1.6.0 for it to get out as soon as possible.

giordano:
> Perhaps those tests could be removed then for now?

For what is worth, I'm ok with disabling this test for Windows
https://github.com/JuliaLang/Pkg.jl/blob/2a0a18d6e05a16dd959817554ce8aafd4b053a3f/test/artifacts.jl#L140
The most important thing is the test below, that the directory can be cleaned up without extra work.

IanButterworth:
Just to explain everywhere this is being discussed.. that test did actually show up a bug, which is now fixed on master https://github.com/JuliaLang/julia/commit/78c448f4d6be2599fa833c309ece2df9acd03033

We should have passing tests in ~30 mins

DilumAluthge:
> FWIW, In my opinion, we shouldn't put this in until 1.6.1. We really need to cut anything non-crucial for 1.6.0 for it to get out as soon as possible.

This sounds like a good plan to me!

PR #2382, opened 2021-02-08
Fix p7zip usage on alpine

staticfloat:
The `p7zip_jll` wrapper defines all sorts of useful environment
variables such as `LD_LIBRARY_PATH` to pick up the `libstdc++` that
Julia bundles with it.  We need to use that if the `7z` we're invoking
comes from that JLL, which it does if we pick it up from the `libexec`
directory of Julia.  So we instead use `p7zip_jll` if it is available on
this platform, and if it's not, we'll pick something up off the `PATH`.


staticfloat:
Fixes https://github.com/JuliaLang/julia/issues/39154

PR #2388, opened 2021-02-10
Disable one of the artifact tests on Windows on GitHub Actions

DilumAluthge:
See this comment: https://github.com/JuliaLang/Pkg.jl/pull/2377#issuecomment-776324614

cc: @giordano


IanButterworth:
This isn't needed. The fix is merged on master. I'm just waiting for nightly to hit that commit

PR #2391, opened 2021-02-10
Make downloads respect IO

### LINKED ISSUE #2328: Artifact downloading does not respect io

IanButterworth:
I'm separating this out of #2355 given in that PR I also tried to fix the way download logging is done, which seems to need further thought.

Makes downloads respect the `io` kwarg
```
julia> Pkg.add("Images", io=devnull)

julia> 
```
Fixes #2328

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1433,7 +1433,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     if ctx.env.pkg !== nothing
         push!(art_pkgs, ctx.env.pkg)
     end
-    Operations.download_artifacts(ctx.env, art_pkgs; platform, verbose)
+    Operations.download_artifacts(ctx.env, art_pkgs; platform, verbose, io=ctx.io)
     # Run build scripts
     Operations.build_versions(ctx, union(UUID[pkg.uuid for pkg in new_apply], new_git); verbose)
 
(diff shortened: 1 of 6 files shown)


DilumAluthge:
Nice!

Is this the last part of Pkg that needs to be fixed to use `io`?

IanButterworth:
> Is this the last part of Pkg that needs to be fixed to use `io`?

I just tried in a fresh depot, and the answer seems to be no
```
julia> Pkg.add("Images", io=devnull)
  Installing known registries into `/tmp/jdepot`
       Added registry `General` to `/tmp/jdepot/registries/General`
```
Plus there were some ephemeral progress bars showing up

IanButterworth:
Let's keep this to fixing #2328

Can someone approve or merge. Thanks

fredrikekre:
`stdout` -> `stderr`?

IanButterworth:
Perhaps the correct thing to do would be to make all of the `io` kwargs default to `DEFAULT_IO[]`, which defaults to `stderr` as the registry functions do?

https://github.com/JuliaLang/Pkg.jl/blob/47105aaefb7eb8bda5c002315ccf1a28be8fdabb/src/Registry/Registry.jl#L42

If people are monitoring `stdout` specifically for Pkg logging, that change would break things, but given `io` can be provided, it would arguably be the wrong way to do it anyway, so ok to break

fredrikekre:
Yea sure, as long as it goes to stderr it is fine. (Already used here for downloads btw: https://github.com/JuliaLang/Pkg.jl/blob/f00fa969d96a50e96a5cd620b5b3dda7fd8ef12b/src/Artifacts.jl#L439)

src/API.jl
```diff
@@ -1443,7 +1444,7 @@ end
 @deprecate status(mode::PackageMode) status(mode=mode)
 
 function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=PKGMODE_PROJECT,
-                io::IO=stdout, kwargs...)
```
fredrikekre, line 1446: This should be `stdout` (this is output the user has asked for).

IanButterworth: or should it use `ctx.io` and remove the `io` kwarg?

IanButterworth: I get your point now.. I'll revert

IanButterworth:
I think i've coalesced the Registry logging and the download progress meters that I was seeing to use `io` now.

At least in a fresh depot I get no output now, with this
```
julia> Pkg.add("VideoIO", io=devnull)

julia> Pkg.add("FFTW", io=devnull)

julia> 
```

PR #2399, opened 2021-02-19
Suggest functional API in warning, consistency in docs

goretkin:
I am guessing the warning is suggesting script writers to anticipate that their scripts might break if the REPL mode asks a question (cf. `apt-get -y`). I thought it would be clearer to indicate what to do that would reduce the need for caution.

I also tried to improve some consistency in terminology in the docs. I was ambivalent about choosing between "REPL mode"  or "REPL interface".

### DIFF
--- a/docs/generate.jl
+++ b/docs/generate.jl
@@ -34,7 +34,7 @@ function generate()
         This section describes available commands in the Pkg REPL.
         The REPL mode is mostly meant for interactive use,
         and for non-interactive use it is recommended to use the
-        "API mode", see [API Reference](@ref API-Reference).
+        "functional API", see [API Reference](@ref API-Reference).
         """)
     # list commands
     println(io, "## `package` commands")
--- a/docs/generate.jl
+++ b/docs/generate.jl
@@ -32,7 +32,7 @@ function generate()
         # [**11.** REPL Mode Reference](@id REPL-Mode-Reference)
 
         This section describes available commands in the Pkg REPL.
-        The REPL mode is mostly meant for interactive use,
+        The Pkg REPL mode is mostly meant for interactive use,
         and for non-interactive use it is recommended to use the
         "functional API", see [API Reference](@ref API-Reference).
         """)
--- a/docs/generate.jl
+++ b/docs/generate.jl
@@ -34,7 +34,7 @@ function generate()
         This section describes available commands in the Pkg REPL.
         The Pkg REPL mode is mostly meant for interactive use,
         and for non-interactive use it is recommended to use the
-        "functional API", see [API Reference](@ref API-Reference).
+        functional API, see [API Reference](@ref API-Reference).
         """)
     # list commands
     println(io, "## `package` commands")
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -1,7 +1,7 @@
 # [**12.** API Reference](@id API-Reference)
 
-This section describes the functional interface,
-for interacting with Pkg.jl. The "functional API" is recommended
+This section describes the functional API for interacting with Pkg.jl.
+It is recommended to use the functional API, rather than the Pkg REPL mode,
 for non-interactive usage, for example in scripts.
 
 ## General API Reference
--- a/docs/src/api.md
+++ b/docs/src/api.md
@@ -16,7 +16,7 @@ For example, `Pkg.add("Example"; io=devnull)` will discard any output produced b
 
 ## Package API Reference
 
-In the REPL mode, packages (with associated version, UUID, URL etc) are parsed from strings,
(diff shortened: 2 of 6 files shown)


docs/generate.jl
```diff
@@ -32,9 +32,9 @@ function generate()
         # [**11.** REPL Mode Reference](@id REPL-Mode-Reference)
 
         This section describes available commands in the Pkg REPL.
-        The REPL mode is mostly meant for interactive use,
+        The Pkg REPL mode is mostly meant for interactive use,
         and for non-interactive use it is recommended to use the
-        "API mode", see [API Reference](@ref API-Reference).
+        "functional API", see [API Reference](@ref API-Reference).
```
fredrikekre, line 37:
```suggestion
        functional API, see [API Reference](@ref API-Reference).
```

docs/src/api.md
```diff
@@ -1,7 +1,7 @@
 # [**12.** API Reference](@id API-Reference)
 
-This section describes the function interface, or "API mode",
-for interacting with Pkg.jl. The function API is recommended
+This section describes the functional interface,
+for interacting with Pkg.jl. The "functional API" is recommended
```
fredrikekre, line 4:
```suggestion
This section describes the functional API interface for interacting with Pkg.jl.
It is recommended to use the functional API, rather than the Pkg REPL mode,
```

docs/src/index.md
```diff
@@ -4,8 +4,8 @@ Welcome to the documentation for Pkg, Julia's package manager.
 The documentation covers many things, for example managing package
 installations, developing packages, working with package registries and more.
 
-Throughout the manual the REPL interface to Pkg is used in the examples.
-There is also a functional API interface, which is preferred when not working
+Throughout the manual the REPL interface to Pkg ("Pkg REPL mode") is used in the examples.
```
fredrikekre, line 7:
```suggestion
Throughout the manual the REPL interface to Pkg, the Pkg REPL mode, is used in the examples.
```

src/REPLMode/REPLMode.jl
```diff
@@ -374,7 +374,7 @@ end
 #############
 function do_cmd(repl::REPL.AbstractREPL, input::String; do_rethrow=false)
     if !isinteractive() && !TEST_MODE[] && !PRINTED_REPL_WARNING[]
-        @warn "The Pkg REPL interface is intended for interactive use, use with caution from scripts."
+        @warn "The Pkg REPL mode is intended for interactive use, use with caution from scripts, or use the functional API."
```
fredrikekre, line 377: above you use "interface" but change it here

docs/src/api.md
goretkin: Do you want to keep "API interface"? It's not uncommon to have those redundancies (e.g. LCD display), but I tried to avoid it.

src/REPLMode/REPLMode.jl
goretkin: Do you prefer mode or interface?

docs/src/api.md
fredrikekre: updated

src/REPLMode/REPLMode.jl
fredrikekre: mode I guess

goretkin: The only place I see where `REPL interface` remains is in the table where it was `REPL version` previously. Is that what you mean? I'll change it to `REPL mode` in the table if so.

goretkin:
@fredrikekre how does this look?

src/REPLMode/REPLMode.jl
fredrikekre, line 377: This sentence is a bit awkward perhaps, maybe something like
```suggestion
        @warn "The Pkg REPL mode is intended for interactive use only and should be used with caution from scripts. It is recommended to use the functional API instead."
```

goretkin: I agree that's better. Thanks.

I kind of guessed at it before, but can you clarify exactly what is to be cautious about? That a script might break because the REPL mode might prompt a "user" for input? (e.g. what the flag is for in `apt-get -y`)

KristofferC: The REPL interface might change in backward-incompatible ways.

goretkin: So can it just prescribe: "The Pkg REPL mode is intended for interactive use only, and should not be used from scripts. It is recommended to use the functional API instead."

KristofferC: Sounds good to me.

goretkin: Done

PR #2414, opened 2021-03-01
only use the stdlib version cache if a custom version is given to the resolver

KristofferC:
some workflows require the ability to inject custom stdlibs when
building Julia, the stdlib cache breaks that. The stdlib cache is now
only used when one wants to resolve something for a custom Julia version
(for example BinaryBuilder).


staticfloat (approved):
Perfect.

DilumAluthge:
Good to merge?

Issue #2407, opened 2021-02-28
Bug: `generate_historical_stdlibs.jl` does not work on macOS


DilumAluthge:
Specifically, this line fails on macOS:
https://github.com/JuliaLang/Pkg.jl/blob/8cbd11723f27ce2feef229aaf944924ab355d864/ext/HistoricaStdlibGenerator/generate_historical_stdlibs.jl#L71-L71

This happens because on macOS, the script downloads `.dmg` files instead of `.tar.gz` files. Then, on line 71, it tries to `untar` a `.dmg` file, which of course fails.

The workaround is to use a Linux machine instead. However, it is mildly annoying to have to SSH into a Linux machine just to be able to run the `generate_historical_stdlibs.jl` script.

PR #2417, opened 2021-03-01
[ext/HSG]: Enable generating historical stdlibs on macOS

staticfloat:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2407

Also, update the historical stdlib list.


DilumAluthge (approved):

DilumAluthge:
If I understand correctly, this will fix issue https://github.com/JuliaLang/Pkg.jl/issues/2407, so I have added the text `Fixes #2407` to the PR description.

Issue #2410, opened 2021-03-01
Bug: The "STDLIBS_BY_VERSION up-to-date" test is currently broken


DilumAluthge:
See also:
- https://github.com/JuliaLang/Pkg.jl/pull/2409

IanButterworth:
I fixed this in https://github.com/JuliaLang/Pkg.jl/pull/2402 but can remove it if it needs its own PR

DilumAluthge:
> I fixed this in #2402 but can remove it if it needs its own PR

Yeah, if you already have a fix for this, can you pull it out and make a separate PR with it?

DilumAluthge:
Also, when you make that PR, can you also change the `@test_broken` back to `@test`?

(I changed the `@test` to `@test_broken` in https://github.com/JuliaLang/Pkg.jl/pull/2409.)

DilumAluthge:
Actually, never mind. #2402 LGTM, so let's just merge that.

DilumAluthge:
I don't think #2402 was sufficient to fix this.

DilumAluthge:
The list of stdlibs needs to have libblastrampoline_jll in it.

But when I add it to the list, everything breaks. See e.g. https://github.com/JuliaLang/Pkg.jl/pull/2411

PR #2418, opened 2021-03-01
[ext/HSG]: Store next release _and_ latest nightly

staticfloat:
In the event that we are in an RC phase where we have three "latest"
branch tips (e.g. latest stable: 1.5.3, latest RC: 1.6.0-rc1 and latest
`master`: 1.7.0-DEV) each with their own set of standard libraries, we
must store this information in Pkg so that we pass tests when running on
latest `master`. It is harmless for this information to be baked into an
older Julia, as we make no guarantees that Julia 1.6.0 final should have
perfect information about what standard libraries 1.7.0 will have.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/2410


staticfloat:
This should be merged after #2417

DilumAluthge:
@staticfloat Can you resolve the merge conflicts?

Alternatively, since I just merged #2417 to master, it might be easier for you to just redo this commit from scratch starting with the latest master.

staticfloat:
Yeah, it's because it bundled the commit you just merged.  Silly `git`, you should know that those two were the same commit!

DilumAluthge:
> Yeah, it's because it bundled the commit you just merged. Silly `git`, you should know that those two were the same commit!

Classic Git.

DilumAluthge:
> > Yeah, it's because it bundled the commit you just merged. Silly `git`, you should know that those two were the same commit!
> 
> Classic Git.

I guess this is the kind of thing that Pijul is supposed to solve.

PR #2420, opened 2021-03-04
If a CI job fails (or times out) on master, post a message in the #stdlib-ci-failures Slack channel

DilumAluthge:


KristofferC:
What's the motivation for this? Do we have any intermittent CI problems in Pkg we don't know the cause of?

DilumAluthge:
Not right now. But we've had test failures on Pkg master in the past. Looking at the past six months of commits on Pkg master, it's really easy to find commits in which at least one CI job failed or timed out.

I figure that setting this up now is a good way to make sure we get notified if/when Pkg tests start failing again in the future.

And if that never happens, then we won't get any Slack notifications, so there's no noise unless the tests start failing.

KristofferC:
> And if that never happens, then we won't get any Slack notifications, so there's no noise unless the tests start failing.

They fail intermittently due to download errors.

fredrikekre:
Doesn't GitHub let you subscribe to things like this if you really want immediate feedback?

DilumAluthge:
> Doesn't GitHub let you subscribe to things like this if you really want immediate feedback?

I don't think GitHub has a way of subscribing to CI failures only on master (i.e. receiving notifications if CI fails on master, but not receiving notifications if CI fails on a PR, and not receiving notifications if CI passes on master).

DilumAluthge:
Based on the feedback here, it sounds like this is probably not something that a lot of Pkg.jl developers would find useful?

I'll leave this open for a few more days if anyone else has thoughts. But I'm inclined to close it.

PR #2439, opened 2021-03-19
Add the `force_latest_compatible_version` keyword argument to `Pkg.test`

### LINKED ISSUE #2445: Adding a test that is pinned to a specific commit of the General registry?
I want to add a test to the Pkg test suite that involves instantiating a project that has a direct dependency (Example.jl) that does not have a `[compat]` entry.

This test is subject to breakage in the future. Therefore, I would like to run this test using a specific commit of the General registry. Since commits are immutable, this ensures that the test will not break in the future.

What is the best way to use a specific commit of the General registry in a `Utils.isolate()` test?

cc: @IanButterworth @KristofferC

DilumAluthge:
Closes #2176 
Fixes #2445 

## Summary

This pull request adds the `force_latest_compatible_version::Bool` keyword argument to the `Pkg.test` function.

If `force_latest_compatible_version` is `true`, then an error will be thrown if any of the direct dependencies is not at the latest compatible version.

The default value is `force_latest_compatible_version = false`.

This is an experimental feature. The behavior is subject to change or removal in minor or patch releases of Julia.

## How It Works

Here is a summary of how this feature works.

1. `Pkg.test` sets up the temporary sandbox project and runs the resolver.
2. If `force_latest_compatible_version` is `false`, we do nothing. If `force_latest_compatible_version` is `true`, we assemble the list of all direct dependencies, and then loop over the list; for each direct dependency `DepName`, we take the following steps:
    1. Inside the temporary sandbox project, we check and see which version of `DepName` the resolver selected.
    2. We assemble the list of all registered versions of `DepName`.
    3. We iterate over each version from step (ii), and for each version, we determine whether or not that version is inside the `[compat]` entry for `DepName`. At the end of this step, we have a list of all registered versions of `DepName` that are compatible with the `[compat]` entry for `DepName`.
    4. We take the `maximum` of the list from step (iii). This gives us the latest registered version of `DepName` that is compatible with the `[compat]` entry for `DepName`.
    5. **We compute the earliest version number `V` such that all changes from `V` to the result of step (iv) are backwards-compatible.**
    6. If the answer from step (i) is strictly less than the answer from step (v), we throw an error. Else, we do nothing.
3. `Pkg.test` runs the `tests/runtests.jl` file.

The algorithm for computing step (v) above is as follows. The input is a version number of the form `x.y.z`.
- If `x > 0`, return `x.0.0`. (This is because, for `x > 0`, all changes from `x.0.0` to `x.y.z` are backwards-compatible.)
- Else, if `y > 0`, return `0.y.0`. (This is because, for `y > 0`, all changes from `0.y.0` to `0.y.z` are backwards-compatible.)
- Else, return `0.0.z`.

## Motivation

Suppose that you have a `[compat]` entry that looks like this:
```toml
[compat]
Foo = "0.1, 0.2, 0.3"
```

And suppose that a bot (e.g. CompatHelper, Dependabot, etc.) opens a pull request to edit the above `[compat]` entry so that it looks like this instead:
```toml
[compat]
Foo = "0.1, 0.2, 0.3, 0.4"
```

When CI runs your package tests on this pull request, you want to make sure that the tests use Foo version `0.4`. (If the tests do not use Foo version `0.4`, then even if the tests pass, you do not know whether or not your package works with Foo version `0.4`.)

However, currently, there is no way to automatically check that your package tests used Foo version `0.4`.

With this pull request, you simply set `force_latest_compatible_version = true`. Then, if your package tests do not use Foo version `0.4`, Pkg will throw an error, and therefore your CI job will fail. This will alert you to the fact that you should probably not yet merge the CompatHelper/Dependabot PR.

## Future Plans

Note the default value is `force_latest_compatible_version = false`.

However, for GitHub Actions CI, we will add a feature to [julia-actions/julia-runtest](https://github.com/julia-actions/julia-runtest) that works as follows:
1. Detect whether or not the current CI job is a pull request job where the pull request was opened by CompatHelper or Dependabot.
2. If the answer to step 1 is "yes", then set `force_latest_compatible_version = true`.
3. If the answer to step 1 is "no", then set `force_latest_compatible_version = false`.
 
## Related Discussions

1. #2176
2. #2234 
2. https://github.com/JuliaRegistries/CompatHelper.jl/issues/160
3. https://github.com/julia-actions/julia-runtest/pull/20


src/force_latest_compatible_version.jl
```diff
@@ -0,0 +1,75 @@
+function _force_latest_compatible_version(ctx::Types.Context; target_name = nothing)
```
KristofferC, line 1: The `_force_latest_compatible_version` seems pretty arbitrarily cut up into 3 functions. Can't you just merge them?

KristofferC: Also, it seems excessive for a whole new file for this. Also, we haven't really used `_` prefix anywhere else in Pkg.

src/force_latest_compatible_version.jl
```diff
@@ -0,0 +1,75 @@
...
+end
+
+function _force_latest_compatible_version(ctx::Types.Context, uuid_list::AbstractVector{Base.UUID};
+                                          target_name = nothing)
+    isempty(uuid_list) && return true
+    results = _force_latest_compatible_version.(Ref(ctx), uuid_list; target_name)
+    return all(results)
+end
+
+function _warn_no_compat(has_compat::Bool,
```
KristofferC, line 15: Just inline this where it is used. The number of lines just to list the arguments is longer than the function body.

src/force_latest_compatible_version.jl
```diff
@@ -0,0 +1,75 @@
+function _force_latest_compatible_version(ctx::Types.Context; target_name = nothing)
+    direct_deps = load_direct_deps(ctx.env)
+    direct_deps_uuids = getproperty.(direct_deps, Ref(:uuid))
+    nonstdlib_direct_deps_uuids = filter(x -> x ∉ keys(stdlibs()), direct_deps_uuids)
```
KristofferC, line 4: Use `is_stdlib` to check if something is an `stdlib` instead.

src/force_latest_compatible_version.jl
```diff
@@ -0,0 +1,75 @@
+function _force_latest_compatible_version(ctx::Types.Context; target_name = nothing)
+    direct_deps = load_direct_deps(ctx.env)
+    direct_deps_uuids = getproperty.(direct_deps, Ref(:uuid))
```
KristofferC, line 3:
```suggestion
    direct_deps_uuids = [dep.uuid for dep in direct_deps]
```

src/force_latest_compatible_version.jl
```diff
@@ -0,0 +1,75 @@
...
+function _get_all_registered_versions(ctx::Types.Context, uuid::Base.UUID)
+    versions = Base.VersionNumber[]
+    for reg in ctx.registries
+        pkg = get(reg, uuid, nothing)
+        pkg == nothing && continue
+        info = Registry.registry_info(pkg)
+        append!(versions, keys(info.version_info))
+    end
+    unique!(versions)
+    sort!(versions)
```
KristofferC, line 73: Is there a need to `sort` here. You could use a `Set{VersionNumber}` if you want to only gather the unique versions.

src/force_latest_compatible_version.jl
```diff
@@ -0,0 +1,75 @@
...
+    end
+    return nothing
+end
+
+function _force_latest_compatible_version(ctx::Types.Context, uuid::Base.UUID;
+                                          target_name= nothing)
+    manifest_deps = load_manifest_deps(ctx.env.manifest)
+    manifest_deps_uuids = getproperty.(manifest_deps, Ref(:uuid))
+    idx = only(findall(manifest_deps_uuids .== uuid))
+    dep = manifest_deps[idx]
```
KristofferC, line 36: Can't you just do `ctx.env.manifest[uuid]` and use the `PackageEntry` it returns?

DilumAluthge: Done.

src/force_latest_compatible_version.jl
DilumAluthge: Done.

src/force_latest_compatible_version.jl
DilumAluthge: Done.

src/force_latest_compatible_version.jl
DilumAluthge: Done!

src/force_latest_compatible_version.jl
DilumAluthge: Done!

src/force_latest_compatible_version.jl
DilumAluthge: 1. So, in the future, I would like to maybe add the ability to only force the latest compatible version of a strict subset of dependencies. So e.g. the user would pass a list of package names as `Pkg.test(; force_latest_compatible_version = ["Foo", "Bar", "Baz"]`, and then we would only enforce this check for those three packages. That's why I split up `_force_latest_compatible_version` into three functions, so that in the future, there will be less refactoring that I need to do in order to add this feature.
2. I'll move this code into the `src/Operations.jl` source code file.
3. The reason that the `_force_latest_compatible_version` function starts with `_` is that the keyword argument to `Pkg.test` is called `force_latest_compatible_version`. So I had to add the `_` to the function name, otherwise we have a name conflict.

00vareladavid:
I am not super clear on the `CompatHelper` situation but would it not be sufficient to simply force the latest version for the single package you wish to check e.g. `pin Foo@4.0` before you run the tests?

DilumAluthge:
> I am not super clear on the `CompatHelper` situation but would it not be sufficient to simply force the latest version for the single package you wish to check e.g. `pin Foo@4.0` before you run the tests?

So, one of the enhancements that I would like to make in the future is to be able to slightly loosen the requirement from "latest compatible version of your dependency" to "latest compatible breaking family of versions of your dependency".

So, for example, suppose that your package has a `[compat]` section that currently looks like this:
```toml
Foo = "0.1"
```

And a bot (CompatHelper, Dependabot, etc.) makes a PR to change this to:
```toml
Foo = "0.1, 0.2"
```

Suppose that the list of versions of the `Foo.jl` package are:
- `0.1.0`
- `0.2.0`
- `0.2.1`
- `0.3.0`

When CI runs your package's test suite on this CompatHelper/Dependabot pull request, we want to make sure that your tests use version 0.2.x of Foo.jl. But, to be honest, we don't actually care if it uses Foo 0.2.0 instead of 0.2.1. According to (Pkg's implementation of) SemVer, if your package is compatible with Foo 0.2.0, then it is also compatible with Foo 0.2.1, because any changes from Foo 0.2.0 -> 0.2.1 must be (by definition) backwards-compatible.

So what we really want is:
1. Make a list of all registered versions of Foo.jl. This list is:
    - `0.1.0`
    - `0.2.0`
    - `0.2.1`
    - `0.3.0`
2. Filter the list to only include versions that are compatible with our new `[compat]` entry for Foo. The list now becomes:
    - `0.1.0`
    - `0.2.0`
    - `0.2.1`
3. Take the maximum of that list. The result of this is `0.2.1`.
4. Compute the earliest version number `V` such that all changes made from `V` to the result of step 3 are backwards compatible by SemVer. For this example, the result is `0.2.0`.
5. When the test suite runs, make sure that the version of Foo used by the tests is greater than or equal to the result from step 4. In this example, we make sure that the version of Foo used in the tests is `≥ 0.2.0`.

I don't think that we can do this with `pin`, can we? I.e. we want to "pin" Foo to any version of the form `0.2.x`.

I haven't implemented this feature yet in this PR, because I want to get it reviewed and merged quickly. But I'll implement it in the future PR. That is really the workflow that we need to support in CompatHelper/Dependabot PRs.

---

The rule for computing step 4 is:
- If the input is `0.0.z`, return `0.0.z`.
- Else, if the input if `0.y.z` where `y` is nonzero, return `0.y.0`. This is because any changes from `0.y.0` to `0.y.z` must be backwards compatible by SemVer.
- Else, the input must be `x.y.z` where `x` is nonzero. Return `x.0.0`. This is because any changes from `x.0.0` to `x.y.z` must be backwards compatible by SemVer.

DilumAluthge:
The natural follow-up question is: "under what circumstances might my package be compatible with Foo 0.1.0 and Foo 0.2.0 but not Foo 0.2.1"

Suppose that our package's `[compat]` section looks like this before the CompatHelper/Dependabot PR:
```toml
Foo = "0.1"
Bar = "0.8"
```

And after the CompatHelper/Dependabot PR, it looks like this:
```toml
Foo = "0.1, 0.2"
Bar = "0.8"
```

Suppose that:
- Foo 0.1.0 is compatible with Bar 0.8.x
- Foo 0.2.0 is compatible with Bar 0.8.x
- Foo 0.2.1 is compatible with Bar 0.9.x

Now, our package will be compatible with Foo 0.1.0 and Foo 0.2.0. But it will not be compatible with Foo 0.2.1.

So, when CI runs our package's test suite, all that we need to ensure is that the tests run with Foo 0.2.x. In this case, they will run with Foo 0.2.0. They can't run with Foo 0.2.1 because Foo 0.2.1 requires Bar 0.9.x, but our package only supports Bar 0.8.x.

But, from a CompatHelper/Dependabot point of view, it is perfectly fine that our tests ran with Foo 0.2.0. Because any changes from Foo 0.2.0 to 0.2.1 (or Foo 0.2.0 to Foo 0.2.x) are by definition backwards compatible.

And fundamentally, what we really want to test when CompatHelper/Dependabot makes a PR to change `Foo = "0.1"` to `Foo = "0.1, 0.2"` is that our package's tests will pass when the tests are run with Foo 0.2.x for some x.

00vareladavid:
> So, one of the enhancements that I would like to make in the future is to be able to slightly loosen the requirement from "latest compatible version of your dependency" to "latest compatible breaking family of versions of your dependency".

Since CompatHelper knows the new version, it can determine this on its own correct? e.g. if the new version is 0.2.3 it can know that the set of compatible versions is 0.2.x

> fundamentally, what we really want to test when CompatHelper/Dependabot makes a PR to change Foo = "0.1" to Foo = "0.1, 0.2" is that our package's tests will pass when the tests are run with Foo 0.2.x for some x.

You can make sure that some 0.2.x is installed with `pkg> add Foo@0.2` and you can then pin the package to make sure that `Pkg.test` preserves it.

It seems like this workflow is already supported? Am I missing something?

00vareladavid:
FWIW I don't mean to seem like I am necessarily against this :sweat_smile: , just that it seems like there is already a way to do it.

DilumAluthge:
So, a couple of notes.

---



> Since CompatHelper knows the new version, it can determine this on its own correct? e.g. if the new version is 0.2.3 it can know that the set of compatible versions is 0.2.x

Just to clarify, CompatHelper isn't actually running the tests itself. This is important to note, because in the future we will deprecate CompatHelper in favor of Dependabot. When we add Julia support to Dependabot, we won't be able to run anything compute-intensive inside Dependabot. For example, we won't be able to run the tests inside Dependabot.

We can add some minimal autodetection to the various CI providers. E.g. for GitHub Actions, we will add some autodetection to [julia-actions/julia-runtest](https://github.com/julia-actions/julia-runtest). And we can do similar things for Travis and AppVeyor. But this auto-detection won't be very fancy. For example, we will probably only detect between the following two cases:
1. This PR is a CompatHelper/Dependabot PR.
2. This PR is NOT a CompatHelper/Dependabot PR.

If this is the only information we have (i.e. "yes this is a Dependabot PR" or "no this is not a Dependabot PR"), we should still be able to have this functionality.
In this example, if this is a dependabot PR, we will pass `force_latest_compatible_version = true`, and if it is not a dependabot PR, we will pass `force_latest_compatible_version = false`.

---

> You can make sure that some 0.2.x is installed with `pkg> add Foo@0.2` and you can then pin the package to make sure that `Pkg.test` preserves it.

Maybe I am misunderstanding, but I'm not able to get this to work.

Here's a real example, based off of the `test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml` project file from this PR.

The `[compat]` entry for `MLJModelInterface` is:
```toml
[compat]
MLJModelInterface = "0.1"
```

So suppose I do `] add MLJModelInterface@0.1` followed by `pin MLJModelInterface@0.1`. When I do this, Pkg ends up giving me `MLJModelInterface v0.1.8`, which it then pins.

Now I run `] test`:
- Expected behavior: the test suite runs the tests with `MLJModelInterface v0.1.0`. This should be allowed, right? I did `pin MLJModelInterface@0.1`, which should mean that I want to restrict to any 0.1.x version of MLJModelInterface. So even though I currently have `MLJModelInterface v0.1.8`, the tiered resolver that runs at the beginning of the test suite should be able to downgrade from `MLJModelInterface v0.1.8` to `MLJModelInterface v0.1.0`, which should be allowed because I wanted to restrict to any 0.1.x version of MLJModelInterface.
- Actual behavior: I get an `Unsatisfiable requirements` error.

Full output:
```julia
(OldOnly2) pkg> instantiate
  Installing known registries into `~/.julia`
   Installed ScientificTypes ─── v0.7.2
   Installed MLJModelInterface ─ v0.1.8
    Updating `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml`
  [e80e1ace] + MLJModelInterface v0.1.8
    Updating `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Manifest.toml`
  [e80e1ace] + MLJModelInterface v0.1.8
  [321657f4] + ScientificTypes v0.7.2

(OldOnly2) pkg> add MLJModelInterface@0.1
   Resolving package versions...
  No Changes to `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml`
  Progress [========================================>]  3/3
3 dependencies successfully precompiled in 4 seconds

(OldOnly2) pkg> pin MLJModelInterface@0.1
   Resolving package versions...
    Updating `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml`
  [e80e1ace] ↓ MLJModelInterface v0.1.8 ⇒ v0.1.8 ⚲
    Updating `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Manifest.toml`
  [e80e1ace] ↓ MLJModelInterface v0.1.8 ⇒ v0.1.8 ⚲

(OldOnly2) pkg> status
     Project OldOnly2 v0.1.0
      Status `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml`
  [e80e1ace] MLJModelInterface v0.1.8 ⚲

(OldOnly2) pkg> test
     Testing OldOnly2
┌ Warning: Could not use exact versions of packages in manifest, re-resolving
└ @ Pkg.Operations ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Operations.jl:1474
ERROR: Unsatisfiable requirements detected for package MLJModelInterface [e80e1ace]:
 MLJModelInterface [e80e1ace] log:
 ├─possible versions are: 0.1.0-0.4.1 or uninstalled
 ├─restricted to versions 0.1 by OldOnly2 [abc047b9], leaving only versions 0.1.0-0.1.8
 │ └─OldOnly2 [abc047b9] log:
 │   ├─possible versions are: 0.1.0 or uninstalled
 │   └─OldOnly2 [abc047b9] is fixed to version 0.1.0
 ├─restricted to versions 0.1.8 by an explicit requirement, leaving only versions 0.1.8
 └─restricted by compatibility requirements with ScientificTypes [321657f4] to versions: 0.1.0 or uninstalled — no versions left
   └─ScientificTypes [321657f4] log:
     ├─possible versions are: 0.1.0-1.1.1 or uninstalled
     └─restricted to versions 0.6 by an explicit requirement, leaving only versions 0.6.0-0.6.1
```

Observe that if I omit the `] pin` step, the tests successfully run with `MLJModelInterface v0.1.0`. Full output:
```julia
(OldOnly2) pkg> instantiate
  Installing known registries into `~/.julia`
   Installed ScientificTypes ─── v0.7.2
   Installed MLJModelInterface ─ v0.1.8
    Updating `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml`
  [e80e1ace] + MLJModelInterface v0.1.8
    Updating `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Manifest.toml`
  [e80e1ace] + MLJModelInterface v0.1.8
  [321657f4] + ScientificTypes v0.7.2

(OldOnly2) pkg> add MLJModelInterface@0.1
   Resolving package versions...
  No Changes to `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml`
  Progress [========================================>]  3/3
3 dependencies successfully precompiled in 4 seconds

(OldOnly2) pkg> status
     Project OldOnly2 v0.1.0
      Status `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2/Project.toml`
  [e80e1ace] MLJModelInterface v0.1.8

(OldOnly2) pkg> test
     Testing OldOnly2
┌ Warning: Could not use exact versions of packages in manifest, re-resolving
└ @ Pkg.Operations ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Operations.jl:1474
      Status `/private/var/folders/jy/7hh5zyw95cg2lh66lfpthqq80000gn/T/jl_SspDg0/Project.toml`
  [e80e1ace] MLJModelInterface v0.1.0
  [abc047b9] OldOnly2 v0.1.0 `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2`
  [321657f4] ScientificTypes v0.6.1
  [8dfed614] Test `@stdlib/Test`
      Status `/private/var/folders/jy/7hh5zyw95cg2lh66lfpthqq80000gn/T/jl_SspDg0/Manifest.toml`
  [e80e1ace] MLJModelInterface v0.1.0
  [abc047b9] OldOnly2 v0.1.0 `~/Downloads/Pkg.jl/test/test_packages/force_latest_compatible_version/OldOnly2`
  [321657f4] ScientificTypes v0.6.1
  [2a0f44e3] Base64 `@stdlib/Base64`
  [b77e0a4c] InteractiveUtils `@stdlib/InteractiveUtils`
  [56ddb016] Logging `@stdlib/Logging`
  [d6f4376e] Markdown `@stdlib/Markdown`
  [9a3f8284] Random `@stdlib/Random`
  [9e88b42a] Serialization `@stdlib/Serialization`
  Progress [========================================>]  3/3
3 dependencies successfully precompiled in 2 seconds
     Testing Running tests...
Test Summary: | Pass  Total
OldOnly2.jl   |    1      1
     Testing OldOnly2 tests passed
```

DilumAluthge:
> I haven't implemented this feature in this PR, because I would like to get this PR reviewed and merged relatively quickly. But I will implement this feature in a future PR. This is really the workflow that we need to support in CompatHelper/Dependabot PRs.

I went ahead and added this functionality to this PR. It took very few lines of code to implement.

src/force_latest_compatible_version.jl
DilumAluthge: I have removed the `_` prefix from all other function names. I had to leave the `_` prefix in the `_force_latest_compatible_version` function name because it conflicts with the name of the `force_latest_compatible_version` kwarg.

DilumAluthge: Okay, I have renamed the `_force_latest_compatible_version` function to `check_force_latest_compatible_version`. Now, this PR do not have any functions that begin with the `_` prefix.

DilumAluthge: @KristofferC Can I mark this as resolved?

I've removed the extra file, and I've removed all the `_` prefixes, so the only outstanding issue is the fact that the function is split up into several different methods, which I'd like to keep to make future refactoring and feature addition easier.

00vareladavid:
> If this is the only information we have (i.e. "yes this is a Dependabot PR" or "no this is not a Dependabot PR"), we should still be able to have this functionality.

Ah, I hadn't realized this. In that case it does seem like you would need to read straight from the compat bounds.

Something that could be done (possibly in a future PR) is to take the set of version ranges you compute and add them during the sandbox resolve step. That way you can error out with a resolver trace of why it was not possible (in the case that the resolve step fails). It seems it would give you a better shot of meeting those requirements since you would be explicitly asking the resolver to look for a solution.

DilumAluthge:
> Something that could be done (possibly in a future PR) is to take the set of version ranges you compute and add them during the sandbox resolve step. That way you can error out with a resolver trace of why it was not possible (in the case that the resolve step fails).

Sure, that would be an enhancement that we could make. In the documentation, I've indicated that this feature is experimental, which will allow us to make modification to its behavior.

Since this PR is already pretty big, I'd prefer to get this merged first, and then make this enhancement in a future PR.

src/force_latest_compatible_version.jl
DilumAluthge: @KristofferC marking this as resolved.

src/Operations.jl
```diff
@@ -1782,4 +1797,114 @@ function status(env::EnvCache, pkgs::Vector{PackageSpec}=PackageSpec[];
...
+    return result
+end
+
+function get_earliest_backwards_compatible_version(ver::Base.VersionNumber)
+    major = ver.major
+    minor = ver.minor
+    patch = ver.patch
+    (major > 0) && return Base.VersionNumber(major, 0, 0)
+    (minor > 0) && return Base.VersionNumber(0, minor, 0)
+    return Base.VersionNumber(0, 0, patch)
```
00vareladavid, line 1879:
```suggestion
    (ver.major > 0) && return Base.VersionNumber(ver.major, 0, 0)
    (ver.minor > 0) && return Base.VersionNumber(0, ver.minor, 0)
    return Base.VersionNumber(0, 0, ver.patch)
```

src/Operations.jl
```diff
@@ -1782,4 +1797,114 @@ function status(env::EnvCache, pkgs::Vector{PackageSpec}=PackageSpec[];
...
+                                               allow_earlier_backwards_compatible_versions::Bool = true)
+    isempty(uuid_list) && return true
+    results = check_force_latest_compatible_version.(
+        Ref(ctx),
+        uuid_list;
+        target_name,
+        allow_earlier_backwards_compatible_versions,
+    )
+    return all(results)
+end
```
00vareladavid, line 1827: Perhaps simpler to merge these two functions:

```suggestion
function check_force_latest_compatible_version(ctx::Types.Context;
                                               target_name = nothing,
                                               allow_earlier_backwards_compatible_versions::Bool = true)
    direct_deps = load_direct_deps(ctx.env)
    direct_deps_uuids = [dep.uuid for dep in direct_deps]
    uuid_list = filter(!is_stdlib, direct_deps_uuids)
    isempty(uuid_list) && return true
    results = check_force_latest_compatible_version.(
        Ref(ctx),
        uuid_list;
        target_name,
        allow_earlier_backwards_compatible_versions,
    )
    return all(results)
end
```

src/Operations.jl
```diff
@@ -1782,4 +1797,114 @@ function status(env::EnvCache, pkgs::Vector{PackageSpec}=PackageSpec[];
...
+function get_all_registered_versions(ctx::Types.Context,
+                                     uuid::Base.UUID)
+    versions = Set{VersionNumber}()
+    for reg in ctx.registries
+        pkg = get(reg, uuid, nothing)
+        if pkg !== nothing
+            info = Registry.registry_info(pkg)
+            for v in keys(info.version_info)
+                push!(versions, v)
+            end
```
00vareladavid, line 1904: I think this can be stated as:
```suggestion
            union!(versions, keys(info.version_info))
```

src/Operations.jl
```diff
@@ -1782,4 +1797,114 @@ function status(env::EnvCache, pkgs::Vector{PackageSpec}=PackageSpec[];
...
+
+function get_latest_compatible_version(ctx::Types.Context,
+                                       uuid::Base.UUID,
+                                       compat_entry_string::AbstractString)
+    all_registered_versions = get_all_registered_versions(ctx, uuid)
+    compat_spec = Pkg.Types.semver_spec(compat_entry_string)
+    compatible_versions = filter(
+        x -> x in compat_spec,
+        all_registered_versions,
+    )
```
00vareladavid, line 1890:
```suggestion
    compatible_versions = filter(in(compat_spec), all_registered_versions)
```

00vareladavid (approved):
I left a few nitpicky suggestions but otherwise looks good to me

src/force_latest_compatible_version.jl
DilumAluthge: Just FYI, I ended up merging some of the methods of the `check_force_latest_compatible_version` function after all. So now there are only two methods for `force_latest_compatible_version`.

DilumAluthge:
Thanks @00vareladavid!

DilumAluthge:
> > Something that could be done (possibly in a future PR) is to take the set of version ranges you compute and add them during the sandbox resolve step. That way you can error out with a resolver trace of why it was not possible (in the case that the resolve step fails).
> 
> Sure, that would be an enhancement that we could make. In the documentation, I've indicated that this feature is experimental, which will allow us to make modification to its behavior.
> 
> Since this PR is already pretty big, I'd prefer to get this merged first, and then make this enhancement in a future PR.

I have implemented this change in #2468.

src/Pkg.jl
```diff
@@ -187,10 +187,22 @@ const update = API.up
   - `coverage::Bool=false`: enable or disable generation of coverage statistics.
   - `julia_args::Union{Cmd, Vector{String}}`: options to be passed the test process.
   - `test_args::Union{Cmd, Vector{String}}`: test arguments (`ARGS`) available in the test process.
+  - `force_latest_compatible_version::Bool=false`: [EXPERIMENTAL] force the latest compatible version of each direct dependency.
+  - `allow_earlier_backwards_compatible_versions::Bool=false`: [EXPERIMENTAL] allow any version that is backwards-compatible with the latest compatible version of a direct dependency. If `force_latest_compatible_version` is `false`, then the value of `force_latest_compatible_version` has no effect.
```
KristofferC, line 191: I don't understand what this means. 

(also very long line).

PR #2448, opened 2021-03-24
allow option for using git CLI for git downloads

KristofferC:


KristofferC:
I don't really understand how the latest commit I added caused CI to start failing... Hmm

DilumAluthge (approved):
LGTM

StefanKarpinski:
This is great. Now if we delete all the stuff that uses LibGit2, we'll really be getting somewhere!

KristofferC:
Hehe, work hard to remove all external download commands in favor of a library when it comes to downloading via HTTP, and then work hard to remove a library for an external command when it comes to downloading via Git :P.

But seriously, I don't think we want to be bundling the Git executable and I think it is nice that the package manager doesn't require git to fully work. Using a library gives us nice progress bars and [a nice API](https://github.com/JuliaLang/Pkg.jl/blob/8ad44253976528e4ede1a1019b6e6c22bd2f3ab4/src/Types.jl#L652-L662), and it works well in the vast majority of cases. Having an escape hatch to the CLI is probably enough.

StefanKarpinski:
Unlike libcurl, which is complex but ultimately well designed and capable, the more I've dug into libgit2, the worse my opinion of it has become. It is, for example, impossible to use it securely without patching it. I think that we should preferentially use a `git` binary if one exists. Yes, that means we lose a progress bar for the rare operations where we still end up downloading something via git, but it seems way better than dealing with libgit2. As to the API, well, opinions are subjective, but I'm not sure that I would agree that the LibGit2 API is nice.

DilumAluthge:
Could we merge this PR now, and then we can make a separate PR that implements this:

> preferentially use a `git` binary if one exists

Then, in that PR, we can continue the discussion about whether or not we want to default to the `git` binary (if one exists) or default to `libgit2`.

But it would be good to get this PR merged so at least people have the ability to opt-in to the Git CLI.

KristofferC:
ENV var should be added to docs when https://github.com/JuliaLang/Pkg.jl/issues/2453 is fixed.

PR #2459, opened 2021-03-27
Add helper notes about syncing with julia repo

IanButterworth:
Just some explanations and convenience links.

Came from a discussion at https://github.com/JuliaPackaging/BumpStdlibs.jl/issues/33

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,19 @@
 # Pkg
 
+Development repository for Julia's package manager,
+shipped with Julia v1.0 and above.
+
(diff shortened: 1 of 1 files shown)


DilumAluthge (approved):

fredrikekre (commented):
Seems like this should be in some devdocs rather than the first thing in the README. It should at least be below the instructions for using the dev version.

IanButterworth:
Oops. Your comment loaded just after I merged. I'll reorder it

PR #2462, opened 2021-03-28
Precompilation: add tip to auto-precomp for getting error details

IanButterworth:
There's been some confusion about how to get information on the errors that are counted during auto-precomp but that aren't thrown or detailed.

This adds a helper tip `To see a full report either call 'pkg> precompile' or load the package(s)` that's seen at the top here

(I added an `error()` to FileIO)

image

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1307,7 +1307,7 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
                 plural1 = length(failed_deps) == 1 ? "y" : "ies"
                 plural2 = length(failed_deps) == 1 ? "" : "s"
                 print(iostr, "\n  ", color_string("$(length(failed_deps))", Base.error_color()), " dependenc$(plural1) errored. ")
-                print(iostr, "To see a full report either call `pkg> precompile` or load the package$(plural2)")
+                print(iostr, "To see a full report either call `import Pkg; Pkg.precompile()` or load the package$(plural2)")
             end
         end
         lock(print_lock) do
--- a/src/API.jl
+++ b/src/API.jl
@@ -1307,7 +1307,7 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
                 plural1 = length(failed_deps) == 1 ? "y" : "ies"
                 plural2 = length(failed_deps) == 1 ? "" : "s"
                 print(iostr, "\n  ", color_string("$(length(failed_deps))", Base.error_color()), " dependenc$(plural1) errored. ")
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -1304,8 +1304,10 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
                 )
             end
             if internal_call && !isempty(failed_deps)
-                plural = length(failed_deps) == 1 ? "y" : "ies"
-                print(iostr, "\n  ", color_string("$(length(failed_deps))", Base.error_color()), " dependenc$(plural) errored")
+                plural1 = length(failed_deps) == 1 ? "y" : "ies"
+                plural2 = length(failed_deps) == 1 ? "" : "s"
+                print(iostr, "\n  ", color_string("$(length(failed_deps))", Base.error_color()), " dependenc$(plural1) errored. ")
+                print(iostr, "To see a full report either call `pkg> precompile` or load the package$(plural2)")
```
KristofferC, line 1310: Not sure if it is better to suggest `Pkg.precompile` over the REPL interface.

IanButterworth: Perhaps it could just say `precompile` and the user can interpret whether they want repl or Pkg. mode?

KristofferC: Imo let's go with the copy-pastable `Pkg.precompile()`? I think we do the same for other suggestions, like if you don't have a certain package

```
julia> using Wat
ERROR: ArgumentError: Package Wat not found in current path:
- Run `import Pkg; Pkg.add("Wat")` to install the Wat package.
```

IanButterworth: Yeah, true that is more standard practice, but it's a shame it's so long.

I've wondered before whether it would be possible to special-case `pkg>` in the julia REPL, so that
```
julia> pkg> add Wat
```
switches to pkg repl and executes `add Wat`. That would mean all the readmes out there could just say:
To install, run `pkg> add Foo`, and it would both work for brand new and familiar users.

IanButterworth: hmmm. Doesn't it seem a bit odd to have the message be this from within the pkg repl?
```
Precompiling project...
  ✗ FileIO
  ✗ TiffImages
  ✗ Netpbm
  ✗ ImageIO
  1 dependency successfully precompiled in 6 seconds (36 already precompiled)
  4 dependencies errored. To see a full report call `import Pkg; Pkg.precompile()` or load the packages.
```

IanButterworth:
Now
```
Precompiling project...
  ✗ FileIO
  ✗ TiffImages
  ✗ Netpbm
  ✗ ImageIO
  1 dependency successfully precompiled in 6 seconds (36 already precompiled)
  4 dependencies errored. To see a full report call `import Pkg; Pkg.precompile()` or load the packages
```

PR #2465, opened 2021-03-29
Remove some unnecessary feedback

### LINKED ISSUE #1629: Uncecessary "exists and looks like the correct package" printout
```
(@BinaryBuilder) pkg> dev ../Registrator.jl/ ../RegistryTools.jl
Path `../Registrator.jl/` exists and looks like the correct package. Using existing path.
Path `../RegistryTools.jl` exists and looks like the correct package. Using existing path.
  Resolving package versions...
   Updating `~/.julia/environments/BinaryBuilder/Project.toml`
  [4418983a] + Registrator v1.1.0 [`../../../JuliaPkgs/Registrator.jl`]
  [d1eb7eb1] + RegistryTools v1.2.1 [`../../../JuliaPkgs/RegistryTools.jl`]
   Updating `~/.julia/environments/BinaryBuilder/Manifest.toml`
  [4418983a] ↑ Registrator v1.0.0 ⇒ v1.1.0 [`../../../JuliaPkgs/Registrator.jl`]
  [d1eb7eb1] ~ RegistryTools v1.2.1 ⇒ v1.2.1 [`../../../JuliaPkgs/RegistryTools.jl`]
```

Don't really see the point of the first messages in this case.

00vareladavid:
Fix #1629

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -477,7 +477,6 @@ function handle_repo_develop!(ctx::Context, pkg::PackageSpec, shared::Bool)
         end
         if isdir(dev_path)
             resolve_projectfile!(ctx.env, pkg, dev_path)
-            println(ctx.io, "Path `$(dev_path)` exists and looks like the correct package. Using existing path.")
             if is_local_path
                 pkg.path = isabspath(dev_path) ? dev_path : relative_project_path(ctx.env.project_file, dev_path)
             else


KristofferC (approved):

fredrikekre:
Should the second one be kept?

00vareladavid:
@fredrikekre IMO the second one reads more like a log message, if anything.

fredrikekre:
That is the one that shows up if you do `dev Example` and you already happen to have an `Example` package in the folder, right? I think that should be displayed since you might otherwise think you get a fresh clone. The message I wanted to remove was just when you explicitly `dev` a path -- then I already know there is apackage there.

00vareladavid:
> That is the one that shows up if you do dev Example and you already happen to have an Example package in the folder, right?

yep

> you might otherwise think you get a fresh clone

yeah, this would probably be confusing if someone does not expect it

PR #2469, opened 2021-03-30
Add the `Pkg.Versions.to_semver_spec` function for converting `VersionSpec`s to strings, with the property that `Pkg.Versions.semver_spec(Pkg.Versions.to_semver_spec(spec::VersionSpec)) == spec`

DilumAluthge:
## Summary

This pull request adds the `Pkg.Versions.to_semver_spec` function, which has the following signature:
```
Pkg.Versions.to_semver_spec(spec::Pkg.Versions.VersionSpec) -> String
```

For `VersionSpec`s in its domain, the `to_semver_spec` function has the following property:
```
Pkg.Versions.semver_spec(Pkg.Versions.to_semver_spec(spec)) == spec
```

That is, if `spec::VersionSpec` is in the domain of `to_semver_spec`, then `to_semver_spec` is the inverse function of `semver_spec`.

## Motivation

I think this is a useful feature in general, but also I would specifically like to use this functionality to simplify #2468.

Here is an example use case from #2468:
```julia
julia> original_compat_entry_string = "1.2.3 - 4.5.6, 7.8.9 - 10.11.12"
"1.2.3 - 4.5.6, 7.8.9 - 10.11.12"

julia> latest_compatible_version = v"10.11.12" # the result of getting all registered versions from the registry, checking them against the compat entry, and taking the maximum
v"10.11.12"

julia> earliest_backwards_compatible_version = Pkg.Operations.get_earliest_backwards_compatible_version(latest_compatible_version)
v"10.0.0"

julia> original_compat_spec = Pkg.Versions.semver_spec(original_compat_entry_string)
VersionSpec("[1.2.3-4.5.6, 7.8.9-10.11.12]")

julia> new_compat_spec = intersect(original_compat_spec, Pkg.Versions.semver_spec("^$(earliest_backwards_compatible_version)"))
VersionSpec("10.0.0-10.11.12")

julia> new_compat_entry_string = Pkg.Versions.to_semver_spec(new_compat_spec)
"10.0.0 - 10.11.12"
```

## Domain of `to_semver_spec`

Almost all `VersionSpec`s are in the domain of `to_semver_spec`. However, there is one type of `VersionSpec` that I cannot figure out how to represent using the SemVer notation:
```julia
julia> lower = Pkg.Versions.VersionBound()
Pkg.Versions.VersionBound((0x00000000, 0x00000000, 0x00000000), 0)

julia> upper = Pkg.Versions.VersionBound(1)
Pkg.Versions.VersionBound((0x00000001, 0x00000000, 0x00000000), 1)

julia> r = Pkg.Types.VersionRange(lower, upper)
VersionRange("0-1.0.0")

julia> r.lower.n
0

julia> r.upper.n
1

julia> spec = Pkg.Types.VersionSpec([r])
VersionSpec("0-1.0.0")
```

I simply cannot figure out how to write a string `str` such that `semver_spec(str)` returns the above `VersionSpec`.

The specific issue here is that `r.lower.n` is zero and `r.upper.n` is nonzero. I can't figure out how to write a range `r` that has this property in SemVer notation.

### DIFF
--- a/src/Versions.jl
+++ b/src/Versions.jl
@@ -179,6 +179,19 @@ function Base.print(io::IO, r::VersionRange)
 end
 Base.show(io::IO, r::VersionRange) = print(io, "VersionRange(\"", r, "\")")
 
+function to_semver_spec(r::VersionRange)
+    m, n = r.lower.n, r.upper.n
+    if (m, n) == (0, 0)
+        return "≥0"
+    elseif m == 0
+        throw(ArgumentError("This version range cannot be represented using SemVer notation"))
+    elseif n == 0
+        return string("≥", join(r.lower.t, "."),)
+    else
+        return string(join(r.lower.t[1:m], "."), " - ", join(r.upper.t[1:n], "."))
+    end
+end
+
 Base.in(v::VersionNumber, r::VersionRange) = r.lower ≲ v ≲ r.upper
 
 Base.intersect(a::VersionRange, b::VersionRange) = VersionRange(stricterlower(a.lower, b.lower), stricterupper(a.upper, b.upper))
@@ -305,6 +318,13 @@ function semver_spec(s::String)
         push!(ranges, range)
     end
     return VersionSpec(ranges)
+end
+
+function to_semver_spec(spec::VersionSpec)
+    ranges = spec.ranges
+    isempty(ranges) && return "1 - 0"
+    specs = to_semver_spec.(ranges)
+    return join(specs, ", ")
 end
 
 function semver_interval(m::RegexMatch)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -39,6 +39,7 @@ include("sandbox.jl")
 include("resolve.jl")
 include("misc.jl")
 include("force_latest_compatible_version.jl")
+include("versions.jl")
 
 # clean up locally cached registry
 rm(joinpath(@__DIR__, "registries"); force = true, recursive = true)
--- a/test/versions.jl
+++ b/test/versions.jl
@@ -12,7 +12,7 @@ function roundtrip_to_semver_spec(str_1::AbstractString)
     spec_2 = Pkg.Versions.semver_spec(str_2)
     str_3 = Pkg.Versions.to_semver_spec(spec_2)
     spec_3 = Pkg.Versions.semver_spec(str_3)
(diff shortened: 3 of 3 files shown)


KristofferC:
To me, having to go `VersionSpec` -> compat string means there is a design flaw somewhere. The compact string should immediately be converted to a `VectionSpec` in Pkg and be used as such throughout everything. The design flaw here is that the `Project` struct store the compat info as a string and it is then converted much later (in the resolver code at https://github.com/JuliaLang/Pkg.jl/blob/a4ade8cea58412a116be5066bb5f84f6d63fba3c/src/Operations.jl#L190).

I have some WIP work for a rewrite of the `Project` / `Manifest` structs and there I store the compat as a `VersionSpec` (but it keeps the compact string around for printing it out to a Project.toml file later (https://github.com/JuliaLang/Pkg.jl/blob/7a4d8a2eb062f9dc8340f052507c28c6e3cd6812/src/Environment/project.jl#L12-L14).

If the structure was like that, it would be simple to modify the `Project` struct and enter whatever `VersionSpec` you want in there. Sure, you couldn't write that compact to a Project.toml but you never need to.

DilumAluthge:
Yeah, my use case here is that I have a `spec::VersionSpec`, and I want to do e.g. `ctx.env.project.compat["Foo"] = spec` where `ctx::Context`.

It sounds like instead of this PR, I should wait for your rewrite of the `Project` struct?

DilumAluthge:
> Sure, you couldn't write that compact to a Project.toml but you never need to.

Yeah I don't need to do that.

The only reason that I need to modify `ctx.env.project.compat["Foo"]` is to force the resolver to use my new compat bounds when resolving.

KristofferC:
> It sounds like instead of this PR, I should wait for your rewrite of the Project struct?

That whole thing will likely take a long time to get in and has to go piece by piece. But just doing the small `Project` change could be done quite easily by tweaking:

https://github.com/JuliaLang/Pkg.jl/blob/a4ade8cea58412a116be5066bb5f84f6d63fba3c/src/Types.jl#L210

It is not enough to store ` Dict{String, VersionSpec}` as the comment suggest because you also need the string (to print it to a TOML file) but keeping both around like I did in my link and allowing to override the `VersionSpec` should be fine.

DilumAluthge:
Closing in favor of implementing the above

PR #2472, opened 2021-03-30
async git installs

oxinabox:
When you have dozens of dependencies that are on gitlab the time to git clone them all is annoying.
I think it's completely safe and reasonable to do them async

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -726,7 +726,7 @@ function download_source(ctx::Context; readonly=true)
     ##################################################
     # Use LibGit2 to download any remaining packages #
     ##################################################
-    for (pkg, urls, path) in missed_packages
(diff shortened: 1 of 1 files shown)


KristofferC:
The progress bars will fight with each other so need to address that I think. Also, there is a variable of how many concurrent downloads to do at once (which defaults to 8). I think `asyncmap` defaults to 100 which is a bit excessive.

KristofferC:
Bump, is there a plan to update this PR?

oxinabox:
I am not going to have time to work on this any time soon. 
Shall  I close it and open an issue?

I still think this would be a great feature to have for people not using github.
Who tend to be closed source.
though it may be unneeded if they have a private PkgServer as well? 
Invenia doesn't.

PR #2477, opened 2021-04-01
Refactor: do not use double constructor for PackageSpec

### LINKED ISSUE #2282: The double constructor PackageSpec is very confusing
For some reason we have the struct `PackageSpec` in `Pkg.Types` but then what we export is a method with the same name 

```jl
julia> using Pkg

julia> PackageSpec
Package (generic function with 4 methods)
```

that acts differently from the constructor to the `PackageSpec` struct.

This leads to very confusing things like:

```jl
julia> Pkg.develop(Pkg.Types.PackageSpec(path="Foo"))
ERROR: name, UUID, URL, or filesystem path specification required when calling `develop`
Stacktrace:
...

julia> Pkg.develop(PackageSpec(path="Foo"))
Path `Foo` exists and looks like the correct package. Using existing path.
  Resolving package versions...
```

00vareladavid:
Fix #2282 

I admit that messing with the constructor was a thoroughly terrible idea. The check could be circumvented anyway by manually setting the fields after the constructor call. The checking should be done inside the API.

### DIFF
--- a/test/artifacts.jl
+++ b/test/artifacts.jl
@@ -316,12 +316,12 @@ end
     # the global artifact namespace and package list, but it should be harmless.
     mktempdir() do project_path
         with_pkg_env(project_path) do
-            copy_test_package(project_path, "ArtifactInstallation")
+            path = git_init_package(project_path, joinpath(@__DIR__, "test_packages", "ArtifactInstallation"))
             add_this_pkg()
             Pkg.add(Pkg.Types.PackageSpec(
                 name="ArtifactInstallation",
                 uuid=Base.UUID("02111abe-2050-1119-117e-b30112b5bdc4"),
-                path=joinpath(project_path, "ArtifactInstallation"),
+                path=path,
             ))
 
             # Run test harness
@@ -635,6 +635,7 @@ end
         # loads overridden package artifacts.
         Pkg.activate(depot_container) do
             copy_test_package(depot_container, "ArtifactOverrideLoading")
+            git_init_and_commit(joinpath(depot_container, "ArtifactOverrideLoading"))
             add_this_pkg()
             Pkg.add(Pkg.Types.PackageSpec(
                 name="ArtifactOverrideLoading",
--- a/test/test_packages/ArtifactOverrideLoading/Project.toml
+++ b/test/test_packages/ArtifactOverrideLoading/Project.toml
@@ -1,5 +1,5 @@
 name = "ArtifactOverrideLoading"
-uuid = "9f5278ee-142d-5d9a-9723-c4aebaa272e1"
+uuid = "7b879065-7f74-5fa4-bdd5-9b7a15df8941"
 version = "0.1.0"
 
 [deps]
--- a/src/API.jl
+++ b/src/API.jl
@@ -143,6 +143,8 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
             Registry.download_default_registries(io)
             ctx = Context()
             kwargs = merge((;kwargs...), (:io => io,))
+            pkgs = deepcopy(pkgs) # don't mutate input
+            foreach(pkg -> handle_package_input!(pkg), pkgs)
             ret = $f(ctx, pkgs; kwargs...)
             $(f in (:add, :up, :pin, :free, :build)) && Pkg._auto_precompile(ctx)
             return ret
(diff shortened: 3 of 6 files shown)


KristofferC:
Could the `handle_package_input!` be added here instead https://github.com/JuliaLang/Pkg.jl/blob/9cf608c85822e781fa70df338951ec5ebfe1f8cc/src/API.jl#L146? And then it is used for everything?

00vareladavid:
There is something weird going on with the artifacts tests. It was giving some bad input to `Pkg.add`. I adjusted them to what I thought was the most likely intention, but I do not really understand those tests yet.

00vareladavid:
Perhaps @staticfloat (or anyone who knows those tests) could take a quick look. I'm almost certain it was just a matter of `git init`ing the test packages, but there was a mismatch between some UUIDs that I'm not so sure about.

test/test_packages/ArtifactOverrideLoading/Project.toml
```diff
@@ -1,5 +1,5 @@
 name = "ArtifactOverrideLoading"
-uuid = "9f5278ee-142d-5d9a-9723-c4aebaa272e1"
+uuid = "7b879065-7f74-5fa4-bdd5-9b7a15df8941"
```
staticfloat, line 2: Why did you need to change the UUID here?

test/artifacts.jl
```diff
@@ -316,12 +316,12 @@ end
...
     mktempdir() do project_path
         with_pkg_env(project_path) do
-            copy_test_package(project_path, "ArtifactInstallation")
+            path = git_init_package(project_path, joinpath(@__DIR__, "test_packages", "ArtifactInstallation"))
             add_this_pkg()
             Pkg.add(Pkg.Types.PackageSpec(
                 name="ArtifactInstallation",
                 uuid=Base.UUID("02111abe-2050-1119-117e-b30112b5bdc4"),
-                path=joinpath(project_path, "ArtifactInstallation"),
+                path=path,
```
staticfloat, line 324: What were the errors you were seeing?  I'm not sure why `git_init_package()` would have any different behavior here from `copy_test_package()`?

00vareladavid: Thanks for reviewing. `Pkg.add` requires the package to be git version controlled because it always clones (it doesn't track the path directly like `Pkg.develop` does).

test/test_packages/ArtifactOverrideLoading/Project.toml
00vareladavid: I changed it to the value of `aol_uuid` because that was the value explicitly given here: https://github.com/JuliaLang/Pkg.jl/blob/f0731405b557b8c04788f7fc6ca03d5a2e673280/test/artifacts.jl#L641. If they are not the same value, Pkg will error out.

staticfloat: huh, why wasn't this erroring before?

00vareladavid: There is a misuse of the API here. API calls were supposed to be made with `Pkg.PackageSpec`, not `Pkg.Types.PackageSpec`. This is needlessly confusing, which is why this PR removes the second constructor.

staticfloat: Right, I'm specifically interested in why this having the wrong UUID didn't cause an issue before.

00vareladavid: Because `Pkg.PackageSpec` did some pre-processing, which was circumvented by using `Pkg.Types.PackageSpec`. In particular, `Pkg.PackageSpec` automatically sets `repo.source`, while `Pkg.Types.PackageSpec` does not. So the package was in some undefined state which just happened to appear like it was tracking a path when it was written to the manifest.

00vareladavid: So it seems like the intention was that they were supposed to be the same UUID? That is the only thing I was not sure of.

staticfloat: Yes, they definitely should have matched

00vareladavid: Ok, thanks. I think this slight adjustment was all that was necessary

DilumAluthge:
It looks like this PR made a breaking change in the public API: 

On Julia 1.6.1:
julia> import Pkg

julia> Pkg.PackageSpec(; name = "Example", uuid = "7876af07-990d-54b4-ab0e-23690620f79a")
PackageSpec(
 name = Example
 uuid = 7876af07-990d-54b4-ab0e-23690620f79a
 version = *
)

On Julia master:
julia> import Pkg

julia> Pkg.PackageSpec(; name = "Example", uuid = "7876af07-990d-54b4-ab0e-23690620f79a")
ERROR: MethodError: Cannot `convert` an object of type String to an object of type Base.UUID
Closest candidates are:
 convert(::Type{T}, ::T) where T at essentials.jl:228
 Base.UUID(::AbstractString) at uuid.jl:87
 Base.UUID(::Any) at uuid.jl:9
Stacktrace:
 [1] convert(#unused#::Type{Union{Nothing, Base.UUID}}, x::String)
  @ Base ./some.jl:36
 [2] Pkg.Types.PackageSpec(name::String, uuid::String, version::Pkg.Versions.VersionSpec, tree_hash::Nothing, repo::Pkg.Types.GitRepo, path::Nothing, pinned::Bool, url::Nothing, rev::Nothing, subdir::Nothing)
  @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:92
 [3] Pkg.Types.PackageSpec(; name::String, uuid::String, version::Pkg.Versions.VersionSpec, tree_hash::Nothing, repo::Pkg.Types.GitRepo, path::Nothing, pinned::Bool, url::Nothing, rev::Nothing, subdir::Nothing)
  @ Pkg.Types ./util.jl:471
 [4] top-level scope
  @ REPL[1]:1

Is this easy to fix? Or should we revert this PR?

PR #2480, opened 2021-04-02
Refactor: use `VersionSpec` as the canonical compat representation

00vareladavid:
Inspired by the discussion in #2469. As stated in the discussion. a VersionSpec is a much more desirable representation because it can be modified more easily.

Changes:
- immediately convert to `VersionSpec` when reading project file
- keep raw string representation to write back out when operations complete

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -171,6 +171,12 @@ end
 #######################################
 # Dependency gathering and resolution #
 #######################################
+get_compat(proj::Project, name::String) = haskey(proj.compat, name) ? proj.compat[name].val : Types.VersionSpec()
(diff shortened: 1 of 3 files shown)


KristofferC (approved):

DilumAluthge:
Out of curiosity, what happens if the `VersionSpec` and the raw `[compat]` entry string get out of sync with each other?

KristofferC:
Then you will have a bug when you write the TOML file. An assert could be added for the project writer that checks that the string is in sync.

DilumAluthge:
> An assert could be added for the project writer that checks that the string is in sync. That is indeed probably a good idea.

That sounds good. @00vareladavid want to do this in this PR?

Alternatively we can merge this PR, and add the assert in the follow-up PR.

00vareladavid:
Ok, I added the check :+1:

PR #2482, opened 2021-04-02
Don't use the `@assert` macro

DilumAluthge:
In the future, we may disable `@assert`s by default. But I think this particular error should always be shown.

### DIFF
--- a/src/project.jl
+++ b/src/project.jl
@@ -151,7 +151,9 @@ function destructure(project::Project)::Dict
 
     # sanity check for consistency between compat value and string representation
     for (name, compat) in project.compat
-        @assert compat.val == semver_spec(compat.str) "inconsistency between compat values and string representation"
+        if compat.val != semver_spec(compat.str)
+            throw(ErrorException("inconsistency between compat values and string representation"))
+        end
     end
 
     # if a field is set to its default value, don't include it in the write


00vareladavid:
It seems the only way there would be a mismatch would be through an API side error (as opposed to a user error), no?

DilumAluthge:
So, something like https://github.com/JuliaLang/Pkg.jl/pull/2469 will intentionally modify `compat.val` without modifying `compat.str`, right? So then we'd want to throw an error if Pkg tries to write a project to file after that project has been modified by https://github.com/JuliaLang/Pkg.jl/pull/2469

00vareladavid:
..yeah. I guess I was misusing `@assert` for this. I agree we should always throw if we detect this

00vareladavid (approved):

DilumAluthge:
@KristofferC @00vareladavid So, when you call `resolve`, it calls `up`, which in turn calls `write_project`. Here is an example stacktrace:

```julia
```

This means that if we ever modify `compat.val`, we then cannot ever call `resolve`. This seems like a problem, right?

So I think we do need some way of turning a `VersionSpec` back into a string.

KristofferC:
> So, when you call resolve, it calls up, which in turn calls write_project.

Yeah, I don't think you can just call `Pkg.resolve` you would need to call something more internal that doesn't write out the resulting project.
I would think you need to do something like this when testing with this new option:

- Change the `compat.val`
- Call resolve (without writing out the project). 
- Set the compact value back to something (what value this is doesn't matter since the resolve is done).
- Print out the new temporary test project / manifest
- Run the test

DilumAluthge:
> Set the compact value back to something (what value this is doesn't matter since the resolve is done).

Hmmm. How would this work if the test suite runs any Pkg operations? E.g. what if someone has `Pkg.add("Foo")` inside their test suite?

DilumAluthge:
Anyway, I have updated #2498 so that it now does the following:
1. Change the `compat.val` as determined by the `compat_modifier` function provided by the user
2. Call `resolve` without writing out the project
3. Set each `compat.val` back to the original value (`Types.semver_spec(compat.str)`)
4. Run the tests

I do think that this approach has some limitations - e.g. if someone runs `Pkg.add("Foo")` inside their test suite, the resolver will use the old `[compat]` values, instead of the new `[compat]` values that are created by the `compat_modifier` function. But we can figure out those edge cases in the future.

PR #2484, opened 2021-04-03
Precompilation: Don't skip precompilation of different versions of loaded modules

IanButterworth:
This is about those pesky yellow `?`'s

I believe the current behavior of `Base.compilecache` skipping precompilation of already loaded modules (what it internally refers to as `concrete_deps`) is because for the code-load precompilation path a different version of a package can't actually be loaded, so it decides to skip it and save the time, and precomp would be picked up in a future session.

During Pkg precomp however I don't believe there's any reason why it shouldn't be able to precompile other versions of loaded modules. After all, the precompilation takes place in a separate child session.

Along with https://github.com/JuliaLang/julia/pull/40345, this PR changes that behavior:

## Master
```
(jl_YqpHCO) pkg> add CSV
...
  [336ed68f] + CSV v0.8.4
...
15 dependencies successfully precompiled in 10 seconds

julia> using CSV

(jl_YqpHCO) pkg> pin CSV@0.8.2
   Resolving package versions...
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_YqpHCO/Project.toml`
  [336ed68f] ↓ CSV v0.8.3 ⚲ ⇒ v0.8.2 ⚲
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_YqpHCO/Manifest.toml`
  [336ed68f] ↓ CSV v0.8.3 ⚲ ⇒ v0.8.2 ⚲
  [2dfb63ee] ↓ PooledArrays v1.2.1 ⇒ v0.5.3
  [9fa8497b] - Future
  Progress [========================================>]  2/2
  ? CSV
1 dependency successfully precompiled in 1 seconds (13 already precompiled)
1 dependency failed but may be precompilable after restarting julia

### The user would then need to restart julia, remember to go to pkg and call a pkg action that triggered precomp 
### which remembers the ?'s between states, otherwise they'd hit code-load precomp (which I think would be common)
```

## This PR
```
(jl_ffq5WF) pkg> add CSV
...
  [336ed68f] + CSV v0.8.4
...
Precompiling project...
  15 dependencies successfully precompiled in 11 seconds

julia> using CSV

(jl_ffq5WF) pkg> pin CSV@0.8.3
   Resolving package versions...
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_ffq5WF/Project.toml`
  [336ed68f] ↓ CSV v0.8.4 ⇒ v0.8.3 ⚲
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_ffq5WF/Manifest.toml`
  [336ed68f] ↓ CSV v0.8.4 ⇒ v0.8.3 ⚲
Precompiling project...
  ✓ CSV
  1 dependency successfully precompiled in 7 seconds (14 already precompiled)
  1 dependency precompiled but a different version is currently loaded. Restart julia to access the new version

### The user would then restart julia and can immediately use the package without hitting code-load precomp
```

What the end looks like in color, where they show up as yellow check marks that stick around at the end:
image

If I understand the `compilecache` process correctly, I think this would eradicate almost all `?`'s, which remain as a handler for `compilecache` throwing internal errors (other than the actual precompilation errors which don't change in this PR). I don't think I ever saw an internal error thrown (on a stable branch) for another reason, only when we explicitly asked `compilecache` to precompile a package that is either loaded or has loaded deps.

Note: This won't pass CI until https://github.com/JuliaLang/julia/pull/40345 is merged upstream

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1099,6 +1099,7 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
     ansi_disablecursor = "\e[?25l"
     n_done::Int = 0
     n_already_precomp::Int = 0
+    n_loaded::Int = 0
 
     function handle_interrupt(err)
         notify(interrupted_or_done)
@@ -1145,15 +1146,16 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
                         bar.max = n_total - n_already_precomp
                         final_loop || print(iostr, sprint(io -> show_progress(io, bar); context=io), "\n")
                         for dep in pkg_queue_show
+                            loaded = haskey(Base.loaded_modules, dep)
                             name = dep in direct_deps ? dep.name : string(color_string(dep.name, :light_black))
                             if dep in precomperr_deps
                                 print(iostr, color_string("  ? ", Base.warn_color()), name, "\n")
                             elseif haskey(failed_deps, dep)
                                 print(iostr, color_string("  ✗ ", Base.error_color()), name, "\n")
                             elseif was_recompiled[dep]
-                                interrupted_or_done.set && continue
-                                print(iostr, color_string("  ✓ ", :green), name, "\n")
-                                @async begin # keep successful deps visible for short period
+                                !loaded && interrupted_or_done.set && continue
+                                print(iostr, color_string("  ✓ ", loaded ? Base.warn_color() : :green), name, "\n")
+                                loaded || @async begin # keep successful deps visible for short period
                                     sleep(1);
                                     filter!(!isequal(dep), pkg_queue)
                                 end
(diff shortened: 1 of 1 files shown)


IanButterworth:
@KristofferC it might be nice to get this into 1.6.1. @vtjnash okayed the julia PR. 

What do you think?

IanButterworth:
https://github.com/JuliaLang/julia/pull/40345 is merged. I'll re-run CI once nightly updates

IanButterworth:
Seems like a preferred approach. I'll merge and watch out for issues on nightly to assess whether it's 1.6.1-ready

KristofferC:
Since 1.6.1 is about to be released, let's wait with this for like 1.6.2 so it gets some more time on master.

IanButterworth:
Sounds good

IanButterworth:
I haven't seen any issues reported on master, but those ?'s still cause confusion https://discourse.julialang.org/t/dependencies-failed-but-may-be-precompilable-after-restarting-julia-what-are-those-question-marks/60236

So it would be good to get this into 1.6.2

PR #2488, opened 2021-04-05
adjust `activate` feedback and hook up to `io` kwarg

### LINKED ISSUE #1761: "Activating environment at ..." message is confusing to new users
I sent a project to a collaborator who is a new-ish Julia user. This included a `Project.toml` / `Manifest.toml`. They were aware that `Manifest.toml` includes the exact versions of things so that its what is needed to fully recreate an environment. 

When they ran `activate project_folder/` to start setting things up, it prints:

```
Activating environment at project_folder/Project.toml
```

However, this confused them because they thought maybe they were supposed to have "activated `Manifest.toml`" rather than `Project.toml`, since its the former that actually contains the package versions they knew were neeed. Although I hadn't thought of this myself, looking through the eyes of a new user, I actually totally agree this is confusing. 

Is there a reason to explicitly highlight `Project.toml` in that message? I wonder if it might better to just say `Activating environment at project_folder/` to avoid confusion?

### LINKED ISSUE #2036: [FR] Quiet environment setup

00vareladavid:
We weren't passing `io` down in `activate` for all cases: this fixes that.
Also adjust `activate` feedback to suggesting in #1761.

Fix #1761, Close #2036

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1496,10 +1496,10 @@ end
 
 function activate(;temp=false, shared=false, io::IO=DEFAULT_IO[])
     shared && pkgerror("Must give a name for a shared environment")
-    temp && return activate(mktempdir())
+    temp && return activate(mktempdir(); io=io)
     Base.ACTIVE_PROJECT[] = nothing
     p = Base.active_project()
-    p === nothing || printpkgstyle(io, :Activating, "environment at $(pathrepr(p))")
+    p === nothing || printpkgstyle(io, :Activating, "project at $(pathrepr(dirname(p)))")
     add_snapshot_to_undo()
     return nothing
 end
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
Seems good. Should we just merge this?

PR #2493, opened 2021-04-07
Suggest uppercased package name when not found

fonsp:
To help users in situations like https://github.com/fonsp/Pluto.jl/issues/1068

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -857,7 +857,6 @@ function ensure_resolved(manifest::Manifest,
                     join(io, uuids, ", ", " or ")
                     print(io, " in manifest but not in project")
                 end
-                println(io, ")")
                 if !isempty(name) && !isuppercase(first(name))
                     println(io, ") Package names are case-sensitive. Did you mean: \"$(uppercasefirst(name))\"?")
                 else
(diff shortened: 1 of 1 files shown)


fredrikekre:
This seems like an somewhat arbitrary thing to suggest (why not PLUTO for example)? There was a PR to Julia which suggested similar names when you try to load a package with a type (e.g. `using pluto` would suggest `using Pluto`) but I can't find it now. I think something like that would be much better.

KristofferC:
Suggesting similar package names based on what is available in the registry would probably work better.

fonsp:
> This seems like an somewhat arbitrary thing to suggest

Good point! Feel free to close this PR, I'm not very invested in it

fredrikekre:
> Suggesting similar package names based on what is available in the registry would probably work better.

Yea thats what I meant.

StefanKarpinski:
Edit distance with a low cost for case changes would be a good way to do this. Now that we have usage stats we can combine with a "language model" that uses package popularity to decide what's most likely.

IanButterworth:
Converted into a feature request. thanks @fonsp

PR #2504, opened 2021-04-11
Add the `resolver_only` kwarg to `add`, `develop`, `free`, `pin`, and `update`

### LINKED ISSUE #2503: Feature request: update a `Manifest.toml` file without downloading anything or running any user code
I would like this feature for CompatHelper.

CompatHelper used to automatically update `Manifest.toml` files. I removed this feature in https://github.com/JuliaRegistries/CompatHelper.jl/pull/288 because of security concerns. Specifically, CompatHelper has access to secure environment variables (`GITHUB_TOKEN`, `DOCUMENTER_KEY`, and/or `COMPATHELPER_PRIV`). The security concern is this: when you run `Pkg.update()`, it will run the build scripts of packages. A malicious package could add some code to its build script that reads these secure environment variables and submits them to a server maintained by the attacked.

So I need a way to update a `Manifest.toml` file without running any user code.

DilumAluthge:
Fixes #2503 
See also: #1718 

This pull request adds the `resolver_only::Bool` keyword argument to the following five functions:
1. `Pkg.add`
2. `Pkg.develop`
3. `Pkg.free`
4. `Pkg.pin`
5. `Pkg.update`

The default value is `resolver_only = false`.

When `resolver_only` is `true`, we skip the following steps:
1. Downloading the source code
2. Downloading the artifacts
3. Running the package build scripts

You must disable auto-precompilation for this feature to work. If `resolver_only` is `true` and auto-precompilation is enabled, an error will be thrown.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -171,7 +171,8 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
 end
 
 function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
-                 preserve::PreserveLevel=PRESERVE_TIERED, platform::AbstractPlatform=HostPlatform(), kwargs...)
+                 preserve::PreserveLevel=PRESERVE_TIERED, platform::AbstractPlatform=HostPlatform(),
+                 resolver_only::Bool=false, kwargs...)
     require_not_empty(pkgs, :develop)
     Context!(ctx; kwargs...)
 
@@ -210,12 +211,13 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
         end
     end
 
-    Operations.develop(ctx, pkgs, new_git; preserve=preserve, platform=platform)
+    Operations.develop(ctx, pkgs, new_git; preserve, platform, resolver_only)
     return
 end
 
 function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PRESERVE_TIERED,
-             platform::AbstractPlatform=HostPlatform(), kwargs...)
+             platform::AbstractPlatform=HostPlatform(), resolver_only::Bool=false,
+             kwargs...)
     require_not_empty(pkgs, :add)
     Context!(ctx; kwargs...)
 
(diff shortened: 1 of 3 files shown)


DilumAluthge:
@fredrikekre This PR implements your suggestion from [this comment](https://github.com/JuliaLang/Pkg.jl/pull/1718#issuecomment-599226107):
> Why not skip downloading the source code as well? Basically a function or option to just run the resolver and output the `toml`s.

KristofferC:
As a side note, this is pretty much exactly the `preview` functionality that we used to have (except it keeps the writing of the toml files). One issue that `preview` (which this inherits) is what should happen with packages that are e.g. added by branch and tracks a git repo? They need to be downloaded/updated for the resolver to give the actual result the command would give. But this kind of goes against the idea that this should be a "pure" operation somehow.

Also, instead of having the option be called something with "resolver" (which is a kind of internal thing) maybe it is better to name it something like `download::Bool=true` because a user is likely to know what `download=false` mean?

DilumAluthge:
> Also, instead of having the option be called something with "resolver" (which is a kind of internal thing) maybe it is better to name it something like `download::Bool=true` because a user is likely to know what `download=false` mean?

Yeah that sounds good to me!

DilumAluthge:
> As a side note, this is pretty much exactly the `preview` functionality that we used to have (except it keeps the writing of the toml files). One issue that `preview` (which this inherits) is what should happen with packages that are e.g. added by branch and tracks a git repo? They need to be downloaded/updated for the resolver to give the actual result the command would give. But this kind of goes against the idea that this should be a "pure" operation somehow.

Yeah, this is tricky. Personally, I would like this to be a "pure" operation, so I would rather not download any packages.

For the CompatHelper use case, we don't really need to update packages that are being tracked by branch, since the main use case here is that we have just bumped a `[compat]` entry, and we want to update the manifest to use the new version of the dependency.

DilumAluthge:
Would this feature be useful to people outside of the CompatHelper use case? If not, I'm inclined to close it. I think CompatHelper will work just fine without manifest updating.

DilumAluthge:
I'm going to close this for two reasons:
1. Doesn't seem like there is a compelling use case. (I don't think we currently need this for CompatHelper.)
2. I'm not sure how to address this:



> One issue that `preview` (which this inherits) is what should happen with packages that are e.g. added by branch and tracks a git repo? They need to be downloaded/updated for the resolver to give the actual result the command would give. But this kind of goes against the idea that this should be a "pure" operation somehow.

PR #2509, opened 2021-04-13
Fix treehash generation

staticfloat:
We missed a `basename()` when comparing `f` to `".git"`, meaning that
embedded `.git` directories get included.

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -274,7 +274,7 @@ function tree_hash(::Type{HashType}, root::AbstractString; debug_out::Union{IO,N
     entries = Tuple{String, Vector{UInt8}, GitMode}[]
     for f in sort(readdir(root; join=true); by = f -> isdir(f) ? f*"/" : f)
         # Skip `.git` directories
-        if f == ".git"
+        if basename(f) == ".git"
             continue
         end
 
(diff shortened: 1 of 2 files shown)


src/GitTools.jl
```diff
@@ -274,7 +274,7 @@ function tree_hash(::Type{HashType}, root::AbstractString; debug_out::Union{IO,N
     entries = Tuple{String, Vector{UInt8}, GitMode}[]
     for f in sort(readdir(root; join=true); by = f -> isdir(f) ? f*"/" : f)
         # Skip `.git` directories
-        if f == ".git"
+        if basename(f) == ".git"
```
giordano, line 277: I thought `Tar.jl` could have been the culprit since the issue was reported only on nightly but on v1.6, and `Tar.jl` had recent changes in permissions.  Interestingly enough, this line was introduced [two years ago](https://github.com/JuliaLang/Pkg.jl/commit/863132441d4be524053f5580012837389a569bb7), so how the error isn't reproducible on v1.6? :confused:

StefanKarpinski: The original version didn't have this bug. The problem is that the new version passes `join=true` to `readdir` which means that an entry named `.git` doesn't match this check.

KristofferC:
Ok, no need to backport since the buggy commit is not on 1.6.

KristofferC:
I pushed a test for this.

staticfloat:
Fixed the test on Windows; looks good to go on my end.

PR #2510, opened 2021-04-13
Check for stdlib compatibility in manifest

IanButterworth:
This adds a check for stdlibs/deps with/without version numbers in the manifests, to catch incompatible manifests from different julia versions.

Not sure if it's a good approach. Tests don't pass.

First, set up an older manifest in 1.5.3
```
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.5.3 (2020-11-09)
|__/                   |

shell> cd ~/testenv
/Users/ian/testenv

(@v1.5) pkg> activate .
 Activating new environment at `~/testenv/Project.toml`

(testenv) pkg> add LibCURL_jll
  Resolving package versions...
Updating `~/testenv/Project.toml`
  [deac9b47] + LibCURL_jll v7.70.0+2
Updating `~/testenv/Manifest.toml`
  [56f22d72] + Artifacts v1.3.0
  [692b3bcd] + JLLWrappers v1.2.0
  [deac9b47] + LibCURL_jll v7.70.0+2
  [29816b5a] + LibSSH2_jll v1.9.0+3
...
```
Then try to activate in 1.7
```
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.7.0-DEV.909 (2021-04-11)
 _/ |\__'_|_|_|\__'_|  |  Commit 8fc38a4d57 (1 day old master)
|__/                   |

julia> import Pkg

shell> cd ~/testenv
/Users/ian/testenv

(Pkg) pkg> activate .
  Activating environment at `~/testenv/Project.toml`
ERROR: Incompatible manifest.toml from different julia version detected. `LibSSH2_jll=29816b5a-b9ab-546f-933c-edad1886dfa8` 
is a stdlib in julia 1.7.0-DEV.909 but the current manifest lists it with a version number. Delete the manifest file to continue
```

### DIFF
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -113,6 +113,27 @@ function validate_manifest(stage1::Dict{String,Vector{Stage1}})
     for (name, infos) in stage1, info in infos
         manifest[info.uuid] = info.entry
     end
+    # check that stdlibs are missing versions, and non-stdlibs have versions
+    for (entry_uuid, entry) in manifest
+        if is_stdlib(entry_uuid)
+            if entry.version != nothing
+                err = string("Incompatible manifest.toml from different julia version detected. ",
+                        "`$(entry.name)=$(entry_uuid)` is a stdlib in julia $(Base.VERSION) ",
+                        "but the current manifest lists it with a version number. ",
+                        "Unexpected errors may occur. Delete the manifest file and re-resolve to fix the issue.")
+                @warn err
+            end
+        else
+            if entry.version == nothing
+                err = string("Incompatible manifest.toml from different julia version detected. ",
+                        "`$(entry.name)=$(entry_uuid)` is a not a stdlib in julia $(Base.VERSION) ",
+                        "but the current manifest lists it without a version number. ",
+                        "Unexpected errors may occur. Delete the manifest file and re-resolve to fix the issue.")
+                @warn err
+            end
+        end
+    end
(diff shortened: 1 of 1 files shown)


DilumAluthge:
It's nice, but I wonder if it just distracts us from the real issue, which is that we need to store the Julia version in the manifest file.

IanButterworth:
Yeah. I wondered if this could be a stop-gap for 1.6.1, before the manifest format is changed in 1.7?

DilumAluthge:
I would go with a warning rather than an error.

IanButterworth:
Looking at how frequently this warning pops up in the tests, it seems non-stdlib dependencies are allowed to not have version numbers (TIL) so this approach doesn't seem valid

DilumAluthge:
> Looking at how frequently this warning pops up in the tests, it seems non-stdlib dependencies are allowed to not have version numbers (TIL) so this approach doesn't seem valid

Yeah, based on the CI logs, it looks like this warning gives a lot of false positives.

DilumAluthge:
I do think that https://github.com/JuliaLang/Pkg.jl/pull/2538 improves the error message that a user gets when trying to instantiate an incompatible manifest.

PR #2517, opened 2021-04-19
Remove `force_latest_compatible_version` and `allow_earlier_backwards_compatible_versions` from the docstring for `Pkg.test`

DilumAluthge:
Since these kwargs are only for use in the CompatHelper/Dependabot use case, we can just make them nonpublic (private/internal).

And if they are not public, we can remove them from the documentation.


DilumAluthge:
Test failures are unrelated. See #2518

PR #2519, opened 2021-04-19
Run CI on both Julia 1.6 and Julia nightly

DilumAluthge:


DilumAluthge:
Hmm, looks like CI fails on 1.6 both with and without the Pkg server.

PR #2522, opened 2021-04-20
Zero out broken_artifact gitsha

staticfloat:
Let's not allow for false-positive results from the PkgServer.


DilumAluthge (approved):

DilumAluthge:
Closes #2518?

PR #2531, opened 2021-04-27
Add the ability to disable re-resolving during `Pkg.test`

DilumAluthge:
This gives users a way to guarantee that the package versions in their `Manifest.toml` file are exactly the package versions that will be used during `Pkg.test`.

This is not public and therefore is not documented.


KristofferC (approved):

DilumAluthge:
It would be very helpful to have this on the LTS, so I have marked this for backport to 1.6.

If it doesn't backport cleanly, let me know, and I'll make a PR directly to `release-1.6` that adds the `allow_reresolve` kwarg.

PR #2534, opened 2021-04-27
Update CODEOWNERS

DilumAluthge:


KristofferC:
Why does this file have to be updated over and over?

DilumAluthge:
It shouldn't need to be. It was fine after https://github.com/JuliaLang/Pkg.jl/commit/f0586c3d19602334b9428df75ea1bec06c61cc88#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7, and it shouldn't have been modified after that.

But then I accidentally made https://github.com/JuliaLang/Pkg.jl/commit/af63ee0b75d185cf3197350026484dad548326c6#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7 on master, when I meant to make it on a throwaway branch. So I need to undo https://github.com/JuliaLang/Pkg.jl/commit/af63ee0b75d185cf3197350026484dad548326c6#diff-3d36a1bf06148bc6ba1ce2ed3d19de32ea708d955fed212c0d27c536f0bd4da7, and then we'll be all set.

Could we put the CI checks as required status checks? That will prevent accidentally committing directly to master.

PR #2537, opened 2021-04-28
Small documentation update for valid package/project names.

BenjaminGalliot:
I think we should update this (old?) falsely restrictive information.

I realized this while reading the documentation for another detail, knowing that some personal project names are not limited to ASCII (see this [old error](https://github.com/JuliaLang/Pkg.jl/issues/1846)).


StefanKarpinski:
Thanks for the addition — this is definitely good to explicitly state.

PR #2540, opened 2021-04-30
Test suite: change permissions before removing directories

DilumAluthge:
If the registry is read-only, then (at least on my machine) the `rm` may sometimes fail even if you pass `force = true`.


DilumAluthge:
I'm having trouble coming up with a MWE that demonstrates the need for this patch.

PR #2541, opened 2021-04-30
Improve the implementation of `force_latest_compatible_version`

DilumAluthge:
This pull request improves the behavior of the following two keyword arguments to the `Pkg.test` function:
1. `force_latest_compatible_version::Bool`
2. `allow_earlier_backwards_compatible_versions::Bool`

These two keyword arguments are:
1. Not part of the public API
2. Only intended for use by CI services that implement auto-detection of CompatHelper/Dependabot pull requests (see e.g. https://github.com/julia-actions/julia-runtest/pull/20 and https://github.com/JuliaRegistries/CompatHelper.jl/issues/298)

## Before this PR

Here is the current behavior:
1. `Pkg.test` creates the temporary sandbox environment.
2. `Pkg.test` runs the resolver.
3. We iterate over each direct dependency that has a `[compat]` entry and check that the direct dependency is at its latest compatible version. If at least one direct dependency is not at its latest compatible version, we throw an error. (For a direct dependency that does not have a `[compat]` entry, we print a warning and do nothing.)
4. `Pkg.test` runs the `test/runtests.jl` file.

This implementation works pretty well, but it does have some flaws:
1. The error message only tells the user that the dependency is not at its latest compatible version. It does not tell the user **why** `Pkg.test` was not able to use the latest compatible version of the dependency. The user will need to clone the repo and run e.g. `] add Foo@0.2` locally in order to figure out why the dependency is being held back.
2. If the package has checked a `Manifest.toml` file into source control, and if the manifest file uses an older version of the dependency, and if `Pkg.test` is able to use the exact same versions as the manifest file, then `Pkg.test` will use the older version of the dependency, and thus the error will be thrown.

## After this PR

This PR changes the implementation as follows:
1. `Pkg.test` creates the temporary sandbox environment.
2. **We iterate over each direct dependency. For each direct dependency, if the dependency has a `[compat]` entry, we modify the `[compat]` entry so that it only includes the latest registered version of the dependency that is compatible with the original `[compat]` entry. If a direct dependency does not have a `[compat]` entry, we print a warning and do nothing.**
3. `Pkg.test` runs the resolver. **At this point, if the resolver is not able to use the latest compatible version of all dependencies, it throws a `ResolveError`.**
4. **We reset all of the `[compat]` entries back to their original values.**
5. `Pkg.test` runs the `test/runtests.jl` file.

Here is an example to illustrate step 2. Suppose that your package has a direct dependency on `Foo.jl`. And suppose that `Foo` has the following registered versions:
- `0.1.0`
- `0.1.1`
- `0.2.0`
- `0.2.1`

Suppose that currently, your package has a `[compat]` entry that looks like this:
```toml
[compat]
Foo = "0.1"
```

And suppose that a bot (e.g. CompatHelper or Dependabot) opens a PR to change your package's `[compat]` entry to this:
```toml
[compat]
Foo = "0.1, 0.2"
```

During step 2, we (roughly speaking) modify the `[compat]` entry to look like this:
```toml
[compat]
Foo = "0.2"
```

This ensures that the resolver uses `Foo` version `0.2.x`.

This new implementation solves the previously mentioned flaws:
1. Instead of printing an error that only tells you that a dependency is not at its latest version, this implementation will cause the resolver to print an error message that details exactly why the resolver was not able to install the latest compatible version of the dependency.
2. This implementation guarantees that `Pkg.test` will use the latest compatible version of each dependency, even if there is a `Manifest.toml` file that uses an older version of a dependency.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -311,7 +311,9 @@ end
 
 function up(ctx::Context, pkgs::Vector{PackageSpec};
             level::UpgradeLevel=UPLEVEL_MAJOR, mode::PackageMode=PKGMODE_PROJECT,
-            update_registry::Bool=true, kwargs...)
+            update_registry::Bool=true,
+            skip_writing_project::Bool=false,
+            kwargs...)
     Context!(ctx; kwargs...)
     if update_registry
         Registry.download_default_registries(ctx.io)
@@ -328,13 +330,13 @@ function up(ctx::Context, pkgs::Vector{PackageSpec};
         manifest_resolve!(ctx.env.manifest, pkgs)
         ensure_resolved(ctx.env.manifest, pkgs)
     end
-    Operations.up(ctx, pkgs, level)
+    Operations.up(ctx, pkgs, level; skip_writing_project)
     return
 end
 
 resolve(; io::IO=stderr_f(), kwargs...) = resolve(Context(;io); kwargs...)
-function resolve(ctx::Context; kwargs...)
-    up(ctx; level=UPLEVEL_FIXED, mode=PKGMODE_MANIFEST, update_registry=false, kwargs...)
+function resolve(ctx::Context; skip_writing_project::Bool=false, kwargs...)
+    up(ctx; level=UPLEVEL_FIXED, mode=PKGMODE_MANIFEST, update_registry=false, skip_writing_project, kwargs...)
     return nothing
 end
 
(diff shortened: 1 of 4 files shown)


DilumAluthge:
BTW, if possible, I'd like to get this into 1.7.

DilumAluthge:
Bump @KristofferC @00vareladavid

DilumAluthge:
The 1.7 feature freeze is on June 1, so I'll merge this at the end of May.

So if there are any changes to make, it would be great to get those changes done during the next two weeks.

DilumAluthge:
I'm going to merge this now.

Feel free to continue reviewing this PR. If there end up being changes that we need to make, we still have several days left until June 1, so we have time to make those changes and merge them before the feature freeze.

PR #2543, opened 2021-04-30
Use the current std(out|err) for printing, fixes #2542.

### LINKED ISSUE #2542: after upgrade to 1.6.1: lost Pkg prompt on Pkg.update() and Pkg.add("...")
After upgrading from 1.6.0 to 1.6.1 I lost the Pkg prompt on Pkg.update() and Pkg.add("...") in my Jupyter notebook. 

Is this intentional or an issue? 

The code runs fast and I believe correct (thanks!) but lacks the prompt reporting updates. So, I can do my work but wanted to report my issue... I searched for a key to switch the prompt on but could not find it.

fredrikekre:


IanButterworth:
@fredrikekre hope you don't mind the push

IanButterworth:
It seems this won't be the full fix for #2542
IJulia needs to set `DEFAULT_IO[]` correctly or to `nothing`
Screen Shot 2021-05-02 at 1 27 17 PM

KristofferC:
I think it is us that need to fix things. Is the problem that stdout gets redirected? Perhaps it should be `default_io()` then which returns stdout unless a custom io is set?

KristofferC:
I think it is on us to fix it :p

IanButterworth:
Yeah, that sounds good. But how can we detect if a custom io is set?
https://github.com/JuliaLang/IJulia.jl/blob/cc2a9bf61a2515596b177339f9a3514de8c38573/src/init.jl#L101-L108

IanButterworth:
I just assumed `DEFAULT_IO` existed for packages like IJulia to override if stdio was redirected? If not, shouldn't everything just point to `stdout/stderr`?

KristofferC:
I think it is just an internal thing to redirect the test output globally. 

When does `DEFAULT_IO` get set in tha I Julia session?

IanButterworth:
> When does DEFAULT_IO get set in tha I Julia session?

I don't think it does. I couldn't find a reference in the package

fredrikekre:
It is just for silencing Pkg output in tests I think.

KristofferC:
But why is it not nothing in the IJulia session then?

fredrikekre:
> But why is it not nothing in the IJulia session then?

It is.

IanButterworth:
It isn't for me (MacOS, 1.6.1, IJulia v1.23.2)
Screen Shot 2021-05-02 at 2 57 37 PM

KristofferC:
Some garbage left from the build process? Perhaps it should be set to `nothing` in `__init__` then?

IanButterworth:
It's already set to `DEFAULT_IO[] = stderr` in Pkg's `__init__` and IJulia doesn't touch it from what I can tell, so I don't understand how it's `nothing` on @fredrikekre's system

fredrikekre:
Have you rebuilt the sysimage or changed the IJulia source code? Note that IJulia won't load your Pkg with changed UUID

KristofferC:
> It's already set to DEFAULT_IO[] = stderr in Pkg's __init__

Okay, but we could change it in this PR to set it to `nothing` maybe?

fredrikekre:
https://github.com/JuliaLang/Pkg.jl/pull/2543/files#diff-1347ee05134e00dcfa1afc9dd8b6fdad9be2afeceee98662a2b76b6d77eddf53R36

IanButterworth:
That's it. Sorry for the noise

KristofferC:
> https://github.com/JuliaLang/Pkg.jl/pull/2543/files#diff-1347ee05134e00dcfa1afc9dd8b6fdad9be2afeceee98662a2b76b6d77eddf53R36

Yeah, but was thinking if something sets it during the Julia build process, then we will have the wrong value. But that probably won't happen.

crstnbr:
Will this be fixed in a Julia 1.6.2 or only 1.7? In the latter case it might be worth fixing this temporarily in IJulia.jl, IMHO.

KristofferC:
It has the backport label so will be in 1.6.2 which hopefully should be out in not too long time.

PR #2544, opened 2021-05-02
add `@debug` logs for artifact download failures

IanButterworth:
I don't think there's a good way currently to identify which errors the artifact downloads should report/throw, so this is a stop-gap to allow debugging errors in the wild.


staticfloat:
Can you manually trigger an error (like downloading an artifact that doesn't exist, or that hashes incorrectly or something) to ensure it looks correct?

IanButterworth:
This is with my internet off

image

IanButterworth:
Just renamed `e` -> `err` to make it a bit clearer

staticfloat:
Looks great, Ian! :D

KristofferC:
Would be good to squash PRs like this at merge time.

PR #2547, opened 2021-05-03
Fix module reference typo in #2307

IanButterworth:
This typo slipped into #2307 :/ 

It would be nice to be able to test  more of the prompt by passing `y/n` into stdin, but I don't know how to do that in CI.


IanButterworth:
Turns out you can preload the `stdin.buffer` before loading the prompt to simulate input. Tests added!

PR #2556, opened 2021-05-08
Fix two `Base.hash` methods and add tests

IanButterworth:
Fixes what must be unused hash methods and adds explicit tests

```
julia> hash(Pkg.Types.Project())
ERROR: UndefVarError: t not defined
Stacktrace:
 [1] #16
   @ ./none:0 [inlined]
 [2] iterate
   @ ./generator.jl:47 [inlined]
 [3] collect(itr::Base.Generator{NTuple{9, Symbol}, Pkg.Types.var"#16#17"})
   @ Base ./array.jl:678
 [4] hash(x::Pkg.Types.Project, h::UInt64)
   @ Pkg.Types /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Types.jl:201
 [5] hash(x::Pkg.Types.Project)
   @ Base ./hashing.jl:18
 [6] top-level scope
   @ REPL[3]:1


julia> hash(Pkg.Types.VersionBound())
ERROR: MethodError: no method matching hash(::UInt64, ::Int64)
Closest candidates are:
  hash(::UInt64, ::UInt64) at hashing.jl:72
  hash(::Real, ::UInt64) at float.jl:489
  hash(::Any) at hashing.jl:18
  ...
Stacktrace:
 [1] hash(r::Pkg.Types.VersionBound, h::UInt64)
   @ Pkg.Types /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Pkg/src/versions.jl:88
 [2] hash(x::Pkg.Types.VersionBound)
   @ Base ./hashing.jl:18
 [3] top-level scope
   @ REPL[2]:1
```

### DIFF
--- a/src/Versions.jl
+++ b/src/Versions.jl
@@ -86,7 +86,7 @@ function isjoinable(up::VersionBound, lo::VersionBound)
     return true
 end
 
-Base.hash(r::VersionBound, h::UInt) = hash(hash(r.t, h), convert(UInt64, r.n))
+Base.hash(r::VersionBound, h::UInt) = hash(r.t, hash(r.n, h))
 
 # Hot code
 function VersionBound(s::AbstractString)
--- a/src/Versions.jl
+++ b/src/Versions.jl
@@ -86,7 +86,7 @@ function isjoinable(up::VersionBound, lo::VersionBound)
     return true
 end
 
-Base.hash(r::VersionBound, h::UInt) = hash(reinterpret(UInt8, [hash(r.t, h)]), convert(UInt64, r.n))
+Base.hash(r::VersionBound, h::UInt) = hash(hash(r.t, h), convert(UInt64, r.n))
 
 # Hot code
 function VersionBound(s::AbstractString)
--- a/src/Versions.jl
+++ b/src/Versions.jl
@@ -86,7 +86,7 @@ function isjoinable(up::VersionBound, lo::VersionBound)
     return true
 end
 
-Base.hash(r::VersionBound, h::UInt) = hash(hash(r.t, h), r.n)
+Base.hash(r::VersionBound, h::UInt) = hash(reinterpret(UInt8, [hash(r.t, h)]), convert(UInt64, r.n))
 
 # Hot code
 function VersionBound(s::AbstractString)
--- a/test/misc.jl
+++ b/test/misc.jl
@@ -12,8 +12,7 @@ end
     @test hash(Pkg.Types.Project()) == 0xd4b8cda960e4de88
     @test hash(Pkg.Types.VersionBound()) == 0xdc46ceef40d7ff7f
     @test hash(Pkg.Resolve.Fixed(VersionNumber(0,1,0))) == 0xea3ee20b2c9d6ccb
-    @test hash(Pkg.Types.VersionSpec()) == 0x4245018905bae555
 
-    # hash isn't stable because the internal `repo` field is a mutable struct
-    hash(Pkg.Types.PackageEntry())
+    hash(Pkg.Types.VersionSpec()) # hash isn't stable
+    hash(Pkg.Types.PackageEntry()) # hash isn't stable because the internal `repo` field is a mutable struct
 end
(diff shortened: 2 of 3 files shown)


test/misc.jl
```diff
@@ -7,3 +7,13 @@ using Pkg
...
 end
+
+@testset "hashing" begin
+    @test hash(Pkg.Types.Project()) == 0xd4b8cda960e4de88
+    @test hash(Pkg.Types.VersionBound()) == 0x873f51b414a315c2
+    @test hash(Pkg.Resolve.Fixed(VersionNumber(0,1,0))) == 0xea3ee20b2c9d6ccb
+
+    # hashes for these empty objects aren't stable
+    hash(Pkg.Types.VersionSpec())
+    hash(Pkg.Types.PackageEntry())
```
IanButterworth, line 18: I don't know why, but these hash test results aren't stable

DilumAluthge: Can you elaborate on what you mean by this?

IanButterworth: They don't give the same result each time

DilumAluthge: @StefanKarpinski Any idea why this might be?

KristofferC: > Huh. That seems.... bad?

Why, there is nothing saying that hashes need to be stable over julia sessions.

fredrikekre: They are `mutable struct`s:
```
julia> struct X x end

julia> mutable struct Y y end

julia> hash(X(1))
0x2211c0ec74129abf

julia> hash(X(1))
0x2211c0ec74129abf

julia> hash(Y(1))
0xeb1f48179582169f

julia> hash(Y(1))
0x006bb8c4d822c6fb
```

DilumAluthge:
I believe we want to use them for the `project_hash` field in the new manifest format.

DilumAluthge:
Would it be easier to `TOML.print` the project file to a string, and then take the hash of that?

DilumAluthge:
Whatever we use for the `project_hash field` in the new manifest format needs to be stable across Julia sessions as long as you are using the exact same commit of Julia.

IanButterworth:
> I believe we want to use them for the project_hash field in the new manifest format.

Correct.

> Would it be easier to TOML.print the project file to a string, and then take the hash of that?

I don't see why. These hash functions are simple

> Whatever we use for the project_hash field in the new manifest format needs to be stable across Julia sessions as long as you are using the exact same commit of Julia.

Correct. `hash(Pkg.Types.Project())` is stable

src/Versions.jl
```diff
@@ -89,7 +89,7 @@ function isjoinable(up::VersionBound, lo::VersionBound)
     return true
 end
 
-Base.hash(r::VersionBound, h::UInt) = hash(hash(r.t, h), r.n)
+Base.hash(r::VersionBound, h::UInt) = hash(hash(r.t, h), convert(UInt64, r.n))
```
KristofferC, line 92: Why convert to `UInt64`?

test/misc.jl
```diff
@@ -7,3 +7,12 @@ using Pkg
     f() = Pkg.Types.UNREGISTERED_STDLIBS
     @inferred f()
 end
+
+@testset "hashing" begin
+    @test hash(Pkg.Types.Project()) == 0xd4b8cda960e4de88
```
KristofferC, line 12: I don't think these should be added. It doesn't make sense to compare a hash value like this imo. You could try `hash(Pkg.Types.Project()) == hash(Pkg.Types.Project())` or something?

src/Versions.jl
IanButterworth: Just to avoid this. There may be a more correct solution
```
julia> hash(Pkg.Types.VersionBound())
ERROR: MethodError: no method matching hash(::UInt64, ::Int64)
Closest candidates are:
  hash(::UInt64, ::UInt64) at hashing.jl:72
  hash(::Real, ::UInt64) at float.jl:489
  hash(::Any) at hashing.jl:18
  ...
Stacktrace:
 [1] hash(r::Pkg.Types.VersionBound, h::UInt64)
   @ Pkg.Types /Users/julia/buildbot/worker/package_macos64/build/usr/share/julia/stdlib/v1.6/Pkg/src/versions.jl:88
 [2] hash(x::Pkg.Types.VersionBound)
   @ Base ./hashing.jl:18
 [3] top-level scope
   @ REPL[2]:1
```

KristofferC: Ah, I think the correct way is to write it as:

```jl
hash(r.t, hash(r.n, h))
```

PR #2557, opened 2021-05-09
WIP: Implement new manifest format with top-level metadata

IanButterworth:
A WIP go at implementing a new manifest format that:
- Adds `julia_version` and `project_hash` metadata fields
- Moves the packages from top-level to a `deps` field

Dependent on:
- https://github.com/JuliaLang/julia/pull/40765 - CI has been limited to linux 64, and that PR's binary is used (ignore appveyor)
- If the project_hash is going to be implemented, the hash method needs fixing https://github.com/JuliaLang/Pkg.jl/pull/2556

Example manifest:
```

julia_version = "1.7.0-DEV.1077"
project_hash = "nothing"

[[deps.Base64]]
uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"

[[deps.Distributed]]
deps = ["LinearAlgebra", "Random", "Serialization", "Sockets"]
uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
```

Notes:
- Backward compatible with the old manifest format. When an old manifest format is detected, the `julia_version` is set to `nothing`
- Needs new format manifest tests added
- The hash checking isn't implemented properly yet. Needs some thought about when the project is hashed and saved to the manifest. Also if/when the user should be warned
- Should version warnings be just for major+minor julia version mismatches?
- Tests passing - Ignore appveyor

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -20,13 +20,13 @@ jobs:
       matrix:
         os:
           - ubuntu-latest
-          - macOS-latest
-          - windows-latest
+          # - macOS-latest
+          # - windows-latest
         julia-arch:
           - 'x64'
-          - 'x86'
+          # - 'x86'
         pkg-server:
-          - ""
+          # - ""
           - "pkg.julialang.org"
         julia-version:
           # - '1.6'
@@ -41,9 +41,10 @@ jobs:
           git config --global core.autocrlf false
           git config --global core.eol lf
       - uses: actions/checkout@v2
-      - uses: julia-actions/setup-julia@latest
+      - name: Install Julia from URL
+        uses: julia-actions/install-julia-from-url@main
         with:
-          version: ${{ matrix.julia-version }}
+          url: https://s3.amazonaws.com/julialangnightlies/assert_pretesting/linux/x64/1.7/julia-88274cb9c2-linux64.tar.gz
       - name: Fix TEMP on windows
         if: matrix.os == 'windows-latest'
         run: |
@@ -67,10 +68,10 @@ jobs:
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v1.0.0
-      - uses: julia-actions/setup-julia@latest
+      - name: Install Julia from URL
+        uses: julia-actions/install-julia-from-url@main
         with:
-          # version: '1.6'
-          version: 'nightly'
+          url: https://s3.amazonaws.com/julialangnightlies/assert_pretesting/linux/x64/1.7/julia-88274cb9c2-linux64.tar.gz
       - name: Generate docs
         run: |
           julia --color=yes -e 'write("Project.toml", replace(read("Project.toml", String), r"uuid = .*?\n" =>"uuid = \"54cfe95a-1eb2-52ea-b672-e2afdf69b78f\"\n"));'
--- a/src/API.jl
+++ b/src/API.jl
@@ -302,7 +302,7 @@ function append_all_pkgs!(pkgs, ctx, mode)
         end
     end
     if mode == PKGMODE_MANIFEST || mode == PKGMODE_COMBINED
-        for (uuid, entry) in ctx.env.manifest
+        for (uuid, entry) in ctx.env.manifest.deps
             push!(pkgs, PackageSpec(name=entry.name, uuid=uuid))
         end
     end
@@ -594,7 +594,7 @@ function gc(ctx::Context=Context(); collect_delay::Period=Day(7), verbose=false,
 
         # Collect the locations of every package referred to in this manifest
         pkg_dir(uuid, entry) = Operations.find_installed(entry.name, uuid, entry.tree_hash)
-        return [pkg_dir(u, e) for (u, e) in manifest if e.tree_hash !== nothing]
+        return [pkg_dir(u, e) for (u, e) in manifest.deps if e.tree_hash !== nothing]
     end
 
     # TODO: Merge with function above to not read manifest twice?
@@ -608,7 +608,7 @@ function gc(ctx::Context=Context(); collect_delay::Period=Day(7), verbose=false,
         end
 
         # Collect the locations of every repo referred to in this manifest
-        return [Types.add_repo_cache_path(e.repo.source) for (u, e) in manifest if e.repo.source !== nothing]
+        return [Types.add_repo_cache_path(e.repo.source) for (u, e) in manifest.deps if e.repo.source !== nothing]
     end
 
     function process_artifacts_toml(path, packages_to_delete)
@@ -973,7 +973,7 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
         if ctx.env.pkg !== nothing
             push!(pkgs, ctx.env.pkg)
         else
-            for (uuid, entry) in ctx.env.manifest
+            for (uuid, entry) in ctx.env.manifest.deps
                 push!(pkgs, PackageSpec(entry.name, uuid))
             end
         end
@@ -992,8 +992,8 @@ function _is_stale(paths::Vector{String}, sourcepath::String)
     return true
 end
 
-function make_pkgspec(man, uuid)
-    pkgent = man[uuid]
+function make_pkgspec(man::Manifest, uuid)
+    pkgent = man.deps[uuid]
     # If we have an unusual situation such as an un-versioned package (like an stdlib that
     # is being overridden) its `version` may be `nothing`.
     pkgver = something(pkgent.version, VersionSpec())
@@ -1024,8 +1024,8 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
         for (name, uuid) in ctx.env.project.deps if !Base.in_sysimage(Base.PkgId(uuid, name))
     ]
 
-    man = Pkg.Types.read_manifest(ctx.env.manifest_file)
-    deps_pair_or_nothing = Iterators.map(man) do dep
+    man = ctx.env.manifest
+    deps_pair_or_nothing = Iterators.map(man.deps) do dep
         pkg = Base.PkgId(first(dep), last(dep).name)
         Base.in_sysimage(pkg) && return nothing
         deps = [Base.PkgId(last(x), first(x)) for x in last(dep).deps]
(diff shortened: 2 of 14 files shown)


DilumAluthge:
It looks like the structure in this PR is this:
- top-level
    - `julia_version`
    - `project_hash`
    - `deps`
        - `dependency_1`
        - `dependency_2`
        - `dependency_3`

How about doing this instead?
- top-level
    - `metadata`
        - `julia_version`
        - `project_hash`
    - `deps`
        - `dependency_1`
        - `dependency_2`
        - `dependency_3`

That way, if we add more metadata in the future, we don't need to add them as top-level keys.

fredrikekre:
Why not have a toplevel key that can never be a name of a Julia package? For example `manifest-metadata` or `.manifest`? That would not require any changes to code loading, and would not be breaking (I think?).

Rada8732 (approved):

DilumAluthge:
> Why not have a toplevel key that can never be a name of a Julia package? For example `manifest-metadata` or `.metadata`? That would not require any changes to code loading, and would not be breaking (I think?).

I think that this would still be breaking.

For example, this manifest instantiates just fine on Julia master:
```toml

[[Example]]
git-tree-sha1 = "46e44e869b4d90b96bd8ed1fdcf32244fddfb6cc"
uuid = "7876af07-990d-54b4-ab0e-23690620f79a"
version = "0.5.3"
```

However, if I take this manifest...
```toml
[manifest-metadata]
julia_version = "123"

[[Example]]
git-tree-sha1 = "46e44e869b4d90b96bd8ed1fdcf32244fddfb6cc"
uuid = "7876af07-990d-54b4-ab0e-23690620f79a"
version = "0.5.3"
```

... and I try to instantiate it, I get this error:
```julia
julia> Pkg.instantiate()
┌ Error: Could not parse entry for `manifest-metadata`
└ @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:153
ERROR: MethodError: no method matching get(::Pair{String, Any}, ::String, ::Nothing)
Closest candidates are:
  get(::IdDict{K, V}, ::Any, ::Any) where {K, V} at iddict.jl:101
  get(::IOContext, ::Any, ::Any) at show.jl:343
  get(::Dict{K, V}, ::Any, ::Any) where {K, V} at dict.jl:504
  ...
Stacktrace:
 [1] Dict{Base.UUID, Pkg.Types.PackageEntry}(raw::Dict{String, Any})
   @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:141
 [2] read_manifest(f_or_io::String)
   @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:177
 [3] Pkg.Types.EnvCache(env::Nothing)
   @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:306
 [4] EnvCache
   @ ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:285 [inlined]
 [5] Pkg.Types.Context()
   @ Pkg.Types ./util.jl:471
 [6] #instantiate#278
   @ ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/API.jl:1399 [inlined]
 [7] instantiate()
   @ Pkg.API ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/API.jl:1399
 [8] top-level scope
   @ REPL[2]:1
```

Similarly, if I take this manifest...
```toml
[".metadata"]
julia_version = "123"

[[Example]]
git-tree-sha1 = "46e44e869b4d90b96bd8ed1fdcf32244fddfb6cc"
uuid = "7876af07-990d-54b4-ab0e-23690620f79a"
version = "0.5.3"
```

... and I try to instantiate it, I get a similar error:
```julia
julia> Pkg.instantiate()
┌ Error: Could not parse entry for `.metadata`
└ @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:153
ERROR: MethodError: no method matching get(::Pair{String, Any}, ::String, ::Nothing)
Closest candidates are:
  get(::IdDict{K, V}, ::Any, ::Any) where {K, V} at iddict.jl:101
  get(::IOContext, ::Any, ::Any) at show.jl:343
  get(::Dict{K, V}, ::Any, ::Any) where {K, V} at dict.jl:504
  ...
Stacktrace:
 [1] Dict{Base.UUID, Pkg.Types.PackageEntry}(raw::Dict{String, Any})
   @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:141
 [2] read_manifest(f_or_io::String)
   @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:177
 [3] Pkg.Types.EnvCache(env::Nothing)
   @ Pkg.Types ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:306
 [4] EnvCache
   @ ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:285 [inlined]
 [5] Pkg.Types.Context()
   @ Pkg.Types ./util.jl:471
 [6] #instantiate#278
   @ ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/API.jl:1399 [inlined]
 [7] instantiate()
   @ Pkg.API ~/dev/repos/JuliaLang/julia/usr/share/julia/stdlib/v1.7/Pkg/src/API.jl:1399
 [8] top-level scope
   @ REPL[2]:1
```

I am pretty sure that all three of those manifests are valid TOML. All three parse just fine using the `TOML` stdlib, and all three validate when pasted into an online TOML validator. But the second and third manifests throw errors when you try to `Pkg.instantiate()` them on Julia master.

DilumAluthge:
I agree that we should redo the format in a breaking way. I.e. there is no way to keep backwards-compatibility.

But I'm still having trouble understanding why `julia_version` and `project_hash` are top-level keys. Wouldn't it be a lot cleaner if the only top-level keys are `deps` and `metadata`? Then `julia_version` and `project_hash would both go under `metadata`. And it leaves room in the future to add other pieces of information under `metadata`.

Is there a specific reason why `julia_version` and `project_hash` have to be top-level keys?

DilumAluthge:
E.g. this is what I'm suggesting:

```toml

[metadata]
julia_version = "1.7.0-DEV.1077"
project_hash = "nothing"

[[deps.Base64]]
uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"

[[deps.Distributed]]
deps = ["LinearAlgebra", "Random", "Serialization", "Sockets"]
uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
```

To me personally, it looks a lot cleaner to just have two top-level keys (`metadata` and `deps`).

DilumAluthge:
If we really want to be concise, we can make the top-level keys `deps` and `meta`, for example:

```toml

[meta]
julia_version = "1.7.0-DEV.1077"
project_hash = "nothing"

[[deps.Base64]]
uuid = "2a0f44e3-6c83-55bd-87e4-b1978d98bd5f"

[[deps.Distributed]]
deps = ["LinearAlgebra", "Random", "Serialization", "Sockets"]
uuid = "8ba89e20-285c-5b6f-9357-94700520ee1b"
```

DilumAluthge:
> I don't see a reason to add an extra "metadata" key.

Presumably, in the future, we want to be able to add more metadata, right? Won't it get messy if each new piece of metadata is a top-level key?

DilumAluthge:
> If this is the format we go with, we should probably backport the support for reading it to 1.6 (which is another reason for trying to keep it as minimal as possible).

Wouldn't having a top-level `metadata` key make it easier to backport support? I.e. if there are only two top-level keys (`deps` and `metadata`), when we backport support to 1.6, we can just tell 1.6 to completely ignore the `metadata` key, and just read the dependencies out of `deps`.

But in contrast, if we have each piece of metadata as a top-level key, then when we backport support to 1.6, we need to teach 1.6 about each individual top-level key and how to handle it.

Whereas if all metadata is under `metadata`, just tell earlier Julia versions to ignore the single top-level `metadata` key.

DilumAluthge:
Also, what if in the future we want to add information to the manifest that is neither dependencies nor metadata? E.g. maybe (just for example, I'm not saying we should actually do this) we start putting information in the manifest about the specific versions of artifacts that were downloaded. That information would not go into the `deps` key. But also, that's not really metadata. It would better fit into its own `artifacts` top-level key.

If we have separate `deps` and `metadata` keys, it leave the door open in the future to add other top-level keys that represent meaningful groupings of data that are neither dependencies nor metadata.

But if we don't have a top-level `metadata` key, then we are basically saying that every top-level key that is not `deps` must be interpreted as metadata, and we make it harder for ourselves to make future non-breaking additions to the manifest format.

IanButterworth:
@KristofferC 
> I think the first PR should only change things to have deps. section and not do anything extra (not add any extra keys, no warnings etc).
If this is the format we go with, we should probably backport the support for reading it to 1.6 (which is another reason for trying to keep it as minimal as possible).

I tried to keep it minimal in https://github.com/JuliaLang/Pkg.jl/pull/2561. See what you think

davidanthoff:
This is great!

Has there been some discussion about the behavior if one tries to load a project with a non-matching Julia version? I'm curios because the stuff I've been doing over at https://github.com/JuliaLang/juliaup would make it relatively trivial to have a situation where the launcher from `juliaup` reads the manifest and then starts the exact (or not) Julia version that was specified in the manifest, see [here](https://github.com/JuliaLang/juliaup/issues/10). But at the moment it is unclear to me what a good design in that space would actually be. The simple solution could be to just always use exactly the Julia version that is recorded in the manifest. But maybe it would actually make more sense to specify a compat bound in `Project.toml` and then decide which Julia version to use based on that? Or something else?

I think that discussion is potentially orthogonal to the more narrow design of this PR, but at the same time clearly related. If you feel this would derail this PR, we could also discuss my question over at https://github.com/JuliaLang/juliaup/issues/10.

IanButterworth:
This PR became a bit more of a playground, but given the general discussion is here, here's an update..

The actual PRs are:

https://github.com/JuliaLang/Pkg.jl/pull/2561

https://github.com/JuliaLang/Pkg.jl/pull/2580

PR #2558, opened 2021-05-09
Docstrings: add references from `Pkg.add` to `Pkg.develop` and vice versa

DilumAluthge:
These two docstrings are right next to each other in the manual. However, if you are just looking at the `Pkg.add` docstring in the REPL (e.g. `?Pkg.add`), you won't see any mention of `Pkg.develop`. This PR fixes that.

The motivation is as follows: sometimes, users will try to `Pkg.add` a path that is not a Git repo. This throws an error. The solution is to instead `Pkg.develop` the desired path. However, currently, the `Pkg.add` docstring does not mention `Pkg.develop`.


GunnarFarneback:
It could also be a good idea to let the error message from `Pkg.add` suggest `Pkg.develop` in case the path is not a git repository but does have a `Project.toml`.

PR #2559, opened 2021-05-10
If the user tries to `Pkg.add(; path = ...)` a path that is not a Git repo, suggest `Pkg.develop` instead

DilumAluthge:


src/Types.jl
```diff
@@ -624,7 +624,10 @@ function handle_repo_add!(ctx::Context, pkg::PackageSpec)
     if !isurl(pkg.repo.source)
         if isdir(pkg.repo.source)
             if !isdir(joinpath(pkg.repo.source, ".git"))
-                pkgerror("Did not find a git repository at `$(pkg.repo.source)`")
+                has_project_file = isfile(joinpath(pkg.repo.source, "Project.toml")) || isfile(joinpath(pkg.repo.source, "JuliaProject.toml"))
+                pkg_develop_suggestion = has_project_file ? ". You may consider using `Pkg.develop` instead" : ""
+                msg = "Did not find a git repository at `$(pkg.repo.source)`$(pkg_develop_suggestion)"
+                pkgerror(msg)
```
fredrikekre, line 630:
```suggestion
                msg = "Did not find a git repository at `$(pkg.repo.source)`"
                if !(isfile(joinpath(pkg.repo.source, "Project.toml")) || isfile(joinpath(pkg.repo.source, "JuliaProject.toml")))
                    msg *= ", perhaps you meant `Pkg.develop`?"
                end
                pkgerror(msg)
```

fredrikekre (commented):
Could use a test

DilumAluthge:
> Could use a test

Done!

DilumAluthge:
@fredrikekre This is the current behavior in the PR:
1. If you have a Project.toml file, we'll print the `Pkg.develop` suggestion.
2. If you don't have a project file, we won't print the suggestion.

Does that look correct?

PR #2561, opened 2021-05-12
New Manifest.toml format: Pt 1. Teach Pkg new format, but default to old format

IanButterworth:
Initial discussion in #2557 

This is Part 1, which is intended to be backported to 1.6.2
- Teach Pkg new manifest format with `julia_version` and `manifest_format` metadata fields, and package deps nested under a `deps` field
- Still defaults to old format
- Maintains format of any existing Manifest.toml
- Warn if unknown `manifest_format` version is detected

Part 2: https://github.com/JuliaLang/Pkg.jl/pull/2580

- [x] Dependent on JuliaLang/julia#40765 - ~~GitHub actions CI has been limited to linux 64 and that PR's binary is used (ignore appveyor)~~
- [x] Needs explicit tests for old/new/unknown manifest formats
- [ ] Needs tests for invalid manifest formats

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -250,6 +250,7 @@ Base.@kwdef mutable struct Manifest
     julia_version::Union{Nothing,VersionNumber} = Base.VERSION
     manifest_format::VersionNumber = v"1.0.0" # default to older flat format
     deps::Dict{UUID,PackageEntry} = Dict{UUID,PackageEntry}()
+    other::Dict{String,Any} = Dict{String,Any}()
 end
 Base.:(==)(t1::Manifest, t2::Manifest) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Manifest))
 Base.hash(m::Manifest, h::UInt) = foldr(hash, [getfield(m, x) for x in fieldnames(Manifest)], init=h)
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -245,7 +245,24 @@ Base.:(==)(t1::PackageEntry, t2::PackageEntry) = t1.name == t2.name &&
     t1.uuid == t2.uuid
     # omits `other`
 Base.hash(x::PackageEntry, h::UInt) = foldr(hash, [x.name, x.version, x.path, x.pinned, x.repo, x.tree_hash, x.deps, x.uuid], init=h)  # omits `other`
-const Manifest = Dict{UUID,PackageEntry}
+
+Base.@kwdef mutable struct Manifest
+    julia_version::Union{Nothing,VersionNumber} = Base.VERSION
+    manifest_format::VersionNumber = v"2.0.0"
+    deps::Dict{UUID,PackageEntry} = Dict{UUID,PackageEntry}()
+end
+Base.:(==)(t1::Manifest, t2::Manifest) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Manifest))
+Base.hash(m::Manifest, h::UInt) = foldr(hash, [getfield(m, x) for x in fieldnames(Manifest)], init=h)
+Base.getindex(m::Manifest, i_or_key) = getindex(m.deps, i_or_key)
+Base.get(m::Manifest, key, default) = get(m.deps, key, default)
+Base.setindex!(m::Manifest, i_or_key, value) = setindex!(m.deps, i_or_key, value)
+Base.iterate(m::Manifest) = iterate(m.deps)
+Base.iterate(m::Manifest, i::Int) = iterate(m.deps, i)
+Base.length(m::Manifest) = length(m.deps)
+Base.empty!(m::Manifest) = empty!(m.deps)
+Base.values(m::Manifest) = values(m.deps)
+Base.keys(m::Manifest) = keys(m.deps)
+Base.haskey(m::Manifest, key) = haskey(m.deps, key)
 
 function Base.show(io::IO, pkg::PackageEntry)
     f = []
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -248,7 +248,7 @@ Base.hash(x::PackageEntry, h::UInt) = foldr(hash, [x.name, x.version, x.path, x.
 
 Base.@kwdef mutable struct Manifest
     julia_version::Union{Nothing,VersionNumber} = Base.VERSION
-    manifest_format::VersionNumber = v"2.0.0"
+    manifest_format::VersionNumber = v"1.0.0" # default to older flat format
     deps::Dict{UUID,PackageEntry} = Dict{UUID,PackageEntry}()
 end
 Base.:(==)(t1::Manifest, t2::Manifest) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Manifest))
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -137,7 +137,7 @@ function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
         if f_or_io isa IO
             @warn "Unknown Manifest.toml format version detected in streamed manifest. Unexpected behavior may occur" manifest_format maxlog = 1
         else
-            @warn "Unknown Manifest.toml format version detected in file `$(path)`. Unexpected behavior may occur" manifest_format maxlog = 1
+            @warn "Unknown Manifest.toml format version detected in file `$(f_or_io)`. Unexpected behavior may occur" manifest_format maxlog = 1
         end
     end
     stage1 = Dict{String,Vector{Stage1}}()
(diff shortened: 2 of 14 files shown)


src/manifest.jl
```diff
@@ -127,40 +127,46 @@ function validate_manifest(stage1::Dict{String,Vector{Stage1}})
...
-            entry.repo.rev    = read_field("repo-rev",      nothing, info, identity)
-            entry.repo.subdir = read_field("repo-subdir",   nothing, info, identity)
-            entry.tree_hash   = read_field("git-tree-sha1", nothing, info, safe_SHA1)
-            entry.uuid        = uuid
-            deps = read_deps(get(info::Dict, "deps", nothing))
-        catch
-            # TODO: Should probably not unconditionally log something
-            @error "Could not parse entry for `$name`"
-            rethrow()
+    if haskey(raw, "deps") # deps field doesn't exist if there are no deps
```
fredrikekre, line 135: Would it make sense to version the format instead of trying to use auto-detection like this?

IanButterworth: Indeed, but I took @KristofferC's suggestion literally; that the only thing done here is to move the packages to a deps field.
This minimal change could also add the `julia_version` field too

src/Types.jl
```diff
@@ -245,7 +245,23 @@ Base.:(==)(t1::PackageEntry, t2::PackageEntry) = t1.name == t2.name &&
     t1.uuid == t2.uuid
     # omits `other`
 Base.hash(x::PackageEntry, h::UInt) = foldr(hash, [x.name, x.version, x.path, x.pinned, x.repo, x.tree_hash, x.deps, x.uuid], init=h)  # omits `other`
-const Manifest = Dict{UUID,PackageEntry}
+
+Base.@kwdef mutable struct Manifest
+    deps::Dict{UUID,PackageEntry} = Dict{UUID,PackageEntry}()
```
IanButterworth, line 250: Also add a dict of top level keys. Cannot be strict/limited

IanButterworth:
Updates:

- Added `julia_version` and `manifest_format` fields
The original idea was to keep this to just moving the packages to under a `deps` field, but I think it makes more sense from the Pkg call to properly _version_ the format, and thus including `manifest_format` as a `VersionNumber` string seems the right way to do it. And introducing this new format as manifest format `2` it didn't seem to make sense to omit the `julia_version` field, given that's the main motivator for making this switch

- Now the format of the manifest file is maintained when re-written.
A `@warn` hints once (via `maxlog=1`) how the user can update to the new manifest format.
image


Beyond reviewing, I think it would be good for others to try this PR out, on top of https://github.com/JuliaLang/julia/pull/40765

src/manifest.jl
```diff
@@ -174,9 +178,23 @@ function read_manifest(f_or_io::Union{String, IO})
         end
         rethrow()
     end
+    if !isempty(raw) && Base.is_v1_format_manifest(raw)
+        raw = convert_flat_format_manifest(raw)
+    end
     return Manifest(raw)
 end
 
+function convert_flat_format_manifest(old_raw_manifest::Dict)
```
KristofferC, line 187: Can't this just be

```jl
new_raw_manifest = Dict{String, Any}(
    "deps" => old_raw_manifest, 
    "julia_version" => nothing,
    "manifest_format" => v"1"
)
```
?

KristofferC:
Some thoughts:

- I am not sure the warning should be added, at least not right away. 
- Perhaps there should be a good error when the manifest version is `>= 3`?

IanButterworth:
> I am not sure the warning should be added, at least not right away.

I feel like a warning should be in 1.7 but maybe this PR isn't exactly what's needed for the backport to 1.6.2

For the backport I think it needs to default to writing to v1 format, and doesn't need to warn if it's maintaining a v2 manifest. 

I can make the goal of this PR be the backport which can then be built on. 


> Perhaps there should be a good error when the manifest version is >= 3?

Perhaps a warning? "Unknown newer Manifest.toml format version detected, unexpected behavior may occur"

> It would be good to add a test that shows that it is now possible to read the new format as well as that writing it works.

Yeah agreed. Also need to add tests to the Base PR

IanButterworth:
Updated based on feedback.

In https://github.com/JuliaLang/Pkg.jl/pull/2557 @davidanthoff suggested `schemaversion` for what is currently `manifest_format`. Would that be a better name?

Currently
```

julia_version = "1.7.0-DEV.1199"
manifest_format = "2.0.0"

[[deps.ArgTools]]
uuid = "0dad84c5-d112-42e6-8d28-ef12dabb789f"
...
```

davidanthoff:
Just a minor question: does it make sense to use three-part versions for the manifest/schema version (whatever the name is, I don't really have a preference)? I think semver is not that useful for format versioning, or at least it is not entirely clear to me how one would use it for that. Maybe better to just use an integer for that and keep it simple?

IanButterworth:
Yeah three part is definitely excessive, but two might be useful. It only prints 3-part because that's the way `VersionNumber`s print in the TOML writer. 

I'll convert to a 2-part string during the `destructure` before write

src/Types.jl
```diff
@@ -245,7 +245,24 @@ Base.:(==)(t1::PackageEntry, t2::PackageEntry) = t1.name == t2.name &&
     t1.uuid == t2.uuid
     # omits `other`
 Base.hash(x::PackageEntry, h::UInt) = foldr(hash, [x.name, x.version, x.path, x.pinned, x.repo, x.tree_hash, x.deps, x.uuid], init=h)  # omits `other`
-const Manifest = Dict{UUID,PackageEntry}
+
+Base.@kwdef mutable struct Manifest
+    julia_version::Union{Nothing,VersionNumber} = Base.VERSION
+    manifest_format::VersionNumber = v"1.0.0" # default to older flat format
```
KristofferC, line 251: Is the `Manifest` constructor actually ever called without these arguments being explicitly given? If not, I suggest the defaults are removed.

src/Types.jl
KristofferC: Is this still a TODO?

src/manifest.jl
```diff
@@ -127,38 +127,45 @@ function validate_manifest(stage1::Dict{String,Vector{Stage1}})
...
     end
-    return manifest
+    return Manifest(; julia_version, manifest_format, deps)
 end
 
 function Manifest(raw::Dict)::Manifest
+    julia_version = isnothing(raw["julia_version"]) ? nothing : VersionNumber(raw["julia_version"])
+    manifest_format = VersionNumber(raw["manifest_format"])
+    if !in(manifest_format.major, 1:2)
+        @warn "Unknown Manifest.toml format version detected. Unexpected behavior may occur" manifest_format maxlog = 1
```
KristofferC, line 137: Might want to give the path to the manifest as well.

src/Types.jl
IanButterworth: It is called, when a manifest file doesn't exist https://github.com/IanButterworth/Pkg.jl/blob/8281b20d232768e0bcc4f28eb85965ffdd8a94e1/src/manifest.jl#L176

src/manifest.jl
IanButterworth: Fixed

IanButterworth:
CI needs re-running when the merged https://github.com/JuliaLang/julia/pull/40765 works its way into the nightlies

src/Types.jl
IanButterworth: Sorry, missed this. After the Pkg call, I couldn't recall why it was necessary for this to be a dict rather than specific fields.
Also, I don't have much free time/bandwidth during the working week at the moment, and I'm aware this is blocking the 1.7 cutoff

I'd happily go with whatever you / @StefanKarpinski think is the right thing to do here

DilumAluthge:
@KristofferC @StefanKarpinski Is this PR good to merge once CI turns green?

KristofferC:
LGTM now

src/Types.jl
KristofferC: I implemented this

DilumAluthge:
Wow, the nightlies are really behind.

DilumAluthge:
Okay, some of the nightlies are starting to catch up.

IanButterworth:
Windows has caught up, but there seems to be an unrelated issue popping up on MacOS sometimes https://github.com/JuliaLang/Pkg.jl/issues/2596

DilumAluthge:
@fredrikekre or @IanButterworth Can you add this PR to the backports PR (https://github.com/JuliaLang/Pkg.jl/pull/2592)?

IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/2575 will need backporting before this. And might as well do https://github.com/JuliaLang/Pkg.jl/pull/2581 too.
I just gave it a go and it needs a little help

KristofferC:
It would be good to clean up the commit message a bit when squash merging something big like this.

PR #2562, opened 2021-05-12
Missing package prompt: Switch default to `y`. Add text wrapping in prompt

IanButterworth:
- Switches the default to `y`
- Add text wrapping to improve the formatting in narrow terminals

Screen Shot 2021-05-11 at 3 54 42 PM
Screen Shot 2021-05-11 at 3 54 19 PM


IanButterworth:
There was general support for moving to `y` on slack, and the formatting implementation can be tweaked later if people find problems, so I'll go ahead

PR #2563, opened 2021-05-12
Test suite: replace one use of `first(...)` with `only(...)`

DilumAluthge:
In this particular test set, for a given package `name`, there should be exactly one corresponding manifest block.


IanButterworth:
Seems fair

PR #2492, opened 2021-04-06
Improvements to generate

pazner:
Following discussion in #1923, un-deprecates `] generate` and makes several improvements:

* Resolves #1917
* Resolves #759
* Partially addresses #1712 (`] generate Tmp.jl` will create a package named `Tmp` in a directory named `Tmp.jl`)

Also, points users to PkgTemplates in the docstrings if they need more than the minimal functionality provided by `generate`.

Regarding #2168, perhaps the issue is not in `generate` and rather in `parse_package_identifier` (specifically [this regex](https://github.com/JuliaLang/Pkg.jl/blob/56d699c2196a42d801e915f172cd71cd7c9c1328/src/REPLMode/argument_parsers.jl#L112) does not match "∂xxxxx", but `Pkg.API.check_package_name("∂xxxxx")` does not give an error). I am not clear on whether "∂xxxxx" is supposed to be a valid package name or not.

PR #2564, opened 2021-05-12
Asking to manually create deps/build.jl in the package directory in creating-packages.md

xgupta:
This is a little confusing when `generate HelloWorld`  did not create deps/ and test/ directory and their files.  See discourse post:  https://discourse.julialang.org/t/deps-and-test-folders-in-package/61003

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -108,7 +108,8 @@ Hello aT157rHV
 ## Adding a build step to the package
 
 The build step is executed the first time a package is installed or when explicitly invoked with `build`.
-A package is built by executing the file `deps/build.jl`.
+A package is built by executing the file `deps/build.jl`, but please don't forget to manually create that 
+before runnning below command.
 
 ```julia-repl
 julia> print(read("deps/build.jl", String))


johnnychen94:
- With the artifacts and BinaryBuilder system , `deps/build.jl` just becomes a legacy feature so I think it's good to not mention it.
- It's better to just let `generate HelloWorld` to create `test/runtests.jl`, isn't it? (Of course, one can just point the users to PkgTemplates.jl ) See also: #2492

xgupta:
> * With the artifacts and BinaryBuilder system , `deps/build.jl` just becomes a legacy feature so I think it's good to not mention it.

It is already mentioned. 

> * It's better to just let `generate HelloWorld` to create `test/runtests.jl`, isn't it? (Of course, one can just point the users to PkgTemplates.jl ) See also: #2492

`generate HelloWorld` is **not** generating  `test/runtests.jl`.
There is already a note on https://pkgdocs.julialang.org/v1/creating-packages/ for PkgTemplates.jl .

xgupta:
And Actually, I don't understand your comment (I am not a native English speaker)

Would you elaborate a little on what to do with this PR?

xgupta:
Waiting...

PR #2567, opened 2021-05-13
Automatically call `Pkg.gc()` after `up`, `pin`, `free` and `rm`

staticfloat:
We want to call `Pkg.gc()` every now and then, so let's automatically
call it after operations that can themselves result in orphaned package
versions (e.g. `rm` or `up`).  We wait the default GC collect interval,
using `mtime(orphanage_path)` as our marker of the last time we have
collected.  After each of the above operations, we query the current
time as compared to the last known GC time, hitting the filesystem only
when there is a possibility that we might need to GC.

Also, we now always write out an orphanage file (even if it's empty) to
ensure that the timestamp is always available.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -600,7 +600,7 @@ end
 ###########
 
 const DEPOT_ORPHANAGE_TIMESTAMPS = Dict{String,Float64}()
-_auto_gc_enabled = Ref{Bool}(true)
+const _auto_gc_enabled = Ref{Bool}(true)
 function _auto_gc(ctx::Types.Context; collect_delay::Period = Day(7))
     if !_auto_gc_enabled[]
         return
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -5,6 +5,7 @@ module Pkg
 import Random
 import REPL
 import TOML
+using Dates
 
 export @pkg_str
 export PackageSpec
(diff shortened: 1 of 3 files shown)


staticfloat:
I originally wrote this to try and prevent the buildup of artifacts within the Yggdrasil CI bot, but it's definitely going to need to be paired with improvements to `manifest_usage.toml` and `artifact_usage.toml` writing as they get corrupted on a weekly basis.  This is not surprising, as we are running 10+ julia processes simultaneously, so they will all be appending to these files simultaneously.

src/Pkg.jl
```diff
@@ -594,6 +595,43 @@ function dir(pkg::String, paths::AbstractString...)
     return abspath(path, "..", "..", paths...)
 end
 
+###########
+# AUTO GC #
+###########
+
+DEPOT_ORPHANAGE_TIMESTAMPS = Dict{String,Float64}()
```
KristofferC, line 602: `const` I guess

src/Pkg.jl
```diff
@@ -594,6 +595,43 @@ function dir(pkg::String, paths::AbstractString...)
...
+_auto_gc_enabled = Ref{Bool}(true)
+function _auto_gc(ctx::Types.Context; collect_delay::Period = Day(7))
+    global DEPOT_ORPHANAGE_TIMESTAMPS
```
KristofferC, line 605: I don't think this does anything since there is no assignment to it.

staticfloat:
Kristoffer, are you satisfied with this?

KristofferC:
Let's put it in and see how it feels!

fredrikekre:
An alternative would be to just compute how much space could be saved and print a message (similar to `apt-get` for example).

fonsp:
Perhaps you can use the same `mtime` trick to make the `Pkg.UPDATED_REGISTRY_THIS_SESSION[]` smarter? We currently update the registry whenever a new process uses Pkg.

KristofferC:
I don't really understand how those are related. When using the Pkg server the following happens:

- Check the version of the registry in the Pkg server
- If that is different from the current version, then download it, otherwise do nothing.

If one uses git it does a `git fetch` which should be quite fast in case of no updates.

What is your proposal?

fonsp:
I see, thanks! That sounds well-optimised already, ignore my suggestion :)

PR #2569, opened 2021-05-19
Precompilation: Adjustments for efficiency

IanButterworth:
1. Use the EnvCache-ed manifest instead of re-reading the manifest file
2. If the cachefile isn't stale, update its mtime. This is already done in base. 
https://github.com/JuliaLang/julia/blob/0bd886297d2757adf75d346a7dbb438fd839f8e6/base/loading.jl#L753-L756
While I'm not sure why it was originally added, now that cachefiles are sorted by mtime during evaluation, there is the additional benefit of keeping the `mtime` of valid cache files up to date

More info: the cache file is `touch`-ed in base during code load to make the `mtime` of the cache more representative of last used time. So it does make sense to update it here too, I think.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1264,7 +1264,13 @@ function precompile(ctx::Context; internal_call::Bool=false, strict::Bool=false,
                         Base.release(parallel_limiter)
                     end
                 else
-                    !is_stale && (n_already_precomp += 1)
+                    if !is_stale
+                        n_already_precomp += 1
+                        try
+                            touch(path_to_try) # update timestamp of precompilation file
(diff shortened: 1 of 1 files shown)


thchr:
I don't think the `path_to_try` variable is defined here (cf. JET): what should this have referenced?

IanButterworth:
Ah.. the danger of a try and empty catch!

That path needs to be extracted from the `_is_stale` , but I'm not sure what to do if that function isn't called. I guess the mtime update can be skipped then. 

Btw this is just a minor optimization, that is supposed to keep the most recently identified cache file the first to be tried during package load.

PR #2571, opened 2021-05-21
Reread registry after updating it.

### LINKED ISSUE #2555: instantiate needs a second try
```
(build-tools) pkg> st
      Status `~/dev/TME061/build-tools/Project.toml`
→ [713c75ef] Franklin v0.6.12
  [16fef848] LiveServer v0.3.6
→ [0c5d862f] Symbolics v0.1.11
        Info packages marked with → not downloaded, use `instantiate` to download

(build-tools) pkg> instantiate
ERROR: expected package `Symbolics [0c5d862f]` to be registered

(build-tools) pkg> st
      Status `~/dev/TME061/build-tools/Project.toml`
→ [713c75ef] Franklin v0.6.12
  [16fef848] LiveServer v0.3.6
→ [0c5d862f] Symbolics v0.1.11
        Info packages marked with → not downloaded, use `instantiate` to download

(build-tools) pkg> instantiate # works
   Installed Formatting ──────────────── v0.4.2
[...]
```

I can only note that the registry *pre-update* might have been since before `Symbolics` was released. But as the output suggest, the registry was updated, so something must have been cached before that I guess.

GunnarFarneback:
Fixes #2555.

When instantiate fails to find a package in the registry it calls `Registry.update()`, but all that does is to update the registry files on disk. This PR also reads the updated registry before proceeding.

Reviewing the other calls to `Registry.update()`, it seems like there may be a few more instances that expects the updated registry to be read in and possibly that should actually happen within `Registry.update()`.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -14,7 +14,7 @@ import ..Artifacts: ensure_artifact_installed, artifact_names, extract_all_hashe
                     artifact_exists, select_downloadable_artifacts
 using Base.BinaryPlatforms
 import ...Pkg
-import ...Pkg: pkg_server, Registry, pathrepr, can_fancyprint, printpkgstyle, stderr_f
+import ...Pkg: pkg_server, Registry, pathrepr, can_fancyprint, printpkgstyle, stderr_f, OFFLINE_MODE, UPDATED_REGISTRY_THIS_SESSION
 
 #########
 # Utils #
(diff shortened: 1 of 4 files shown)


KristofferC:
Yes, it seems there is similar problems elsewhere. Perhaps there should be an:

```
function Pkg.update_registries(ctx::Context, ...) 
     Registry.update_registries()
     copy!(ctx.registries, Registry.reachable_registries())
end
```

GunnarFarneback:
Did you actually mean a top level `Pkg` function or something like in the revised commit?

src/Operations.jl
```diff
@@ -1051,6 +1051,12 @@ function update_package_add(ctx::Context, pkg::PackageSpec, entry::PackageEntry,
     return pkg
 end
 
+# Update registries AND read them back in.
+function update_registries(ctx::Context, registries...; kwargs...)
```
GunnarFarneback, line 1055: Maybe just skip the `registries` argument? It doesn't really match up with the `copy!`.

KristofferC:
I tweaked this a bit, looks ok to you?

GunnarFarneback:
Looks fine to me.

PR #2576, opened 2021-05-28
Fixup some registry things

KristofferC:


DilumAluthge:
Does this PR close issue https://github.com/JuliaLang/julia/issues/40981?

IanButterworth:
I don't think so. The function in base that passes modules to the pkg install hook still needs to be taught to ignore non-package modules

PR #2580, opened 2021-05-30
New Manifest.toml format: Pt 2. Default to new format, warn if old format is being maintained

### LINKED ISSUE #1458: Additional Manifest entries

### LINKED ISSUE #2508: Add metadata to manifest
It seems a fair few issues with upgrading to 1.6 have been around using old manifests that fail in unclear/unpleasant ways (because of changes to stdlibs for one), and the advice is "manifests aren't guaranteed to work across julia versions, delete the manifest and try again".

For 1.7 given there's no guarantee of cross-version manifest compatibility, I was wondering if a clean break could be made and change the structure of the manifest & Manifest.toml to something a bit more flexible than a `Dict{UUID,PackageEntry}` so that metadata can be added, like julia version and a hash of the Project.toml so that Pkg can be sure that the manifest is a valid resolve of the current project.

A heuristic could be added to 1.6.1 to `read_manifest` to detect if the manifest is of the new format, to warn/instruct the user to delete the manifest to proceed.

The new manifests would break if users tried to use them with older julia versions. But perhaps that's part of the _no guarantee_?

I had a search and couldn't find another issue to track this

IanButterworth:
New Manifest.toml format: Pt 2. (For 1.7 only, not to be backported to 1.6.2)
- Defaults to the new manifest format
- Warns the user (once) if an old format manifest is being maintained

Continuation after part 1 (https://github.com/JuliaLang/Pkg.jl/pull/2561)

Fixes #1458 
Fixes #2508 
Closes #2557

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1670,4 +1670,17 @@ function handle_package_input!(pkg::PackageSpec)
     pkg.uuid = pkg.uuid isa String ? UUID(pkg.uuid) : pkg.uuid
 end
 
+function upgrade_manifest(ctx::Context = Context())
+    before_format = ctx.env.manifest.manifest_format
+    if before_format == v"2.0"
+        pkgerror("Format of manifest file at `$(ctx.env.manifest_file)` already up to date: manifest_format == $(before_format)")
+    elseif before_format != v"1.0"
+        pkgerror("Format of manifest file at `$(ctx.env.manifest_file)` version is unrecogized: manifest_format == $(before_format)")
+    end
+    ctx.env.manifest.manifest_format = v"2.0"
+    Types.write_manifest(ctx.env)
+    printpkgstyle(ctx.io, :Updated, "Format of manifest file at `$(ctx.env.manifest_file)` updated from v$(before_format.major).$(before_format.minor) to v2.0")
+    return nothing
+end
+
 end # module
--- a/src/API.jl
+++ b/src/API.jl
@@ -1678,6 +1678,7 @@ function upgrade_manifest(ctx::Context = Context())
         pkgerror("Format of manifest file at `$(ctx.env.manifest_file)` version is unrecogized: manifest_format == $(before_format)")
     end
     ctx.env.manifest.manifest_format = v"2.0"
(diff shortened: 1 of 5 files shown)


KristofferC:
One problem with deleting and runing `resolve` is that you lose all the information in the manifest. Alternatively, we could have a very simple function `Pkg.upgrade_manifest()` and suggest that, which would keep all the dependency versions.

DilumAluthge:
@IanButterworth @KristofferC @StefanKarpinski While we are waiting for CI on https://github.com/JuliaLang/Pkg.jl/pull/2561, is there anything else that needs to be done for this PR?

IanButterworth:
> Alternatively, we could have a very simple function `Pkg.upgrade_manifest()` and suggest that, which would keep all the dependency versions.

Nice. Done!

DilumAluthge:
https://github.com/JuliaLang/Pkg.jl/pull/2561 has been merged, so I think this needs a rebase now? 

Also looks like there are some merge conflicts.

DilumAluthge:
@StefanKarpinski @KristofferC Is this PR good to merge once CI turns green?

KristofferC:
I made it so that the upgrade manifest function also set the version field (I think that was omitted by accident).

IanButterworth:
> I made it so that the upgrade manifest function also set the version field (I think that was omitted by accident)

@KristofferC it was intentional. I should've added a comment. I think the version needs to represent which julia version the resolve happened in, and given the tool just updates the format I don't think it's accurate to set the version field.

KristofferC:
That is true, let's take that back then :)

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/2599

PR #2594, opened 2021-06-04
Standardize the CI workflow for checking and changing the UUID

DilumAluthge:
I want to standardize the CI workflow for checking and changing the UUID, so that it is exactly the same across all stdlibs that live in external repos.

That way, it is really easy for someone to set this up on a new external stdlib:
1. Copy two lines from the `.github/workflows/ci.yml` file in an existing external stdlib and paste those two lines into the `.github/workflows/ci.yml` file  in your new external stdlib
2. Copy two files from the `.ci/` directory in an existing external stdlib and paste those two lines into the `.github/workflows/ci.yml` file  in your new external stdlib
3. In the `.ci/test_uuid_is_unchanged.jl` file in your new external stdlib, change line 8 (`correct_uuid = "..."`) to contain the correct UUID for new external stdlib

### DIFF
--- a/.ci/change_uuid.jl
+++ b/.ci/change_uuid.jl
@@ -1,0 +1,5 @@
+using UUIDs
+
+project_filename = joinpath(dirname(@__DIR__), "Project.toml")
+
+write("Project.toml", replace(read("Project.toml", String), r"uuid = .*?\n" =>"uuid = \"$(uuid4())\"\n"))
--- a/.ci/test_uuid_is_unchanged.jl
+++ b/.ci/test_uuid_is_unchanged.jl
@@ -1,7 +1,7 @@
 using Pkg
 using Test
 
-@testset "Pkg UUID" begin 
+@testset "Test that the UUID is unchanged" begin 
     project_filename = joinpath(dirname(@__DIR__), "Project.toml")
     project = Pkg.TOML.parsefile(project_filename)
     uuid = project["uuid"]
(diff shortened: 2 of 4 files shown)


.ci/change_uuid.jl
```diff
@@ -0,0 +1,5 @@
+using UUIDs
+
+project_filename = joinpath(dirname(@__DIR__), "Project.toml")
+
+write("Project.toml", replace(read("Project.toml", String), r"uuid = .*?\n" =>"uuid = \"$(uuid4())\"\n"))
```
fredrikekre, line 5: https://github.com/JuliaLang/Pkg.jl/pull/2379

fredrikekre (commented):
Couldn't the `change_uuid.jl` and `test_uuid_is_unchanged.jl` be the same file?

DilumAluthge:
> Couldn't the `change_uuid.jl` and `test_uuid_is_unchanged.jl` be the same file?

Oh good catch, definitely! They used to be in two different folders, which is why they were two different files. And then when I moved them into the same folder, I completely missed that they can and should just be a single file.

.ci/change_uuid.jl
DilumAluthge: Ah good catch. Yes we should just use the same UUID every time.

DilumAluthge: This will be fixed in #2595

DilumAluthge:
> Couldn't the `change_uuid.jl` and `test_uuid_is_unchanged.jl` be the same file?

This will be implemented in #2595

PR #2599, opened 2021-06-07
dont set version on upgraded manifest

KristofferC:


IanButterworth (commented):
This looks good, but additionally I may have not got the point that the version field set right. It is currently set when a new manifest is created, but not during the resolve. Perhaps it should be only set during the resolve and be `nothing` by default?

PR #2610, opened 2021-06-09
New manifest format: Fix `upgrade_manifest` and raw dict entry for nothing

### LINKED ISSUE #2609: Pkg.upgrade_manifest broke my Manifest
```
julia> using Pkg; Pkg.upgrade_manifest()
     Updated Format of manifest file at `/home/vtjnash/.julia/environments/v1.1/Manifest.toml` updated from v1.0 to v2.0

(@v1.1) pkg> up
ERROR: ArgumentError: invalid version string: nothing
Stacktrace:
  [1] parse
    @ ./version.jl:126 [inlined]
  [2] VersionNumber(v::String)
    @ Base ./version.jl:130
  [3] Pkg.Types.Manifest(raw::Dict{String, Any}, f_or_io::String)
    @ Pkg.Types /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:134
  [4] read_manifest(f_or_io::String)
    @ Pkg.Types /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/manifest.jl:198
  [5] Pkg.Types.EnvCache(env::Nothing)
    @ Pkg.Types /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:338
  [6] EnvCache
    @ /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:317 [inlined]
  [7] Pkg.Types.Context()
    @ Pkg.Types ./util.jl:471
  [8] up(pkgs::Vector{Pkg.Types.PackageSpec}; io::Base.TTY, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/API.jl:145
  [9] up(pkgs::Vector{Pkg.Types.PackageSpec})
    @ Pkg.API /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/API.jl:144
 [10] do_cmd!(command::Pkg.REPLMode.Command, repl::REPL.LineEditREPL)
    @ Pkg.REPLMode /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/REPLMode/REPLMode.jl:407
 [11] do_cmd(repl::REPL.LineEditREPL, input::String; do_rethrow::Bool)
    @ Pkg.REPLMode /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/REPLMode/REPLMode.jl:385
 [12] do_cmd
    @ /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/REPLMode/REPLMode.jl:376 [inlined]
 [13] (::Pkg.REPLMode.var"#24#27"{REPL.LineEditREPL, REPL.LineEdit.Prompt})(s::REPL.LineEdit.MIState, buf::IOBuffer, ok::Bool)
    @ Pkg.REPLMode /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/Pkg/src/REPLMode/REPLMode.jl:549
 [14] #invokelatest#2
    @ ./essentials.jl:716 [inlined]
 [15] invokelatest
    @ ./essentials.jl:714 [inlined]
 [16] run_interface(terminal::REPL.Terminals.TextTerminal, m::REPL.LineEdit.ModalInterface, s::REPL.LineEdit.MIState)
    @ REPL.LineEdit /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/REPL/src/LineEdit.jl:2493
 [17] run_frontend(repl::REPL.LineEditREPL, backend::REPL.REPLBackendRef)
    @ REPL /data/vtjnash/julia/usr/share/julia/stdlib/v1.7/REPL/src/REPL.jl:1227
 [18] (::REPL.var"#49#54"{REPL.LineEditREPL, REPL.REPLBackendRef})()
    @ REPL ./task.jl:411
```
```
$ head Manifest.toml 

julia_version = "nothing"
manifest_format = "2.0"

[[deps.ATK_jll]]
deps = ["Artifacts", "Glib_jll", "JLLWrappers", "Libdl", "Pkg"]
git-tree-sha1 = "a5a8f114e4d70bee6cf82ed28b488d57f1fa9467"
uuid = "7b86fcea-f67b-53e1-809c-8f1719c154e8"
version = "2.36.0+0"
```

DilumAluthge:
Fixes #2609

### DIFF
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -131,7 +131,7 @@ function validate_manifest(julia_version::Union{Nothing,VersionNumber}, manifest
 end
 
 function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
-    julia_version = isnothing(raw["julia_version"]) ? nothing : VersionNumber(raw["julia_version"])
+    julia_version = ((raw["julia_version" === nothing) || (raw["julia_version" == "nothing")) ? nothing : VersionNumber(raw["julia_version"])
     manifest_format = VersionNumber(raw["manifest_format"])
     if !in(manifest_format.major, 1:2)
         if f_or_io isa IO
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -131,7 +131,7 @@ function validate_manifest(julia_version::Union{Nothing,VersionNumber}, manifest
 end
 
 function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
-    julia_version = ((raw["julia_version"] === nothing) || (raw["julia_version"] == "nothing")) ? nothing : VersionNumber(raw["julia_version"])
+    julia_version = (raw["julia_version"] == "nothing") ? nothing : VersionNumber(raw["julia_version"])
     manifest_format = VersionNumber(raw["manifest_format"])
     if !in(manifest_format.major, 1:2)
         if f_or_io isa IO
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -131,7 +131,7 @@ function validate_manifest(julia_version::Union{Nothing,VersionNumber}, manifest
 end
 
 function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
-    julia_version = ((raw["julia_version" === nothing) || (raw["julia_version" == "nothing")) ? nothing : VersionNumber(raw["julia_version"])
+    julia_version = ((raw["julia_version"] === nothing) || (raw["julia_version"] == "nothing")) ? nothing : VersionNumber(raw["julia_version"])
     manifest_format = VersionNumber(raw["manifest_format"])
     if !in(manifest_format.major, 1:2)
         if f_or_io isa IO
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -131,7 +131,7 @@ function validate_manifest(julia_version::Union{Nothing,VersionNumber}, manifest
 end
 
 function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
-    julia_version = ((isnothing(raw["julia_version"])) || (raw["julia_version"] == "nothing")) ? nothing : VersionNumber(raw["julia_version"])
+    julia_version = raw["julia_version"] == "nothing" ? nothing : VersionNumber(raw["julia_version"])
     manifest_format = VersionNumber(raw["manifest_format"])
     if !in(manifest_format.major, 1:2)
         if f_or_io isa IO
@@ -201,7 +201,7 @@ end
 function convert_flat_format_manifest(old_raw_manifest::Dict)
     new_raw_manifest = Dict{String, Any}(
             "deps" => old_raw_manifest,
-            "julia_version" => nothing,
+            "julia_version" => "nothing", # must be a string here to match raw dict
             "manifest_format" => "1.0.0" # must be a string here to match raw dict
         )
     return new_raw_manifest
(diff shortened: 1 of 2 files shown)


IanButterworth:
Shoot. I thought this was covered by tests. I'll push a test

DilumAluthge:
@vtjnash If you get a chance, could you try this PR branch with your manifest from #2609 and see if it works? It should, but it'd be good to double-check.

DilumAluthge:
Also @IanButterworth does this need backporting to both 1.6 and 1.7, or just 1.7?

IanButterworth:
Just 1.7

DilumAluthge:
> Shoot. I thought this was covered by tests. I'll push a test

You can push directly to this branch, if that works for you.

src/manifest.jl
```diff
@@ -131,7 +131,7 @@ function validate_manifest(julia_version::Union{Nothing,VersionNumber}, manifest
 end
 
 function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
-    julia_version = isnothing(raw["julia_version"]) ? nothing : VersionNumber(raw["julia_version"])
+    julia_version = ((raw["julia_version"] === nothing) || (raw["julia_version"] == "nothing")) ? nothing : VersionNumber(raw["julia_version"])
```
KristofferC, line 134: How is the first case even possible?

DilumAluthge: I left it in only because the original code handled it (with `isnothing(raw["julia_version"])`).

I'm not familiar enough with the TOML parser - will it ever return `nothing` as a value?

DilumAluthge:
CI is passing now.

src/manifest.jl
KristofferC: I think the bug came from the original code being buggy, TOML can not contain `nothing`.

As a side note, on Julia 1.7 there should be no advantage with `=== nothing` over `== nothing` and `isnothing`.

vtjnash: `== nothing` can still return missing, so it is conceptually and fundamentally different still. Calling `isnothing` should now be the same as `===` in nearly all cases however.

DilumAluthge: > so it should be fine to remove.

I tried taking out the `isnothing` check, but tests fail without it, so I've put it back.

I've switched to doing `isnothing` instead of `=== nothing`.

IanButterworth:
Wait. There's a bug

src/manifest.jl
IanButterworth: There was a bug. Fixed in https://github.com/JuliaLang/Pkg.jl/pull/2610/commits/c5a3717185ee0c45ff2a78efcd80e20486f83416

DilumAluthge:
Good to merge now?

StefanKarpinski:
Alternatively, could we just leave this key out when we don't know the Julia version? Using `"nothing"` like Julia's `nothing` feels like it is a bit of a violation of the TOML abstraction: TOML data should be meaningful and interpretable without knowing about Julia's type system; using the string `"nothing"` like this leaks the Julia data model into TOML land in a kind of unfortunate way. It seems like it would be cleaner for the `julia_version` entry to be absent instead.

IanButterworth:
That sounds good to me. It wouldn't break the base heuristic
https://github.com/JuliaLang/julia/blob/15c19c8a6b1c1a87e6dd24fd058de5192826b250/base/loading.jl#L560-L570

But we would want to get that change out before people start generating manifests with `julia_version = "nothing"` in it, so could block the 1.7 beta?

StefanKarpinski:
I wouldn't worry about the people who are using master, just tell them how to fix it when it breaks.

PR #2614, opened 2021-06-10
Enable installation of versioned historical stdlibs

### LINKED ISSUE #2479: `Pkg.add("GMP_jll"; julia_version=v"1.6")` doesn't install `GMP_jll` for v1.6
```julia
julia> Pkg.add("GMP_jll"; julia_version=v"1.5")
  [...]

(jl_uYNCOr) pkg> st
      Status `/tmp/jl_uYNCOr/Project.toml`
  [781609d7] GMP_jll v6.1.2+6

julia> Pkg.add("GMP_jll"; julia_version=v"1.6")
   Resolving package versions...
    Updating `/tmp/jl_uYNCOr/Project.toml`
  [781609d7] ~ GMP_jll v6.1.2+6 ⇒ 
    Updating `/tmp/jl_uYNCOr/Manifest.toml`
  [...]

(jl_uYNCOr) pkg> st
      Status `/tmp/jl_uYNCOr/Project.toml`
  [781609d7] GMP_jll
```
`Pkg.add("GMP_jll"; julia_version=v"1.6")` doesn't work as expected, I guess because `GMP_jll` is a standard library in v1.6.  As you can guess, this is needed for BinaryBuilder :upside_down_face: 

Spotted in https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/129#issuecomment-812259489

staticfloat:
Fixes #2479

When installing stdlibs, if we are requesting a versioned stdlib, record
its version in the manifest and download/install it.  This allows for
tools such as BinaryBuilder to download the correct JLL for e.g.
`GMP_jll` when targeting Julia v1.6, for example.

This does slightly change manifest writing; `Manifest.toml` files now
record the version of versioned stdlibs (such as `Pkg`, `Tar`, all JLL
stdlibs, `NetworkOptions`, etc...) however this has no impact on code
loading (as they still do not contain `git-tree-sha1` entries).

### DIFF
--- a/test/resolve.jl
+++ b/test/resolve.jl
@@ -435,7 +435,6 @@ end
             return versions[idx]
         end
 
-        # DISABLED FOR NOW BECAUSE THE JLL INTO STDLIBS BROKE IT
         # First, we're going to resolve for specific versions of Julia, ensuring we get the right dep versions:
         Pkg.Registry.download_default_registries(Pkg.stdout_f())
         ctx = Pkg.Types.Context(;julia_version=v"1.5")
--- a/test/resolve.jl
+++ b/test/resolve.jl
@@ -435,7 +435,6 @@ end
             return versions[idx]
         end
 
-        #=
         # DISABLED FOR NOW BECAUSE THE JLL INTO STDLIBS BROKE IT
         # First, we're going to resolve for specific versions of Julia, ensuring we get the right dep versions:
         Pkg.Registry.download_default_registries(Pkg.stdout_f())
@@ -469,7 +468,6 @@ end
         mpfr = find_by_name(versions, "MPFR_jll")
         @test mpfr !== nothing
         @test mpfr.version.major == 4 && mpfr.version.minor == 0
-        =#
     end
 end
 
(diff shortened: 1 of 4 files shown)


staticfloat:
Fixes #2479

This was originally intended to target v1.7, which will really help with BB tasks that need to build against historical versions of Julia (e.g. they need to link against `libgmp`, so they must be built per-julia-version, to deal with the fact that `libgmp` itself is shipped with Julia).

If it's too late to make it in, that's okay, we can just run Yggdrasil on Julia `master` for a while until 1.8 is released.

staticfloat:
To the reviewers: there is one slightly unusual part of this change, which is that if you try to install an stdlib for a different version of Julia, (e.g. `Pkg.add("GMP_jll"; julia_version=v"1.6)` on `v1.7`) it won't actually install the package `GMP_jll` because of [this check](https://github.com/JuliaLang/Pkg.jl/blob/b2b215e098550c22e0e694890f964ed14299e4ba/src/Operations.jl#L652) which discards stdlibs.  We could change that, which would be what "truly" fixes #2479, however it would require a slightly more involved change where we disambiguate between `julia_version` being set (in which case we always download and install stdlibs) and not being set (the default behavior).

As it stands, this at least gives us the information to allow BB to download and install the packages it needs, even if `Pkg` isn't downloading them for us.

KristofferC:
Could https://github.com/JuliaLang/Pkg.jl/blob/0de93211d96c57c1617000a683b70a70154439d5/test/resolve.jl#L438-L472 be enabled again now?

staticfloat:
Looks like we can!  Great!

test/resolve.jl
```diff
@@ -435,7 +435,6 @@ end
             return versions[idx]
         end
 
-        #=
         # DISABLED FOR NOW BECAUSE THE JLL INTO STDLIBS BROKE IT
```
KristofferC, line 438:
```suggestion
```

:P

DilumAluthge:
Removing the backport label, as this has been reverted on the release-1.7 branch.

Issue #2616, opened 2021-06-11
Old manifest warning printed for tmp test manifest


fredrikekre:
```
$ pkg test
     Testing IOCapture
┌ Warning: The active manifest file at `/tmp/jl_v5cov0/Manifest.toml` has an old format that is being maintained.
│ To update to the new format run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.
└ @ Pkg.Types ~/julia-master/usr/share/julia/stdlib/v1.8/Pkg/src/manifest.jl:287
```

IanButterworth:
I don't recognize your repl setup.. and can't reproduce it with:

```
(@v1.8) pkg> add IOCapture
   Resolving package versions...
   Installed IOCapture ─ v0.2.1
    Updating `~/.julia/environments/v1.8/Project.toml`
  [b5f81e59] + IOCapture v0.2.1
    Updating `~/.julia/environments/v1.8/Manifest.toml`
  [b5f81e59] + IOCapture v0.2.1
Precompiling project...
  1 dependency successfully precompiled in 1 seconds (42 already precompiled)

(@v1.8) pkg> test IOCapture
     Testing IOCapture
      Status `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_xkthg9/Project.toml`
  [b5f81e59] IOCapture v0.2.1
  [56ddb016] Logging `@stdlib/Logging`
  [8dfed614] Test `@stdlib/Test`
      Status `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_xkthg9/Manifest.toml`
  [b5f81e59] IOCapture v0.2.1
  [2a0f44e3] Base64 `@stdlib/Base64`
  [b77e0a4c] InteractiveUtils `@stdlib/InteractiveUtils`
  [56ddb016] Logging `@stdlib/Logging`
  [d6f4376e] Markdown `@stdlib/Markdown`
  [9a3f8284] Random `@stdlib/Random`
  [9e88b42a] Serialization `@stdlib/Serialization`
  [8dfed614] Test `@stdlib/Test`
     Testing Running tests...
Test Summary: | Pass  Total
IOCapture.jl  |   78     78
     Testing IOCapture tests passed
```

fredrikekre:
Hmm. I was running tests inside the IOCapture repo, probably with an old manifest there. Can you replicate it like that?

[...]

PR #2617, opened 2021-06-11
Automatically upgrade the sandbox manifest format when preserving

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2616


IanButterworth:
I recreated the issue in #2616 and this does indeed fix it

PR #2620, opened 2021-06-12
New manifest: Set `julia_version` during resolve only, and warn if instantiating a manifest that was resolved in a different or unknown julia version

IanButterworth:
The intention is to use the new manifest format `julia_version` to guide the user when the manifest is (potentially) incompatible with the julia version.

- The `julia_version` field in the manifest is now only set during resolve
- Adds a `@warn` (maxlog = 1 per manifest path) when loading the manifest (any action that uses `Context()` plus a direct instantiate of a manifest)

For old format manifests:
```
┌ Warning: The active manifest file is an older format with no julia version entry. Dependencies may have been resolved with a different julia version.
└ @ ~/Documents/GitHub/Pkg.jl/test/manifest/formats/v1.0/Manifest.toml:0
```
For new format manifests with the `julia_version` field:
```
┌ Warning: The active manifest file has dependencies that were resolved with a different julia version (1.7.0-DEV.1199). Unexpected behavior may occur.
└ @ ~/Documents/GitHub/Pkg.jl/test/manifest/formats/v2.0/Manifest.toml:0
```

Questions:
When should the warning also be shown? After activate?

What if julia is started up via `--project`. Should there be a warning in the julia repl?

On master/non-released versions the full version check could create a lot of noise. Should the version check just be for `major.minor`? Edit: Implemented

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -261,7 +261,7 @@ Base.:(==)(t1::PackageEntry, t2::PackageEntry) = t1.name == t2.name &&
 Base.hash(x::PackageEntry, h::UInt) = foldr(hash, [x.name, x.version, x.path, x.pinned, x.repo, x.tree_hash, x.deps, x.uuid], init=h)  # omits `other`
 
 Base.@kwdef mutable struct Manifest
-    julia_version::Union{Nothing,VersionNumber} = Base.VERSION
+    julia_version::Union{Nothing,VersionNumber} = nothing # only set to VERSION when resolving
     manifest_format::VersionNumber = v"2.0.0"
     deps::Dict{UUID,PackageEntry} = Dict{UUID,PackageEntry}()
     other::Dict{String,Any} = Dict{String,Any}()
--- a/src/API.jl
+++ b/src/API.jl
@@ -1415,6 +1415,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     end
     if !isfile(ctx.env.project_file) && isfile(ctx.env.manifest_file)
         _manifest = Pkg.Types.read_manifest(ctx.env.manifest_file)
+        Types.check_warn_manifest_julia_version_compat(_manifest, ctx.env.manifest_file)
         deps = Dict{String,String}()
         for (uuid, pkg) in _manifest
             if pkg.name in keys(deps)
@@ -1433,6 +1434,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     if !isfile(ctx.env.manifest_file) && manifest == true
         pkgerror("expected manifest file at `$(ctx.env.manifest_file)` but it does not exist")
     end
+    Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
     Operations.prune_manifest(ctx.env)
     for (name, uuid) in ctx.env.project.deps
         get(ctx.env.manifest, uuid, nothing) === nothing || continue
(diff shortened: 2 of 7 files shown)


DilumAluthge:
> On master/non-released versions the full version check could create a lot of noise. Should the version check just be for `major.minor`?

That makes sense to me.

DilumAluthge:
> What if julia is started up via `--project`. Should there be a warning in the julia repl?

Yeah I think this makes sense.

DilumAluthge:
> When should the warning also be shown? After activate?

Doing the warning on `Pkg.activate` seems fine to me.

IanButterworth:
Updated to just check major.minor


> Doing the warning on Pkg.activate seems fine to me.

AFAICT `Pkg.activate` currently doesn't load the manifest, so to add the warning at that step that would need to happen. `Context()` might be fast enough to not be noticeable. Not sure.

(github gripe.. it would be nice to reply to your messages as threads..)

DilumAluthge:
> AFAICT `Pkg.activate` currently doesn't load the manifest, so to add the warning at that step that would need to happen. `Context()` might be fast enough to not be noticeable. Not sure.

Hmmm. What are our alternatives here? We could do nothing on `Pkg.activate`, and just wait for the user to call some other `Pkg` operation? (`Pkg.add`, `Pkg.update`, `Pkg.status`, etc.)

IanButterworth:
If the weird problems are only happening happen during precompilation, then the warning could be added to `Context()` and just triggered whenever that first loads. 

But I think problems also occur during package load and regular usage.. so I think it does need to happen immediately after `Pkg.actívate` and julia startup after `--project`.

Perhaps in this PR we could add it to `Context()` and try that out, and see if more is needed?

IanButterworth:
If this goes ahead, I think it should be backported to both 1.7 and 1.6. 
With 1.6 the warning and tests will need a little tweaking during the rebase though

IanButterworth:
@KristofferC at a minimum I think we should get the first commit here into 1.6.2 and the first 1.7 beta to fix #2621

KristofferC:
> warn if instantiating a manifest that was resolved in a different or unknown julia version 

This doesn't really make sense to me. Why would you warn when you are resolving? At that point you are already doing the right thing (resolving on the new Julia version), or?

IanButterworth:
I'm not suggesting warning when you resolve. Instantiating doesn't re-resolve, right? That's the time you expect the install to be repeatable, so seems like a good time to warn

KristofferC:
My bad, I totally misread that. Yes, doing it during instantiate makes sense. 👍

src/Types.jl
```diff
@@ -337,6 +337,8 @@ function EnvCache(env::Union{Nothing,String}=nothing)
     write_env_usage(manifest_file, "manifest_usage.toml")
     manifest = read_manifest(manifest_file)
 
+    Types.check_warn_manifest_julia_version_compat(manifest, manifest_file)
```
KristofferC, line 340: I don't think this one should be here. This will trigger even when doing "API" style use (like in Pluto).

KristofferC: I removed this.

KristofferC: Oh, gotta update test as well :)

IanButterworth: Ok so this PR now only warns during instantiate. It's obviously better than before, but not sure how many manifest compat issues it will catch given instantiate isn't that common, I believe

KristofferC:
I just want to say that we should be quite careful with adding warnings. Warning spam can be a bit annoying and a warning should (in my opinion) pretty much only be printed when there is a direct immediate action you should take as a response to fix the warning.

I'm getting a little bit worried with all the logic for these added warnings. Ideally, it should be kept to a minimum.

IanButterworth:
I totally agree. I don't think I figured out a good design, but this PR now has the minimal example via `instantiate` so at least people can try that out

fredrikekre:
Looks like this is already on release-1.7?

PR #2622, opened 2021-06-14
fix not warning about old manifest during Pkg.test

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/2617 was supposed to fix this but I just tried `Pkg.test` a random package and the warning still showed up. In my opinion, this manifest upgrade should happen unconditionally, and not just when testing the active project.


IanButterworth (approved):
Indeed. I think I misunderstood what that conditional block was for. 

The comment needs moving too

PR #2628, opened 2021-06-16
Backports for 1.6, take 2.

KristofferC:


IanButterworth:
I tested this locally against https://github.com/JuliaLang/julia/pull/40702 and got the same error
```
undo redo functionality: Test Failed at /Users/ian/Documents/GitHub/Pkg.jl/test/pkg.jl:700
  Expression: !(haskey(Pkg.dependencies(), unicode_uuid))
```

No clue..

IanButterworth:
@KristofferC I've looked at this a few times and still don't have any idea what broke the undo tests

IanButterworth:
2018f15559001f70fde4aa7d8a0d6bd3324be14b is the first bad commit for the undo tests, testing with https://github.com/JuliaLang/julia/pull/40702

Also, 27ab0709c0c6180329e9f250469f86302957410f seems like a bad backport as `Pkg.upgrade_manifest()` wasn't backported to 1.6 but that backport add a test for it, so that'll fail.
I don't think what was in that original PR is needed at all as there are later commits that fix the other changes it made. So I'd just drop it.

IanButterworth:
I think the problem is that the manifest equality is failing here https://github.com/JuliaLang/Pkg.jl/blob/kc/1.6/src/API.jl#L1539

I guess during the undo tests the `original_manifest` that's deepcopied during the intial `Context()` is missing the `julia_version` which is added in the initial resolve during the first `Pkg.add`, so every comparison afterwards back to that will result in `ctx.manifest == ctx.original_manifest` being `false`, which means the undo snapshot will always be added.

Perhaps `==` should just check the `deps`? https://github.com/JuliaLang/Pkg.jl/blob/kc/1.6/src/Types.jl#L229

Though, I don't yet understand why the undo tests aren't failing on master..

IanButterworth:
The agreed fix with @KristofferC was to change the undo functionality to just consider equality wrt. the manifest deps, rather than change the equality method for the manifest type to do that

IanButterworth (approved):

DilumAluthge:
@KristofferC @IanButterworth @fredrikekre Is this good to go?

PR #2635, opened 2021-06-18
release1.7: Revert "Enable installation of versioned historical stdlibs (#2614)"

KristofferC:
This reverts commit e9b6ffb5bf67a6bb0a590f8ed8ddb72cc83d8f2e.

Adding `version` fields to all stdlibs is too scary to add this late in the release process (ref https://github.com/JuliaLang/julia/issues/41266). I will revert this on 1.7 and then we could potentially add it back if it could be done in a way that does not change the manifest.


DilumAluthge:
Should this be merged into `master` or into `release-1.7`?

KristofferC:
That is up to @staticfloat when he looks at https://github.com/JuliaLang/julia/issues/41266. I just want release-1.7 to be decent so we can finally announce the beta.

KristofferC:
Ok now I see your point lol.

PR #2636, opened 2021-06-18
Revert "release1.7: Revert "Enable installation of versioned historical stdlibs (#2614)""

KristofferC:
Reverts JuliaLang/Pkg.jl#2635


KristofferC:
All these forced CI checks are very annoying....

DilumAluthge:
Want to just change it to only require the "docs" check?

The main purpose is to prevent accidentally committing directly to master, and we only need to require a single check for that to work.

DilumAluthge:
> Want to just change it to only require the "docs" check?

I've done this.

KristofferC:
Yeah, docs is fast enough I think.

PR #2639, opened 2021-06-22
Add documentation of do-block form of Pkg.activate

st--:
The `do`-block form of Pkg.activate() can be very useful but is not actually documented in the docstring.
This PR also clarifies that `shared` is only allowed when also providing a name for the environment.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -408,7 +408,7 @@ const status = API.status
 
 """
     Pkg.activate([s::String]; shared::Bool=false, io::IO=stderr)
-    Pkg.activate(; temp::Bool=false, shared::Bool=false, io::IO=stderr)
+    Pkg.activate(; temp::Bool=false, io::IO=stderr)
 
 Activate the environment at `s`. The active environment is the environment
 that is modified by executing package commands.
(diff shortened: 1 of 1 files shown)


fredrikekre:
I don't think it is good to advertise this, changing environments dynamically like this is not a good idea. I believe the method exist only for testing purposes.

st--:
@fredrikekre thanks for the quick response. Dynamically changing environments seems like a reasonable use-case for build scripts, such as the make.jl for building docs in KernelFunctions.jl (see discussion on https://github.com/JuliaGaussianProcesses/KernelFunctions.jl/pull/303#discussion_r652790022). If you think we shouldn't be using the do-form, what would you propose instead?

fredrikekre:
Well, it doesn't work like you think (?) it does. If you have already loaded a version of a package you can't load another version of a package.

st--:
Thanks for the clarification and help with improving our workflow! If this form should not be used and is only used for the tests, how about moving it to some kind of private helper function then, instead of exposing it as part of the API? (Would be happy to make that PR if you like.)

fredrikekre:
I'd like to remove it yea, but it seems that it is used a bit (https://juliahub.com/ui/Search?q=Pkg.activate%28.%2a%29%5Csdo&type=code&r=true) so probably it has to be kept around (but not advertised).

fonsp:
Pluto.jl also uses this behaviour to manage multiple environments from a single process. I was planning to copy-paste this function because it is not documented.

st--:
> I'd like to remove it yea, but it seems that it is used a bit (https://juliahub.com/ui/Search?q=Pkg.activate%28.%2a%29%5Csdo&type=code&r=true) so probably it has to be kept around (but not advertised).

How about deprecating it then, at least then everyone relying on it will find out about it, and you could add an explicitly marked-internal (and non-exported) function for the testing usecase ?

fredrikekre:
Fine with me

IanButterworth:
Replaced by https://github.com/JuliaLang/Pkg.jl/pull/2912

Issue #190, opened 2018-03-12
Windows, fails to unpack


SimonDanisch:
```julia
pkg> add PackageCompiler
┌ Info: Pkg3 is still under development, please file issues at `https://github.com/JuliaLang/Pkg3.jl`.
└ Pkg3 is running without precompile statements, first action will be slow.
[ Info: Resolving package versions
ERROR: Error when installing packages:
failed process: Process(`'C:\Users\sdani\AppData\Local\Julia-0.7.0-DEV\bin\7z.exe' x -si -y -ttar '-oC:\Users\sdani\AppData\Local\Temp\4quPbMZVkekT'`, ProcessExited(2)) [2]
Stacktrace:
 [1] error(::String, ::Base.Process, ::String, ::Int64, ::String) at .\error.jl:42
 [2] pipeline_error at .\process.jl:712 [inlined]
 [3] pipeline_error(::Base.ProcessChain) at .\process.jl:725
 [4] #run#537(::Bool, ::Function, ::Base.OrCmds, ::Tuple{Base.DevNullStream,Base.DevNullStream,Base.DevNullStream}, ::Vararg{Tuple{Base.DevNullStream,Base.DevNullStream,Base.DevNullStream},N} where N) at .\process.jl:670
 [5] run(::Base.OrCmds, ::Tuple{Base.DevNullStream,Base.DevNullStream,Base.DevNullStream}, ::Vararg{Tuple{Base.DevNullStream,Base.DevNullStream,Base.DevNullStream},N} where N) at .\process.jl:668
 [6] install_archive(::Array{String,1}, ::VersionNumber, ::String) at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\site\v0.7\Pkg3\src\Operations.jl:350
 [7] macro expansion at C:\cygwin\home\Administrator\buildbot\worker\package_win64\build\usr\share\julia\site\v0.7\Pkg3\src\Operations.jl:451 [inlined]
 [8] (::getfield(Pkg3.Operations, Symbol("##21#24")){Pkg3.Types.Context,Dict{Base.UUID,Array{String,1}}})() at .\event.jl:92
```

The file:

![grafik](https://user-images.githubusercontent.com/1010467/37286418-2d4f077e-2602-11e8-94ac-4c9645be895c.png)

SimonDanisch:
Other packages work, PackageCompiler even after deleting all Pkg3 folders, reproducably fails...

[...]

PR #2641, opened 2021-06-23
download/unpack: some code cleanup

StefanKarpinski:
- Remove an unused import.
- Remove a try/catch that is no longer needed: The try/catch was introduced to deal with https://github.com/JuliaLang/Pkg.jl/issues/190 which is closed and is now handled automatically by `Tar.extract` which automatically detects whether it can create symlinks or not and simulates symlinks by copying when it can't create them.
- Move `list_tarball_files` into test utils since it's only used in tests

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -374,14 +374,6 @@ function unpack(
     Tar.extract(`$(exe7z()) x $tarball_path -so`, dest, copy_symlinks = copy_symlinks())
 end
 
-function list_tarball_files(tarball_path::AbstractString)
-    names = String[]
-    Tar.list(`$(exe7z()) x $tarball_path -so`) do hdr
-        push!(names, hdr.path)
-    end
-    return names
-end
-
 """
     package(src_dir::AbstractString, tarball_path::AbstractString)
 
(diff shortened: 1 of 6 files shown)


StefanKarpinski:
I was going to do some more here but it turned into a yak shave, so I might as well merge this by itself.

src/PlatformEngines.jl
```diff
@@ -40,14 +40,13 @@ function find7z()
     error("7z binary not found")
 end
 
-function probe_platform_engines!(;verbose::Bool = false)
```
KristofferC, line 43: https://juliahub.com/ui/Search?q=probe_platform_engines%21&type=symbols

Maybe let them break?

KristofferC (approved):

src/PlatformEngines.jl
StefanKarpinski: Yeah, they can pretty easily code around the absence or presence of this function. It will put a cap on compatible Julia versions for old releases of those packages, which is mildly annoying, but newer versions of the package can still support old and new Julias. Otoh, we could just leave this 🤷🏻‍♂️.

StefanKarpinski: I merged and opened issue on the packages that allow issues. NeXLCore doesn't allow issues so they don't get one.

PR #2656, opened 2021-07-01
Update Documenter and adjust CSS for dark theme.

fredrikekre:


DilumAluthge:
Out of curiosity, why `"~0.27"`? Isn't `"0.27"` equivalent here, since the major version is zero?

fredrikekre:
Yes, smaller diff.

PR #2658, opened 2021-07-02
Delete the `Pkg.Versions.:≳` function

DilumAluthge:
If I understand correctly, the current definitions of `Pkg.Versions.:≳` are incorrect, right?

Right now, `≳` is defined as such:
- `v ≳ b` if and only if `v ≲ b`
- `b ≳ v` if and only if `b ≲ v`

This PR instead defines `≳` as such:
- `v ≳ b` if and only if `b ≲ v`
- `b ≳ v` if and only if `v ≲ b`

### DIFF
--- a/src/Versions.jl
+++ b/src/Versions.jl
@@ -38,9 +38,6 @@ function ≲(b::VersionBound, v::VersionNumber)
     return (v.major, v.minor, v.patch) >= (b[1], b[2], b[3])
 end
 
-≳(v::VersionNumber, b::VersionBound) = v ≲ b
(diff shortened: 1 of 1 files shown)


KristofferC:
Is this used anywhere? Otherwise, can it be deleted?

DilumAluthge:
> Is this used anywhere? Otherwise, can it be deleted?

Oh, good point. I don't think it's used anywhere, because otherwise this incorrect definition would have shown up as a bug earlier.

And it's not part of the public API.

Let me try deleting it, and let's see if CI passes.

KristofferC (approved):

Issue #2657, opened 2021-07-02
`]precompile` in an empty environment updates the registry


giordano:
Quick way to reproduce the issue, at least with Julia v1.6.1:
```
(@v1.6) pkg> activate --temp
  Activating new environment at `/tmp/jl_kBXDD9/Project.toml`

(jl_kBXDD9) pkg> precompile
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
  No Changes to `/tmp/jl_kBXDD9/Project.toml`
  No Changes to `/tmp/jl_kBXDD9/Manifest.toml`
```
I wasn't expecting `precompile` to update the registry at all

DilumAluthge:
What if you do `Pkg.precompile(; allow_build = false)`?

PR #2660, opened 2021-07-04
During instantiate don't update the registry if there are no env deps

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2657

```
(Pkg) pkg> activate --temp
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_HuSQWs`

(jl_HuSQWs) pkg> precompile
  No Changes to `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_HuSQWs/Project.toml`
  No Changes to `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_HuSQWs/Manifest.toml`
```


IanButterworth:
I think this is good to go, but I'll give it another day or so in case someone can review

Issue #2633, opened 2021-06-18
Race condition in writing manifest_usage.toml?


Keno:
My manifest_usage.toml looks like it may have had a race condition in appending entries, corrupting the file and preventing `Pkg.gc` from running properly:
```
 [["/home/keno/.julia/environments/v1.1/Manifest.toml"]]
 time = 2020-03-03T15:27:50.063Z
 [["/home/keno/VeronaCompiler/Manifest.toml"]]
 time = 2020-03-03T15:27:50.065Z
 7:50.07Z
```
We should use advisory locking to prevent this from happening.

DilumAluthge:
Is the issue here that multiple threads of the same Julia process are writing to the same file? Or are multiple different Julia processes writing to the same file?

Keno:
Multiple different julia processes presumably.

staticfloat:
We write strings out in one big chunk, here:

https://github.com/JuliaLang/Pkg.jl/blob/1685f44ecc6f152c10341930d916a6cc64e1f1bd/src/Types.jl#L467-L476

Does anyone have a better idea on how to do this?

vtjnash:
Duplicate of https://github.com/JuliaLang/Pkg.jl/issues/684? (and others?)

FWIW, "append" does not actually put entries at the end of the file as it is emulated differently than the guarantees provided by posix (https://github.com/JuliaLang/julia/issues/3374)

IanButterworth:
As an alternative idea.. instead of writing a single toml, maintain a folder of symlinks to manifests that have been used and `touch` them each time the manifests are used, and remove old ones based on `mtime` on the symlinks? The symlink names could be either a doctored version of the path to remove slashes, or just a hash string of the path

KristofferC:
Is it really that hard to write a string to a file atomically? Instead of looking at workarounds, can't we just fix the problem?

staticfloat:
> As an alternative idea.. instead of writing a single toml, maintain a folder of symlinks to manifests that have been used and `touch` them each time the manifests are used, and remove old ones based on `mtime` on the symlinks? The symlink names could be either a doctored version of the path to remove slashes, or just a hash string of the path

I do wish we could use symlinks, but we also could jus that’s normal files with hashed file names and the path to the manifest as the content. That being said, I don’t like the idea of having a folder with hundreds of tiny files in it when we could have just one file.

IanButterworth:
I believe this is what @vtjnash is suggesting?
[...]

PR #2661, opened 2021-07-04
Make `write_env_usage` atomic

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2633

As per @StefanKarpinski's suggestion in https://github.com/JuliaLang/Pkg.jl/issues/2633#issuecomment-865861720

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -468,11 +468,28 @@ function write_env_usage(source_file::AbstractString, usage_filepath::AbstractSt
         TOML.print(io, Dict(source_file => [Dict("time" => now())]))
     end
 
-    # Append entry to log file in one chunk
+    ## Atomically write usage file
     usage_file = joinpath(logdir(), usage_filepath)
-    open(usage_file, append=true) do io
-        write(io, entry)
+    while true
+        # read existing usage file
+        existing_usage = isfile(usage_file) ? read(usage_file, String) : ""
+
+        # Write to a temp file in the same directory as the destination
+        temp_usage_file = tempname(logdir())
+        open(temp_usage_file, "w") do io
+            write(io, existing_usage)
+            write(io, entry)
+        end
+
+        # Move the temp file into place, replacing the original
+        mv(temp_usage_file, usage_file, force = true)
+
+        # Check that the new file has what we want in it
+        new_usage = read(usage_file, String)
+        occursin(entry, new_usage) && break
+        # If not, try again
     end
+
 end
 
 function read_package(path::String)
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -459,34 +459,57 @@ end
 function write_env_usage(source_file::AbstractString, usage_filepath::AbstractString)
     # Don't record ghost usage
     !isfile(source_file) && return
-
     # Ensure that log dir exists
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
How about also filtering it down either when we load it or when we write it back to only keep the most recent usage of each manifest path? Otherwise I think that as this file gets large this will start to be problematically slow. If you're just appending (which isn't atomic, as we've learned), then it makes sense to keep all the old entries, but since we're not doing that any more, let's clean up as we go.

IanButterworth:
I think that's implemented now. It tidied up my usage file nicely, and sorted it

src/Types.jl
```diff
@@ -459,20 +459,60 @@ end
...
+
+    ## Atomically write usage file
+    while true
+        # read existing usage file
+        usage = if isfile(usage_file)
+            content = read(usage_file)
+            Base.TOML.reinit!(p, String(content); filepath=usage_file)
+            Base.TOML.parse(p)
+        else
+            Dict()
```
KristofferC, line 477:
```suggestion
            Dict{String, Any}()
```

src/Types.jl
```diff
@@ -459,20 +459,60 @@ end
...
     usage_file = joinpath(logdir(), usage_filepath)
-    open(usage_file, append=true) do io
-        write(io, entry)
+    timestamp = now()
+
+    ## Atomically write usage file
+    while true
+        # read existing usage file
+        usage = if isfile(usage_file)
+            content = read(usage_file)
```
KristofferC, line 473: Isn't this block just `TOML.parse(p, usage_file)`?

KristofferC: I also think it is fine to just use the `TOML.parse(usage_file` API here. In the vast majority of cases you will only read this file twice so it isn't very useful to save the parser object.

src/Types.jl
```diff
@@ -463,15 +463,50 @@ function write_env_usage(source_file::AbstractString, usage_filepath::AbstractSt
...
+            TOML.parsefile(usage_file)
+        else
+            Dict{String, Any}()
+        end
+
+        # record new usage
+        usage[source_file] = [Dict("time" => timestamp)]
+
+        # keep only latest usage info
+        for kv in usage
```
KristofferC, line 482: `for k in keys(usage)`?

KristofferC (approved):

StefanKarpinski:
Just want to add that this is really nicely done and should make Pkg much more useable on shared file systems.

IanButterworth:
welp

StefanKarpinski:
I'm pretty sure that all that was required to fix this was adding `force=true` on this line:

https://github.com/JuliaLang/julia/blob/c5f348726cebbe55e169d4d62225c2b1e587f497/base/file.jl#L328

IanButterworth:
That'd be nice and simple!

Before that I'll try to add a multi-process test to Pkg master that reproduces the problem

KristofferC:
> I'm pretty sure that all that was required to fix this was adding force=true on this line:

Nah. See https://github.com/JuliaLang/julia/pull/42253#issuecomment-920335890.

Issue #2638, opened 2021-06-18
Versioned stdlibs are broken on `Pkg#master`


antimon2:
I built and tried Julia-1.7.0-beta1, and Pkg went wrong.
When I run `up`, it's explicitly added the version number `1.6.0` to Pkg in Manifest.toml, and then `resolve` shows an error.

````julia
(@v1.7) pkg> resolve
ERROR: Unsatisfiable requirements detected for package Pkg [44cfe95a]:
 Pkg [44cfe95a] log:
 ├─possible versions are: 1.7.0 or uninstalled
 └─restricted to versions 1.6.0 by an explicit requirement — no versions left
````

I tried to set the version number to `1.7.0` by directly editing Manifest.toml, but it reverts to `1.6.0` when I run `resolve` or `up` again.

What's wrong?

versioninfo():

```julia
Julia Version 1.7.0-beta1
Commit a15fbbc809 (2021-06-17 16:39 UTC)
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-12.0.0 (ORCJIT, skylake)
```

Is there any other information I should show?

KristofferC:
I guess this is fixed.

PR #2662, opened 2021-07-04
Add historical stdlib check CI job

IanButterworth:
The historical stdlib entries are currently out of date but tests are passing. So the tests don't seem to cover all requirements.
Here's a CI job to run `generate_historical_stdlibs.jl` and check for changes.

I'm not sure this should be run every commit though?

Perhaps the smart thing to do would be to cron this and automatically open a PR with the changes, but I didn't want to create that noise before checking this was the right thing to do

Related to https://github.com/JuliaLang/Pkg.jl/issues/2638


ext/HistoricaStdlibGenerator/Project.toml
```diff
@@ -1,6 +1,6 @@
 [deps]
 Downloads = "f43a241f-c20a-4ad4-852c-f6b1247861c6"
 JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
-Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78e"
+Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
```
IanButterworth, line 4: Note that I changed it to use the script in the nightly Pkg rather than the checked-out version. I think that's ok?

staticfloat:
Yeah, I think this is desirable!  I'm not sure why the tests are insufficient though....  What is the situation that can arise to break things?

IanButterworth:
I thought this was the issue behind https://github.com/JuliaLang/Pkg.jl/issues/2638?

staticfloat:
No, I don't think so.  I think that issue is purely because some parts of the resolver don't like stdlibs having version numbers.

KristofferC:
I am not sure about that, the issue says:

>  When I run up, it's explicitly added the version number 1.6.0 to Pkg in Manifest.toml, and then resolve shows an error.

But that was on version 1.7 (where Pkg is at version 1.7). So something seems out of date.

IanButterworth:
Closing given better fix in #2670

PR #2663, opened 2021-07-05
Only warn about a manifest being the old format once

IanButterworth:
This uses `maxlog = 1` and assigning all the manifest warnings the same `_id` to ensure the user only gets one warning per manifest file

Note that I'm setting the `_file` of the warning to the manifest file, so that the path at the bottom of the warning indicates the manifest path.
This is what the `@warn`'s in `check_warn_manifest_julia_version_compat` do already. Unfortunately setting `_line = 0` doesn't hide the line suffix.. that would be nice.

I'm activating my 1.6 env here because it's the old format.

Previously this would've warned twice

```julia
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.0-DEV.123 (2021-07-02)
 _/ |\__'_|_|_|\__'_|  |  Commit 5584620db8 (2 days old master)
|__/                   |

julia> import Pkg

(Pkg) pkg> activate @v1.6
  Activating project at `~/.julia/environments/v1.6`

julia> using CSV
 │ Package CSV not found, but a package named CSV is available from a registry. 
 │ Install package?
 │   (@v1.6) pkg> add CSV 
 └ (y/n) [y]: 
   Resolving package versions...
┌ Warning: The active manifest file has an old format that does not record the julia version. The dependencies
│ resolved in this manifest may be incompatible with the current julia version.
│ To update to the new format run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.
└ @ ~/.julia/environments/v1.6/Manifest.toml:0
    Updating `~/.julia/environments/v1.6/Project.toml`
  [336ed68f] + CSV v0.8.5
    Updating `~/.julia/environments/v1.6/Manifest.toml`
  [336ed68f] + CSV v0.8.5
  [9a962f9c] + DataAPI v1.7.0
  [e2d170a0] + DataValueInterfaces v1.0.0
  [e2ba6199] ↑ ExprTools v0.1.3 ⇒ v0.1.4
  [82899510] + IteratorInterfaceExtensions v1.0.0
  [2dfb63ee] + PooledArrays v1.2.1
  [91c51154] + SentinelArrays v1.3.3
  [3783bdb8] + TableTraits v1.0.1
  [bd369af6] + Tables v1.4.4
  [9fa8497b] + Future
  [8dfed614] + Test

julia> 


```

### DIFF
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -284,8 +284,9 @@ function write_manifest(env::EnvCache)
 end
 function write_manifest(manifest::Manifest, manifest_file::AbstractString)
     if manifest.manifest_format.major == 1
-        @warn """The active manifest file at `$(manifest_file)` has an old format that is being maintained.
-            To update to the new format run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.""" maxlog = 1 _id = Symbol(manifest_file)
+        @warn """The active manifest file has an old format that does not record the julia version. The dependencies
+            resolved in this manifest may be incompatible with the current julia version.
+            To update to the new format run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.""" maxlog = 1 _id = Symbol(manifest_file) _file = manifest_file _line = 0 _module = nothing
     end
     return write_manifest(destructure(manifest), manifest_file)
 end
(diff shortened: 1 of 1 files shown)


IanButterworth:
Maybe the guidance should be different here though..
```
To update to the new format run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.
```
That would put the user into a state where they have the new manifest format, but the `julia_version` would still be missing, as a re-resolve hasn't happened.

Perhaps
```
To update to the new format run `Pkg.upgrade_manifest()` which will upgrade the format, then re-resolve with `Pkg.resolve()`.
```

src/manifest.jl
```diff
@@ -284,8 +284,9 @@ function write_manifest(env::EnvCache)
 end
 function write_manifest(manifest::Manifest, manifest_file::AbstractString)
     if manifest.manifest_format.major == 1
-        @warn """The active manifest file at `$(manifest_file)` has an old format that is being maintained.
-            To update to the new format run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.""" maxlog = 1 _id = Symbol(manifest_file)
+        @warn """The active manifest file has an old format that does not record the julia version. The dependencies
```
IanButterworth, line 287: Perhaps allow an override?
```suggestion
        do_warn = tryparse(Int, get(ENV, "JULIA_PKG_OLD_MANIFEST_WARN", "0")) == 1
        do_warn && @warn """The active manifest file has an old format that does not record the julia version. The dependencies
```

IanButterworth:
Bump @KristofferC

IanButterworth:
Evidently this isn't a pain point so closing

PR #2668, opened 2021-07-11
Allow control of --bounds-check in Pkg.test

IanButterworth:
Forcing bounds checking to be on can severely slow down some test code, so it would be useful to be able to allow bounds checking to be the julia default automatic that observes in-code declarations (set here via `check_bounds = nothing` which omits the `--check-bounds` option entirely).

There may be some use cases where forcing it off entirely may be desired too (`check_bounds = false`), so all three are supported

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -403,7 +403,7 @@ function free(ctx::Context, pkgs::Vector{PackageSpec}; all_pkgs::Bool=false, kwa
 end
 
 function test(ctx::Context, pkgs::Vector{PackageSpec};
-              coverage=false, test_fn=nothing,
+              coverage=false, check_bounds=true, test_fn=nothing,
               julia_args::Union{Cmd, AbstractVector{<:AbstractString}}=``,
               test_args::Union{Cmd, AbstractVector{<:AbstractString}}=``,
               force_latest_compatible_version::Bool=false,
@@ -427,6 +427,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
         ctx,
         pkgs;
         coverage,
+        check_bounds,
         test_fn,
         julia_args,
         test_args,
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1358,6 +1358,7 @@ end
 
 function gen_test_code(testfile::String;
         coverage=false,
+        check_bounds=true,
         julia_args::Cmd=``,
         test_args::Cmd=``)
     code = """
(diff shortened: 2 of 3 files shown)


DilumAluthge:
Instead of adding another keyword argument to the `Pkg.test` function, could we do either of the following:
1. Just read the value of `--bounds-check` that was passed to the original (parent) Julia process.
2. Allow the user to pass `--bounds-check` to the `julia_args` keyword argument that we already have.

IanButterworth:
> Just read the value of --bounds-check that was passed to the original (parent) Julia process.

This would be a significant (breaking?) change, as the parent process is likely run by default with `--check-bounds` omitted, i.e. default checking that observes code declarations

> Allow the user to pass --bounds-check to the julia_args keyword argument that we already have.

Users can currently do this, but it's not possible to set it to the default, as that requires omitting `--check-bounds` entirely.


If Base was changed to accept `--check-bounds=auto` then the latter could work

DilumAluthge:
> If Base was changed to accept `--check-bounds=auto` then the latter could work

Regardless of what we end up doing in this PR, I do think it would be nice for Julia to accept `--check-bounds=auto`.

IanButterworth:
Yeah and it would be nice to get that into 1.7

fredrikekre:
Closing in favor of https://github.com/JuliaLang/julia/pull/41551 then? In any case, if you need special behavior like this you can always start your own process from within the `test/runtests.jl` file. Many packages do that already for thigns like this.

KristofferC:
> Users can currently do this, but it's not possible to set it to the default, as that requires omitting --check-bounds entirely.

I don't understand why this is not possible. In case you want it as the default you just omit it in `julia_args`?

IanButterworth:
By default I mean julia default, which respects `@inbounds`, which is a third state that requires omitting `--check-bohnds` entirely. 

> you can always start your own process from within the test/runtests.jl file

Yes but that makes managing test dependencies a pain, as there's no convenient way to merge a test Project and the parent in the same way that the sandbox does it, AFAIK

fredrikekre:
> Yes but that makes managing test dependencies a pain, as there's no convenient way to merge a test Project and the parent in the same way that the sandbox does it, AFAIK

You just launch julia with the env Pkg gave `test/runtests.jl`

KristofferC:
> By default I mean julia default, which respects @inbounds, which is a third state that requires omitting --check-bohnds entirely.

I understand that. But again, why isn't `julia_args` fine here, and just leave out the argument if you want the julia default?

IanButterworth:
> just leave out the argument if you want the julia default?
```
% julia --check-bounds
ERROR: option `--check-bounds` is missing an argument
% julia --project --check-bounds=
ERROR: julia: invalid argument to --check-bounds={yes|no} ()
```
or am I misunderstanding your suggestion?

> You just launch julia with the env Pkg gave test/runtests.jl

The env Pkg gave `test/runtests.jl` is a merge of the two envs in a temp dir sandbox. How would I access that? or do you just mean launch with `--project` in `/test`?

KristofferC:
> or am I misunderstanding your suggestion?

Yes, I think so. Just don't add the whole `--check-bounds` bit at all.

IanButterworth:
Are you factoring in that it's currently hardcoded before `julia_args` in the sandbox? 
https://github.com/JuliaLang/Pkg.jl/blob/cb9441fa7af3e82f9dbb0d3123cf1a0d5e772b23/src/Operations.jl#L1374

KristofferC:
No, that would make it harder indeed :laughing:.

IanButterworth:
Yeah. I'm happy for that PR to be the one

PR #2670, opened 2021-07-12
Push forward "versioned stdlibs" work

staticfloat:
This updates the historical stdlibs list, and threads the `version` field through a bit more, including changing `Pkg.Types.stdlibs()` to return a dict mapping `uuid => (name, version)` to match.  This makes it a lot easier to test whether the historical stdlibs list has fallen out of date, and should solve https://github.com/JuliaLang/Pkg.jl/issues/2638 by causing the Pkg tests to fail much earlier.


KristofferC:
What I didn't get about the original issue is that it was running with no explicitly given `julia_version` and thereby this cache shouldn't have been used and the version should have been read from the Project file (which did have the 1.7 version and not 1.6). So is there somewhere we read the data here even though we shouldn't?

KristofferC:
Bump. I am still confused how this could be broken on default Julia usage (when this information should not be used, AFAIU).

IanButterworth:
@staticfloat I missed this when doing https://github.com/JuliaLang/Pkg.jl/pull/2795 which appears to fully cover this PR? I guess this can be closed?

PR #2673, opened 2021-07-20
Give examples of using `julia_args`, `test_args`

johnomotani:
In the docstring for `Pkg.test()`, give usage examples for the `julia_args` and `test_args` keyword arguments, as discussed here https://discourse.julialang.org/t/how-to-pass-use-julia-args-argument-to-pkg-test/64274/2.

Maybe it would be good for the `test_args` example to be more concise? The current one was the reason why I wanted `test_args`, so it's what I can think of.


src/Pkg.jl
```diff
@@ -218,7 +218,32 @@ test = ["Test"]
...
+end
+```
+
+Then some tests that take a long time to run could be marked with the macro like `@long @test foo == bar`
+and they would only be run if the argument `--long` was passed either to Julia on the command line (or by
+modifying `ARGS`), or by passing it using `test_args`, e.g.
+
+```
+Pkg.test("foo"; test_args=["--long"])
+```
```
IanButterworth, line 246: I think a simpler example would be better. 
```suggestion
tests in some way. For example, the tests could have optional additional tests:

```
if "extended" in ARGS
    @test some_function()
end
```
which could be enabled by testing with
```
Pkg.test("foo"; test_args=["extended"])
```
```

PR #2676, opened 2021-07-23
Docs: Mention automatic precompilation and link to more info

Datseris:
(in the docstring of `Pkg.update`, whose call is the most likely first place someone will see precompilation happen.)


KristofferC:
Isn't `add` the first place they would see this happening?

Datseris:
Yeah I guess you are right. So, should I put it in both `add` and `update`, or only `add`? (I'd vote for both)

KristofferC:
I am not sure. It feels a little bit excessive to add this to a bunch of functions since the precompile step runs after almost anything (`instantiate`, `build`, `add`, `test`) etc. It is already documented in `precompile` so perhaps it is enough to just add a bread crumb to get people to look there?

Datseris:
Then I think the smallest we can do is to add "`Pkg.update` automatically calls [`Pkg.precompile`](`@ref`) after its call." as it cross links to the docstring. I'd say it is not so excessive to add this small sentence to the most common scenarios, which (without much knowledge) I'd say are `add, update, isntantiate`.

IanButterworth:
@Datseris I hope you don't mind, I pushed what I believe is agreed here

PR #2678, opened 2021-07-28
Enable switching to the previous environment

IanButterworth:
Adds `pkg> activate -` and `Pkg.activate(prev=true)` to switch to the previously active environment. Matches `cd -` and `git checkout -` behavior.

There was a suggestion to instead do this with a blank `pkg> activate` but for me at least that's muscle memory for getting to the `@x.x` env.

```julia
(Pkg) pkg> activate --temp
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_d8KZ5n`

(jl_d8KZ5n) pkg> activate -
  Activating project at `~/Documents/GitHub/Pkg.jl`

(Pkg) pkg> activate -
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_d8KZ5n`

(jl_d8KZ5n) pkg> activate
  Activating project at `~/.julia/environments/v1.8`

(@v1.8) pkg> activate -
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_d8KZ5n`

(jl_d8KZ5n) pkg> activate -
  Activating project at `~/.julia/environments/v1.8`

(@v1.8) pkg> activate .
  Activating project at `~/Documents/GitHub/Pkg.jl`

(Pkg) pkg> activate -
  Activating project at `~/.julia/environments/v1.8`

(@v1.8) pkg> activate -
  Activating project at `~/Documents/GitHub/Pkg.jl`

(Pkg) pkg>
```

Note that the previously used environment is looked up from the manifest usage file on startup, which is quick for my 6KB usage file. 
And this should be quick generally given the usage file is now minimized with https://github.com/JuliaLang/Pkg.jl/pull/2661
```julia
julia> @btime Pkg.get_last_used_env_from_usage()
  85.058 μs (910 allocations: 55.55 KiB)
"/Users/ian/Documents/GitHub/Pkg.jl"
```

Will add tests if this goes forward

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -12,7 +12,7 @@ using Serialization
 
 import ..depots, ..depots1, ..logdir, ..devdir, ..printpkgstyle
 import ..Operations, ..GitTools, ..Pkg, ..Registry
-import ..can_fancyprint, ..pathrepr, ..isurl
+import ..can_fancyprint, ..pathrepr, ..isurl, ..PREV_ENV_PATH
 using ..Types, ..TOML
 using ..Types: VersionTypes
 using Base.BinaryPlatforms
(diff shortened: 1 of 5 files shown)


KristofferC:
I like this but I think using the previous session is a bit overkill, or? That seems like it would never be used.

IanButterworth:
I just pushed a tweak on where the try-catch is, but happy to remove the previous session stuff. I'll leave it a few days

KristofferC:
Could perhaps use a test and then should be ready to go.

KristofferC (approved):

PR #2680, opened 2021-07-31
Mention other contibutors in LICENSE

oxinabox:
Other places in our stdlibs give a few contributors then a list.
It seemed odd that this mentioned only @StefanKarpinski .
Contrast [Statistics.jl](https://github.com/JuliaLang/Statistics.jl/blob/master/LICENSE.md)

> Copyright (c) 2012-2016: Jeff Bezanson, Stefan Karpinski, Viral B. Shah, Dahua Lin, Simon Byrne, Andreas Noack, Douglas Bates, John Myles White, Simon Kornblith, and other contributors.

and

[Julia itself](https://github.com/JuliaLang/julia/blob/master/LICENSE.md):

> Copyright (c) 2009-2021: Jeff Bezanson, Stefan Karpinski, Viral B. Shah, and other contributors: https://github.com/JuliaLang/julia/contributors

Could go and list a whole bunch of names.
But  @KristofferC  and  @StefanKarpinski   are by a nontrivial margin the largest contributors, on all metrics.
After that it is a bit more mixed depending what metric.

Could also just say `Copyright Pkg.jl contributors` I guess?

### DIFF
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,6 @@
 The Pkg.jl package is licensed under the MIT "Expat" License:
 
-> Copyright (c) 2017-2021: Kristoffer Carlsson, Stefan Karpinski, and other contributors:
+> Copyright (c) 2017-2021: Stefan Karpinski, and  contributors:
 > https://github.com/JuliaLang/Pkg.jl/graphs/contributors
 >
 > Permission is hereby granted, free of charge, to any person obtaining a copy
--- a/LICENSE.md
+++ b/LICENSE.md
@@ -1,6 +1,7 @@
 The Pkg.jl package is licensed under the MIT "Expat" License:
 
-> Copyright (c) 2017: Stefan Karpinski.
+> Copyright (c) 2017-2021: Kristoffer Carlsson, Stefan Karpinski, and other contributors:
+> https://github.com/JuliaLang/Pkg.jl/graphs/contributors
 >
 > Permission is hereby granted, free of charge, to any person obtaining a copy
 > of this software and associated documentation files (the "Software"), to deal


KristofferC:
I'm not sure having any names at all is useful? Is there some legal requirement to have names? I think just having "contributors" is the best, if that is OK.

LICENSE.md
```diff
@@ -1,6 +1,7 @@
 The Pkg.jl package is licensed under the MIT "Expat" License:
 
-> Copyright (c) 2017: Stefan Karpinski.
+> Copyright (c) 2017-2021: Kristoffer Carlsson, Stefan Karpinski, and other contributors:
```
oxinabox, line 3:
```suggestion
> Copyright (c) 2017-2021: Pkg.jl contributors:
```

StefanKarpinski:
Feels weird to remove a name from the attribution (I think I'd fee the same way even if it wasn't mine), so maybe just add "and contributors"? Changing it to just "contributors" would also be ok though.

LICENSE.md
oxinabox:
```suggestion
> Copyright (c) 2017-2021: Stefan Karpinski, and  contributors:
```

oxinabox:
Someone with merge rights, pick the version you wanted (I put both options as suggestions) and merge if happy?

DilumAluthge:
I would go with "Stefan Karpinski and contributors".

KristofferC:
Let's just add the top 10 contributor names in alphabetic order then or something? Or should it be in the order of number of commits? But then it has to be updated. I just don't see a reason to have explicit names here, you can just click on the contributor list to see the level of contributions.

DilumAluthge:
I think it's not uncommon for packages to have "Copyright (c) 20xx CREATOR and contributors", where CREATOR is the original creator of the package.

DilumAluthge:
I don't feel particular strongly, so we could just change it to "Copyright (c) 20xx Pkg contributors".

I do agree that we don't want to have to continuously update this.

DilumAluthge:
> Feels weird to remove a name from the attribution — I think I'd feel the same way even if it wasn't mine, so maybe just add "and contributors"? That said, changing it to just "contributors" would also be ok with me.

I also think it's a little weird to remove a name.

But yeah I don't feel too strongly either way.

oxinabox:
This is such a bike shed.
Noone cares strongly either way, as long as we have the link to the contributors list.
And yet here is a 10 comment thread discussing options.

DilumAluthge:
I've merged this with the premise that "good enough" is better than nothing. We can obviously tweak it further.

PR #2687, opened 2021-08-09
Fix list formatting in PackageSpec documentation

KnutAM:
Fix of formatting error in [PackageSpec](https://pkgdocs.julialang.org/dev/api/#Pkg.PackageSpec) docstring breaking the list layout


KristofferC:
Nice catch!

PR #2695, opened 2021-08-11
Fix invalidations from `Pair(obj)` methods

timholy:
xref https://github.com/JuliaGraphs/LightGraphs.jl/pull/1581

These are mostly julia#15276-related, with some julia#36454 thrown in for good measure.


timholy:
CC @ChrisRackauckas

KristofferC (approved):

KristofferC:
I'll just never use a `do` block again :)

timholy:
`do` blocks are wonderful, so I'd say use 'em if you want. julia#15276 is a bummer for sure, but we should just write some tooling to detect them automatically.

PR #2700, opened 2021-08-13
Also include changes to HistoricalStdlibs

timholy:
This is a forgotten-followup (and fix :sweat_smile: ) to #2699 (sorry).


src/API.jl
```diff
@@ -1387,7 +1387,7 @@ function recall_precompile_state()
         if isfile(fpath)
             open(fpath) do io
                 try
-                    pkgspecs = deserialize(io)
+                    pkgspecs = deserialize(io)    # what type does this have?
```
KristofferC, line 1390: I think a vector of `PackageSpec`. @IanButterworth probably knows.

KristofferC (approved):

src/API.jl
timholy: Bump @IanButterworth

IanButterworth: Yes sorry, didn't get notified for some reason. It's always a `Vector{PackageSpec}`

Issue #341, opened 2018-06-05
Pkg REPL compatibility bounds


staticfloat:
We need a way to specify compatibility bounds when adding packages, interactively through the REPL mode.  I'm thinking something like `pkg> add Compat@[0.55.0+]`, or perhaps `pkg> add Compat` followed by `pkg> restrict Compat 0.55.0-0.66.0`

What do you guys think?

KristofferC:
I mean, you can already do something like

```
(Foo) pkg> add Compat@0.55
  Updating registry at `~/.julia/registries/Uncurated`
  Updating git-repo `https://github.com/JuliaRegistries/Uncurated.git`
 Resolving package versions...
  Updating `Project.toml`
  [34da2185] + Compat v0.55.2
```

Perhaps that should also put `Compat = "0.55"` into the compatibility section?

KristofferC:
Should probably show the `compatibility` bounds in the status output as well

staticfloat:
Maybe we should allow “add Compat@0.55.0-“ to set a lower bound, as that’s the most common use case. Maybe go whole hog and allow “add Compat@0.55.0-0.66.0”

StefanKarpinski:
`Compat = "0.55"` in the compatibility section means `Compat = "^0.55.0"`.

oxinabox:
I propose that the API for this be based around the versions of packages currently being used in the Manifest.
This would be partnered with #1607 which would allow things to ignore the current compat settings.

For sake of example, lets call this function `write_compat`
And it will take options:
 - `extend::Bool` which if true (default?) will *add* the compatibility to what is already there, and if false will clear all current compat sections and just write what you have to be there.
 - `specificy` being one of `^, ~, ==, <=, >=, ` etc (with `^` being the default), which controls what restrictor is written.
- `packages` (optional) which packages to include. Defaults to all packages in Project.toml

So if I just created a new project for some data science I might do:
```
pkg"add CSV, DataFrames, Plots"
write_compat()
```
which will give me a Project.toml withat has those packages and normal (carat) bound to the current versions. Which is what i normally do when I start a new project.

If I want to later restrict to also support a newly released CSV v0.6 I would do:
```
pkg"add CSV@0.6 --ignore-compat"
write_compat("CSV")

pkg"resolve"  # make sure that we no longer are needing to `--ignore-compat`
```

If I became paranoid about thing breaking, and i wanted to lock all the versions down to what i know currently works.
```
[...]

Issue #2463, opened 2021-03-29
Automatically add `compat` entries according to current installed versions


dpsanders:
`compat` entries should automatically be added according to the current minor version of the relevant package whenever that package is `Pkg.add`ed to an environment.

ufechner7:
This cannot be so difficult... What is missing? Do we want that this works in the way described by dpsanders? 
I think that would be very useful for beginners to avoid that they upgrade packages and brake their code because they added a package.

ufechner7:
Is this a duplicate, or what is different: https://github.com/JuliaLang/Pkg.jl/issues/776 ?

PR #2702, opened 2021-08-15
Add project compat entry viewing & editing

IanButterworth:
Updated: Was `-e|--edit` and defaulted to the static version.

Adds a Project compat section viewer via `pkg> compat` and an interactive editor via `pkg> compat -l|--list`.

Also some API endpoints
- `Pkg.compat!()`: equivalent to `pkg> compat` (errors early if run in a non-interactive terminal)
- `Pkg.compat()`: equivalent to `pkg> compat -l`
- `Pkg.compat("Foo")`: Return compat entry for Foo (`nothing` if no entry)
- `Pkg.compat!("Foo", "1.2")`: Set entry for Foo (use `nothing` or `""` to remove the entry)

I think there a few benefits to adding this:
- compat entries can be edited without leaving the repl
- it validates that new entries are valid compat version specifiers before they're written to file
- could be expanded to do more stuff i.e.
   - an option to add the current installed version(s) (https://github.com/JuliaLang/Pkg.jl/issues/2463)
   - an option to add missing versions from the registry, like CompatHelper but local
   - show available versions of packages in the editor for reference
   - highlight gaps in compat entry lists to reduce the chance of accidental restriction i.e. `0.4,0.5,0.7,0.8`
   - add tips and/or a link to docs on [version specifier formats](https://pkgdocs.julialang.org/v1/compatibility/#Version-specifier-format)  

Todo (if approved of)
- [x] Tests (the interactive editor isn't testable so the codecov diff misses)
- [x] Edit entry rather than retype. Implemented via raw stdin manipulation

Screen Shot 2021-08-15 at 12 04 40 AM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,12 @@ Pkg v1.8 Release Notes
 
 - New `outdated::Bool` kwarg to `Pkg.status` (`--outdated` or `-o` in the REPL mode) to show
   information about packages not at the latest version.
-- Pkg now only tries to download pacakges from the package server in case the
+- New `compat::Bool` kwarg to `Pkg.status` (`--compat` or `-c` in the REPL mode) to show any [compat]
+  entries in the Project.toml.
+- New `pkg> compat` (and `Pkg.compat`) mode for setting Project compat entries. Provides an interactive editor
+  via `pkg> compat`, or direct entry manipulation via `pkg> Foo 0.4,0.5` which can load current entries via tab-completion.
+  i.e. `pkg> compat Fo<TAB>` autocompletes to `pkg> Foo 0.4,0.5` so that the existing entry can be edited.
+- Pkg now only tries to download packages from the package server in case the
   server tracks a registry that contains the package.
 
 Pkg v1.7 Release Notes
@@ -14,7 +19,7 @@ Pkg v1.7 Release Notes
   Julia 1.6.2 is compatible with the new format.
 - Registries downloaded from the Pkg Server (not git) are no longer uncompressed into files but instead read directly from the compressed tarball into memory. This improves performance on
   filesystems which do not handle a large number of files well. To turn this feature off, set the environment variable `JULIA_PKG_UNPACK_REGISTRY=true`.
-- It is now possible to use an external `git` executable instead of the default libgit2 library for 
+- It is now possible to use an external `git` executable instead of the default libgit2 library for
   the downloads that happen via the Git protocol by setting the environment variable `JULIA_PKG_USE_CLI_GIT=true`.
 - Registries downloaded from the Pkg Server (not git) is now assumed to be immutable. Manual changes to their files might not be picked up by a running Pkg session.
 - The number of packags precompiled in parallel are now limited to 16 unless the
--- a/src/API.jl
+++ b/src/API.jl
@@ -9,6 +9,7 @@ using Dates
 import LibGit2
 import Logging
 using Serialization
+using REPL.TerminalMenus
 
 import ..depots, ..depots1, ..logdir, ..devdir, ..printpkgstyle
 import ..Operations, ..GitTools, ..Pkg, ..Registry
@@ -1522,8 +1523,14 @@ end
 
 @deprecate status(mode::PackageMode) status(mode=mode)
 
-function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=PKGMODE_PROJECT, outdated::Bool=false, io::IO=stdout, kwargs...)
-    Operations.status(ctx.env, ctx.registries, pkgs; mode, git_diff=diff, io, outdated)
+function status(ctx::Context, pkgs::Vector{PackageSpec}; diff::Bool=false, mode=PKGMODE_PROJECT, outdated::Bool=false, compat::Bool=false, io::IO=stdout, kwargs...)
+    if compat
+        diff && pkgerror("Compat status has no `diff` mode")
+        outdated && pkgerror("Compat status has no `outdated` mode")
+        Operations.print_compat(ctx, pkgs; io)
+    else
+        Operations.status(ctx.env, ctx.registries, pkgs; mode, git_diff=diff, io, outdated)
+    end
     return nothing
 end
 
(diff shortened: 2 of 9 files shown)


johnnychen94:
It would be easier to realize the edit mode if there's one line description after/before `pkg> compat` output, e.g., "To interactively edit entries values, use `pkg> compat -e`"

Another question: is the compat edit affecting immediately, or it saves onto some in-memory struct and only changes the "Project.toml" until you quit the edit mode?

IanButterworth:
Yeah I wondered about adding a note on edit mode. It's not done for any other `pkg>` mode though, so it might feel OTT. I'll add one to see how it feels

> Another question: is the compat edit affecting immediately, or it saves onto some in-memory struct and only changes the "Project.toml" until you quit the edit mode? I'm thinking of the second one because it provides some rollback mechanism (sort of dry run) besides fewer IO operations.

It acts immediately on the file but complies with the `pkg> undo` functionality, so everything can be rolled back or forward with `pkg> redo`. That's the standard practice across Pkg.

IanButterworth:
I think this will get annoying quickly
Screenshot from 2021-08-15 16-46-18

Perhaps julia/pkg could have a tip mechanism that only tips the first time a function is used. There could be a dict stored on disk.. but that's a whole thing

StefanKarpinski:
This is very cool and useful. Let's punt on the tip thing for now. People can RTFM or be told about `compat -e`. Alternatively, make edit mode the default (since it does view if you do nothing) and have a different flag to tell the `compat` command just to dump the current compat entries.

IanButterworth:
Yeah that sounds good. I've switched `pkg> compat` over to the interactive edit mode, and `pkg> compat -l|--list` for the static list.
Also dropped the tip

IanButterworth:
Seems worth a go? My only remaining thought is whether the API occupies space that might be better used for something else?

DilumAluthge:
Personally, I can't think of another meaning for `] compat` than either "show `[compat]` entries" or "edit `[compat]` entries".

DilumAluthge:
I know that @GunnarFarneback has built similar functionality in a package. He might be able to comment on any lessons or best practices that have been learned from the package and its users.

DilumAluthge:
Also, there is at least one open issue requesting this feature, so I think that Paul request would close that issue.

IanButterworth:
> Also, there is at least one open issue requesting this feature, so I think that this pull request would close that issue.

Maybe https://github.com/JuliaLang/Pkg.jl/issues/341 ?
https://github.com/JuliaLang/Pkg.jl/issues/2463 is related

Otherwise I couldn't find anything else

DilumAluthge:
Yeah I think this closes #341.

GunnarFarneback:
> I know that @GunnarFarneback has built similar functionality in a package. He might be able to comment on any lessons or best practices that have been learned from the package and its users.

Yes, PackageCompatUI.

The feedback can mostly be summarized as 50 likes on the [discourse announcement](https://discourse.julialang.org/t/ann-packagecompatui/58118), 23 stars on the [Github repo](https://github.com/GunnarFarneback/PackageCompatUI.jl) and an appreciation on slack.

IanButterworth:
Oh cool! The version selection through radio menu and automatic collapsing of version spec is really nice! It would've been nice if I'd found that sooner 😬

I'm not sure what to do now. Is there a reason you didn't consider developing that as part of Pkg @GunnarFarneback ?

Some thoughts if this PR goes forward:
- The way this PR prints the info as part of the menu seems like a better fit for how Pkg displays info
- Incorporating the version selection radio menu system could be good, but I think also just letting the user edit the text and have that validated might be ok as a default?
- Highlighting when a compat entry doesn't include the latest version of the package is great, and that should be added

KristofferC:
Interactivity is cool but for the specific case of adding compat, I don't really see that big of an advantage over just specifying the package and the compat string directly (or just using a normal text editor). Let's take `rm` as an example, you just give a list of package names to it. It doesn't provide a list where you interactively mark what packages to remove etc. With tab-completion such an interface seems a lot quicker than a menu-driven one (and it is easier to test etc)

Some other thoughts:

One question is if `compat` should be `status --compat`?

> Highlighting when a compat entry doesn't include the latest version of the package is great, and that should be added

The `status --outdated` shows this as well so it should be fairly easy to use the implementation there.

GunnarFarneback:
The main reason for not doing it inside Pkg is that I wanted the feature to show the registration date of package versions, which seemed unreasonable to include in Pkg as it requires a git clone of the registry and is somewhat slow, plus more freedom to try things out.

> Incorporating the version selection radio menu system could be good, but I think also just letting the user edit the text and have that validated might be ok as a default?

The primary advantage of the menu selection is that the user gets a correct compat entry without knowing the version specifier format. The main disadvantage is that some things can't really be expressed that way, e.g. compat with Julia 1.7 before it's released.

IanButterworth:
`pkg> status --compat` is a good suggestion. I'll do that. 

I'll add the basic `pkg> compat Foo 0.2,0.3,0.4` editing form too. 

I still think there's a benefit to prompting the edit with the current string for manipulation, as it would be a lot to expect the user to retype/copy-paste the current entry in, to insert a new version. It may also also be error-prone. 

Perhaps a form `pkg> compat Foo` that would launch the prompt directly would be worth adding too.

All of those additions could be CI tested. Come to think of it, so can the full menu, by pre-loading the stdin buffer before running the menu.

IanButterworth:
> Perhaps a form `pkg> compat Foo` that would launch the prompt directly would be worth adding too.

Tab completion! That's the way to do it probably. `pkg> compat Foo<tab>` completes the current entry for editing.

StefanKarpinski:
> I still think there's a benefit to prompting the edit with the current string for manipulation, as it would be a lot to expect the user to retype/copy-paste the current entry in, to insert a new version. It may also also be error-prone.

Strongly agree with this. I'd be happy to run this to capture the current versions as lower bounds but I'd be annoyed to have to look up what version each dep is on and add the bound manually.

IanButterworth:
Update
`pkg> status --compat` now prints the full compat entries, or `Pkg.status(compat = true)`
`pkg> status --compat Foo Bar julia` now prints the specific entries for Foo, Bar, julia

`pkg> compat` launches the interactive editor
`pkg> compat Foo 0.8,0.9` now supports tab completion, so you can type `pkg> compat Fo` and it will complete to `pkg> compat Foo 0.8,0.9` so you can edit the entry without retyping it
`pkg> compat Foo` will remove the entry, equivalent to `Pkg.compat("Foo", nothing)`

IanButterworth:
Mergeable?

IanButterworth:
I think there's support in this current form, so I'll merge this at the weekend if I don't hear otherwise

Issue #2698, opened 2021-08-13
Pkg puts in wrong dependencies for versioned stdlibs 


KristofferC:
If you start a new environment and add `p7zip_jll` you get an entry in the manifest:

```toml
[[deps.p7zip_jll]]
deps = ["Artifacts", "JLLWrappers", "Libdl", "Pkg"]
uuid = "3f19e933-33d8-53b3-aaab-bd5110c3b7a0"
version = "16.2.1+1"
```

But the `p7zip_jll`  does not depend on Pkg: https://github.com/JuliaLang/julia/blob/master/stdlib/p7zip_jll/Project.toml.

What seems to happen is that this check:

https://github.com/JuliaLang/Pkg.jl/blob/b1ec6de5417da990a75c186a24f88fb801657d12/src/Operations.jl#L416

fails and then Pkg goes and looks in the registry for compat info for the package.

cc @staticfloat since I am quite sure this got introduced in https://github.com/JuliaLang/Pkg.jl/pull/2614 with the extra `stdlib_version === nothing` check added.

PR #2707, opened 2021-08-17
ensure stdlibs do not take dependencies from the registry

KristofferC:
Since stdlibs are versioned now, the presence of a version cannot be used to determine if one wants to treat an stdlib as a normal package.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/2698

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -2767,6 +2767,13 @@ end
         @test linalg_block["uuid"] == linalg_uuid
         @test !haskey(linalg_block, "version")
     end
+
+    isolate(loaded_depot=true) do
+        # Next, test that stdlibs do not get dependencies from the registry
+        Pkg.add("p7zip_jll")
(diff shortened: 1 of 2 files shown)


test/new.jl
```diff
@@ -2767,6 +2767,13 @@ end
         @test linalg_block["uuid"] == linalg_uuid
         @test !haskey(linalg_block, "version")
     end
+
+    isolate(loaded_depot=true) do
+        # Next, test that stdlibs do not get dependencies from the registry
```
staticfloat, line 2772:
```suggestion
        # Next, test that stdlibs do not get dependencies from the registry
        # NOTE: this test depends on the fact that in Julia v1.6+ we added
        # "fake" JLLs that do not depend on Pkg while the "normal" p7zip_jll does.
        # A future p7zip_jll in the registry may not depend on Pkg, so be sure
        # to verify your assumptions when updating this test.
```

staticfloat (approved):

staticfloat:
Thanks for finding this Kristoffer!

PR #2710, opened 2021-08-19
Some missed compat -> outdated replaces from #2284

IanButterworth:
This is mainly just one docstring fix that was missed in the switch in #2284, 
but also some comment changes given that "compat" may be used to mean another thing in the status API shortly


KristofferC:
Thanks!

PR #2712, opened 2021-08-24
Docs: add comment for what `.` in `activate .` means

ToucheSir:
This came up recently on discourse. Though the dot is reasonably visible, it may be skimmed over or disregarded if there's no mention of why it is important.

### DIFF
--- a/docs/src/environments.md
+++ b/docs/src/environments.md
@@ -14,7 +14,7 @@ julia> mkdir("MyProject")
 julia> cd("MyProject")
 /Users/kristoffer/MyProject
 
-(v1.0) pkg> activate .
+(v1.0) pkg> activate . # "." means the current directory
 
 (MyProject) pkg> st
     Status `Project.toml`
--- a/docs/src/environments.md
+++ b/docs/src/environments.md
@@ -11,10 +11,10 @@ In order to create a new project, create a directory for it and then activate th
 ```julia-repl
 julia> mkdir("MyProject")
 
-julia> cd("MyProject")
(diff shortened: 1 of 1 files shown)


StefanKarpinski (approved):

docs/src/environments.md
```diff
@@ -14,7 +14,7 @@ julia> mkdir("MyProject")
 julia> cd("MyProject")
 /Users/kristoffer/MyProject
 
-(v1.0) pkg> activate .
+(v1.0) pkg> activate . # "." means the current directory
```
KristofferC, line 17: One problem is that now this command does not work to copy-paste any more.

ToucheSir: My mistake, should have tested in-REPL. How about adding a comment to the previous line like so?
```julia
 julia> cd("MyProject") # we can now use "." instead of a longer relative or full path
```
Might need some wordsmithing, but would be copy-paste friendly.

docs/src/environments.md
```diff
@@ -11,7 +11,7 @@ In order to create a new project, create a directory for it and then activate th
 ```julia-repl
 julia> mkdir("MyProject")
 
-julia> cd("MyProject")
+julia> cd("MyProject") # we can now use "." instead of a longer relative or full path
 /Users/kristoffer/MyProject
 
 (v1.0) pkg> activate .
```
IanButterworth, line 17: This works with paste mode
```suggestion
julia> cd("MyProject")
/Users/kristoffer/MyProject

# we can now use "." instead of a longer relative or full path:
(v1.0) pkg> activate .
```

PR #2715, opened 2021-08-26
Improve the local optimizer at the end of resolve

### LINKED ISSUE #1949: ]up only downgrades packages

carlobaldassi:
This is an improvement of the resolver heuristic, intended to fix #1949.

It is a strict improvement over the previous code, in terms of solution quality. It may be a tad slower, although I don't think it should be noticeable.
It fixed the problems mentioned in that issue that I could reproduce.
The PR is still not complete (there's still commented-out debugging stuff...), but the code is functional. So I'm posting here to encourage testing this (and reviewing, of course).

Still todo: polishing, tests, improve logging (?)


KristofferC:
> We could compress it but that would introduce some testing dependencies, right?

Pkg has the ability to decompress so there is no extra test dependency. I pushed a commit that does this.

> The test is also kind of slow (a few minutes). Is it too much? Should I try to reduce it as much as I can? (That may take a while...)

We've worked quite a bit to reduce the test time, a full run takes about 8 minutes now. So a few minutes extra is a quite large percentage increase. But if it is a lot of work to reduce that test time I rather have this in now and get the improved test time later.

carlobaldassi:
So I have some good news. I found out that my previous attempt was still suboptimal, in that it was missing the fact that in some cases we could have higher versions by installing additional packages. I have now further improved the optimizer to deal with this situation.
While doing that, I found out about a bug which went surprisingly unnoticed so far. Under very particular circumstances (such as some of those I used for testing this new code) it would produce a subtly inconsistent state, and potential crashes.
I have also reduced the test case as much as I could. It's now only a few kb and pretty fast. I checked that the new test would have failed with the previous optimizer.

So we should be all good hopefully, I have also cleaned up the code and added comments.

KristofferC:
> in that it was missing the fact that in some cases we could have higher versions by installing additional packages

Yes, I've noticed that when this tend to happen, usually additional packages are installed.

KristofferC:
I added a commit with a few fixes to things `@code_warntype` complained about.

src/Resolve/Resolve.jl
```diff
@@ -411,11 +372,182 @@ function enforce_optimality!(sol::Vector{Int}, graph::Graph)
         staged = staged_next
     end
 
+    # @info "clearing up: $([pkgID(p0, graph) for p0 in findall(uninst) if sol[p0] ≠ spp[p0]])"
```
KristofferC, line 375: Is this one (and the comment below) intended to be left in (to ease future debugging perhaps)?

carlobaldassi: Ah no sorry, I wasn't intending to leave this one here. I'll clean it up.

KristofferC:
Is it ok to squash this or do you prefer to keep the commits? If no squash, perhaps some of the commits should be merged together?

carlobaldassi:
Squash away

PR #2724, opened 2021-09-07
Precompile: Enable specifying packages

### LINKED ISSUE #2713: Allow `precompile` to take packages as argument
```jl
(@v1.6) pkg> precompile Plots
ERROR: Wrong number of arguments
```

would be nice to be able to do. What do you think @IanButterworth?

IanButterworth:
Fixes #2713

Adds
```
pkg> precompile Foo
...
pkg> precompile Foo Bar
...
```

- [ ] Add tests


KristofferC:
An alternative implementation is to make it so that `precompile` with no packages calls precompile with all packages in the project. Then there is no need to special-case `isempty` etc in the core precompile function.

IanButterworth:
> An alternative implementation is to make it so that precompile with no packages calls precompile with all packages in the project. Then there is no need to special-case isempty etc in the core precompile function.

Lovely idea but I've spent more time than I spent on this trying to figure out how to do that.. The problem is that the depsmap is a much nicer map to work with than the manifest, so filtering that earlier (which is what you're proposing) requires rethinking a fair amount of the mechanics, I think.

I'd happily let another PR take over from this one, if it's clear to you. Or maybe the idea will eat away at me until I just figure it out.. who knows

IanButterworth:
Should be ready now

PR #2731, opened 2021-09-16
release-1.7 Revert "Make `write_env_usage` atomic (#2661)"

KristofferC:
This reverts commit ebac45fdb43621a3009bd5cf09ee1f53d8b3122f.

While this was supposed to increase the reliability of multiple Pkg instances running at the same time, it made the situation worse (https://github.com/JuliaLang/julia/pull/42253#issuecomment-920346866). Since the release is imminent, I will revert this and it can be improved on master.

cc @Sacha0

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -29,8 +29,7 @@ jobs:
           - ""
           - "pkg.julialang.org"
         julia-version:
-          # - '1.6'
-          - 'nightly'
+          - "^1.7.0-0"
         exclude:
           - os: macOS-latest
             julia-arch: x86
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -29,8 +29,7 @@ jobs:
           - ""
           - "pkg.julialang.org"
         julia-version:
-          # - '1.6'
-          - 'nightly'
```
DilumAluthge, line 33: Any reason why we shouldn't test on Julia nightly?

DilumAluthge, line 33: Any reason why we shouldn't test on Julia nightly?

DilumAluthge: 🤦 

This is on the release-1.7 branch.

Sacha0 (approved):
Looks great, thanks Kristoffer! :)

Sacha0:
CI looks happy, so this patch should be in shape to merge? :) (I am not familiar with practices on this repo, so will wait on input.)

PR #2732, opened 2021-09-19
Add atomic write_env_usage test. Fix write_env_usage

IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/2661 intended to fix https://github.com/JuliaLang/Pkg.jl/issues/2633 but introduced different failure modes and was evidently missing tests.

Update: This PR
- Adds a test for atomic `write_env_usage` across processes
- Fixes `write_env_usage`
- Uncovered a similar bug updating the registry from different processes on windows https://github.com/JuliaLang/Pkg.jl/issues/2746

### DIFF
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -363,6 +363,56 @@ temp_pkg_dir() do project_path
         @test any(x -> startswith(x, manifest), keys(usage))
     end
 
+    @testset "test atomicity of write_env_usage with $(Sys.CPU_THREADS) parallel processes" begin
+        tasks = Task[]
+        iobs = IOBuffer[]
+        Sys.CPU_THREADS == 1 && error("Cannot test for atomic usage log file interaction effectively with only Sys.CPU_THREADS=1")
+        run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg"`) # to precompile Pkg given we're in a different depot
+        flag_start_dir = tempdir() # once n=Sys.CPU_THREADS files are in here, the processes can proceed to the concurrent test
+        flag_end_file = tempname() # use creating this file as a way to stop the processes early if an error happens
+        for i in 1:Sys.CPU_THREADS
+            iob = IOBuffer()
+            t = @async run(pipeline(`$(Base.julia_cmd()[1]) --project="$(pkgdir(Pkg))"
+                -e "import Pkg;
+                Pkg.UPDATED_REGISTRY_THIS_SESSION[] = true;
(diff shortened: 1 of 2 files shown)


IanButterworth:
Currently suffering from https://github.com/JuliaLang/Pkg.jl/issues/2733 first

test/pkg.jl
```diff
@@ -363,6 +363,29 @@ temp_pkg_dir() do project_path
         @test any(x -> startswith(x, manifest), keys(usage))
     end
 
+    @testset "atomic write_env_usage with $(Sys.CPU_THREADS) processes for 10 seconds" begin
+        tasks = Task[]
+        iobs = IOBuffer[]
+        Sys.CPU_THREADS == 1 && error("Cannot test for atomic usage log file interaction effectively with only Sys.CPU_THREADS=1")
+        for i in 1:Sys.CPU_THREADS
+            iob = IOBuffer()
+            t = @async run(pipeline(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))"
```
IanButterworth, line 372: Why would these subprocesses be precompiling Pkg?? Windows is failing because of atomic issues with `compilecache`, which shouldn't be needed here.. the master process should've precompiled Pkg.. 🤔

IanButterworth:
Might not be ready for review yet. Just trying to get every platform failing first

DilumAluthge:
Ooof. Sorry, I jumped the gun!

IanButterworth:
I think this is in good shape now. For some reason some ubuntu instances survive 1000 iterations, but the rest Gail quickly

IanButterworth:
Ok, the test seems stable across all platforms and is now marked as `@test_broken`. 

Should I merge this as-is, or keep this test on the shelf if someone comes up with a fix?

cc. @KristofferC @StefanKarpinski

StefanKarpinski:
I'm generally in favor of merging broken tests rather than leaving PRs around and then opening an issue to track fixing the broken tests. Of course, unless the tests are reliably broken and don't hang, that's a bad idea, but if the broken tests aren't flaky, then I'd merge them.

fredrikekre:
Can we vendor https://github.com/vtjnash/Pidfile.jl instead?

IanButterworth:
Seems like a good idea to me and it's not much code. @vtjnash have you ever considered absorbing Pidfile into Base?

vtjnash:
Sure, that seems great

IanButterworth:
Superseded by https://github.com/JuliaLang/Pkg.jl/pull/2793

PR #2734, opened 2021-09-20
generate: use joinpath instead of /

visr:
A nit, but the mixed slashes here don't look so pretty:

```
(jl_Nd3gOk) pkg> generate MyPackage
  Generating  project MyPackage:
    MyPackage\Project.toml
    MyPackage\src/MyPackage.jl
```

I checked and the `genfile` function defined higher in the same file doesn't care whether this dir is in the second or third argument.


KristofferC:
👍

PR #2744, opened 2021-09-27
Allow watchers for the active project

timholy:
This allows packages (with Revise as the intended target) to register
themselves as watchers for when the active project changes.
This will make it possible to effectively switch which manifest files
are analyzed for changes to the running source.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -15,7 +15,7 @@ import ..depots, ..depots1, ..logdir, ..devdir, ..printpkgstyle
 import ..Operations, ..GitTools, ..Pkg, ..Registry
 import ..can_fancyprint, ..pathrepr, ..isurl, ..PREV_ENV_PATH
 using ..Types, ..TOML
-using ..Types: VersionTypes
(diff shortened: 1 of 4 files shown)


timholy:
Companion Revise PR: https://github.com/timholy/Revise.jl/pull/649

timholy:
I'm going to go ahead and merge that Revise PR once CI passes; while it interacts with this PR, there's nothing externally visible and it does nothing if it doesn't find the expected names. So there is no rush on this.

KristofferC:
Arguably it would be more correct to put this in Base and make a function that changes the active project and fire this watcher and make Pkg (and everything else) use that function. There is nothing really specific about Pkg other than it is a common "user" of this variable.

DilumAluthge:
Closing in favor of https://github.com/JuliaLang/julia/pull/42731

Note that Pkg will need to be patched to use the new `Base.set_active_project` function once https://github.com/JuliaLang/julia/pull/42731 is merged.

PR #2750, opened 2021-10-05
Report if any package precomp did any downloading

IanButterworth:
There's been a couple of instances of packages downloading things during precompilation which can silently make precompilation take a very long time on a slow internet connection.
One confirmed & fixed case: https://github.com/FluxML/Flux.jl/issues/1597

I don't know if this is worth merging, but it provides a way to identify when that happens

Dependent on:
https://github.com/IanButterworth/julia/tree/ib/report_precomp_download
https://github.com/IanButterworth/Downloads.jl/tree/ib/session_download_flag


KristofferC:
Hm, this feels a little bit too special cased to me personally.

- This message will come after the precompilation, while users sit annoyed during the precompilation.
- This message will be mostly shown to users where the "not advisable" info is not actionable.
- If it is not advisable, what is the correct thing to do that the packages that currently download things during precompilation should do?

IanButterworth:
Yeah thats all fair. It's more of a debug tool. There might also be a simpler way to flush any downloads out

PR #2756, opened 2021-10-09
fix missing passing along `allow_autoprecom` in an instance

KristofferC:
The current way of `instantiate` calling precompile and `precompile` calling instantiate is kind of brittle. This prevents two auto precompiles from running on a `Pkg.up


IanButterworth:
I think this could/should be backported to 1.6 too?

PR #2753, opened 2021-10-06
Precompilation: Fix recognizing when new versions of already loaded packages are already precompiled

IanButterworth:
## Issue
A MWE should be easy for this issue that @kleinschmidt posted on slack
```
add Makie
   Resolving package versions...
    Updating `~/stuff/Project.toml`
  [ee78f7c6] + Makie v0.15.2
  No Changes to `~/stuff/Manifest.toml`
Precompiling project...
```
i.e. The specific Makie version was already in the Manifest where no changes were needed, however precompilation still did work.
I've also seen this in the wild.

## Fix
I believe this PR should be the fix, but I've not tested it yet

This disables the `Base.stale_cachefile` loaded module shortcut during stale cache checks
Depends on https://github.com/IanButterworth/julia/tree/ib/stale_cachefile_ignore_loaded

Basically `Base.stale_cachefile` was shortcutting the stale check if the package was loaded already because there used to be no reason to do anything else, so once users see those yellow check marks, they keep seeing them any time `Pkg.precompile` is called, until restarting the session.

PR #2760, opened 2021-10-11
Backports release 1.7

KristofferC:


IanButterworth:
Are you thinking save https://github.com/JuliaLang/Pkg.jl/pull/2753 for 1.7.1? I think that's fair

PR #2765, opened 2021-10-12
Allow the `Pkg.PlatformEngines.download_verify` and `Pkg.PlatformEngines.download_verify_unpack` functions to accept the optional `headers` keyword argument

DilumAluthge:
1. The `Pkg.PlatformEngines.download_verify` function can now accept the optional `headers` keyword argument, which it then forwards to `Pkg.PlatformEngines.download`.
2. The `Pkg.PlatformEngines.download_verify_unpack` function can now accept the optional `headers` keyword argument, which it then forwards to `Pkg.PlatformEngines.download_verify`.

This pull request is necessary for #2766.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -278,6 +278,7 @@ end
         hash::Union{AbstractString, Nothing},
         dest::AbstractString;
         verbose::Bool = false,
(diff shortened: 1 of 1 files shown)


DilumAluthge:
This is no longer needed for #2766.

We may end up needing it in the future for something else. But I think for now this can be closed.

PR #2766, opened 2021-10-12
Send the `Julia-Pkg-Allow-Incomplete-Registries` header unconditionally with every Pkg Server request

DilumAluthge:
## Summary

This is the client-side portion of the fix for https://github.com/JuliaRegistries/General/issues/16777.

## Description

This pull request sends the `Julia-Pkg-Allow-Incomplete-Registries` header unconditionally with every Pkg Server request.

The value of the header is set based on the `JULIA_PKG_ALLOW_INCOMPLETE_REGISTRIES` environment variable. Here are the details:

1. If the environment variable is not set, the header is set to `default_value`.
2. If the environment variable is set to the empty string (`""`), the header is set to `default_value`.
3. If the environment variable consists only of whitespace, the header is set to `default_value`.
4. If the environment variable is set to `true` or `1`, the header is set to `true`.
5. If the environment variable is set to `false` or `0`, the header is set to `false`.
6. If the environment variable is set to something that cannot be parsed as a `Bool`, a warning is printed, and the header is set to `default_value`.

## Default value

Currently, I have set `default_value = false`.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -232,7 +232,28 @@ function get_metadata_headers(url::AbstractString)
     end
     push!(headers, "Julia-CI-Variables" => join(ci_info, ';'))
     push!(headers, "Julia-Interactive" => string(isinteractive()))
+    push!(headers, "Julia-Pkg-Allow-Incomplete-Registries" => "$(pkg_server_allow_incomplete_registries())")
     return headers
+end
+
+function pkg_server_allow_incomplete_registries()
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
This feels overly complicated. I would send this header unconditionally with every pkg server request, which allows putting the header logic in the same place as all the other header logic. I can take a crack at that.

DilumAluthge:
Closing in favor of #2767.

PR #2767, opened 2021-10-13
Pkg requests: set Julia-Registry-Preference header based on JULIA_PKG_REGISTRY_PREFERENCE

StefanKarpinski:
Refs https://github.com/JuliaRegistries/General/issues/16777
Alternative to https://github.com/JuliaLang/Pkg.jl/pull/2766

This passes whatever value the environment variable `JULIA_PKG_REGISTRY_PREFERENCE` has through as a header. That's a bit unconstrained, but that means that if at some point a server supports other options here, we don't need to change the client.

@staticfloat is working on the pkg server side, which will ignore any value besides `complete` and `incomplete`, the former of which will cause `/registries` to redirect to `/registries.complete` and the latter of which will cause to redirect to `/registries.incomplete`, and those end-points will respectively respond with tree hashes for versions of the registry that are either already complete (pkg server has processed all the packages and artifacts) or incomplete (they haven't yet been processed). In the future, we could add more options without needing any client-side changes.

If this is used to talk to a server that doesn't understand this header, it will simply get ignored and the server will serve whatever registry versions is knows how to serve. If the variable isn't set at all, then the header doesn't get sent and the server chooses whatever registry version it prefers. That also means that we can change whether complete or incomplete is the default with purely server-side changes without modifying clients at all.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -232,8 +232,8 @@ function get_metadata_headers(url::AbstractString)
     end
     push!(headers, "Julia-CI-Variables" => join(ci_info, ';'))
     push!(headers, "Julia-Interactive" => string(isinteractive()))
-    registry_pref = get(ENV, "JULIA_PKG_REGISTRY_PREFERENCE", nothing)
-    if registry_pref !== nothing
+    registry_pref = get(ENV, "JULIA_PKG_REGISTRY_PREFERENCE", "")
+    if !isempty(strip(registry_pref))
         push!(headers, "Julia-Registry-Preference" => registry_pref)
     end
     return headers


StefanKarpinski:
I suppose it's pretty low risk. @KristofferC, thoughts? This is the entire client-side change required for allowing people to choose between complete and incomplete registries once the servers support that.

DilumAluthge:
> I suppose it's pretty low risk. @KristofferC, thoughts? This is the entire client-side change required for allowing people to choose between complete and incomplete registries once the servers support that.

Also pinging @fredrikekre and @IanButterworth for their thoughts on the backport decision.

StefanKarpinski:
Another thing we want is for the default for CI requests to be `incomplete` regardless of what the default for other clients it, but I don't think encoding that logic in the client is the best way to go. I'd prefer for the server to implement that logic by examining the `Julia-CI-Variables` header if this header is absent and defaulting to `incomplete` if any of those variables are set to `t` for "true" or `o` for "other"; the other possibilities are `n` for "not set" and `f` for "false". @staticfloat, do you think we can configure nginx to implement the logic required for that? Namely for the resource `/registries`:

- Check for the `Julia-Registry-Preference` header
- If present and set to `incomplete`, redirect to `/registries.incomplete`
- If present and set to `complete`, redirect to `/registries.complete`
- Otherwise check for the `Julia-CI-Variables` header
- If present and the value matches `r"=[to]"`, redirect to `/registries.incomplete` (CI default)
- Otherwise redirect to `/registries.complete` (non-CI default)

This seems like pretty simple logic to me, but I have no experience with nginx configuration.

DilumAluthge:
> I don't think encoding that logic in the client is the best way to go.

The less we encode on the client side, the happier I'll be.

StefanKarpinski:
I mean, yes, that is objectively the right strategy for this kind of thing.

StefanKarpinski:
Code for that would be something like this:
```jl
for (key, val) in ENV
     m = match(r"^JULIA_PKG_HEADER_([A-Z0-9_]+)$"i, key)
     m === nothing && continue
     hdr = join(map(titlecase, split(m.captures[1], '_')), '-')
     push!(headers, "Julia-$hdr" => val)
end
```
It's kind of appealing because of how future-proof it is.

staticfloat:
> Code for that would be something like this:
> 
> ```julia
> for (key, val) in ENV
>     m = match(r"^JULIA_PKG_HEADER_([A-Z0-9_]+)$"i, key)
>     m === nothing && continue
>     words = split(m.captures[1], '_', keepempty=false)
>     isempty(words) && continue
>     hdr = "Julia-" * join(map(titlecase, words), '-')
>     any(hdr == k for (k, v) in headers) && continue
>     push!(headers, hdr => val)
> end
> ```
> 
> It's kind of appealing because of how future-proof it is.

Yeah, I kind of dig this.  Very simple, easy-to-explain behavior.

> So `eager` and `complete`?

I like `eager`!  One thing I don't like about `complete` that much, is that even with the `complete` registry endpoint, we don't absolutely _guarantee_ that every resource is available.  We just guarantee that we've _tried_ to process every resource.  If a server is giving us 404's for some artifact, etc... eventually we'll give up trying to fetch that resource for that registry.  So a name other than `complete` might be a little clearer; `processed`? `itriedmybest`?

staticfloat:
Ahhh I like conservative!

staticfloat:
When processing resources, each resource's status is one of `pending`, `success`, `soft-fail`, or `hard-fail`:

* `pending`: A new registry was downloaded, its contents were explored, it generated a bunch of Package/Artifact resources to store, and this resource has yet to be tended to by a worker.
* `success`: The resource was successfully downloaded, repackaged and uploaded to S3.
* `soft-fail`: Something went wrong during processing, but we should try again later (e.g. the server did not respond, the download was interrupted halfway through, etc...)
* `hard-fail`: Something went wrong during processing, and we should never try again (e.g. the artifact is a 100GB data corpus that we should not bother with repackaging)

The two registry types are therefor defined as:
* `eager`: no guarantee on the state of all resources referred to by this registry; some may be in `success` state, but others can be `pending` still.
* `conservative`: All resources are guaranteed to be at least `success` or `soft/hard-fail`.  Nothing will be `pending`.

There is another option for a `complete` registry, which is one where we guarantee that every resource is either `success` or `hard-fail`, but practically speaking I think that's not very useful, as there are a good number of artifacts that refer to dangling links, etc...  We could try to clean those up, but I think it's just not that worth it, since they're obviously not very high in demand.  I think just `eager` and `conservative` are good enough for now.

DilumAluthge:
Closing in favor of https://github.com/JuliaLang/Pkg.jl/pull/2770

PR #2770, opened 2021-10-13
Set headers in Pkg server requests from JULIA_PKG_SERVER_*

StefanKarpinski:
Yet another alternative to #2767 and #2766.

This looks for environment variables with names that start with `JULIA_PKG_HEADER_` and uses the rest of the variable name and its value to populate HTTP headers sent along with requests to Pkg servers, prefixing each header with `Julia-`. For example, if the environment variable `JULIA_PKG_HEADER_REGISTRY_PREFERENCE` is set to `complete` then each Pkg server request will include the header

     Julia-Registry-Preference: complete

Details:

- header is not set if the value of the variable is all whitespace
- header name is determined by splitting on underscores and titlecasing each word, then joining words with dashes; `Julia-` is prefix
- if a header name is already set, it is not set again; this prevents using this mechanism to change standard header values and means that if multiple environment variables set the same header (e.g. multiple or  trailing underscores), then only the first one is sent

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -233,7 +233,7 @@ function get_metadata_headers(url::AbstractString)
     push!(headers, "Julia-CI-Variables" => join(ci_info, ';'))
     push!(headers, "Julia-Interactive" => string(isinteractive()))
     for (key, val) in ENV
-        m = match(r"^JULIA_PKG_HEADER_([A-Z0-9_]+)$"i, key)
+        m = match(r"^JULIA_PKG_SERVER_([A-Z0-9_]+)$"i, key)
         m === nothing && continue
         val = strip(val)
         isempty(val) && continue


fredrikekre:
`JULIA_PKG_HEADER_XX` -> `JULIA_PKG_SERVER_XX` maybe? `Julia-Pkg-Server` is already set to the `JULIA_PKG_SERVER` variable so this would simply "extend" that.

StefanKarpinski:
I do like @fredrikekre's idea. The term "server" probably is clearer to more people than "header" (header of what?). So then setting `JULIA_PKG_SERVER_FOO=bar` would cause `Julia-Pkg-Server-Foo: bar` to be sent as a header. In the specific motivating case `JULIA_PKG_SERVER_REGISTRY_PREFERENCE=complete` would cause `Julia-Pkg-Server-Registry-Preference: complete` to be sent as a header.

DilumAluthge:
> The term "server" probably is clearer to more people than "header" (header of what?).

Yeah, that's a good point.

DilumAluthge:
> Seems like an implementation detail that the setting is used as a HTTP header, and `SERVER` already kind of implies that it is a setting for the server.

Yeah, that makes sense!

Issue #2525, opened 2021-04-23
`Pkg.dev` chooses a registry inconsistently with `Pkg.add`


ericphanson:
If you have a package registered with the same name and UUID in two registries, `Pkg.add(name=...)` will choose whichever registry has a later version. However, `Pkg.dev` will choose some other way (whatever is first in some list it checks, I think). Instead, `Pkg.dev` should choose the same registry as `Pkg.add` would. This mismatch can be very confusing for users.

fredrikekre:
Could prompt the user in interactive mode and error otherwise.

ericphanson:
I think that would be an improvement over the current situation, but why not just use the same registry resolution as `add`? Since if you aren’t passing a path or url it needs to use a registry so why not make that consistent.

fredrikekre:
But `Pkg.add` doesn't really choose a registry, it just collapses them all.

ericphanson:
Ah ok, I didn’t realize that. Then I guess it’s a bit harder to implement my suggestion. As a user it still seems like the consistent way for it to behave though.

ericphanson:
But I think a prompt would be a big improvement already. Currently when users dev and get the old version of a package it can make them think something is configured wrong on their end or they have some internal state that is out of date (especially because configuring multiple registries and auth for private registries is not always perfectly smooth). So even if it seems like Pkg should know which registry is the right one, at least giving the options is a big improvement over silently giving the wrong one.

fredrikekre:
With multiple repo urls there is not really a good way to find the canonical one though.

ericphanson:
Well I'm saying there is: in the scenario Pkg is given a name or name + uuid and no URL, then it is hopeless to find a URL unless the package is registered in a registry, in which case it has version numbers in that registry. The case of multiple URLs is when it is registered in multiple registries. Which is the canonical one? The registry which contains the highest version number.

fredrikekre:
> Which is the canonical one? The registry which contains the highest version number.

Why? This seems almost as arbitrary as the current behavior. For example, you have have a private repo and private registry. You decide to open start making releases to a public registry instead, but you probably still want `Pkg.dev` to track your private repo.

Prompting seems simple enough. `Pkg.dev` is inheritly an interactive function so erroring in the non-interactive case seems good anyways.

ericphanson:
> Why?

Since when you `dev`, you get the latest copy of the code. Where is the latest copy of the code likely to be? The one with the latest release.

[...]

PR #2777, opened 2021-10-17
Make adding by branch or dev'ing use repo with latest version

ericphanson:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2525 by using @fredrikekre's suggestion https://github.com/JuliaLang/Pkg.jl/issues/2525#issuecomment-827492474. ~~It seems to work in some brief testing locally, but I should still figure out how to add the scenario to the tests here.~~ added some tests!

This gives a menu like this:
```julia
Repository for package with UUID `7876af07-990d-54b4-ab0e-23690620f79a` found in multiple registries, choose one:
 > General: https://github.com/JuliaLang/Example.jl.git
   RegistryFoo: https://github.com/JuliaLang/Example1.jl.git
   RegistryFoo: https://github.com/JuliaLang/Example2.jl.git:subdir
```


ericphanson:
I realized that we don't care which registry we get a repo from if two registries have the *same* repo; no need to prompt or error in that case (which would be a regression from the current behavior). So I rearranged things to use a `Dict` to collect the unique sets of `(repo, subdir)` and which registries they come from (which is just passed to the user so they understand better where the multiple repos are coming from).

Now, in the admittedly complicated scenario tested, we get this menu:
```
Multiple repositories found for package with UUID `7876af07-990d-54b4-ab0e-23690620f79a`, choose one:
 > General: https://github.com/JuliaLang/Example.jl.git
   RegistryFoo1, RegistryFoo2: https://github.com/JuliaLang/OtherExampleRepo.jl.git
   RegistryFoo3: https://github.com/JuliaLang/OtherExampleRepo.jl.git:subdir
```
Here there are 4 registries; 2 have the same (repo, subdir), namely RegistryFoo1 and RegistryFoo2. RegistryFoo3 has a subdir parameter, and of course General. In particular if we only have RegistryFoo1 and RegistryFoo2, then there's no prompt, since it's unambiguous which repo to use.

ericphanson:
> To me it feels like if you know which registry you want (you know the correct URL) why not just add it via that URL?

Isn’t that also true for name clashes, where a name is in multiple registries? There we could error and print the UUIDs but instead give an interactive menu (untested iiuc, unlike this code). And it seems like this feature is well-received and folks cite it to show how well Pkg handles multiple registries.

KristofferC:
> Isn’t that also true for name clashes, where a name is in multiple registries?

Yes, I agree. That should probably be removed as well and be replaced with a nice error message instead.

> And it seems like this feature is well-received and folks cite it to show how well Pkg handles multiple registries.

Not because of the interactive selector in the rare case of a name collision though?

PR #2787, opened 2021-10-21
Always allow stdlibs to update during a resolve

IanButterworth:
@staticfloat I believe this might be a bug, and should be `nothing` by default?

See discussion in https://github.com/JuliaRegistries/General/issues/47158

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -337,7 +337,11 @@ function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryIn
     for pkg in pkgs
         names[pkg.uuid] = pkg.name
     end
-    reqs = Resolve.Requires(pkg.uuid => VersionSpec(pkg.version) for pkg in pkgs)
+
+    # Unless using the unbounded or historical resolver, always allow stdlibs to update. Helps if the previous resolve
+    # happened on a different julia version / commit and the stdlib version in the manifest is not the current stdlib version
+    unbind_stdlibs = julia_version === VERSION
+    reqs = Resolve.Requires(pkg.uuid => is_stdlib(pkg.uuid) && unbind_stdlibs ? VersionSpec("*") : VersionSpec(pkg.version) for pkg in pkgs)
     graph, compat_map = deps_graph(env, registries, names, reqs, fixed, julia_version)
     Resolve.simplify_graph!(graph)
     vers = Resolve.resolve(graph)


staticfloat:
Hmmm, setting `julia_version` to `nothing` has a larger effect than you're probably thinking.  Setting it to `nothing` causes Julia version bounds to be completely ignored; e.g. you can instantiate a manifest that has a Julia v1.5- compatible package, and a Julia v1.6+ compatible package at the same time.

IanButterworth:
> Perhaps what is missing is a special-case in get_last_stdlibs() that checks if julia_version == VERSION, and uses stdlibs() (which loads the stdlibs from the current stdlib folder) instead of searching through STDLIBS_BY_VERSION.

Now implemented

DilumAluthge:
Just to be safe, instead of checking that the major and minor versions match, can we check that the major, minor, and patch versions match?

IanButterworth:
I think we guarantee that the stdlibs won't change version between patch versions? I'm trying to think through what people would expect. People that check in a manifest would likely expect the manifest to _just work_ across all patches of a given major-minor julia version? That seems to be how General handles it? Anything more granular than that would cause similar headaches each patch release

DilumAluthge:
> I think we guarantee that the stdlibs won't change version between patch versions?

Do we? I know we try to guarantee it, but is that an absolute statement? E.g. in theory, if we need to apply a bugfix as part of a patch release, isn't it possible that we might need to update the stdlib version as part of that bugfix.

IanButterworth:
Yeah, actually.. I'm not sure why I was thinking that would be the case. After all a patch is a patch, no reason why a stdlib wouldn't patch bump during a julia patch bump

IanButterworth:
But I think, unless this whole idea is flawed (it could be) the right thing to do is to allow stdlibs to be updated across patch releases and downwards, so adding the patch check as you suggest above make that logic too specific

KristofferC:
I'm quite out of the loop when it comes to versioned stdlibs and compat with them, setting `julia_version` for usage with BinaryBuilder etc. @staticfloat implemented most of that so his review is more valuable than mine.

As a side note, it is a bit unfortunate how this special case for BinaryBuilder has kind of ballooned in scope with all the surrounding infrastructure that is needed around it. Large dictionaries that need to be kept in sync with all Julia versions and even nightly with automated workflows to update them, arguments that need to be passed along more or less the whole codebase etc etc. I wonder if this really was the best way to do this.

IanButterworth:
> Large dictionaries that need to be kept in sync with all Julia versions and even nightly with automated workflows to update them

Just a note, nightly is no longer dependent on the hardcoded dictionaries as of https://github.com/JuliaLang/Pkg.jl/pull/2795 
But there are bots to keep it updated. Perhaps those could be only retriggered on new tags to Julia base @DilumAluthge

DilumAluthge:
Right now the bot is triggered only manually, but yeah we could do something fancier like automatically run it when there's a new Julia tag.

Anyway, for now, to run the bot manually, go here: https://github.com/JuliaLang/Pkg.jl/actions/workflows/run_hsg.yml

src/Operations.jl
```diff
@@ -337,7 +337,11 @@ function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryIn
     for pkg in pkgs
         names[pkg.uuid] = pkg.name
     end
-    reqs = Resolve.Requires(pkg.uuid => VersionSpec(pkg.version) for pkg in pkgs)
+
+    # Unless using the unbounded or historical resolver, always allow stdlibs to update. Helps if the previous resolve
+    # happened on a different julia version / commit and the stdlib version in the manifest is not the current stdlib version
+    unbind_stdlibs = julia_version === VERSION
+    reqs = Resolve.Requires(pkg.uuid => is_stdlib(pkg.uuid) && unbind_stdlibs ? VersionSpec("*") : VersionSpec(pkg.version) for pkg in pkgs)
```
IanButterworth, line 344: @staticfloat it just occurred to me, should it be this instead? Maybe `VersionSpec("*")` would've allowed updating to latest always, beyond what's allowed for the given julia version?
```suggestion
    reqs = Resolve.Requires(pkg.uuid => is_stdlib(pkg.uuid) && unbind_stdlibs ? VersionSpec(stdlibs()[uuid][2]) : VersionSpec(pkg.version) for pkg in pkgs)
```

staticfloat: I think in practice it doesn't matter, because as long as the compat ranges that we declare here match what is already in the system image, the resolver should choose those versions, right?  The issue isn't that we're trying to install a different version than exists, it's that the resolver isn't being allowed to use the versions that are already installed, I thought?

IanButterworth: Sounds reasonable to me. I'll merge

DilumAluthge:
@KristofferC @IanButterworth Can we backport this to Julia 1.6 and 1.7?

fredrikekre:
Doesn't backport cleanly to 1.6 (can make a PR to #3033 if this is still needed).

Issue #2746, opened 2021-09-29
Registry downloads not atomic across processes: IOError: unlink("registries\\General.tar.gz"): resource busy or locked


IanButterworth:
Seen in #2732 where multiple processes are updating the registry at the same time. https://github.com/JuliaLang/Pkg.jl/pull/2732/checks?check_run_id=3739676185
Might just be a problem on windows?

```
ERROR: IOError: unlink("tmp\\registries\\General.tar.gz"): resource busy or locked (EBUSY)
Stacktrace:
  [1] uv_error
    @ .\libuv.jl:97 [inlined]
  [2] unlink(p::String)
    @ Base.Filesystem .\file.jl:970
  [3] rm(path::String; force::Bool, recursive::Bool)
    @ Base.Filesystem .\file.jl:281
  [4] checkfor_mv_cp_cptree(src::String, dst::String, txt::String; force::Bool)
    @ Base.Filesystem .\file.jl:328
  [5] #mv#15
    @ .\file.jl:423 [inlined]
  [6] download_registries(io::Base.DevNull, regs::Vector{Pkg.Registry.RegistrySpec}, depot::String)
    @ Pkg.Registry D:\a\Pkg.jl\Pkg.jl\src\Registry\Registry.jl:182
  [7] download_registries
    @ D:\a\Pkg.jl\Pkg.jl\src\Registry\Registry.jl:161 [inlined]
  [8] download_default_registries(io::Base.DevNull; only_if_empty::Bool)
    @ Pkg.Registry D:\a\Pkg.jl\Pkg.jl\src\Registry\Registry.jl:108
  [9] download_default_registries
    @ D:\a\Pkg.jl\Pkg.jl\src\Registry\Registry.jl:96 [inlined]
 [10] add(pkgs::Vector{Pkg.Types.PackageSpec}; io::Base.DevNull, kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
    @ Pkg.API D:\a\Pkg.jl\Pkg.jl\src\API.jl:145
 [11] #add#27
    @ D:\a\Pkg.jl\Pkg.jl\src\API.jl:143 [inlined]
 [12] #add#26
    @ D:\a\Pkg.jl\Pkg.jl\src\API.jl:142 [inlined]
 [13] top-level scope
    @ none:4
```

Issue #2346, opened 2021-01-18
`write_env_usage()` can self-clobber


staticfloat:
When cleaning up the Yggdrasil CI bots, I found that the `manifest_usage.toml` and `artifact_usage.toml` files were clobbered rather often due to multiple Julia processes writing out at once.

I believe this may be due to `TOML.print(io, ::Dict)` partially printing then yielding.  It would probably be better to have this construct the string first, then append it in one `write()` call.

https://github.com/JuliaLang/Pkg.jl/blob/94d3c45f1b7c6c7a1e887d97f4e87e3662801d0f/src/Types.jl#L415

fredrikekre:
Yep, that is what we do for the project and manifest (https://github.com/JuliaLang/Pkg.jl/pull/1203)

fredrikekre:
But the print to file is done in one chunk already?

staticfloat:
Hmmm, you're right.  That must not be the issue then.

fredrikekre:
Same with Scratch.jl: https://github.com/JuliaPackaging/Scratch.jl/blob/2b527ef57d94f10240ac45a2992e2cbffe69c210/src/Scratch.jl#L156

fredrikekre:
https://github.com/JuliaLang/Pkg.jl/blob/94d3c45f1b7c6c7a1e887d97f4e87e3662801d0f/src/API.jl#L518-L520 maybe

Issue #1664, opened 2020-02-11
Excessive writes to `manifest_usage.toml`


ericphanson:
I instantiated a manifest (on Julia 1.3.1) and it added 9 entries (for the same manifest at very close timestamps) to `.julia/logs/manifest_usage.toml`.

I had a bunch of corrupted entries in there (probably from interrupting Pkg operations) which caused `Pkg.gc()` to fail, and it seems like maybe there would be fewer opportunities for corrupted entries if Pkg didn't write to that file so much.

(I first reported this on Slack).

ericphanson:
It might be nice also if interrupt exceptions during these writes were handled somehow so as to not corrupt the toml file.

KristofferC:
Yeah, and also handling corrption in the file better (trying to recover)

KristofferC:
@IanButterworth, how many writes are done to `manifest_usage.toml` now?

IanButterworth:
The number of writes is the same, but each manifest is only entered once, and the corruption won't happen. 

If that doesn't mean this issue can be closed, feel free to reopen.

ericphanson:
I actually tried to reproduce this last night to see if it should be reopened but instantiating a Manifest (on Pkg 1.3) only added one entry to that file, not 9. So I’m not sure what was different the first time. So I think it’s ok for this to stay closed unless there’s a reproducible issue w/excessive writes.

Thanks a lot for the locking fix btw, the corruption was definitely the most annoying part. Will ctrl-c at the wrong time cause corruption still, do you think?

Issue #684, opened 2018-08-23
manifest_usage.toml gets corrupted in a distributed environment


ghost:
When launching multiple julia processes `manifest_usage.toml` gets corrupted by concurrent writes that leads to `gc` having trouble:

```
pkg> gc
ERROR: Pkg.TOML.ParserError(594, 594, "expected a key but found an empty string")

...and 1 more exception(s).

Stacktrace:
 [1] parse(::Base.GenericIOBuffer{Array{UInt8,1}}) at /home/src/julia.7/usr/share/julia/stdlib/v1.0/Pkg/ext/TOML/src/TO
ML.jl:33                                                                                                              
 [2] parse at /home/src/julia.7/usr/share/julia/stdlib/v1.0/Pkg/ext/TOML/src/TOML.jl:40 [inlined]
 [3] #gc#45(::Base.Iterators.Pairs{Union{},Union{},Tuple{},NamedTuple{(),Tuple{}}}, ::Function, ::Pkg.Types.Context) at
 /home/src/julia.7/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:292                                                     
 [4] gc at /home/src/julia.7/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:273 [inlined]
```

removing the file clears the issue.

StefanKarpinski:
Definitely an issue. We should do some file locking here. @vtjnash, what's the current best way?

vtjnash:
Use atomic file operations. (Never modify or truncate an existing file.)

KristofferC:
I think this should be fixed now.

timholy:
This just happened to me.
```julia
(@v1.5) pkg> gc
ERROR: Pkg.TOML.ParserError(2128571, 2128572, "expected `=`, but found `:`")
Stacktrace:
 [1] parse(::Base.GenericIOBuffer{Array{UInt8,1}}) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.5/Pkg/ext/TOML/src/TOML.jl:37
 [2] parse at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.5/Pkg/ext/TOML/src/TOML.jl:44 [inlined]
 [3] (::Pkg.API.var"#collect_usage!#83")(::Dict{String,Dates.DateTime}, ::String) at /home/tim/src/julia-master/usr/share/julia/stdlib/v1.5/Pkg/src/API.jl:353
[...]

PR #2793, opened 2021-10-24
Fix manifest_usage.toml and Registry process concurrency via pid locking

IanButterworth:
Implements pidfile locking for manifest_usage.toml and Registry add/update operations

Fixes https://github.com/JuliaLang/Pkg.jl/issues/2875
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2346
Fixes https://github.com/JuliaLang/Pkg.jl/issues/684
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2746
Fixes https://github.com/JuliaLang/Pkg.jl/issues/1664

Includes the process concurrency test from https://github.com/JuliaLang/Pkg.jl/pull/2732 which would cause Pkg master to error in two ways
- the env usage file mv breaking
- the registry update mv breaking

cc. @vtjnash


IanButterworth:
The MacOS failure is https://github.com/JuliaLang/Pkg.jl/issues/2596

DilumAluthge:
> ERROR: LoadError: ArgumentError: Package Pkg does not have FileWatching in its dependencies:

fredrikekre (approved):
Great! Could Pkg also use a lockfile for precompilation?

DilumAluthge:
> ERROR: LoadError: ArgumentError: Package Pkg does not have FileWatching in its dependencies:

@IanButterworth Could you update the manifest file at `ext/Historical.../Manifest.toml`?

IanButterworth:
> Could Pkg also use a lockfile for precompilation?

Could be, but there's a currently [a bug](https://github.com/JuliaLang/Pkg.jl/issues/2648) with `ctrl-c`-ing pkg precompilation that causes julia to fatally error, that I've not been able to track down. and if we pidlocked the precompile process, I believe that would leave the lock file around until it was deemed stale, which might be hard to find the right balance for.

Perhaps that could be a followup PR, ideally after that bug has been squashed?

IanButterworth:
Needs to wait for https://github.com/JuliaLang/julia/pull/44367 to land on the nightlies

src/Registry/Registry.jl
```diff
@@ -250,6 +253,7 @@ function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=d
             end
         end
     end
+    end
```
fredrikekre, line 256: This is so far from the start so maybe nice to add a comment:
```suggestion
    end # mkpidlock
```

src/Registry/Registry.jl
```diff
@@ -163,6 +164,8 @@ function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=d
     populate_known_registries_with_urls!(regs)
     regdir = joinpath(depot, "registries")
     isdir(regdir) || mkpath(regdir)
+    FileWatching.mkpidlock(joinpath(regdir, ".pid"), stale_age = 10) do
+    # only allow one julia process to download and install registries at a time
```
fredrikekre, line 168:
```suggestion
    # only allow one julia process to download and install registries at a time
    FileWatching.mkpidlock(joinpath(regdir, ".pid"), stale_age = 10) do
```

src/Registry/Registry.jl
```diff
@@ -348,6 +352,8 @@ function update(regs::Vector{RegistrySpec} = RegistrySpec[]; io::IO=stderr_f(),
     for reg in unique(r -> r.uuid, find_installed_registries(io, regs); seen=Set{UUID}())
         let reg=reg, errors=errors
             regpath = pathrepr(reg.path)
+            FileWatching.mkpidlock(joinpath(dirname(dirname(reg.path)), ".$(reg.name).pid"), stale_age = 10) do
```
fredrikekre, line 355: Shouldn't this use the same lock as above?

src/Registry/Registry.jl
```diff
@@ -348,6 +352,8 @@ function update(regs::Vector{RegistrySpec} = RegistrySpec[]; io::IO=stderr_f(),
...
+            # only allow one julia process to update a registry at a time
```
fredrikekre, line 356:
```suggestion
            # only allow one julia process to update a registry at a time
            FileWatching.mkpidlock(joinpath(dirname(dirname(reg.path)), ".$(reg.name).pid"), stale_age = 10) do
```

src/Registry/Registry.jl
```diff
@@ -439,6 +445,7 @@ function update(regs::Vector{RegistrySpec} = RegistrySpec[]; io::IO=stderr_f(),
                     end
                 end
             end
+            end
```
fredrikekre, line 448:
```suggestion
            end # mkpidlock
```

src/Registry/Registry.jl
```diff
@@ -163,6 +164,8 @@ function download_registries(io::IO, regs::Vector{RegistrySpec}, depot::String=d
     populate_known_registries_with_urls!(regs)
     regdir = joinpath(depot, "registries")
     isdir(regdir) || mkpath(regdir)
+    FileWatching.mkpidlock(joinpath(regdir, ".pid"), stale_age = 10) do
```
fredrikekre, line 167: Perhaps `.pidlock` or `.lock` as extension? Is there a standard naming?

IanButterworth: I originally had something more descriptive but changed it to .pid because it appears to be the standard https://stackoverflow.com/questions/8296170/what-is-a-pid-file-and-what-does-it-contain

fredrikekre: Yea, I thought there were more info in the file here about the lock, but looks like it is jsut the pid.

mbauman:
Could this be backported to 1.8?

KristofferC:
I don't think we have PID locking available in Base on 1.8.

PR #2803, opened 2021-10-30
Pkg test suite: do not try to write to or modify the directory containing the Pkg source code

DilumAluthge:
In certain situations (e.g. when we download the Julia binaries at the beginning of the Buildkite `tester_*` jobs), the `stdlib/Pkg` directory may be read-only. So instead of trying to write to the `stdlib/Pkg/test` directory, let's just use a temp directory instead.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -21,11 +21,6 @@ Pkg.DEFAULT_IO[] = IOBuffer()
 
 include("utils.jl")
 
-# Clean slate. Make sure to not start with an outdated registry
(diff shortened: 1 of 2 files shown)


IanButterworth:
Perhaps to be sure there's no other writing you could check the repo treehash at the start vs. end of tests?

DilumAluthge:
I think this will need #2808.

DilumAluthge:
> Perhaps to be sure there's no other writing you could check the repo treehash at the start vs. end of tests?

Yeah that's not a bad idea.

PR #2815, opened 2021-11-01
Add project hash to manifest metadata and warn when manifest and project are out of sync

IanButterworth:
One of the reasons I was a fan of adding extensible metadata to the manifest was to be able to save a hash of the project state in the manifest, to make it easy to identify when the manifest and project are out of sync. i.e. a project dep entry has been manually added/removed in the toml file, or a compat entry added/changed/removed

Note that the project hash used here only hashes the deps and compat fields of the project, the rest shouldn't affect a resolve, so should be free to change.

One thing this would allow is CI to easily check whether checked-in manifests are valid, as it's easy to forget to update them.
Currently Pkg.instantiate doesn't check compat bounds and just does what the manifest says. This adds a warning.

Todo:

- [x] Add tests

```
julia> import Pkg

(Pkg) pkg> activate --temp
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2QJmtW`

(jl_2QJmtW) pkg> add CSV
   Resolving package versions...
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2QJmtW/Project.toml`
  [336ed68f] + CSV v0.9.10
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2QJmtW/Manifest.toml`
  [336ed68f] + CSV v0.9.10
...

(jl_2QJmtW) pkg> compat CSV 0.8
      Compat entry set:
  CSV = "0.8"

(jl_2QJmtW) pkg> instantiate
┌ Warning: The project hash recorded into the manifest when it was resolved does not match the hash of the current project.
│ A project dependency has been added/removed or compat entry has changed. The environment may need to be updated
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:1509

(jl_2QJmtW) pkg> up
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2QJmtW/Project.toml`
  [336ed68f] ↓ CSV v0.9.10 ⇒ v0.8.5
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2QJmtW/Manifest.toml`
  [336ed68f] ↓ CSV v0.9.10 ⇒ v0.8.5
  [944b1d66] - CodecZlib v0.7.0
  [48062228] - FilePathsBase v0.9.13
  [842dd82b] - InlineStrings v1.0.1
  [69de0a69] ↓ Parsers v2.1.1 ⇒ v1.1.2
  [3bb67fe8] - TranscodingStreams v0.9.6
  [ea10d353] - WeakRefStrings v1.4.1
  [cf7118a7] - UUIDs
  [83775a58] - Zlib_jll v1.2.12+1

(jl_2QJmtW) pkg> instantiate

(jl_2QJmtW) pkg> compat CSV 0.9
      Compat entry set:
  CSV = "0.9"

(jl_2QJmtW) pkg> instantiate
┌ Warning: The project hash recorded into the manifest when it was resolved does not match the hash of the current project.
│ A project dependency has been added/removed or compat entry has changed. The environment may need to be updated
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:1509

(jl_2QJmtW) pkg> rm CSV
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2QJmtW/Project.toml`
  [336ed68f] - CSV v0.8.5
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2QJmtW/Manifest.toml`
  [336ed68f] - CSV v0.8.5
...

(jl_2QJmtW) pkg> instantiate

julia> Pkg.is_manifest_current()
true

(jl_2QJmtW) pkg> add CSV
...

(jl_2QJmtW) pkg> compat CSV 0.8
      Compat entry set:
  CSV = "0.8"

julia> Pkg.is_manifest_current()
false

(jl_2QJmtW) pkg> instantiate
┌ Warning: The project hash recorded into the manifest when it was resolved does not match the hash of the current project.
│ A project dependency has been added/removed or compat entry has changed. The environment may need to be updated
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:1509

```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -439,16 +439,17 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     return
 end
 
-function is_manifest_current(ctx::Context = Context())
-    if haskey(ctx.env.manifest.other, "project_hash")
-        recorded_hash = ctx.env.manifest.other["project_hash"]
-        current_hash = Types.project_resolve_hash(ctx.env.project)
+function is_manifest_current(env::EnvCache)
+    if haskey(env.manifest.other, "project_hash")
+        recorded_hash = env.manifest.other["project_hash"]
+        current_hash = Types.project_resolve_hash(env.project)
         return recorded_hash == current_hash
     else
         # Manifest doesn't have a hash of the source Project recorded
         return nothing
     end
 end
+is_manifest_current(ctx::Context = Context()) = is_manifest_current(ctx.env)
 
 const UsageDict = Dict{String,DateTime}
 const UsageByDepotDict = Dict{String,UsageDict}
@@ -1505,9 +1506,10 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     end
     Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
 
-    if is_manifest_current(ctx) === false
-        @warn """The project hash recorded into the manifest when it was resolved does not match the hash of the current project.
-            A project dependency has been added/removed or compat entry has changed. The environment may need to be updated"""
+    if is_manifest_current(ctx.env) === false
+        @warn """The project and manifest may be out of sync. \
+        The project hash recorded into the manifest when it was resolved does not match the hash of the current project. \
+        A project dependency has been added/removed or compat entry has changed. The environment may need to be updated"""
     end
 
     Operations.prune_manifest(ctx.env)
--- a/src/API.jl
+++ b/src/API.jl
@@ -1497,9 +1497,9 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
 
     if Operations.is_manifest_current(ctx.env) === false
-        @warn """The project and manifest may be out of sync.
-            A project dependency has been added/removed or compat entry has changed since the manifest was last resolved.
-            The environment may need to be updated. Try `Pkg.resolve()` or `Pkg.update()`."""
+        @warn """The project and manifest may be out of sync as either project dependencies have been \
+        added/removed or compat entries have changed since the manifest was last resolved.
+        Try `Pkg.resolve()` or consider `Pkg.update()` if necessary."""
     end
 
     Operations.prune_manifest(ctx.env)
--- a/src/API.jl
+++ b/src/API.jl
@@ -1042,10 +1042,10 @@ precompile(pkg::String; kwargs...) = precompile(Context(), pkg; kwargs...)
 precompile(ctx::Context, pkg::String; kwargs...) = precompile(ctx, [pkg]; kwargs...)
 precompile(pkgs::Vector{String}=String[]; kwargs...) = precompile(Context(), pkgs; kwargs...)
 function precompile(ctx::Context, pkgs::Vector{String}=String[]; internal_call::Bool=false,
-                    strict::Bool=false, warn_loaded = true, kwargs...)
+                    strict::Bool=false, warn_loaded = true, already_instantiated = false, kwargs...)
     Context!(ctx; kwargs...)
     internal_call || resolve(ctx, silent_no_change = true)
-    instantiate(ctx; allow_autoprecomp=false, kwargs...)
+    already_instantiated || instantiate(ctx; allow_autoprecomp=false, kwargs...)
     time_start = time_ns()
 
     # Windows sometimes hits a ReadOnlyMemoryError, so we halve the default number of tasks. Issue #2323
@@ -1520,7 +1520,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     end
     # check if all source code and artifacts are downloaded to exit early
     if Operations.is_instantiated(ctx.env)
-        allow_autoprecomp && Pkg._auto_precompile(ctx)
+        allow_autoprecomp && Pkg._auto_precompile(ctx, already_instantiated = true)
         return
     end
 
@@ -1577,7 +1577,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     # Run build scripts
     allow_build && Operations.build_versions(ctx, union(new_apply, new_git); verbose=verbose)
 
-    allow_autoprecomp && Pkg._auto_precompile(ctx)
+    allow_autoprecomp && Pkg._auto_precompile(ctx, already_instantiated = true)
 end
 
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -1497,9 +1497,9 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
 
     if Operations.is_manifest_current(ctx.env) === false
-        @warn """The project and manifest may be out of sync. \
-        The project hash recorded into the manifest when it was resolved does not match the hash of the current project. \
-        A project dependency has been added/removed or compat entry has changed. The environment may need to be updated"""
+        @warn """The project and manifest may be out of sync.
+            A project dependency has been added/removed or compat entry has changed since the manifest was last resolved.
+            The environment may need to be updated. Try `Pkg.resolve()` or `Pkg.update()`."""
     end
 
     Operations.prune_manifest(ctx.env)
--- a/src/API.jl
+++ b/src/API.jl
@@ -439,6 +439,17 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     return
 end
 
+function is_manifest_current(ctx::Context = Context())
+    if haskey(ctx.env.manifest.other, "project_hash")
+        recorded_hash = ctx.env.manifest.other["project_hash"]
+        current_hash = Types.project_resolve_hash(ctx.env.project)
+        return recorded_hash == current_hash
+    else
+        # Manifest doesn't have a hash of the source Project recorded
+        return nothing
+    end
+end
+
 const UsageDict = Dict{String,DateTime}
 const UsageByDepotDict = Dict{String,UsageDict}
 
@@ -1493,6 +1504,12 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
         pkgerror("expected manifest file at `$(ctx.env.manifest_file)` but it does not exist")
     end
     Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
+
+    if is_manifest_current(ctx) === false
+        @warn """The project hash recorded into the manifest when it was resolved does not match the hash of the current project.
+            A project dependency has been added/removed or compat entry has changed. The environment may need to be updated"""
+    end
+
     Operations.prune_manifest(ctx.env)
     for (name, uuid) in ctx.env.project.deps
         get(ctx.env.manifest, uuid, nothing) === nothing || continue
--- a/src/API.jl
+++ b/src/API.jl
@@ -439,17 +439,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     return
 end
 
-function is_manifest_current(env::EnvCache)
-    if haskey(env.manifest.other, "project_hash")
-        recorded_hash = env.manifest.other["project_hash"]
-        current_hash = Types.project_resolve_hash(env.project)
-        return recorded_hash == current_hash
-    else
-        # Manifest doesn't have a hash of the source Project recorded
-        return nothing
-    end
-end
-is_manifest_current(ctx::Context = Context()) = is_manifest_current(ctx.env)
+is_manifest_current(ctx::Context = Context()) = Operations.is_manifest_current(ctx.env)
 
 const UsageDict = Dict{String,DateTime}
 const UsageByDepotDict = Dict{String,UsageDict}
@@ -1506,7 +1496,7 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     end
     Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
 
-    if is_manifest_current(ctx.env) === false
+    if Operations.is_manifest_current(ctx.env) === false
         @warn """The project and manifest may be out of sync. \
         The project hash recorded into the manifest when it was resolved does not match the hash of the current project. \
         A project dependency has been added/removed or compat entry has changed. The environment may need to be updated"""
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -142,6 +142,7 @@ function update_manifest!(env::EnvCache, pkgs::Vector{PackageSpec}, deps_map, ju
         env.manifest[pkg.uuid] = entry
     end
     prune_manifest(env)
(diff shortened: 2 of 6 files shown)


DilumAluthge:
> One thing this would allow is CI to easily check whether checked-in manifests are valid, as it's easy to forget to update them.

This is a good use case for this PR. Although, I am guessing that most people don't check the CI logs unless the CI job fails, so maybe a warning isn't sufficient for CI? We could do something fancy like check for the presence of the `CI` environment variable, and if we are running on CI, we throw a fatal error instead of just printing a warning.

IanButterworth:
Yeah, that's why I exposed `Pkg.is_manifest_current()` as part of the API. So you can do
```
@test Pkg.is_manifest_current()
```
but that requires people to add that to their tests.

Detecting `CI` and throwing sounds good to me.. but I'm not sure whether people will be happy with that

src/Types.jl
```diff
@@ -236,6 +237,9 @@ end
 Base.:(==)(t1::Project, t2::Project) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Project))
 Base.hash(t::Project, h::UInt) = foldr(hash, [getfield(t, x) for x in fieldnames(Project)], init=h)
 
+# only hash the deps and compat fields as they are the only fields that affect a resolve
+project_resolve_hash(t::Project, h::UInt = UInt(0)) = string(foldr(hash, [t.deps, t.compat], init=h))
```
fredrikekre, line 241: Wouldn't this be better as a SHA1?

src/Pkg.jl
```diff
@@ -612,6 +612,16 @@ Upgrades the format of the manifest file from v1.0 to v2.0 without re-resolving.
...
 
+"""
+    is_manifest_current(ctx::Context = Context())
+
+Returns whether the active manifest was resolved from the active project state.
+For instance, if the project had compat entries changed, but the manifest wasn't re-resolved, this would return false.
+
+If the manifest doesn't have the project hash recorded, `nothing` is returned.
+"""
+const is_manifest_current = API.is_manifest_current
```
fredrikekre, line 623: Perhaps this can simply be indicated by Pkg.status instead of a separate API function?

IanButterworth: Adding a warning to `Pkg.status` for visual review is a good idea. But one purpose of exposing this is to return a testable Bool. Perhaps we could make `Pkg.status(test_manifest_current = true)` return a Bool, but that seems like cramming functionality in.

src/Types.jl
IanButterworth: ah ok. Would you mind helping me with a code suggestion?

fredrikekre: Maybe something like this?
```
julia> using SHA

julia> io = IOBuffer();

julia> write(io, "write everything in a canonical order");

julia> bytes2hex(sha1(seekstart(io)))
"a28efaf9173c5c61bfe48ed1402e2c3408a69b3b"
```

IanButterworth: thanks!

IanButterworth:
Now with a warning in Pkg.status as well

```
(jl_hHdFrB) pkg> st
Status `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_hHdFrB/Project.toml`
  [336ed68f] CSV v0.9.10
Warning The project and manifest may be out of sync. The project hash recorded into the manifest when it was resolved does not match the hash of the current project. A project dependency has been added/removed or compat entry has changed. The environment may need to be updated

(jl_hHdFrB) pkg> instantiate
┌ Warning: The project and manifest may be out of sync. The project hash recorded into the manifest when it was resolved does not match the hash of the current project. A project dependency has been added/removed or compat entry has changed. The environment may need to be updated
└ @ Pkg.API ~/Documents/GitHub/Pkg.jl/src/API.jl:1510
```

IanButterworth:
Perhaps the warning message could be shorter. This might do
```
Warning: The project and manifest may be out of sync. A project dependency has been added/removed or compat entry has changed since the manifest was last resolved. The environment may need to be updated.
```

KristofferC:
Just to point out another way where things can get "out of date" is if you dev a package, and then you work on the deved package and i.e. add a dependency, the original project that deved the package will now have the wrong dependency list in its Manifest and will fail to load the newly added dependency of the deved package.

IanButterworth:
Yeah. That issue's covered at package load time by the `x doesn't have y in its dependencies` error, I guess?

The issue this tries to overcome is having a manifest which is a false resolve of a project, and instantiate doing as it's told and instantiating the manifest without warning, then the package loading without error, but during runtime hitting some different runtime behavior in a dep because that dep compat narrowed in the project but wasn't re-resolved.

src/Types.jl
```diff
@@ -236,6 +237,13 @@ end
 Base.:(==)(t1::Project, t2::Project) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Project))
 Base.hash(t::Project, h::UInt) = foldr(hash, [getfield(t, x) for x in fieldnames(Project)], init=h)
 
+# only hash the deps and compat fields as they are the only fields that affect a resolve
+function project_resolve_hash(t::Project, h::UInt = UInt(0))
+    iob = IOBuffer()
+    println(iob, t.deps)
```
fredrikekre, line 243: This relies on printing of this dict (which also isn't ordered). Perhaps something like:
```suggestion
    foreach(((name, uuid),) -> println(io, name, "=", uuid), sort!(collect(t.deps); by=first))
```

src/Types.jl
```diff
@@ -236,6 +237,13 @@ end
...
+    println(iob, t.compat)
```
fredrikekre, line 244: Same here. The compat entries here also include the string representation that was originally in the file, so changing
```
[compat]
Example = "1, 2"
```
to
```
[compat]
Example = "1,2"
```
would give a different hash.

Perhaps something like
```suggestion
    foreach(((name, compat),) -> println(name, "=", compat.val), sort!(collect(t.compat); by=first))
```

src/Operations.jl
```diff
@@ -1993,6 +1999,22 @@ function status(env::EnvCache, registries::Vector{Registry.RegistryInstance}, pk
     if mode == PKGMODE_MANIFEST || mode == PKGMODE_COMBINED
         print_status(env, old_env, registries, header, filter_uuids, filter_names; diff, ignore_indent, io, outdated, silent_no_change)
     end
+    if is_manifest_current(env) === false
+        printpkgstyle(io, :Warning, """The project and manifest may be out of sync. \
+        The project hash recorded into the manifest when it was resolved does not match the hash of the current project. \
```
KristofferC, line 2004: This feels like an implementation detail of how the detection is made. No need to show that to a user?

src/Operations.jl
```diff
@@ -1993,6 +1999,22 @@ function status(env::EnvCache, registries::Vector{Registry.RegistryInstance}, pk
...
+        A project dependency has been added/removed or compat entry has changed. The environment may need to be updated""", ignore_indent; color=Base.warn_color())
```
KristofferC, line 2005: It would be good to suggest `resolve` here.

src/Types.jl
```diff
@@ -219,6 +219,7 @@ Base.@kwdef mutable struct Compat
     val::VersionSpec
     str::String
 end
+Base.hash(t::Compat, h::UInt) = foldr(hash, [t.val, t.str], init=h)
```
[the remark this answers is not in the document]
fredrikekre: I also don't think this method is used for the updates SHA1 hashing so can probably be removed.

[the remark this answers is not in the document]
IanButterworth:
```suggestion
```

src/Operations.jl
IanButterworth: Yeah. I added "`Pkg.resolve()` or `Pkg.update()`" as the resolve will fail if the changed compat bounds require an update

src/API.jl
```diff
@@ -1493,6 +1495,13 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
         pkgerror("expected manifest file at `$(ctx.env.manifest_file)` but it does not exist")
     end
     Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
+
+    if Operations.is_manifest_current(ctx.env) === false
+        @warn """The project and manifest may be out of sync.
+            A project dependency has been added/removed or compat entry has changed since the manifest was last resolved.
+            The environment may need to be updated. Try `Pkg.resolve()` or `Pkg.update()`."""
```
IanButterworth, line 1502: @KristofferC what about this?
```suggestion
            The environment may need to be updated. Try `Pkg.resolve()` or consider `Pkg.update()`."""
```

src/Pkg.jl
fredrikekre: Who will use this function and when? What do you do with the result?

IanButterworth: During CI you could choose to check `@test Pkg.is_manifest_current()`.

Alternatively we could go with @DilumAluthge's suggestion of turning the `Pkg.instantiate()` warning into an error if `ENV["CI"]` is detected?

IanButterworth: An alternative name is `Pkg.is_environment_synced()`

IanButterworth: @fredrikekre what do you think?

IanButterworth: Can we unblock this?

KristofferC: It's just an internal function for now so let's just go with this for now and we can decide later if it is put in the manual.

GunnarFarneback:
> Yeah, that's why I exposed Pkg.is_manifest_current() as part of the API. So you can do
> ```
> @test Pkg.is_manifest_current()
> ```

Is that actually working? I rather get the impression that it tests if the manifest in the temporary test environment is current, which it kind of trivially is.

```
julia> using Pkg

julia> Pkg.generate("HashTest")
  Generating  project HashTest:
    HashTest/Project.toml
    HashTest/src/HashTest.jl
Dict{String, Base.UUID} with 1 entry:
  "HashTest" => UUID("21f82d62-6511-4653-97ca-6e7cc2914814")

julia> mkdir("HashTest/test")
"HashTest/test"

julia> write("HashTest/test/runtests.jl",
             """
             using Test, HashTest, Pkg
             @test Pkg.is_manifest_current()
             """)
58

julia> write("HashTest/Project.toml",
             """
             name = "HashTest"
             uuid = "21f82d62-6511-4653-97ca-6e7cc2914814"
             version = "0.1.0"

             [deps]

             [extras]
             Pkg = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"

             [targets]
             test = ["Pkg", "Test"]
             """)
225

julia> write("HashTest/Manifest.toml",
             """
             julia_version = "1.9.3"
             manifest_format = "2.0"
             project_hash = "0000000000000000000000000000000000000000"
             """)
106

julia> Pkg.activate("HashTest")
  Activating project at `/tmp/HashTest`

julia> Pkg.test()
┌ Warning: The project dependencies or compat requirements have changed since the manifest was last resolved.
│ It is recommended to `Pkg.resolve()` or consider `Pkg.update()` if necessary.
└ @ Pkg.API ~/.julia/juliaup/julia-1.9.3+0.x64.linux.gnu/share/julia/stdlib/v1.9/Pkg/src/API.jl:1687
     Testing HashTest
      Status `/tmp/jl_Gy6PhH/Project.toml`
  [21f82d62] HashTest v0.1.0 `/tmp/HashTest`
  [44cfe95a] Pkg v1.9.2 `@stdlib/Pkg`
  [8dfed614] Test `@stdlib/Test`
      Status `/tmp/jl_Gy6PhH/Manifest.toml`
  [21f82d62] HashTest v0.1.0 `/tmp/HashTest`
  [0dad84c5] ArgTools v1.1.1 `@stdlib/ArgTools`
  [56f22d72] Artifacts `@stdlib/Artifacts`
  [2a0f44e3] Base64 `@stdlib/Base64`
  [ade2ca70] Dates `@stdlib/Dates`
  [f43a241f] Downloads v1.6.0 `@stdlib/Downloads`
  [7b1f6079] FileWatching `@stdlib/FileWatching`
  [b77e0a4c] InteractiveUtils `@stdlib/InteractiveUtils`
  [b27032c2] LibCURL v0.6.3 `@stdlib/LibCURL`
  [76f85450] LibGit2 `@stdlib/LibGit2`
  [8f399da3] Libdl `@stdlib/Libdl`
  [56ddb016] Logging `@stdlib/Logging`
  [d6f4376e] Markdown `@stdlib/Markdown`
  [ca575930] NetworkOptions v1.2.0 `@stdlib/NetworkOptions`
  [44cfe95a] Pkg v1.9.2 `@stdlib/Pkg`
  [de0858da] Printf `@stdlib/Printf`
  [3fa0cd96] REPL `@stdlib/REPL`
  [9a3f8284] Random `@stdlib/Random`
  [ea8e919c] SHA v0.7.0 `@stdlib/SHA`
  [9e88b42a] Serialization `@stdlib/Serialization`
  [6462fe0b] Sockets `@stdlib/Sockets`
  [fa267f1f] TOML v1.0.3 `@stdlib/TOML`
  [a4e569a6] Tar v1.10.0 `@stdlib/Tar`
  [8dfed614] Test `@stdlib/Test`
  [cf7118a7] UUIDs `@stdlib/UUIDs`
  [4ec0a83e] Unicode `@stdlib/Unicode`
  [deac9b47] LibCURL_jll v7.84.0+0 `@stdlib/LibCURL_jll`
  [29816b5a] LibSSH2_jll v1.10.2+0 `@stdlib/LibSSH2_jll`
  [c8ffd9c3] MbedTLS_jll v2.28.2+0 `@stdlib/MbedTLS_jll`
  [14a3606d] MozillaCACerts_jll v2022.10.11 `@stdlib/MozillaCACerts_jll`
  [83775a58] Zlib_jll v1.2.13+0 `@stdlib/Zlib_jll`
  [8e850ede] nghttp2_jll v1.48.0+0 `@stdlib/nghttp2_jll`
  [3f19e933] p7zip_jll v17.4.0+0 `@stdlib/p7zip_jll`
Precompiling project...
  1 dependency successfully precompiled in 0 seconds. 3 already precompiled.
     Testing Running tests...
     Testing HashTest tests passed 
```

GunnarFarneback:
Btw, I think
```
@test Pkg.is_manifest_current(dirname(@__DIR__))
```
would be an acceptable API.

IanButterworth:
Oh you're right.. yes, not having a default arg would be a reasonable bug fix I think.

GunnarFarneback:
If you just remove the default argument all you have left is `is_manifest_current(::Context)`, which doesn't seem very user friendly?

IanButterworth:
Maybe we could leave the default arg for the `Context` method but error if it is run on an unnamed `Project.toml` and point people to the correct form? That might be nicer as a bugfix

GunnarFarneback:
I don't have a strong opinion about it, but to me it seems useful to be able to verify the manifest-up-to-dateness for any environment, not only packages.

GunnarFarneback:
See how you like #3701 and if something needs to be changed. It would probably be good with some additional and/or revised tests but I'll wait with that until we are happy with the design changes.

PR #2816, opened 2021-11-01
Precompile: don't instantiate again after an instantiate

IanButterworth:
Calling `Pkg.instantiate` automatically calls `Pkg.precompile` which calls `Pkg.instantiate` (which then knows not to call `Pkg.precompile` again)

This tells precompile that instantiate has already been called


fredrikekre:
Doesn't backport cleanly on 1.6.

PR #2819, opened 2021-11-01
[CI] re-enable the "Check Historical Stdlib Generator" CI check

DilumAluthge:


DilumAluthge:
Let's close and reopen several times to make sure the job always passes and never fails or times out.

DilumAluthge:
And indeed it timed out this time.

PR #2820, opened 2021-11-01
[CI] Actually use the correct arch when downloading Julia, and update one of the tests to use Permutations v0.4.9

DilumAluthge:
Currently, our "32-bit" CI jobs are not actually using 32-bit builds of Julia.

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -286,6 +286,7 @@ end
     isolate(loaded_depot=true) do
         Pkg.add(name="Permutations", version="0.3.2")
         if Sys.WORD_SIZE == 32
+            # The Permutations.jl v0.3.2 tests are known to fail on 32-bit Julia
             @test_skip Pkg.test("Permutations")
         else
             Pkg.test("Permutations")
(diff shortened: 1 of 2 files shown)


DilumAluthge:
The test failures should be fixed by https://github.com/scheinerman/Permutations.jl/pull/20

test/new.jl
```diff
@@ -284,7 +284,7 @@ end
 
 @testset "test: fallback when no project file exists" begin
     isolate(loaded_depot=true) do
-        Pkg.add(name="Permutations", version="0.3.2")
+        Pkg.add(name="Permutations", version="0.4.9")
```
fredrikekre, line 287: But now this version have a project file and this test is useless.

DilumAluthge: Ooof. Brutal.

Should we make a dummy example package ourselves, and use that for the test?

fredrikekre: Why not just guard the test here based on the arch instead?

DilumAluthge: Sure, we can do that instead.

DilumAluthge: Done.

staticfloat (approved):

PR #2824, opened 2021-11-03
More forcefully encourage `ArtifactOverrideLoading` to recompile when loaded the second time

### LINKED ISSUE #2806: The artifacts override tests fail for me locally
Steps to reproduce:

```julia
delete!(ENV, "JULIA_PKG_SERVER");
empty!(Base.DEPOT_PATH)
pushfirst!(Base.DEPOT_PATH, mktempdir(; cleanup = true))
import Pkg
cd(joinpath(dirname(dirname(pathof(Pkg))), "test"))
include("utils.jl")
include("artifacts.jl")
```

The error that I get:

```
Override.toml: Test Failed at /Users/dilum/local/julia/share/julia/stdlib/v1.8/Pkg/test/artifacts.jl:690
Expression: arty_path == barty_override_path
 Evaluated: "/var/folders/jy/7hh5zyw95cg2lh66lfpthqq80000gn/T/jl_25KvpZ/depot2/artifacts/64d0b4f8d9c004b862b38c4acfbd74988226995c" == "/var/folders/jy/7hh5zyw95cg2lh66lfpthqq80000gn/T/jl_25KvpZ/a_wild_barty_appears"
Stacktrace:
[1] macro expansion
 @ ~/local/julia/share/julia/stdlib/v1.8/Test/src/Test.jl:460 [inlined]
[2] (::Main.ArtifactTests.var"#126#142"{String})()
 @ Main.ArtifactTests ~/local/julia/share/julia/stdlib/v1.8/Pkg/test/artifacts.jl:690
Test Summary: | Pass  Fail  Total
Override.toml |   20     1     21
ERROR: LoadError: Some tests did not pass: 20 passed, 1 failed, 0 errored, 0 broken.
in expression starting at /Users/dilum/local/julia/share/julia/stdlib/v1.8/Pkg/test/artifacts.jl:1
```

staticfloat:
I'm not 100% certain why this worked before and doesn't now, I suspect a change in Julia internals around loading.

Fixes #2806
Closes #2810
Closes #2823

### DIFF
--- a/test/artifacts.jl
+++ b/test/artifacts.jl
@@ -637,7 +637,7 @@ end
             copy_test_package(depot_container, "ArtifactOverrideLoading")
             git_init_and_commit(joinpath(depot_container, "ArtifactOverrideLoading"))
             add_this_pkg()
-            Pkg.add(Pkg.Types.PackageSpec(
+            Pkg.develop(Pkg.Types.PackageSpec(
                 name="ArtifactOverrideLoading",
                 uuid=aol_uuid,
                 path=joinpath(depot_container, "ArtifactOverrideLoading"),
@@ -673,7 +673,9 @@ end
 
         # Force Julia to re-load ArtifactOverrideLoading from scratch
         pkgid = Base.PkgId(aol_uuid, "ArtifactOverrideLoading")
+        delete!(Base.module_keys, Base.loaded_modules[pkgid])
         delete!(Base.loaded_modules, pkgid)
+        touch(joinpath(depot_container, "ArtifactOverrideLoading", "src", "ArtifactOverrideLoading.jl"))
 
         # Verify that the hash-based overrides (and clears) worked
         @test artifact_path(foo_hash) == barty_override_path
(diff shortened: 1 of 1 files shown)


staticfloat:
This should fix https://github.com/JuliaLang/Pkg.jl/issues/2806

test/artifacts.jl
```diff
@@ -637,7 +637,7 @@ end
             copy_test_package(depot_container, "ArtifactOverrideLoading")
             git_init_and_commit(joinpath(depot_container, "ArtifactOverrideLoading"))
```
DilumAluthge, line 638:
```suggestion
```

If we are going to `Pkg.develop` the package (instead of `Pkg.add`ing it), there's no need for it to be a Git repo.

DilumAluthge (approved):
This fixes #2806 for me locally.

fredrikekre:
This approach doesn't work on release-1.6 so if this is crucial to have on LTS it needs another fix.

fredrikekre:
This approach doesn't work on release-1.6 so if this is crucial to have on LTS it needs another fix.

PR #2825, opened 2021-11-05
[CI] Re-enable the "Check Historical Stdlib" CI check

DilumAluthge:
With the latest version of Downloads.jl that is now available on Julia nightly, I think the timeouts should be fixed.


DilumAluthge:
And, of course, now it times out.

PR #2826, opened 2021-11-07
Fix Project equality when compat entries are present

IanButterworth:
The compat field of a Project appears to breaking equality even when the projects are identical.

```julia
julia> Pkg.Types.Compat(Pkg.Types.VersionSpec("0.1"), "0.1") == Pkg.Types.Compat(Pkg.Types.VersionSpec("0.1"), "0.1")
false
```

This was observed in https://github.com/JuliaRegistries/General/pull/47659#issuecomment-961932857 where an unchanged Project was being re-written to disk, without any actual changes to the file even though [this check](https://github.com/JuliaLang/Pkg.jl/blob/6f8e36db88ed0c7727b5616b2baa58787cdc4e95/src/Types.jl#L1014) should prevent that 

I'm adding the tests first to prove the problem exists first.

Also, @staticfloat the pkgserver tests are failing with
```
│   exception = HTTP/1.1 502 Bad Gateway while requesting https://pkg.julialang.org/registries
└ @ Pkg.Registry ~/Documents/GitHub/Pkg.jl/src/Registry/Registry.jl:66
```

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -219,6 +219,7 @@ Base.@kwdef mutable struct Compat
     val::VersionSpec
     str::String
 end
+Base.:(==)(t1::Compat, t2::Compat) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Compat))
 
 Base.@kwdef mutable struct Project
     other::Dict{String,Any} = Dict{String,Any}()
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -219,7 +219,7 @@ Base.@kwdef mutable struct Compat
     val::VersionSpec
     str::String
 end
-Base.:(==)(t1::Compat, t2::Compat) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Compat))
+Base.:(==)(t1::Compat, t2::Compat) = t1.val == t2.val
 
 Base.@kwdef mutable struct Project
     other::Dict{String,Any} = Dict{String,Any}()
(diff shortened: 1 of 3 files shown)


IanButterworth:
@staticfloat are the failures here expected?

staticfloat:
PkgServer breakage has been fixed.

src/Types.jl
```diff
@@ -219,6 +219,7 @@ Base.@kwdef mutable struct Compat
     val::VersionSpec
     str::String
 end
+Base.:(==)(t1::Compat, t2::Compat) = all(x -> (getfield(t1, x) == getfield(t2, x))::Bool, fieldnames(Compat))
```
KristofferC, line 222: Do you need to define `hash` now as well? Also, seems easier to just write out the actual fields here.

fredrikekre: Should probably just compare the `val`?

IanButterworth: Sounds good. Both changes made

fredrikekre:
Is this equality check used for something other than https://github.com/JuliaLang/Pkg.jl/pull/2202 ? That PR didn't implement it so I guess so? Possibly another internal function should be used instead of overloading `==` and `hash` then.

IanButterworth:
Sorry, I don't follow. I may be being slow though.. 

This PR specifically fixes a bug in the behavior introduced in #2202 where it will alway re-write a Project when compat fields are present. And because the equality check #2202 uses is just `==(::Project, ::Project)` I don't see how fixing this  could be anything other than defining `==` and `hash` for `Compat`?

fredrikekre:
I just meant that probably `==` as used in #2202 could be switched to use something like `have_same_toml_representation(p1, p2)::Bool` function instead. But since #2202 didn't add the equality check maybe it is used for something else?

KristofferC:
I don't think this applies to 1.6.

PR #2832, opened 2021-11-09
Use explicit module name in trailing comment

adigitoleo:
Change the trailing comment produced by `generate` to use the explicit module name.
This might be helpful to avoid confusion when scanning the file,
e.g. if there are submodules.

### DIFF
--- a/src/generate.jl
+++ b/src/generate.jl
@@ -66,7 +66,7 @@ function entrypoint(io::IO, pkg::AbstractString, dir)
 
             greet() = print("Hello World!")
 
-            end # module
+            end # module $pkg
             """
         )
     end


KristofferC:
From what I have seen the convention is to have a `# module` in the end of modules.

https://github.com/JuliaLang/julia/blob/8fc7e1b8c207fd73a9af7f6d9eae272878bbcd98/base/mpfr.jl#L1077
https://github.com/JuliaLang/julia/blob/8fc7e1b8c207fd73a9af7f6d9eae272878bbcd98/base/meta.jl#L453
https://github.com/JuliaLang/julia/blob/8fc7e1b8c207fd73a9af7f6d9eae272878bbcd98/base/libc.jl#L485

Perhaps `end # module $pkg` would be ok?

adigitoleo:
Thanks for the links. I guess those packages were generated with the current `generate` function so it makes sense. But your suggestion works well I think.

KristofferC:
> Thanks for the links. I guess those packages were generated with the current generate function so it makes sense.

Nah, they are files in the Julia Base so they were made long before `generate` existed.

PR #2833, opened 2021-11-09
Extend Pkg.offline to optionally handle offline registries. Use offline mode during precompile_script

IanButterworth:
- Extends `Pkg.offline()` with a `offline_registry` kwarg to set whether the registries have offline-accessible paths such that they can still be used to install packages in offline mode
- Uses this extended offline mode during `precompile_script` given the mock registry that is set up there uses local source paths. (Using the old `Pkg.offline(true)` wasn't enough as it disallowed using the registry to `Pkg.add` a package that wasn't already installed)

Instead of making this a build-specific hack, I thought there may be situations where users have locally linked registries that have local paths, that could be functional during _offline_ mode.

Fixes https://github.com/JuliaLang/julia/issues/43006

Previously lots of server requests were happening. None of these warnings happen now.

```
Generating REPL precompile statements... 25/36
#### inputting statement: ####
"Pkg.add(\"TestPkg\")"
####
julia> Pkg.add("TestPkg")
   Resolving package versions...
┌ Warning: could not download https://pkg.julialang.org/registries
│   exception = Could not resolve host: pkg.julialang.org while requesting https://pkg.julialang.org/registries
└ @ Pkg.Registry ~/Documents/GitHub/julia/usr/share/julia/stdlib/v1.8/Pkg/src/Registry/Registry.jl:76
     Cloning [bddd00b5-e60b-4520-a64a-f93bb8a01829] TestPkg from /var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_FmuO5s/TestPkg.jl
   Installed TestPkg ─ v0.1.0
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_FmuO5s/Project.toml`
  [bddd00b5] + TestPkg v0.1.0
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_FmuO5s/Manifest.toml`
  [bddd00b5] + TestPkg v0.1.0
  [ade2ca70] + Dates
  [de0858da] + Printf
  [4ec0a83e] + Unicode

Generating REPL precompile statements... 26/36
#### inputting statement: ####
"Pkg.develop(Pkg.PackageSpec(path=\"TestPkg.jl\"))"
####
julia> Pkg.develop(Pkg.PackageSpec(path="TestPkg.jl"))
   Resolving package versions...
┌ Warning: could not download https://pkg.julialang.org/registries
[...]

PR #2834, opened 2021-11-09
prevent internet access during precompile script

KristofferC:
I tried running this with internet turned off and I did not get any warnings. Can anyone confirm? Fixes JuliaLang/julia#43006

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -614,6 +614,7 @@ Upgrades the format of the current or specified manifest file from v1.0 to v2.0 
 const upgrade_manifest = API.upgrade_manifest
 
 function __init__()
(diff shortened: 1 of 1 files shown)


IanButterworth:
I tried this before the other PR and hit the registry info fetch inside `download_source` failure many times as it's never cached. I don't know why you didn't see those request failure warnings

KristofferC:
Maybe 1.7 vs master difference (https://github.com/JuliaLang/Pkg.jl/pull/2833#issuecomment-964141583).

KristofferC:
Let's put this minimal thing into 1.7 (since it is easy to backport etc) and then we can continue thinking about https://github.com/JuliaLang/Pkg.jl/pull/2833 which might very well be a good idea.

PR #2837, opened 2021-11-09
Remove "STDLIBS_BY_VERSION up-to-date" test

IanButterworth:
These tests caused a bit of an awkard circularity given Base CI now runs Pkg tests, so bumping a stdlib on base made these tests fail until you update Pkg, which wouldn't happen in that PR.
https://buildkite.com/julialang/julia-master/builds/5729

Given we have https://github.com/JuliaLang/Pkg.jl/blob/master/.github/workflows/check_hsg.yml running routinely here which is a more comprehensive test, just run that here and remove these tests.

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -2739,18 +2739,6 @@ using Pkg.Types: is_stdlib
     @test is_stdlib(pkg_uuid, nothing)
 end
 
-@testset "STDLIBS_BY_VERSION up-to-date" begin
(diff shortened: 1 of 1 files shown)


DilumAluthge:
When was this test first added? Do we need to backport this PR?

DilumAluthge (approved):

IanButterworth:
> When was this test first added?

https://github.com/JuliaLang/Pkg.jl/pull/2319

DilumAluthge:
Looks like we'll need to remove this test from both 1.6 and 1.7, then.

PR #2843, opened 2021-11-10
Backports release 1.6

KristofferC:
Just did the ones that worked and didn't require extensive backporting. Feel free to push any I miss here. (@IanButterworth).


IanButterworth:
Cross-referencing, I see these ones not here:
- https://github.com/JuliaLang/Pkg.jl/pull/2753
- https://github.com/JuliaLang/Pkg.jl/pull/2816
- https://github.com/JuliaLang/Pkg.jl/pull/2620

I'll look into them

IanButterworth:
I can't figure out why `ArtifactOverrideLoading` isn't found like it is on master, but I also can't see where the `path` field of the `PackageSpec` is used either here or on master..

KristofferC:
Let's try without that one then..

PR #2845, opened 2021-11-11
CI: Re-enable the "Check Historical Stdlib" CI check

DilumAluthge:


DilumAluthge:
Just as an FYI, if you notice that this CI check is failing (either on a pull request or on a commit on master), you can update the list of historical stdlibs by triggering a manual workflow dispatch run of the "Update Historical Stdlibs" GitHub Actions workflow.

The instructions are:
1. Go to https://github.com/JuliaLang/Pkg.jl/actions/workflows/run_hsg.yml
2. In the blue row that says "This workflow has a workflow_dispatch event trigger", click on the gray "Run workflow" button.
3. Click the green "Run button".

If for some reason the above URL does not work, here are the alternate instructions:1
1. Go to https://github.com/JuliaLang/Pkg.jl
2. In the row of tabs that says "Code", "Issues", "Pull Requests", etc., click on the "Actions" tab.
3. In the left-hand column, under "Workflows", click on "Update Historical Stdlibs".
4. In the blue row that says "This workflow has a workflow_dispatch event trigger", click on the gray "Run workflow" button.
5. Click the green "Run button".

PR #2847, opened 2021-11-12
Add example of deving local file path

logankilpatrick:
There is currently no example for deving a local package via its file path. This simple example adds one. We can switch this to show Kristoffer's file path if we want as well.

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -182,6 +182,10 @@ When tracking a path, the package manager will never modify the files at that pa
 If `dev` is used on a local path, that path to that package is recorded and used when loading that package.
 The path will be recorded relative to the project file, unless it is given as an absolute path.
 
+```julia-repl
+(v1.0) pkg> dev /Users/logankilpatrick/Desktop/Example
+```
+
 To stop tracking a path and use the registered version again, use `free`:
 
 ```julia-repl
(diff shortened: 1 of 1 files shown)


logankilpatrick:
@KristofferC Can I get a review on this?

docs/src/managing-packages.md
```diff
@@ -182,6 +182,10 @@ When tracking a path, the package manager will never modify the files at that pa
 If `dev` is used on a local path, that path to that package is recorded and used when loading that package.
 The path will be recorded relative to the project file, unless it is given as an absolute path.
 
+```julia-repl
+(v1.0) pkg> dev /Users/logankilpatrick/Desktop/Example
```
fredrikekre, line 186: Perhaps use the same user throughout the same file?

logankilpatrick: updated

Vectornaut:
The example added in ae8c21e seems to have been removed from the documentation, and the ability to use `dev` with a local path is no longer mentioned in "Developing Packages": it's now isolated in "Adding a local package." I don't know why these changes were made, because I can't find the commit that made them, but both of them made the docs harder for me to use: to learn that I could use `dev` with a local path, I had to dig up this issue!

Maybe the example could be put back into "Adding a local package," and a prominent reference to that section could be added to "Developing Packages"? I could write a draft and PR it if that would be helpful.

PR #2851, opened 2021-11-20
use Base.LOADING_CACHE during Pkg.precompile for stale checks

### LINKED ISSUE #2850: Auto-precompilation still performs excessive file system checks
Originally reported by @GunnarFarneback 

---

In https://github.com/JuliaLang/julia/pull/40890, we introduced a cache to reduce the number of excessive file system checks that were performed during calls to `require` as part of code loading.

However, automatic precompilation does not use that cache, which means that we are still performing excessive file system checks (e.g. calls to `stat`) during auto-precomp.

@GunnarFarneback was able to confirm that when auto-precomp is disabled (e.g. with `export JULIA_PKG_PRECOMPILE_AUTO=false`), the number of observed `stat` calls decreases drastically, which supports the hypothesis that auto-precomp is the culprit here.

IanButterworth:
Closes #2850 

@KristofferC does this look right?

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1236,6 +1236,7 @@ function precompile(ctx::Context, pkgs::Vector{String}=String[]; internal_call::
         end
     end
     tasks = Task[]
+    Base.LOADING_CACHE[] = Base.LoadingCache()
     for (pkg, deps) in depsmap # precompilation loop
         paths = Base.find_all_in_cache_path(pkg)
         sourcepath = Base.locate_package(pkg)
@@ -1341,6 +1342,8 @@ function precompile(ctx::Context, pkgs::Vector{String}=String[]; internal_call::
         wait(interrupted_or_done)
     catch err
         handle_interrupt(err)
+    finally
+        Base.LOADING_CACHE[] = nothing
     end
     notify(first_started) # in cases of no-op or !fancyprint
     save_precompile_state() # save lists to scratch space


KristofferC:
Yes, but would be good to confirm with `strace` that this helps I guess. Precompilation will have to stat many files but perhaps this will take down on the number of times e.g. Project.toml is statted.

IanButterworth:
I can't find a drop-in for strace on MacOS so I had to use the Instruments app

This is for a no-op `Pkg.precompile` with
```
(@v1.8) pkg> st
Status `~/.julia/environments/v1.8/Project.toml`
  [6e4b80f9] BenchmarkTools v1.2.0
  [295af30f] Revise v3.1.20
  [d2c73de3] GR_jll v0.62.0+0
```
using
```
--startup-file=no --project=/Users/ian/Documents/GitHub/Pkg.jl -e 'import Pkg; Pkg.activate(); Pkg.precompile()'
```

The timing isn't a fair comparison because the test with this PR isn't baked into the sysimage

# Master
Screen Shot 2021-11-20 at 2 32 30 PM

# This PR
Screen Shot 2021-11-20 at 2 34 08 PM

KristofferC:
Looks a lot better, right. Also, I apparently gotta learn and use Instruments more.

IanButterworth:
Yeah. Looks good! I'll merge

Instruments is great, but I really wanted to be able to count stat calls grouped by path, and that doesn't seem possible

Issue #2528, opened 2021-04-25
Error message for missing project file could be improved


logankilpatrick:
I am trying to do: 

```Julia
(@v1.6) pkg> add https://github.com/danielsuo/Crypto.jl
    Updating git-repo `https://github.com/danielsuo/Crypto.jl`
ERROR: could not find project file in package at `https://github.com/danielsuo/Crypto.jl` maybe `subdir` needs to be specified
```
and it fails. I tried a git clone locally and it worked without issue. My guess is that the project structure is causing the issues?

logankilpatrick:
It looks like per here: https://discourse.julialang.org/t/error-could-not-find-project-file-in-package-at-https-github-com/44059 you can't add a package without a Project file, perhaps the error message wording could be improved?

logankilpatrick:
Ehh, maybe that error message is fine, I still have not had coffee yet so will let it sink in and revisit

DilumAluthge:
We could change `project file` to `Project.toml file`, if that helps?

logankilpatrick:
Yeah, that will do the trick

KristofferC:
It could be called `JuliaProject.toml` as well which I guess maybe is why it is just called `project file`. On the other hand, no one uses that so perhaps it is fine to just hard code `Project.toml`.

PR #2854, opened 2021-11-23
Update error message to specify project.toml file

logankilpatrick:
Closes: https://github.com/JuliaLang/Pkg.jl/issues/2528 

Do I also need to update: https://github.com/JuliaLang/Pkg.jl/blob/b963d05a89cfacaa0fe57752de5f3fb439ccfa43/src/Operations.jl#L203 ?

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -796,7 +796,7 @@ end
 
 function resolve_projectfile!(env::EnvCache, pkg, project_path)
     project_file = projectfile_path(project_path; strict=true)
-    project_file === nothing && pkgerror(string("could not find project file in package at `",
+    project_file === nothing && pkgerror(string("could not find project.toml file in package at `",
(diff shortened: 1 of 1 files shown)


src/Types.jl
```diff
@@ -796,7 +796,7 @@ end
 
 function resolve_projectfile!(env::EnvCache, pkg, project_path)
     project_file = projectfile_path(project_path; strict=true)
-    project_file === nothing && pkgerror(string("could not find project file in package at `",
+    project_file === nothing && pkgerror(string("could not find project.toml file in package at `",
```
fredrikekre, line 799: I think just "project file" is used quite a lot, but perhaps:
```suggestion
    project_file === nothing && pkgerror(string("could not find project file (Project.toml or JuliaProject.toml) in package at `",
```

Issue #2705, opened 2021-08-16
Consistency in environment variable names


KristofferC:
I think all environment variables that one set for Pkg should start with `JULIA_PKG_`. Most of them already do but some don't. Perhaps it is worth making this consistent. We can keep the old env names around for backwards compat.

StefanKarpinski:
I think the only exceptions currently are:

* `JULIA_NUM_PRECOMPILE_TASKS` which is arguably not a Pkg-specific setting since precompilation could (is?) also be done in parallel when loading code, not just when Pkg installs it.
* `JULIA_PKGRESOLVE_ACCURACY` which could be changed to `JULIA_PKGRESOLVE_ACCURACY` but would ideally go a away at some point.

IanButterworth:
Assuming you meant:
- `JULIA_PKGRESOLVE_ACCURACY` which could be changed to `JULIA_PKG_RESOLVE_ACCURACY` ...

simonbyrne:
From @giordano 
> Another inconsistency is that some boolean variables take values true/false, others 0/1

KristofferC:
Should be fixed

PR #2858, opened 2021-11-24
Consistently parse all boolean environment variables

giordano:
Note: this is basically only checking the truthy values, so values like `"truw"` would silently be parsed as a falsy value, but as far as I can tell that's what's happening at the moment anyways, so I don't think there is any practical change.

Address https://github.com/JuliaLang/Pkg.jl/issues/2705#issuecomment-978147571.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -6,6 +6,7 @@ using Artifacts
 import Artifacts: artifact_names, ARTIFACTS_DIR_OVERRIDE, ARTIFACT_OVERRIDES, artifact_paths,
                   artifacts_dirs, pack_platform!, unpack_platform, load_artifacts_toml,
                   query_override, with_artifacts_directory, load_overrides
+import ..bool_env
 import ..set_readonly
 import ..GitTools
 import ..TOML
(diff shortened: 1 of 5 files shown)


StefanKarpinski:
I think there's some spots in PlatformEngines where we do this too. Might want to allow passing whether the default should be true or false? Although if the default is always false, I guess no need.

giordano:
> I think there's some spots in PlatformEngines where we do this too.

Oh yes, I saw
https://github.com/JuliaLang/Pkg.jl/blob/0d6a726e91a6a29b3f47fca2396ca5ea7fa5511b/src/PlatformEngines.jl#L228-L231
but that uses a three-value logic,  I wasn't sure whether to add that to the helper function `bool_env`, it'd require all other places to guard `nothing` values.

giordano:
Bump :slightly_smiling_face:

src/Artifacts.jl
```diff
@@ -357,7 +358,7 @@ function download_artifact(
             # Since tree hash calculation is still broken on some systems, e.g. Pkg.jl#1860,
             # and Pkg.jl#2317 so we allow setting JULIA_PKG_IGNORE_HASHES=1 to ignore the
             # error and move the artifact to the expected location and return true
-            ignore_hash = get(ENV, "JULIA_PKG_IGNORE_HASHES", nothing) == "1"
+            ignore_hash = bool_env("JULIA_PKG_IGNORE_HASHES")
```
IanButterworth, line 361: Perhaps
```suggestion
            ignore_hash = get_bool_env("JULIA_PKG_IGNORE_HASHES")
```

giordano: Not sure what you mean, is this a suggestion to rename the function?

IanButterworth: Yeah. Just a minor suggestion

IanButterworth (approved):

IanButterworth:
I think we'd want env vars that exist for shared functionality in both LTS and new releases to behave the same, so I'm going to mark this for backport

PR #2859, opened 2021-11-25
Rename `JULIA_PKGRESOLVE_ACCURACY` to `JULIA_PKG_RESOLVE_ACCURACY`

giordano:
The new name overrides the old one, which is still kept for compatibility.

Address https://github.com/JuliaLang/Pkg.jl/issues/2705#issuecomment-899720958.

### DIFF
--- a/src/Resolve/maxsum.jl
+++ b/src/Resolve/maxsum.jl
@@ -13,8 +13,11 @@ mutable struct MaxSumParams
                           # step
 
     function MaxSumParams()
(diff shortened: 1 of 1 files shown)


giordano:
Bump

IanButterworth:
For the same reason as https://github.com/JuliaLang/Pkg.jl/pull/2858#issuecomment-997225130 I'm marking this for backport

PR #2865, opened 2021-11-29
Warn if no registry is available during pkg install prompt

IanButterworth:
Redo at #2862 

This instead warns once if no registries are installed.

If someone is operating in a mode where they intentionally have no registry installed (I assume quite rare?), they could hide this always with `import Pkg; Pkg.REG_WARNED[] = true`

```
julia> using CSV
 │ Attempted to search for missing packages in Pkg registries but no registries are installed.
 └ Use Pkg mode to install a registry. Both `pkg> add` and `pkg> update` will install the default registries.

ERROR: ArgumentError: Package CSV not found in current path.
- Run `import Pkg; Pkg.add("CSV")` to install the CSV package.
Stacktrace:
 [1] macro expansion
   @ ./loading.jl:1007 [inlined]
 [2] macro expansion
   @ ./lock.jl:221 [inlined]
   @ Base ./loading.jl:988

julia> using CSV
ERROR: ArgumentError: Package CSV not found in current path.
- Run `import Pkg; Pkg.add("CSV")` to install the CSV package.
Stacktrace:
 [1] macro expansion
   @ ./loading.jl:1007 [inlined]
 [2] macro expansion
   @ ./lock.jl:221 [inlined]
   @ Base ./loading.jl:988
```

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -652,9 +652,20 @@ Some commands have an alias, indicated below.
 end
 
 const help = gen_help()
+const REG_WARNED = Ref{Bool}(false)
 
 function try_prompt_pkg_add(pkgs::Vector{Symbol})
     ctx = Context()
+    if isempty(ctx.registries)
+        if !REG_WARNED[]
+            printstyled(ctx.io, " │ "; color=:green)
+            printstyled(ctx.io, "Attempted to search for missing packages in Pkg registries but no registries are installed.\n")
+            printstyled(ctx.io, " └ "; color=:green)
+            printstyled(ctx.io, "Use Pkg mode to install a registry. Both `pkg> add` and `pkg> update` will install the default registries.\n\n")
+            REG_WARNED[] = true
+        end
+        return false
+    end
     available_uuids = [Types.registered_uuids(ctx.registries, String(pkg)) for pkg in pkgs] # vector of vectors
     available_pkgs = pkgs[isempty.(available_uuids) .== false]
     isempty(available_pkgs) && return false


src/REPLMode/REPLMode.jl
```diff
@@ -652,9 +652,20 @@ Some commands have an alias, indicated below.
...
+const REG_WARNED = Ref{Bool}(false)
 
 function try_prompt_pkg_add(pkgs::Vector{Symbol})
     ctx = Context()
+    if isempty(ctx.registries)
+        if !REG_WARNED[]
+            printstyled(ctx.io, " │ "; color=:green)
+            printstyled(ctx.io, "Attempted to search for missing packages in Pkg registries but no registries are installed.\n")
+            printstyled(ctx.io, " └ "; color=:green)
+            printstyled(ctx.io, "Use Pkg mode to install a registry. Both `pkg> add` and `pkg> update` will install the default registries.\n\n")
```
fredrikekre, line 664: Would be better to say something like
```
Run `using Pkg; Pkg.Registry.add()` to install default package registries.
```

IanButterworth: But that implies that the way to install default registries is `using Pkg; Pkg.Registry.add()`, but it's usually seamlessly managed by `pkg> add/update` etc. It feels odd to direct these people to the `using Pkg; Pkg...` style of interaction.

Also, it's worth noting that this only happens in interactive mode, where the Pkg repl is available. 
AFAICT the reason to give the `import Pkg; Pkg..` style action, like package loading errors do, is to cover when the pkg repl mode isn't available. i.e. in IJulia, which this user flow doesn't include

fredrikekre: Okay, suggest `pkg> registry add` then? `pkg> add<Enter>` and `pkg> update<Enter>` installs the default registries I think?

DilumAluthge:
Here is what I think should happen.

If someone does `using Foo`, and the `Foo` package is not installed, and no registries are available, we print a prompt that says `No package registries are available. Would you like to install the default registries? (y/n) [y]:`.

- If the user presses enter (without typing anything), we download the default registries.
- If the user types `y` and then presses enter, we download the default registries.
- If the user types `n` and then presses enter, we do not download any registries.

In other words, I think we should exactly mirror the prompt that we use for auto-installing packages. That way, if the user simply presses enter repeatedly, they get the "happy path": the default registries are downloaded, and all missing packages are installed.

src/REPLMode/REPLMode.jl
DilumAluthge: If we want to give the user the "most minimal" command for downloading the default registries, how about we suggest that they do `import Pkg; Pkg.status()`.

`Pkg.status()` will download the default registries if no registries are available. However, unlike both `Pkg.add` and `Pkg.update`, `Pkg.status()` will make no modifications to the active environment. In contrast, both `Pkg.add` and `Pkg.update` can make changes to the active environment, which might not be something a user wants if they only want to download the available registries.

fredrikekre: And even if it did, why would you suggest `Pkg.status()` as a way to install registries when there is an actual command for installing registries `Pkg.Registry.add()`?

DilumAluthge: I thought Ian didn't want to recommend `Pkg.Registry.add()`?

DilumAluthge: I just tested it in a fresh depot, and `Pkg.status()` definitely downloads the registries.

I agree that `Pkg.Registry.add()` is the best thing to recommend. I was just under the impression that Ian specifically didn't want to recommend `Pkg.Registry.add()`, and that we needed somerhing else.

fredrikekre: > I just tested it in a fresh depot, and Pkg.status() definitely downloads the registries.

Okay this is new in 1.7 then. Why?

DilumAluthge: Ahhh, was the debate just about recommending `import Pkg; Pkg.Registry.add()` versus recommending `] registry add`?

I don't really feel strongly between the two of those.

DilumAluthge:
Sure. But simply pressing enter twice (first for the download registries, second for the intall package) is a lot faster and easier than:
1. You see a message that says "please do `import Pkg; Pkg.Registry.add()`".
2. You copy that to your clipboard, paste it, and press enter.
3. Now you do `using Foo` again.

src/REPLMode/REPLMode.jl
DilumAluthge:
```suggestion
printstyled(ctx.io, "Use Pkg mode to install a registry. In the Pkg REPL mode, `pkg> registry add` will install the default registries.\n\n")
```

DilumAluthge: > Okay this is new in 1.7 then. Why?

Yeah it must be a new feature. I'm not sure why it was added.

Anyway, I went back through this discussion, and I realize I misunderstood the issue with `Pkg.Registry.add()` - it was just that we wanted to recommend the REPL mode instead.

fredrikekre:
Honestly, what is the reason for this warning at all? The correct way to install the missing package is already in the error message.

DilumAluthge:
> Honestly, what is the reason for this warning at all? The correct way to install the missing package is already in the error message.

That's a good point. The user can just do `import Pkg; Pkg.add("CSV")` and it will both download the registries and install the missing package.

However, my thinking is that the auto-install feature is really convenient, so not make that feature available to users that don't have any registries available.

IanButterworth:
> The correct way to install the missing package is already in the error message.

I think that's the rub. They've been told by the 1.7 blog post that _the correct way to install missing packages_ is just to write `using Foo`. 

It looks like 1.7.0 is going out. I guess we can see how it goes without this warning and adjust as necessary

fredrikekre:
> They've been told by the 1.7 blog post that the correct way to install missing packages is just to write using Foo.

The correct way to install packages is to use the package manager, no?

The only one-time inconvenience here is for users that install Julia *for the first time* and do `using Package` before using the package manager. I feel like that will be very few people.

> However, my thinking is that the auto-install feature is really convenient, so not make that feature available to users that don't have any registries available.

Yes, and they will have to do this *once* and after that the convenience feature is available to new users just like old users.

IanButterworth:
> The correct way to install packages is to use the package manager, no?

I should have said "a" correct way.

Issue #2583, opened 2021-06-01
Compatibility debugging


fonsp:
In addition to https://github.com/JuliaLang/Pkg.jl/issues/2582, @staticfloat also proposed adding "compatibility debugging" features to IDEs ([like Pluto.jl](https://github.com/fonsp/Pluto.jl/pull/844)): if a package has a more recent version but it _cannot_ be installed, show the user _why_ not (i.e. which other direct dep is holding it back). 

I think that this is also valuable directly in the Pkg REPL. For example:

```julia
pkg> add Revise
... Revise 2 added! ...

pkg> # hmmm I heard that Revise 3 is available, let's try update!

pkg> update Revise
... No changes to Manifest ...
... Revise 3 is available but Example restricts it to 2 ...
```

Currently, I can achieve this by copying my environment to a temporary directory, and calling `add Revise@3.2.1` (if `3.2.1` is latest according to the registry). The resulting error message will be the debugging information I need. However, like in https://github.com/JuliaLang/Pkg.jl/issues/2582, a "dry run" option to `Pkg.add` would be very useful, allowing me to quickly call `Pkg.add` without downloading source or binaries.

In the future, perhaps this behaviour can be implemented into Pkg itself?

KristofferC:
I did some WIP here https://github.com/JuliaLang/Pkg.jl/pull/2284. Might be able to take something from there.

KristofferC:
Cosed by #2284

KristofferC:
You can probably use the `status_compat_info` as an API and then display the result of that.

PR #2866, opened 2021-11-29
Warn if non-latest version is installed after explicit `up`

### LINKED ISSUE #1655: Optionally warn if `update` cannot install latest version of a package
The following has bitten me and colleagues many times: we want to install the latest version of a package (e.g. `CxxWrap`), so I do `] update CxxWrap`, and I see this:
```
(v1.3) pkg> update CxxWrap
 Resolving package versions...
  Updating `~/.julia/environments/v1.3/Project.toml`
 [no changes]
  Updating `~/.julia/environments/v1.3/Manifest.toml`
 [no changes]
```

Odd, I wonder, nothing happened... Did I already have the latest version? Hm, let's check:
```
(v1.3) pkg> status CxxWrap
    Status `~/.julia/environments/v1.3/Project.toml`
  [b99e7846] BinaryProvider v0.5.8
  [1f15a43c] CxxWrap v0.8.2
```

Oh no, that's *not* the latest. Which leads me to my first question:

1. *Is there any way to find out what the latest version of a package from the Julia REPL? Or get a list of all available package versions?*

What's going on? Well, after some searching around, I may discover that some other package has a dependency on CxxWrap with an upper bound that prevents any newer version of CxxWrap installed. The only way I found to do that was to manually go through each installed/active package, and look at its dependencies. So my second question:

2. *Is there any way to get a list of active packages that depend on a given package?*

And finally, the feature request from the headline of this issue:

3. Would it be possible/acceptable/... to change the behavior of `update` to print a warning if a package is not updated to the latest due to a constraint?

I can simulate this by removing the package and then adding it back with a version constraint:
```
(v1.3) pkg> rm CxxWrap
  Updating `~/.julia/environments/v1.3/Project.toml`
  [1f15a43c] - CxxWrap v0.8.2
  Updating `~/.julia/environments/v1.3/Manifest.toml`
 [no changes]

(v1.3) pkg> add CxxWrap@0.9.0
 Resolving package versions...
ERROR: Unsatisfiable requirements detected for package Singular [bcd08a7b]:
 Singular [bcd08a7b] log:
 ├─possible versions are: 0.1.0 or uninstalled
 ├─restricted to versions 0.1.0 by an explicit requirement, leaving only versions 0.1.0
 └─restricted by compatibility requirements with CxxWrap [1f15a43c] to versions: uninstalled — no versions left
   └─CxxWrap [1f15a43c] log:
     ├─possible versions are: [0.7.0-0.7.4, 0.8.0-0.8.2, 0.9.0] or uninstalled
     └─restricted to versions 0.9.0 by an explicit requirement, leaving only versions 0.9.0
```

But that requires me to know that a newer version is available (see question 1); I guess if there was a "pseudo version" to indicate the latest, that'd also work; something like `add CxxWrape@latest` ?

Also, it seems I have to remove the package first? At least I couldn't figure out a way to make a version specifier work with `update`; doing `]update CxxWrap@0.9.0` does the same as `update CxxWrap`.

CC @fieker

fonsp:
This is just a draft PR to get the ball rolling to fix #1655. This only warns that a more recent version is available, it does not say why (I don't know how to implement that: #2583).

# Demo
```julia
(Pkg) pkg> activate --temp
  Activating new project at `/var/folders/v_/fhpj9jn151d4p9c2fdw2gv780000gn/T/jl_9ciGWf`

(jl_9ciGWf) pkg> add Pluto@0.16.4 Configurations

# This installs Configurations@0.16 because of a bound by this old version of Pluto. 0.17 is available.

(jl_9ciGWf) pkg> up Configurations

  No Changes to `/private/var/folders/v_/fhpj9jn151d4p9c2fdw2gv780000gn/T/jl_WblATC/Project.toml`
  No Changes to `/private/var/folders/v_/fhpj9jn151d4p9c2fdw2gv780000gn/T/jl_WblATC/Manifest.toml`

┌ Warning: Configurations (at 0.16.6) could not be updated to the latest version (0.17.0). Run `Pkg.add(name="Configurations", version=v"0.17.0")` to learn why.
└ @ Pkg.Operations ~/Documents/PkgClone.jl/src/Operations.jl:1309
```

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1300,6 +1300,13 @@ function up(ctx::Context, pkgs::Vector{PackageSpec}, level::UpgradeLevel;
     download_artifacts(ctx.env, julia_version=ctx.julia_version, io=ctx.io)
     write_env(ctx.env; skip_writing_project) # write env before building
     show_update(ctx.env, ctx.registries; io=ctx.io, silent_no_change)
+    for pkg in pkgs
+        latest_available = get_latest_compatible_version(ctx, pkg.uuid, Pkg.Types.VersionSpec())
+        currently_installed = manifest_info(ctx.env.manifest, pkg.uuid).version
+        if latest_available > currently_installed
+            @warn """$(pkg.name) (at $(currently_installed)) could not be updated to the latest version ($(latest_available). Run `Pkg.add(name="$(pkg.name)", version=$(latest_available))` to learn why."""
+        end
+    end
     build_versions(ctx, union(new_apply, new_git))
 end
 
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1292,6 +1292,8 @@ function up(ctx::Context, pkgs::Vector{PackageSpec}, level::UpgradeLevel;
     for pkg in pkgs
         up_load_manifest_info!(pkg, manifest_info(ctx.env.manifest, pkg.uuid))
     end
+    
+    oringal_pkgs = pkgs
     pkgs = load_direct_deps(ctx.env, pkgs; preserve = (level == UPLEVEL_FIXED ? PRESERVE_NONE : PRESERVE_DIRECT))
     check_registered(ctx.registries, pkgs)
     deps_map = resolve_versions!(ctx.env, ctx.registries, pkgs, ctx.julia_version)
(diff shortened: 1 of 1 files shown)


KristofferC:
The `--outdated` flag to status shows this information already. What could be added is a small note very similar to the download icon that gets added when a package is not installed:

image

There could be a similar icon added to packages that are not on the latest version with a "hook" towards the `--outdated` flag to `status` for more information.

IanButterworth:
@fonsp check out #2906, I think it might address the issue, in combination with the `--outdated` flag.
I'd be interested to see what you make of it

IanButterworth:
I think https://github.com/JuliaLang/Pkg.jl/pull/2906 covers this now

PR #2876, opened 2021-12-05
Use `Base.set_active_project`

timholy:
This will allow Revise to track changes in the running source when
users switch environments. Replaces #2744.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1603,7 +1603,7 @@ function activate(;temp=false, shared=false, prev=false, io::IO=stderr_f())
     if !isnothing(Base.active_project())
         PREV_ENV_PATH[] = Base.active_project()
     end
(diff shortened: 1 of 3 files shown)


fredrikekre:
Just for my understanding, this is needed because Revise does stuff asynchronously in the background now?

fredrikekre:
Okay, but Revise is still only triggered on <kbd>Enter</kbd> (and `Revise.revise()`) as before (and not in the callback)? In that case, why doesn't Revise just check the result of `Base.active_project()` when triggered? In any case, what you really want is to somehow monitor `Base.load_path()` (https://github.com/JuliaLang/julia/issues/37318), right?

DilumAluthge:
@timholy Is this PR still needed?

PR #2877, opened 2021-12-06
tests: use parent testsets to summarize tests

IanButterworth:
I think it's helpful to show the test summary at the end.

Pushing a verbose version first to show timing. I'll turn verbose off once that's run


DilumAluthge:
Can you post some screenshots before this PR versus after this PR?

IanButterworth:
The screenshots would be bigger than a single browser viewport. 

The reports themselves can be reviewed on the two commits in this PR.

Issue #2888, opened 2021-12-09
Improve docs around overriding artifacts and specific libraries


odow:
I'm writing docs for JuMP on how to replace JLL solvers with custom ones: https://github.com/jump-dev/JuMP.jl/pull/2824

Spurred by this discussion on Slack:
```
Oscar Dowson Yesterday at 5:58 PM
Is it possible to override a single file within an artifact? For example, if I have a JLL, can I override the location of libfoo but keep the default foo executable? I'm writing some documentation for JuMP: https://github.com/jump-dev/JuMP.jl/pull/2823

Kristoffer Carlsson  13 hours ago
I don't think so

Oscar Dowson  13 hours ago
Okay. I can see why it would be difficult. In which case, the docs are clear enough.

Stefan Karpinski  5 hours ago
I think that @staticfloat had a plan for this involving preferences, but I’m not sure if it ever got implemented

Elliot Saba:horse:  4 hours ago
Yes, JLLWrappers allows for this

Elliot Saba:horse:  4 hours ago
https://github.com/JuliaPackaging/JLLWrappers.jl/pull/27

Elliot Saba:horse:  4 hours ago
No real docs so far, because it won't work (yet) for any stdlibs

Oscar Dowson  39 minutes ago
Awesome! Works a treat. I'll draft some docs for JuMP. Would you want docs for it? If so, where? The JLLWrappers README?

Stefan Karpinski  31 minutes ago
I feel like having some docs in Pkg for what to do when you need to use system libraries rather than artifacts would be good
```

I'll have a go at improving the current Pkg docs. Opening this issue to remind me.

PR #2889, opened 2021-12-10
Add docs on how to override a library within a JLL

odow:
Closes https://github.com/JuliaLang/Pkg.jl/issues/2888. 

The Pkg docs assume a fairly high level of understanding, so I didn't go into detail like I did in https://jump.dev/JuMP.jl/dev/developers/custom_solver_binaries/. 

Can expand if needed.


giordano:
Should this be in https://docs.binarybuilder.org/stable/jll/ instead?  I'm not sure this has much to do with Pkg directly.

odow:
Ah, yes, the binary builder docs might be a better fit. I'll move this and then add a link to the relevant binary builder docs from this section. 

I've been testing the JuMP documentation on different solvers and it works seamlessly. We're going to drop 1.0 support in the next round of releases and get rid of all vestiges of BinaryProvider `build.jl`.

PR #2893, opened 2021-12-13
Add environment variables for disabling auto-GC and/or disable the writing of environment usage

DilumAluthge:
1. If you want to record environment usage, and you want to automatically run Pkg GC:
    - Do nothing. This is the default behavior.
2. If you want to record environment usage, but you do not want to automatically run Pkg GC:[^1]
    - `export JULIA_PKG_AUTO_GC="false"`
3. If you do not want to record environment usage, and you do not want to automatically run Pkg GC:
    - `export JULIA_PKG_WRITE_ENV_USAGE="false"`

Intentionally undocumented for now, so we can give it a try without having to commit to the API.

[^1]: For example, this use case would make sense if you do not want to automatically run Pkg GC, but you might still want to manually run `Pkg.gc()`.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -672,7 +672,7 @@ end
 const DEPOT_ORPHANAGE_TIMESTAMPS = Dict{String,Float64}()
 const _auto_gc_enabled = Ref{Bool}(true)
 function _auto_gc(ctx::Types.Context; collect_delay::Period = Day(7))
-    if !_auto_gc_enabled[]
+    if !(_auto_gc_enabled[] && Types.env_var_auto_gc() && Types.env_var_write_env_usage())
         return
     end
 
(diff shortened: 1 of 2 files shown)


KristofferC:
Is the desire for this based on the fact that the current behavior of the auto gc is buggy with the broken usage TOML file?

DilumAluthge:
Yeah. And I want to give people a short-term escape hatch, especially people running a lot of Julia processes simultaneously.

But hopefully we'll fix the issues soon. So I want to leave these environment variables undocumented, so that we can remove them once the fancier solution (pidfile or whatever) has landed.

KristofferC:
The GC thing doesn't error, it just prints a logging message so I don't see what the big deal is with that. At least not big enough to warrant adding extra API for it.

PR #2896, opened 2021-12-17
package info from name.

oscardssmith:
This provides an easy way to get the package info from only it's name.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -48,6 +48,10 @@ function Base.:(==)(a::PackageInfo, b::PackageInfo)
         a.source == b.source && a.dependencies == b.dependencies
 end
 
+function package_info(pkg::AbstractString)::PackageInfo
+    dependencies()[Base.identify_package(name).uuid]
+end
+
 function package_info(env::EnvCache, pkg::PackageSpec)::PackageInfo
     entry = manifest_info(env.manifest, pkg.uuid)
     if entry === nothing
--- a/src/API.jl
+++ b/src/API.jl
@@ -49,7 +49,7 @@ function Base.:(==)(a::PackageInfo, b::PackageInfo)
 end
 
 function package_info(pkg::AbstractString)::PackageInfo
-    dependencies()[Base.identify_package(pkg).uuid]
+    filter(p -> p.second.name == pkg, dependencies()) |> values |> only
 end
 
 function package_info(env::EnvCache, pkg::PackageSpec)::PackageInfo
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -48,6 +48,10 @@ function Base.:(==)(a::PackageInfo, b::PackageInfo)
         a.source == b.source && a.dependencies == b.dependencies
 end
 
+function package_info(pkg::AbstractString)::PackageInfo
+    dependencies()[Base.identify_package(name).uuid]
```
KristofferC, line 52: `identify_package` can return nothing, also I don't see a point in calling into full code loading when you have all the data in `dependencies()`.

oscardssmith: without `identify_package`, how can I get the UUID from the name?

DilumAluthge:
```julia
filter(p -> p.second.name == "Example", dependencies()) |> keys |> collect
```

If there is only one UUID that matches the desired name, you can do:

```julia
filter(p -> p.second.name == "Example", dependencies()) |> keys |> collect |> only
```

src/API.jl
```diff
@@ -48,6 +48,10 @@ function Base.:(==)(a::PackageInfo, b::PackageInfo)
         a.source == b.source && a.dependencies == b.dependencies
 end
 
+function package_info(pkg::AbstractString)::PackageInfo
+    dependencies()[Base.identify_package(pkg).uuid]
```
DilumAluthge, line 52:
```suggestion
    filter(p -> p.second.name == pkg, dependencies()) |> values |> only
```

KristofferC:
This is just a trivial filtering from the output of `dependencies()`. I don't think it needs to be explicitly added to Pkg.

oscardssmith:
While it is a relatively simple filtering, I think it is a common one to want to do. I made this PR after I saw it done badly in the [h2oai db-benchmark](https://github.com/h2oai/db-benchmark/blob/c7421051af7530951d16b1505158371aebc0d2c1/_helpers/helpers.jl#L5). The other solution to this that I would be OK with is if `Pkg.status(pkg::String)` returned data rather than only printing it. Basically, what I want is to make `status` more useful for use by scripting, since it currently is only useful interactively.

zsz00:
what is status of this pr , optimize or give up ?

oscardssmith:
if there is a version of this PR that works be merged, I would write it, but it's not clear to me that there is.

PR #2897, opened 2021-12-17
Don't return "julia" as an available missing package

IanButterworth:
@KristofferC perhaps this is a better fix for https://github.com/JuliaLang/julia/issues/43440 than https://github.com/JuliaLang/julia/pull/43449


KristofferC:
No idea really what is best. Merge whichever :P

IanButterworth:
I think this one is. Plus it has a test

PR #2905, opened 2021-12-30
Remove precompile resolve

### LINKED ISSUE #2904: `precompile()` should not call `resolve()` if the environment is already resolved
Originally reported by @giordano:

```
(gtk) pkg> precompile
    Updating `~/Downloads/gtk/Project.toml`
  [77ec8976] ↑ GTK3_jll v3.24.30+0 ⇒ v3.24.30+1
    Updating `~/Downloads/gtk/Manifest.toml`
```

@IanButterworth suggested the following, which I think is the right way to fix this:

> I think it makes more sense to replace that resolve with an `isresolved` check, and a warning

IanButterworth:
A resolve was previously added to direct calls of `Pkg.precompile` to try to prevent the unexpected behavior users saw when the manifest was out of sync with the project.

The manifest out of sync warning introduced in https://github.com/JuliaLang/Pkg.jl/pull/2815 actually serves as a warning if the manifest isn't resolved/in-sync with the project, so this resolve can be removed.

Closes #2904 

But removing the resolve does expose the risk of not detecting circular deps fully, as demonstrated by two resolves needing to be added to the circular dep tests.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1502,9 +1502,8 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     Types.check_warn_manifest_julia_version_compat(ctx.env.manifest, ctx.env.manifest_file)
 
     if Operations.is_manifest_current(ctx.env) === false
-        @warn """The project and manifest may be out of sync as either project dependencies have been \
-        added/removed or compat entries have changed since the manifest was last resolved.
-        Try `Pkg.resolve()` or consider `Pkg.update()` if necessary."""
+        @warn """The project dependencies or compat requirements have changed since the manifest was last resolved.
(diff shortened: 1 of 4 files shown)


DilumAluthge:
> The manifest out of sync warning introduced in #2815 actually serves as a warning if the manifest isn't resolved/in-sync with the project, so this resolve can be removed.

Does that warning print on `] precompile`, though? I was under the impression that the warning only printed on `] instantiate` or `] st`.

IanButterworth:
A direct `Pkg.precompile()` calls `instantiate`. See just below the first change in the diff

DilumAluthge:
@IanButterworth Is this good to merge?

IanButterworth:
I'm a little concerned that this will resurface undetected circular dep hangs, so it might also need a fallback detector for that.

But that could be a different PR

PR #2912, opened 2022-01-02
deprecate Pkg.activate do-block syntax from public API

IanButterworth:
This is what was decided in https://github.com/JuliaLang/Pkg.jl/pull/2639
cc. @st-- @fredrikekre 

```
% ../julia/julia --project --depwarn=yes
...
julia> import Pkg

julia> Pkg.activate("/tmp/123") do
       Pkg.status()
       end
┌ Warning: `activate(f::Function, new_project::AbstractString)` is deprecated, use `Operations.activate(f::Function, new_project::AbstractString)` instead.
│   caller = top-level scope at REPL[2]:1
└ @ Core REPL[2]:1
Status `/private/tmp/123/Project.toml` (empty project)
```

But does it achieve the right message? It's not saying "don't use this", just "use this other function".
Perhaps `@deprecate` needs a `warn_against = true` mode that says something like:

```
┌ Warning: `activate(f::Function, new_project::AbstractString)` is deprecated, this functionality should not be used but  `Operations.activate(f::Function, new_project::AbstractString)` does exist as a replacement.
│   caller = top-level scope at REPL[2]:1
└ @ Core REPL[2]:1
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1671,15 +1671,9 @@ function activate(path::AbstractString; shared::Bool=false, temp::Bool=false, io
     add_snapshot_to_undo()
     return nothing
 end
-function activate(f::Function, new_project::AbstractString)
-    old = Base.ACTIVE_PROJECT[]
-    Base.ACTIVE_PROJECT[] = new_project
-    try
-        f()
-    finally
-        Base.ACTIVE_PROJECT[] = old
-    end
-end
+
+@deprecate activate(f::Function, new_project::AbstractString) Operations.activate(f::Function, new_project::AbstractString)
+
 
 function compat(ctx::Context; io = nothing)
     io = something(io, ctx.io)
@@ -1899,7 +1893,7 @@ end
 function upgrade_manifest(man_path::String)
     dir = mktempdir()
     cp(man_path, joinpath(dir, "Manifest.toml"))
-    Pkg.activate(dir) do
+    Pkg.Operations.activate(dir) do
         Pkg.upgrade_manifest()
     end
     mv(joinpath(dir, "Manifest.toml"), man_path, force = true)
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1003,6 +1003,17 @@ end
 ##############
 # Operations #
 ##############
+
+function activate(f::Function, new_project::AbstractString)
+    old = Base.ACTIVE_PROJECT[]
+    Base.ACTIVE_PROJECT[] = new_project
+    try
+        f()
+    finally
+        Base.ACTIVE_PROJECT[] = old
+    end
+end
+
 function rm(ctx::Context, pkgs::Vector{PackageSpec}; mode::PackageMode)
     drop = UUID[]
     # find manifest-mode drops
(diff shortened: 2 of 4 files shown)


st--:
@IanButterworth yes it'd be great if `@deprecate` had a more explicit warning against usage as you suggest. That's unlikely to happen anytime soon, so how about the following instead? Add an (unexported)
```julia
function _deprecated_activate(f::Function, new_project)
    @warn "You should not use this form [maybe short explanation why]"
    Operations.activate(f, new_project)
end
```
and then
```julia
@deprecate activate(f::Function, new_project::AbstractString) _deprecated_activate(f::Function, new_project::AbstractString)
```
(This is based on my understanding that `Operations.activate` _is_ needed by Pkg-internal code and hence _shouldn't_ raise a warning.)

st--:
> > That's unlikely to happen anytime soon
> 
> Why? If people like the idea we could add the feature for 1.8

Well, "soon" is relative - extrapolating from the last few releases, it'd be about mid to late summer, I just thought you might want to merge this PR in first so wanted to provide a low-effort suggestion that might work well enough until a `@denounce` macro is available.:)

IanButterworth:
I don't think deprecation like this would be backported, so the earliest this would land would be 1.8, so both can happen at the same time

giordano:
For what is worth, we use this `Pkg.activate` method in BinaryBuilder: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/e1345f7ee6c57b9e263723ce8d02970cbc7d2bfd/src/Prefix.jl#L453

fredrikekre:
> For what is worth, we use this Pkg.activate method in BinaryBuilder:

It is invalid if the user doesn't have the default `LOAD_PATH`, so that code should probably be fixed then.

fredrikekre:
Same with the two uses inside of Pkg here btw.

IanButterworth:
What if instead we fixed it with something like
```
function activate(f::Function, new_project::AbstractString)
    activate(new_project)
    try
        f()
    finally
        activate(prev = true)
    end
end
```

fonsp:
Not sure if it's too late to object, but there is currently no public API to query/manipulate an environment that is not the current environment. I believe this is very useful, and our particular use case is the Pluto server process, which maintains multiple project environments in parallel.

The argument for removing this behaviour is that [changing environments dynamically is not recommended](https://github.com/JuliaLang/Pkg.jl/pull/2639#issuecomment-865965532). But then why do we have `Pkg.activate(::String)`?

Ideally, I would like `Pkg.Types.Context` to be public API. (In internal Pkg API, `Context` is always the first argument to any function, [e.g. `Pkg.API.add(context, ["Plots", "Test"])`](https://github.com/JuliaLang/Pkg.jl/blob/16c3d6a7ad0075a56c8da300811b0e5249ab9e39/src/API.jl#L226).) This is easy API to manage multiple environments, and it does not modify `Base.LOAD_PATH`.

fonsp:
Please don't remove `Pkg.activate`! 👀 That's not what I meant at all

PR #2917, opened 2022-01-03
Reduce noise of test suite

### LINKED ISSUE #1097: turn off noisy test output
When we run the integration tests on JuliaLang, these tests are generating a lot of very noisy output in the build logs. Generally we have ensured there that only a failing test should produce output.

IanButterworth:
Trying to help with https://github.com/JuliaLang/Pkg.jl/issues/1097

Fixes #1097

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -973,7 +973,7 @@ function build_versions(ctx::Context, uuids::Set{UUID}; verbose=false)
         fancyprint && show_progress(ctx.io, bar)
 
         sandbox(ctx, pkg, source_path, builddir(source_path), build_project_override) do
-            flush(stdout)
+            flush(ctx.io)
             ok = open(log_file, "w") do log
                 std = verbose ? ctx.io : log
                 success(pipeline(gen_build_code(buildfile(source_path)),
@@ -1659,7 +1659,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
             status(sandbox_ctx.env, sandbox_ctx.registries; mode=PKGMODE_COMBINED, io=sandbox_ctx.io, ignore_indent = false)
             Pkg._auto_precompile(sandbox_ctx, warn_loaded = false)
             printpkgstyle(ctx.io, :Testing, "Running tests...")
-            flush(stdout)
+            flush(ctx.io)
             cmd = gen_test_code(testfile(source_path); coverage=coverage, julia_args=julia_args, test_args=test_args)
             p = run(pipeline(ignorestatus(cmd), stdout = sandbox_ctx.io))
             ctx.io isa IOBuffer && (ctx.io.writable = true) # the sandbox julia proccess `closewrite`'s the io
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1662,6 +1662,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
             flush(stdout)
             cmd = gen_test_code(testfile(source_path); coverage=coverage, julia_args=julia_args, test_args=test_args)
             p = run(pipeline(ignorestatus(cmd), stdout = sandbox_ctx.io))
+            ctx.io isa IOBuffer && (ctx.io.writable = true) # the sandbox julia proccess `closewrite`'s the io
             if success(p)
                 printpkgstyle(ctx.io, :Testing, pkg.name * " tests passed ")
             else
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1661,7 +1661,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
             printpkgstyle(ctx.io, :Testing, "Running tests...")
             flush(stdout)
             cmd = gen_test_code(testfile(source_path); coverage=coverage, julia_args=julia_args, test_args=test_args)
-            p = run(ignorestatus(cmd))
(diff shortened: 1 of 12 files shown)


src/Operations.jl
```diff
@@ -1661,7 +1661,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
             printpkgstyle(ctx.io, :Testing, "Running tests...")
             flush(stdout)
             cmd = gen_test_code(testfile(source_path); coverage=coverage, julia_args=julia_args, test_args=test_args)
-            p = run(ignorestatus(cmd))
+            p = run(pipeline(ignorestatus(cmd), stdout = sandbox_ctx.io))
             if success(p)
                 printpkgstyle(ctx.io, :Testing, pkg.name * " tests passed ")
```
IanButterworth, line 1666: After changing this the `printpkgstyle` can't print to `ctx.io`. Does the sub julia process close the io or something?
```
  ArgumentError: ensureroom failed, IOBuffer is not writeable
  Stacktrace:
    [1] ensureroom_slowpath(io::IOBuffer, nshort::UInt64)
      @ Base ./iobuffer.jl:303
    [2] ensureroom
      @ ./iobuffer.jl:325 [inlined]
    [3] unsafe_write(to::IOBuffer, p::Ptr{UInt8}, nb::UInt64)
      @ Base ./iobuffer.jl:424
    [4] write
      @ ./strings/io.jl:244 [inlined]
    [5] print
      @ ./strings/io.jl:246 [inlined]
    [6] with_output_color(f::Function, color::Symbol, io::IOBuffer, args::String; bold::Bool, underline::Bool, blink::Bool, reverse::Bool, hidden::Bool)
      @ Base ./util.jl:79
    [7] #printstyled#884
      @ ./util.jl:129 [inlined]
    [8] printpkgstyle(io::IOBuffer, cmd::Symbol, text::String, ignore_indent::Bool; color::Symbol)
      @ Pkg ~/Documents/GitHub/Pkg.jl/src/utils.jl:5
    [9] printpkgstyle (repeats 2 times)
      @ ~/Documents/GitHub/Pkg.jl/src/utils.jl:3 [inlined]
   [10] (::Pkg.Operations.var"#109#111"{Bool, Cmd, Cmd, Nothing, Pkg.Types.Context, Vector{Tuple{String, Base.Process}}, String, Pkg.Types.PackageSpec})()
      @ Pkg.Operations ~/Documents/GitHub/Pkg.jl/src/Operations.jl:1666
```

src/REPLMode/REPLMode.jl
```diff
@@ -453,7 +454,7 @@ struct MiniREPL <: REPL.AbstractREPL
     t::REPL.Terminals.TTYTerminal
 end
 function MiniREPL()
-    MiniREPL(TextDisplay(stdout), REPL.Terminals.TTYTerminal(get(ENV, "TERM", Sys.iswindows() ? "" : "dumb"), stdin, stdout, stderr))
+    MiniREPL(TextDisplay(stdout_f()), REPL.Terminals.TTYTerminal(get(ENV, "TERM", Sys.iswindows() ? "" : "dumb"), stdin, stdout_f(), stderr_f()))
 end
```
IanButterworth, line 458: I don't know why the `pkg""` tests are still printing after changing this ^ https://github.com/JuliaLang/Pkg.jl/blob/0dc6150bd56fa89c37a0b86da8036d68df1bdcea/test/repl.jl#L17

KristofferC: I guess one has to investigate 

https://github.com/JuliaLang/Pkg.jl/blob/0dc6150bd56fa89c37a0b86da8036d68df1bdcea/src/REPLMode/REPLMode.jl#L421

and

https://github.com/JuliaLang/Pkg.jl/blob/0dc6150bd56fa89c37a0b86da8036d68df1bdcea/src/REPLMode/REPLMode.jl#L441

and figure out where that gets printed in the end.

IanButterworth: Looks like `Pkg.REPLMode.minirepl[]` just needed resetting because the default io changed after it was intiially set in `REPLMode.__init__()`

IanButterworth:
I added an env var for controlling whether lot messages are suppressed in Pkg tests. `JULIA_PKG_TEST_QUIET` which defaults to "true". 

For `make testall` that means tests are much quieter. 
Pkg CI could be set to `JULIA_PKG_TEST_QUIET="false"` , or is it ok to rely on `@test` results alone?

Things that remain are some spurious git cli and gzip output, and some redefinition warnings

src/utils.jl
```diff
@@ -6,7 +6,7 @@ function printpkgstyle(io::IO, cmd::Symbol, text::String, ignore_indent::Bool=fa
     println(io, " ", text)
 end
 
-function linewrap(str::String; io = stdout, padding = 0, width = Base.displaysize(io)[2])
+function linewrap(str::String; io = stdout_f(), padding = 0, width = Base.displaysize(io)[2])
```
KristofferC, line 9: `io` looks unused here?

KristofferC (approved):

src/utils.jl
IanButterworth: It's used in the `width` kwarg default. I think I set it up this way so that it was easy to just specify an io, rather than have to figure out the width for yourself, but also allow overriding the width directly. There might be a better way?

KristofferC: Ah, indeed. Just leave it for now I guess.

PR #2919, opened 2022-01-04
CI: Fix git and 7zip not using default io

IanButterworth:
I think these are the last remaining noisy tests.

I also added the env var to the github action to make it easier to turn logs back on


DilumAluthge:
Just to clarify, if the environment variable is not defined, what does it default to? Noisy or quiet?

DilumAluthge:
And maybe add a comment to the CI YAML file to that effect.

DilumAluthge (approved):
The CI change looks fine. I did not look at the non-CI changes.

staticfloat (approved):

staticfloat:
It looks to me like some more jiggering around of the IOS stuff is necessary, but in principle, this looks nice.

IanButterworth:
Yeah. Is it possible to tell `pipeline` to not make IOBuffer's unwritable?

staticfloat:
I don't know, that's a @vtjnash question

DilumAluthge:
Instead of `IOBuffer`, could you use `Pipe`?

DilumAluthge:
`IOBuffer`:

```julia
julia> io = IOBuffer();

julia> run(pipeline(`pwd`; stdout = io));

julia> run(pipeline(`pwd`; stdout = io));
┌ Warning: Process I/O error
│   exception =
│    ArgumentError: ensureroom failed, IOBuffer is not writeable
│    Stacktrace:
│     [1] ensureroom_slowpath(io::IOBuffer, nshort::UInt64)
│       @ Base ./iobuffer.jl:303
│     [2] ensureroom
│       @ ./iobuffer.jl:325 [inlined]
│     [3] unsafe_write(to::IOBuffer, p::Ptr{UInt8}, nb::UInt64)
│       @ Base ./iobuffer.jl:424
│     [4] unsafe_write
│       @ ./io.jl:675 [inlined]
│     [5] write
│       @ ./io.jl:698 [inlined]
│     [6] write(to::IOBuffer, from::Base.PipeEndpoint)
│       @ Base ./io.jl:745
│     [7] (::Base.var"#709#710"{IOBuffer, Bool, Base.PipeEndpoint, IOBuffer, Base.PipeEndpoint})()
│       @ Base ./task.jl:466
└ @ Base process.jl:278
```

`Pipe`:

```julia
julia> p = Pipe();

julia> run(pipeline(`pwd`; stdout = p));

julia> run(pipeline(`pwd`; stdout = p));

julia> run(pipeline(`pwd`; stdout = p));

julia> run(pipeline(`pwd`; stdout = p));

julia> close(p.in)

julia> read(p, String)
"/Users/dilum/Downloads\n/Users/dilum/Downloads\n/Users/dilum/Downloads\n/Users/dilum/Downloads\n"
```

The downside is that you have to close the `Pipe` at the end before you can read from it.

vtjnash:
A BufferStream is probably the easiest, and will auto-convert to a Pipe when you use it. So basically anything but IOStream will work.

DilumAluthge:
This is fantastic! It looks like you got all of them now.

DilumAluthge:
```
Testing Running tests...
Test Summary:                                  | Pass  Total  Time
Test that we have imported the correct package |    1      1  0.1s
┌ Warning: Module Pkg with build ID 159589902205 is missing from the cache.
│ This may mean Pkg [54cfe95a-1eb2-52ea-b672-e2afdf69b78f] does not support precompilation but is imported by a module that does.
└ @ Base loading.jl:1160
┌ Warning: Module Pkg with build ID 159589902205 is missing from the cache.
│ This may mean Pkg [54cfe95a-1eb2-52ea-b672-e2afdf69b78f] does not support precompilation but is imported by a module that does.
└ @ Base loading.jl:1160
Test Summary: | Pass  Broken  Total     Time
Pkg           | 2411       3   2414  9m43.2s
Testing Pkg tests passed
```

Should we be worried about those two warnings?

IanButterworth:
These two are coming from `test/artifacts.jl`

```
┌ Warning: Module Pkg with build ID 54283914900 is missing from the cache.
│ This may mean Pkg [54cfe95a-1eb2-52ea-b672-e2afdf69b78f] does not support precompilation but is imported by a module that does.
└ @ Base loading.jl:1160
┌ Warning: Module Pkg with build ID 54283914900 is missing from the cache.
│ This may mean Pkg [54cfe95a-1eb2-52ea-b672-e2afdf69b78f] does not support precompilation but is imported by a module that does.
└ @ Base loading.jl:1160
```
Seems like they need fixing rather than hiding

DilumAluthge:
What's the `git blame` on those tests? @staticfloat, presumably?

DilumAluthge:
> Seems like they need fixing rather than hiding

Agreed.

DilumAluthge:
Let me try to see if I can at least narrow down which test set it is within the `test/artifacts.jl`.

IanButterworth:
It's these
https://github.com/JuliaLang/Pkg.jl/blob/fff928aadc54872237984d35fed4c79ec9531c57/test/artifacts.jl#L670

https://github.com/JuliaLang/Pkg.jl/blob/fff928aadc54872237984d35fed4c79ec9531c57/test/artifacts.jl#L711

DilumAluthge:
Anyway don't let that derail this PR; we can figure out those two warnings in a subsequent PR.

From looking at the CI logs, this PR looks good to go.

DilumAluthge:
> It's these
> 
> https://github.com/JuliaLang/Pkg.jl/blob/fff928aadc54872237984d35fed4c79ec9531c57/test/artifacts.jl#L670
> 
> https://github.com/JuliaLang/Pkg.jl/blob/fff928aadc54872237984d35fed4c79ec9531c57/test/artifacts.jl#L711

Of course it is 😂 

Those are the same tests that gave me problems when I was trying to get the Pkg tests to pass on Buildkite.

IanButterworth:
ok, let's merge, but I'm actually not 100% convinced that all the logs should be suppressed on Pkg repo CI.
For `make testall` it does make sense.

Easy to turn back on if someone makes a case for it.

DilumAluthge:
I would say:
1. User runs `make testall` locally: definitely hide logs by default 
2. Pkg.jl GitHub Actions CI on this repository: definitely show logs by default
3. Base Julia CI on Buildkite: you could make a case for either.

DilumAluthge:
I'll take a crack at those two warnings.

.github/workflows/test.yml
```diff
@@ -61,6 +61,7 @@ jobs:
       - run: julia --project --color=yes --check-bounds=yes -e 'import Pkg; Pkg.build(); Pkg.test(; coverage=true)'
         env:
           JULIA_PKG_SERVER: ${{ matrix.pkg-server }}
+          JULIA_PKG_TEST_QUIET: "true" # "true" is the default. i.e. tests are quiet when this env var isn't set
```
DilumAluthge, line 64:
```suggestion
          JULIA_PKG_TEST_QUIET: "false" # "true" is the default. i.e. tests are quiet when this env var isn't set
```

DilumAluthge: It's nice to leave the template in, so if people want to toggle it temporarily in a PR, it's easy to do.

test/runtests.jl
```diff
@@ -43,6 +46,7 @@ Logging.with_logger(quiet ? Logging.NullLogger() : Logging.current_logger()) do
             "force_latest_compatible_version.jl",
             "manifests.jl",
             ]
+            @info "=== Testing `test/$f` ==="
```
DilumAluthge, line 49: Should we only print this message if `JULIA_PKG_TEST_QUIET` is set to `false`?

IanButterworth: That's the case already. It's inside the `with_logger`

DilumAluthge:
Should `verbose` always be defined as `!quiet` (or vice versa)?

IanButterworth:
Let's give that a go

DilumAluthge:
@IanButterworth Do you want to bump Pkg on master now, or wait for #2921? It might be a while before #2921 is ready.

IanButterworth:
No rush. This is very minor from a julia repo perspective

DilumAluthge:
Sounds good'

PR #2920, opened 2022-01-05
Improve handling of preferences for sandboxes used in Pkg

### LINKED ISSUE #2500: `Pkg.test()` drops Preferences
While attempting to move FFTW over to using `Preferences` instead of environment variables/mutating its own state to record which backend is currently in use, I have discovered that `Pkg` generates a new project when running tests, and that new project drops unknown metadata.  I believe the culprit is this section:

https://github.com/JuliaLang/Pkg.jl/blob/0fe89b17771a76e42373d2ec4477b36ba37d05db/src/Operations.jl#L1516-L1534

Since users (and other packages) are storing pieces of information inside of `Project.toml` files, I think we need to preserve as much information in the `Project.toml` as possible.

For preferences in particular, we also need to copy any `LocalPreferences.toml` files to the temporary directory where the project is stored as well.  I am currently working around this by generating an explicit `test/Project.toml`, but this is obviously sub-optimal.

vchuravy:
Fixes #2500, when we set up the preferences for test/build we want
to propagate the preferences from either `test/LocalPreferences.toml` if `test/Project.toml` exists
or from `Project.toml` if the old style is being used. We gather the dependencies up and then write
them out to the temp directory. 

Similarily for the `.pkg/select_artifacts.jl` functionality we need to make the current preferences
available so that we can augment the platform we are downloading.

Co-authored-by: @staticfloat

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -25,6 +25,7 @@ p7zip_jll = "3f19e933-33d8-53b3-aaab-bd5110c3b7a0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 
 [targets]
-test = ["Test"]
+test = ["Test", "Preferences"]
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -604,8 +604,10 @@ function collect_artifacts(pkg_root::String; platform::AbstractPlatform=HostPlat
 
             # If there is a dynamic artifact selector, run that in an appropriate sandbox to select artifacts
             if isfile(selector_path)
-                select_cmd = Cmd(`$(gen_build_code(selector_path)) $(triplet(platform))`)
-                addenv(select_cmd, "JULIA_LOAD_PATH" => "@stdlib")
+                # Despite the fact that we inherit the project, since the in-memory manifest
+                # has not been updated yet, if we try to load any dependencies, it may fail.
+                # Therefore, this project inheritance is really only for Preferences, not dependencies.
+                select_cmd = Cmd(`$(gen_build_code(selector_path; inherit_project=true)) $(triplet(platform))`)
                 meta_toml = String(read(select_cmd))
                 push!(artifacts_tomls, (artifacts_toml, TOML.parse(meta_toml)))
             else
@@ -896,7 +898,7 @@ function dependency_order_uuids(env::EnvCache, uuids::Vector{UUID})::Dict{UUID,I
     return order
 end
 
-function gen_build_code(build_file::String)
+function gen_build_code(build_file::String; inherit_project::Bool = false)
     code = """
         $(Base.load_path_setup_code(false))
         cd($(repr(dirname(build_file))))
@@ -906,6 +908,7 @@ function gen_build_code(build_file::String)
         $(Base.julia_cmd()) -O0 --color=no --history-file=no
         --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
         --compiled-modules=$(Bool(Base.JLOptions().use_compiled_modules) ? "yes" : "no")
+        $(inherit_project ? `--project=$(Base.active_project())` : ``)
         --eval $code
         ```
 end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -910,6 +910,17 @@ function gen_build_code(build_file::String)
         ```
 end
 
+with_load_path(f::Function, new_load_path::String) = with_load_path(f, [new_load_path])
+function with_load_path(f::Function, new_load_path::Vector{String})
+    old_load_path = copy(Base.LOAD_PATH)
+    copy!(Base.LOAD_PATH, new_load_path)
+    try
+        f()
+    finally
+        copy!(LOAD_PATH, old_load_path)
+    end
+end
+
 const PkgUUID = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 pkg_scratchpath() = joinpath(depots1(), "scratchspaces", PkgUUID)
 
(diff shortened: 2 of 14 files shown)


vchuravy:
https://github.com/JuliaLang/Pkg.jl/pull/2920/commits/52a2b4c5b3ea0f6e276eaceb6353cd3fe9a93ede depends on  https://github.com/JuliaLang/julia/commit/0b7827abe08ec8c2054912d1c5623542179a5567 

We might be able to backport https://github.com/JuliaLang/Pkg.jl/pull/2920/commits/78963efc165d393c7089193b0714fcde05dacaee

vchuravy:
Depends on https://github.com/JuliaLang/julia/pull/43684

src/Operations.jl
```diff
@@ -941,9 +944,18 @@ function build_versions(ctx::Context, uuids::Set{UUID}; verbose=false)
...
         # compatibility shim
-        build_project_override = isfile(projectfile_path(builddir(source_path))) ?
-            nothing :
-            gen_target_project(ctx.env, ctx.registries, pkg, source_path, "build")
+        local build_project_override, build_project_preferences
+        if isfile(projectfile_path(builddir(source_path)))
+            build_project_override = nothing
+            Pkg.activate(projectfile_path(builddir(source_path))) do
+                build_project_preferences = Base.get_preferences()
+            end
```
staticfloat, line 952: @KristofferC is it okay for us to use `Pkg.activate()` here, or should we manually just set/reset `Base.LOAD_PATH` like I do in [these tests here](https://github.com/JuliaLang/julia/blob/7a8613f1c1aac6eda0e6b7e93c24d67bdb187a5a/test/loading.jl#L878-L880)?  I'm not sure if there's extra work that `Pkg.activate()` is doing beyond just manipulating `LOAD_PATH`.

src/Operations.jl
```diff
@@ -1648,12 +1669,21 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     pkgs_errored = Tuple{String, Base.Process}[]
     for (pkg, source_path) in zip(pkgs, source_paths)
         # compatibility shim between "targets" and "test/Project.toml"
-        test_project_override = isfile(projectfile_path(testdir(source_path))) ?
-            nothing :
-            gen_target_project(ctx.env, ctx.registries, pkg, source_path, "test")
+        local test_project_preferences, test_project_override
+        if isfile(projectfile_path(testdir(source_path)))
+            test_project_override = nothing
+            Pkg.activate(projectfile_path(testdir(source_path))) do
```
staticfloat, line 1675: Whatever Kristoffer says above, we'll copy here as well.

staticfloat (approved):

src/Operations.jl
fredrikekre: It is invalid for non-default load paths.

Project.toml
```diff
@@ -25,6 +25,7 @@ p7zip_jll = "3f19e933-33d8-53b3-aaab-bd5110c3b7a0"
 
 [extras]
 Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+Preferences = "21216c6a-2e73-6563-6e65-726566657250"
 
 [targets]
-test = ["Test"]
+test = ["Test", "Preferences"]
```
KristofferC, line 31: I'm not sure how the Julia CI tests will work with an stdlib depending on non-stdlib packages.

src/Operations.jl
```diff
@@ -596,8 +596,10 @@ function collect_artifacts(pkg_root::String; platform::AbstractPlatform=HostPlat
 
             # If there is a dynamic artifact selector, run that in an appropriate sandbox to select artifacts
             if isfile(selector_path)
-                select_cmd = Cmd(`$(gen_build_code(selector_path)) $(triplet(platform))`)
-                addenv(select_cmd, "JULIA_LOAD_PATH" => "@stdlib")
+                # Despite the fact that we inherit the project, since the in-memory manifest
+                # has not been updated yet, if we try to load any dependencies, it may fail.
+                # Therefore, this project inheritance is really only for Preferences, not dependencies.
```
KristofferC, line 601: I must say I don't understand what this comments is meant to convey.

KristofferC:
I don't really get why this is not just as simple as just not removing existing preferences from the Project file when creating the temp project file and copy an existing `LocalPreferences.toml` to the temp environment.

src/Operations.jl
staticfloat: https://github.com/JuliaLang/Pkg.jl/pull/2024 added the ability for packages to give a little snippet of code that should be run in order to help `Pkg` determine which artifact to install at `Pkg.add()` time.  The canonical example is CUDA version (e.g. CUDA 9 vs CUDA 10); that's not something Julia natively knows about (as opposed to processor architecture, e.g. `x86_64` vs. `aarch64`).  The eventual design was to have a script that prints out TOML chunks that represent the artifacts that need to be downloaded.

Valentin and Simon have been working on a way to improve the MPI situation, and their current thinking is to use `Preferences` to allow setting a particular MPI backend, which then is used to coordinate which artifacts get downloaded by downstream MPI-aware packages.  The mechanism by which this happens is that the `select_artifacts.jl` hook inspects the preferences of the current environment, and uses that to inform the artifact selection machinery accordingly.

We run the `select_artifacts.jl` file in a new Julia process and without any particular project, using the same machinery as `Pkg.build()`.  Because Preferences reading relies on the `Base.LOAD_PATH` stack, it's important to pass `--project=$(Base.active_project())` to the Julia subprocess so that it sees the same preferences as when you are actually running the package.  However, because this is happening in the middle of `Pkg.add()`, this comment is meant to educate nosy users that despite the fact that we're running with a project that claims to have a bunch of dependencies, those dependencies may not be loadable via `using Foo`, because we haven't finished the `Pkg.add()` operation yet.

staticfloat:
> I don't really get why this is not just as simple as just not removing existing preferences from the Project file when creating the temp project file and copy an existing LocalPreferences.toml to the temp environment.

It's because Preferences merge across stacked environments.  E.g. if you have `MPIBackend = "foo"` set in your `~/.julia/environments/v1.X/LocalPreferences.toml`, that needs to be reflected in the `LocalPreferences.toml` you set up in the temp environment as well.  The easiest way to do that is to use `Base.get_preferences()` to flatten everything down into a single Dict, then just write that out as a TOML.

src/Operations.jl
staticfloat: What is invalid?

fredrikekre: `Pkg.activate`

staticfloat: Alright, we'll do the `Base.LOAD_PATH` mcjigggery then

Project.toml
fredrikekre: Could `Pkg.add` it from within `runtests.jl` instead

vchuravy: Indeed a problem:

```
  LoadError: ArgumentError: Package Preferences not found in current path.
  - Run `import Pkg; Pkg.add("Preferences")` to install the Preferences package.
  Stacktrace:
```

with:
```
JULIA_DEPOT_PATH=(mktemp -d) make -C test Pkg
```

vchuravy:
Ok this now passes tests :)

staticfloat (approved):

staticfloat:
IMO this is ready to merge.  Any objections?

staticfloat (approved):

staticfloat:
It's not me, it was @vchuravy!  He's the one with the patience and panache to push this through, and he deserves all the credit.

timholy:
Ahh, fooled by the "merged by" again! (Though I see you did help out.) Thanks @vchuravy!!

PR #2924, opened 2022-01-07
CI: don't skip intermediate builds on the `master` or `release-*` branches

DilumAluthge:


.github/workflows/test.yml
```diff
@@ -13,9 +13,9 @@ defaults:
   run:
     shell: bash
 concurrency:
-  # Skip intermediate builds: always.
-  # Cancel intermediate builds: only if it is a pull request build.
-  group: ${{ github.workflow }}-${{ github.ref }}
+  # Skip intermediate builds: all builds except for builds on the `master` or `release-*` branches
+  # Cancel intermediate builds: only pull request builds
+  group: ${{ github.workflow }}-${{ github.ref }}-${{ github.ref != 'refs/heads/master' || startsWith(github.ref, 'refs/heads/release-') || github.run_number }}
   cancel-in-progress: ${{ startsWith(github.ref, 'refs/pull/') }}
```
staticfloat, line 19: Wouldn't it be better to put this into `cancel-in-progress` instead of `group`?

DilumAluthge: I'm not sure I follow.

If two builds have the same group, then the older build will get skipped in favor of the new build.

So if the group consists only of `group: ${{ github.workflow }}-${{ github.ref }}`, then if there are multiple builds on master, some of those builds will get skipped. (Because all those builds will have the same `group`, because `${{ github.ref }}` will be `refs/heads/master` for all of those builds.)

The goal here is that we never want to skip a build on master.

DilumAluthge: I might be misunderstanding what you are asking.

DilumAluthge:
h/t @fingolfin for showing me how to do this, based on an example from the https://github.com/oscar-system/Oscar.jl repository.

.github/workflows/test.yml
staticfloat: I was reading docs for buildkite instead of GHA..... 😅 

Carry on.

staticfloat (approved):

.github/workflows/test.yml
DilumAluthge: I will say, Buildkite's implementation of this feature is much easier to use.

Issue #2922, opened 2022-01-05
Interrupting Pkg.test sometimes orphans the test sandbox process


IanButterworth:
`^C`-ing a `Pkg.test` run doesn't seem to reliably interrupt the child session, leaving it orphaned and running even when you close julia

Observed on ubuntu

IanButterworth:
This just happened to me interrupting Pkg tests on master, though it's reliable interrupting other test suites like `Dates`

```
Version 1.9.0-DEV.706 (2022-06-02)
Commit bd8dbc388c (2 days old master)
```

PR #2933, opened 2022-01-09
Make interrupting `Pkg.test` more graceful

IanButterworth:
Currently `^C`-ing a `Pkg.test` can result in a big mess of mixed-up InterruptException stack traces from the child and parent process, and it's common for the child process to not actually get interrupted given there's only one shot for the `^C` to get through to the child before the parent is interrupted and closes the signal forwarding, making it orphaned by the parent and continue to run tests while the parent process tries to go back to the repl, no matter how many times the user `^C`'s, requiring the user to find the orphaned process and kill it.

This change tries to make the process more graceful:
- Catches the interrupt in the parent and prints a nicer Pkg message immediately
- Doesn't show the stacktrace of the interrupted parent process
- Waits a short while for the child process to exit gracefully, throwing a few more interrupts at it
- If not exited it then kills the child process cleanly (no signal termination message)
- Only releases back to the `pkg>` prompt once the child process is stopped, to avoid the prompt being printed-over

This PR
Screenshot from 2022-01-08 23-58-30

Closes https://github.com/JuliaLang/Pkg.jl/issues/2922


IanButterworth:
I think this is a definite UX improvement, but would appreciate a second opinion

ericphanson:
+1, the current behavior is annoying and feels buggy as a user

src/Operations.jl
```diff
@@ -1661,10 +1661,27 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
...
+            try
+                wait(p)
+            catch e
+                if e isa InterruptException
+                    interrupted = true
+                    print("\n")
+                    printpkgstyle(ctx.io, :Testing, "Tests interrupted. Exiting the test process\n", color = Base.error_color())
+                    # Give some time for the child interrupt handler to print a stacktrace and exit,
+                    # then kill the process if still running
+                    if timedwait(() -> !process_running(p), 4) == :timed_out
```
vtjnash, line 1675: Using `timedwait` is often a general design problem, where you can usually use a `Timer(4)` and `wait(p)` instead though

vtjnash, line 1675: Using `timedwait` is often a general design problem, where you can usually use a `Timer(4)` and `wait(p)` instead though

PR #2947, opened 2022-01-17
Use precompile context to find package location

simonbyrne:
Precompile currently calls `Base.locate_package`, which which looks in the current environment, rather than the environment specified via the context. This changes the logic to look up the path via the context environment, and fall back on `Base.locate_package` if this was not successful.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1241,8 +1241,13 @@ function precompile(ctx::Context, pkgs::Vector{String}=String[]; internal_call::
     Base.LOADING_CACHE[] = Base.LoadingCache()
     for (pkg, deps) in depsmap # precompilation loop
         paths = Base.find_all_in_cache_path(pkg)
-        sourcepath = Base.locate_package(pkg)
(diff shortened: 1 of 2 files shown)


KristofferC:
Doesn't code loading (in the precompile process) look at the active project at all?

simonbyrne:
Perhaps? In that case, I guess this effort is doomed.

simonbyrne:
This seems to work.

simonbyrne:
I think you're right: `Base.compilecache` is hardcoded to use the current `load_path`:
https://github.com/JuliaLang/julia/blob/e2a8a4e6b3bb333fdab5a5c9a023fe96e2f39c92/base/loading.jl#L1631

quinnj:
This can be closed?

simonbyrne:
Can we fix Base.compilecache?

KristofferC:
I guess the load path could be made into an argument and Pkg passes something in there?

PR #2946, opened 2022-01-17
Make tests robust against init.defaultBranch being set

vchuravy:

PR #2948, opened 2022-01-18
Support default branch on Git prior to 2.28

vchuravy:
Fixes https://github.com/JuliaLang/Pkg.jl/pull/2946#issuecomment-1015104029

### DIFF
--- a/test/subdir.jl
+++ b/test/subdir.jl
@@ -14,6 +14,20 @@ function gitcmd(path::AbstractString)
          "-c", "user.email=ci@juliacomputing.com"])
 end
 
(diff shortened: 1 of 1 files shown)


vchuravy:
As @fredrikekre just pointed out this will not work if the user has manually set that config option.

https://github.com/JuliaLang/Pkg.jl/pull/2946#issuecomment-1015466684

fredrikekre:
It doesn't seem so difficult to parse the version number though?

DilumAluthge:
I'm not sure this is the best approach. Let me try to make a different PR.

PR #2949, opened 2022-01-19
Make tests robust against Git's `init.defaultBranch` option being set, while still supporting Git 1.2.0+ and Git 2.0.0+

DilumAluthge:
Follow-up to #2946
Replaces and closes #2948

## Description

After we make the first commit to a new repository, we check and see what the branch name is. If the branch name is `master`, we do nothing. If the branch name is not `master`, we run the following commands:

1. `git branch -f master`
2. `git checkout master`
3. `git branch -D $(old_branch_name)`

After running this command, the branch name is now `master`.

This supports Git 1.2.0+ and Git 2.0.0+. If someone really needs to support a version of Git older than 1.2.0, then we'll need to rewrite these tests to use the LibGit2 stdlib instead.

#### Git 1.2.0: `branch -f`

https://github.com/git/git/blob/v1.2.0/Documentation/git-branch.txt#L28-L29

```
-f::
	Force a reset of <branchname> to <start-point> (or current head).
```

#### Git 1.2.0: `branch -D`

https://github.com/git/git/blob/v1.2.0/Documentation/git-branch.txt#L22-L26

```
-d::
	Delete a branch. The branch must be fully merged.

-D::
	Delete a branch irrespective of its index status.
```

#### Git 2.0.0: `branch -f`

https://github.com/git/git/blob/v2.0.0/Documentation/git-branch.txt#L94-L97

```
-f::
--force::
	Reset <branchname> to <startpoint> if <branchname> exists
	already. Without `-f` 'git branch' refuses to change an existing branch.
```

#### Git 2.0.0: `branch -D`

https://github.com/git/git/blob/v2.0.0/Documentation/git-branch.txt#L77-L84

```
-d::
--delete::
	Delete a branch. The branch must be fully merged in its
	upstream branch, or in `HEAD` if no upstream was set with
	`--track` or `--set-upstream`.

-D::
	Delete a branch irrespective of its merged status.
```

https://github.com/git/git/blob/v2.0.0/Documentation/git-branch.txt#L64-L66

```
With a `-d` or `-D` option, `<branchname>` will be deleted.  You may
specify more than one branch for deletion.  If the branch currently
has a reflog then the reflog will also be deleted.
```

### DIFF
--- a/test/subdir.jl
+++ b/test/subdir.jl
@@ -127,8 +127,7 @@ end
 # Note: we cannot use `git init -b`, because that requires Git 2.28.0+, and we want to
 # support older versions of Git. Therefore, we instead use `git branch -f` and `git branch -D`,
 # which only require Git 1.2.0+ or Git 2.0.0+.
-function fix_default_branch(; dir::String)
-    new_branch_name = "master"
+function fix_default_branch(; dir::String, new_branch_name::String = "master")
     old_branch_name = _current_branch_name(; dir)
     git = gitcmd(dir)
     if old_branch_name != new_branch_name
--- a/test/subdir.jl
+++ b/test/subdir.jl
@@ -44,9 +44,10 @@ function setup_packages_repository(dir)
         """)
 
     git = gitcmd(dir)
-    run(pipeline(`$git init -b master -q`, stdout = stdout_f(), stderr = stderr_f()))
+    run(pipeline(`$git init -q`, stdout = stdout_f(), stderr = stderr_f()))
     run(pipeline(`$git add .`, stdout = stdout_f(), stderr = stderr_f()))
     run(pipeline(`$git commit -qm 'Create repository.'`, stdout = stdout_f(), stderr = stderr_f()))
+    fix_default_branch(; dir)
     package_tree_hash = readchomp(`$git rev-parse HEAD:julia`)
     dep_tree_hash = readchomp(`$git rev-parse HEAD:dependencies/Dep`)
     return package_tree_hash, dep_tree_hash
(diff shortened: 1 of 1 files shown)


fredrikekre:
Whats wrong with #2948? That should work on any git version. This PR won't work on e.g. Ubuntu < 20.

DilumAluthge:
What's the oldest version of Git that you want to support?

The Git docs (https://git-scm.com/docs/git-branch/2.4.12) say that `git branch --move` is available in Git 2.4.12.

It may be available even earlier; I could only find the docs going back to 2.4.12.

I just put "Git 2.20" in the PR title because that is what we run on Buildkite. But this PR works on any Git version that supports `git branch --move`.

DilumAluthge:
> Whats wrong with #2948?

Personally, I think that trying to parse the version number of Git is going to be unreliable and subject to error.

DilumAluthge:
https://github.com/git/git/blob/v2.0.0/Documentation/git-branch.txt#L99-L101:

```
-m::
--move::
	Move/rename a branch and the corresponding reflog.
```

So it looks like Git 2.0.0 supports `git branch --move`.

Does the Pkg.jl test suite need to support versions of Git older than Git 2.0.0?

DilumAluthge:
https://github.com/git/git/blob/v1.5.0/Documentation/git-branch.txt#L57-L58:

```
-m::
	Move/rename a branch and the corresponding reflog.
```

staticfloat:
Yeah, I think I prefer this as well (could you add a note saying _why_ we're using this instead of `-b`?)

Git v1.5 is plenty old for us; the oldest thing we would ever need to support is v1.8, which is CentOS 6's default `git` version.

DilumAluthge:
Yeah I'll add a comment explaining why we are doing this, and another comment listing the minimum required Git version.

Also, we can go older than Git 1.5 if I manually create the new branch, check it out, and delete the old branch.

DilumAluthge:
I've added some comments explaining the motivation for the `fix_default_branch()` function and documenting the minimum required Git version.

I switched from using `branch -m` to using `branch -f` and `branch -D`, so we now support Git 1.2.0+ and Git 2.0.0+.

test/subdir.jl
```diff
@@ -104,9 +105,46 @@ function setup_registry(dir, packages_dir_url, package_tree_hash, dep_tree_hash)
...
+# 2. `git checkout master`
+# 3. `git branch -D $(old_branch_name)`
+#
+# Note: this requires Git 1.2.0+ or Git 2.0.0+
+#
+# Note: we cannot use `git init -b`, because that requires Git 2.28.0+, and we want to
+# support older versions of Git. Therefore, we instead use `git branch -f` and `git branch -D`,
+# which only require Git 1.2.0+ or Git 2.0.0+.
+function fix_default_branch(; dir::String)
+    new_branch_name = "master"
```
staticfloat, line 131:
```suggestion
function fix_default_branch(; dir::String, new_branch_name::String = "master")
```

staticfloat (approved):

Issue #2938, opened 2022-01-13
Error `no method matching length(::Pkg.Versions.VersionSpec)` when passing context to `Pkg.add` if the version is a string


giordano:
```julia
julia> using Pkg

julia> Pkg.add(Pkg.Types.Context(), [PackageSpec(;name="Example", version="0.5.3")])
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
   Resolving package versions...
ERROR: MethodError: no method matching length(::Pkg.Versions.VersionSpec)
Closest candidates are:
  length(::Union{Base.KeySet, Base.ValueIterator}) at ~/repo/julia/usr/share/julia/base/abstractdict.jl:58
  length(::Union{LinearAlgebra.Adjoint{T, S}, LinearAlgebra.Transpose{T, S}} where {T, S}) at ~/repo/julia/usr/share/julia/stdlib/v1.8/LinearAlgebra/src/adjtrans.jl:171
  length(::LinRange) at ~/repo/julia/usr/share/julia/base/range.jl:755
  ...
Stacktrace:
  [1] union!(s::Set{Any}, itr::Pkg.Versions.VersionSpec)
    @ Base ./abstractset.jl:99
  [2] intersect!(s::Set{Any}, itr::Pkg.Versions.VersionSpec)
    @ Base ./abstractset.jl:190
  [3] _shrink(shrinker!::Function, itr::String, itrs::Tuple{Pkg.Versions.VersionSpec})
    @ Base ./array.jl:2661
  [4] intersect(itr::String, itrs::Pkg.Versions.VersionSpec)
    @ Base ./array.jl:2665
  [5] resolve_versions!(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, julia_version::VersionNumber)
    @ Pkg.Operations ~/repo/julia/usr/share/julia/stdlib/v1.8/Pkg/src/Operations.jl:327
  [6] targeted_resolve(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, preserve::Pkg.Types.PreserveLevel, julia_version::VersionNumber)
    @ Pkg.Operations ~/repo/julia/usr/share/julia/stdlib/v1.8/Pkg/src/Operations.jl:1187
[...]

Issue #2942, opened 2022-01-16
`Pkg.add(Pkg.Types.Context(;julia_version), [PackageSpec(...)])` failing since Julia v1.7


giordano:
Julia v1.6:
```
julia> using Pkg, UUIDs

julia> Pkg.add(Pkg.Types.Context(;julia_version=v"1.6.3"), [PackageSpec(;name="libjulia_jll", uuid=UUID("5ad3ddd2-0711-543a-b040-befd59781bbf"), version=Pkg.Types.VersionSpec("*"))])
   Resolving package versions...
    Updating `/tmp/jl_aPeyH5/Project.toml`
  [5ad3ddd2] + libjulia_jll v1.7.0+8
    Updating `/tmp/jl_aPeyH5/Manifest.toml`
  [692b3bcd] + JLLWrappers v1.4.0
  [21216c6a] + Preferences v1.2.3
  [5ad3ddd2] + libjulia_jll v1.7.0+8
  [0dad84c5] + ArgTools
  [56f22d72] + Artifacts
  [2a0f44e3] + Base64
  [ade2ca70] + Dates
  [f43a241f] + Downloads
  [b77e0a4c] + InteractiveUtils
  [b27032c2] + LibCURL
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [ca575930] + NetworkOptions
  [44cfe95a] + Pkg
  [de0858da] + Printf
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA
  [9e88b42a] + Serialization
  [6462fe0b] + Sockets
  [fa267f1f] + TOML
  [a4e569a6] + Tar
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
  [deac9b47] + LibCURL_jll
  [a83860b7] + LibOSXUnwind_jll
  [29816b5a] + LibSSH2_jll
  [183b4373] + LibUV_jll
  [745a5e78] + LibUnwind_jll
  [c8ffd9c3] + MbedTLS_jll
  [14a3606d] + MozillaCACerts_jll
  [83775a58] + Zlib_jll
  [8f36deef] + libLLVM_jll
  [8e850ede] + nghttp2_jll
  [3f19e933] + p7zip_jll
```

Julia v1.7:
```
julia> using Pkg, UUIDs

[...]

Issue #2930, opened 2022-01-07
`Pkg.add([PackageSpec("GMP_jll")]; julia_version=v"1.6")` doesn't install anything


giordano:
Trying to [make `BinaryBuilderBase` work with Julia nightly](https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/129) I found that `Pkg.add([PackageSpec("GMP_jll")]; julia_version=v"1.6")` does _nothing_:
```
julia> using Pkg, UUIDs

(@v1.8) pkg> activate --temp
  Activating new project at `/tmp/jl_KrcKAp`

julia> Pkg.add([PackageSpec("GMP_jll")]; julia_version=v"1.6")
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
   Resolving package versions...
    Updating `/tmp/jl_KrcKAp/Project.toml`
  [781609d7] + GMP_jll v6.2.0+5
    Updating `/tmp/jl_KrcKAp/Manifest.toml`
  [56f22d72] + Artifacts
  [8f399da3] + Libdl
  [781609d7] + GMP_jll v6.2.0+5
```
If I use `Pkg.Operations.find_installed` to get the path where `v6.2.0+5` is installed I get
```
julia> Pkg.Operations.find_installed("GMP_jll", UUID("781609d7-10c4-51f6-84f2-b8444358ff6d"), Base.SHA1("40388878122d491a2e55b0e730196098595d8a90"))
"/home/mose/.julia/packages/GMP_jll/IA6Pi"
```
but this directory doesn't exist:
```
shell> ls $(Pkg.Operations.find_installed("GMP_jll", UUID("781609d7-10c4-51f6-84f2-b8444358ff6d"), Base.SHA1("40388878122d491a2e55b0e730196098595d8a90")))
ls: cannot access '/home/mose/.julia/packages/GMP_jll/IA6Pi': No such file or directory

shell> ls -l ~/.julia/packages/GMP_jll/
total 4
drwxr-xr-x 3 mose mose 4096 Jan  7 19:32 i3hkK
```
The only directory in `packages/GMP_jll` is `6.1.2+6`:
```
julia> Pkg.Operations.find_installed("GMP_jll", UUID("781609d7-10c4-51f6-84f2-b8444358ff6d"), Base.SHA1("15abc5f976569a1c9d651aff02f7222ef305eb2a"))
[...]

PR #2952, opened 2022-01-23
Make `f(ctx` methods part of API

IanButterworth:
As @KristofferC pointed out in https://github.com/JuliaLang/Pkg.jl/issues/2938#issuecomment-1012105819 the `add(ctx..` etc entry points are considered internals.

BinaryBuilder needs to set `add(Context(; julia_version), ...`) for the julia version resolver stuff.

So it seems best to make that format formally part of the API, and pass them through the necessary processing steps.

This also distinguishes the API entry points from internals.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/2938

Todo:
- [ ] Add tests for all the added API entrypoints

cc. @giordano

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -139,12 +139,15 @@ end
 
 # Provide some convenience calls
 for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status)
+    f_internal = Symbol(string("_",f))
     @eval begin
         $f(pkg::Union{AbstractString, PackageSpec}; kwargs...) = $f([pkg]; kwargs...)
         $f(pkgs::Vector{<:AbstractString}; kwargs...)          = $f([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
         function $f(pkgs::Vector{PackageSpec}; io::IO=$(f === :status ? :stdout_f : :stderr_f)(), kwargs...)
             Registry.download_default_registries(io)
-            ctx = Context()
+            return $f(Context(), pkgs; kwargs...)
+        end
+        function $f(ctx::Context, pkgs::Vector{PackageSpec}; io::IO=$(f === :status ? :stdout_f : :stderr_f)(), kwargs...)
             # Save initial environment for undo/redo functionality
             if !saved_initial_snapshot[]
                 add_snapshot_to_undo(ctx.env)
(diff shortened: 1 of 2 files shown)


KristofferC:
> BinaryBuilder needs to set add(Context(; julia_version), ...) for the julia version resolver stuff.

Why? They can just use `Pkg.add("Package"; julia_version=v"1.6")` for example.

DilumAluthge:
Making `Context` (and methods that accept it) part of the public API removes our ability to make breaking changes to it in the future.

If BinaryBuilder needs to use Pkg internals, that's totally fine. We should just make sure that BinaryBuilder CI includes a job that runs on Julia nightly, so that when internals change, we catch it sooner, and fix it.

But that doesn't mean we should make those internals part of the public API.

IanButterworth:
> Why? They can just use Pkg.add("Package"; julia_version=v"1.6") for example.

I was under the impression specifying the context was necessary. Perhaps @giordano could expand/correct me.

giordano:
This is the code: https://github.com/JuliaPackaging/BinaryBuilderBase.jl/blob/74fd093d3ab33f9410a64cd42b28662654a014fe/src/Prefix.jl#L535-L665.  We use `ctx` for other things in the function, not just `Pkg.add`, if we pass `julia_version` only to `Pkg.add` then `julia_version` isn't retained in `ctx`.  If the function can be improved, please do.

I tend to agree with the sentiment that freezing the internal functions just for one package that happens to hook into them may not be a great idea, but, well, we do need to do weird things with package resolution.  I opened https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/129 almost one year ago exactly to try and catch failures in nightly early, but there were already failures and took forever to work around them (in the meantime _other_ stuff broke), asked for help in `#pkg-dev` on Slack several times, didn't get much.  @staticfloat recently managed to reworked `BinaryBuilderBase.setup_dependencies` to be compatible with v1.7+ (but now it doesn't work anymore in v1.6, fine, I don't care much about supporting older versions).  However now there is the issue with `LibOSXUnwind_jll` which isn't a stdlib any longer and apparently this throws package installation out of the window (#2942 and #2930).  It's unfortunate that that happened, but, well, it did happen and I guess there is no way back, but it doesn't change the fact we do need to be able to install `libjulia_jll` for v1.6, that's needed for [all packages using CxxWrap](https://juliahub.com/ui/Packages/libcxxwrap_julia_jll/9VnAb/0.9.0+0?page=2), which at the moment are stuck.  So I'd appreciate very much some help with fixing either #2942 or #2930 when installing libjulia_jll in the `julia_version=v"1.6.0"` context.  Also yet another workaround in BB would be ok, but at the moment I have zero clue of what to do.

giordano:
> They can just use Pkg.add("Package"; julia_version=v"1.6") for example.

BTW, no, we can't: #2930

KristofferC:
> if we pass julia_version only to Pkg.add then julia_version isn't retained in ctx.

That sounds like a bug.

I still don't know personally what exactly the whole `julia_version` business is about and what it is supposed to do when it is passed. If it is to simulate a resolve on a different julia version you need to bundle the project file for every stdlib for every julia version (alt have some sort of mini stdlib registry that encodes the same information).

IanButterworth:
Closing as unpopular.

Although, the fact tests failed here may point to an issue in existing tests, as AFAICT this shouldn't be breaking

PR #2966, opened 2022-02-03
Inferrability: eliminate more Core.Box

timholy:
These stem from julia#15276, i.e.,
https://docs.julialang.org/en/v1/manual/performance-tips/#man-performance-captured
They were identified by scanning all compiled MethodInstances with
`hasbox` in the newly-released MethodAnalysis 0.4.5.

`Core.Box` often causes "follow-on" inference problems,
but for these cases there were relatively few, which may be why
these didn't show up earlier during the Great Invalidation Hunt.
Still, there doesn't seem to be any particular reason not to fix them.


KristofferC:
`do` syntax strikes again.

timholy:
They're close to gone now, though. Most of what's left is https://github.com/JuliaLang/julia/issues/40990, which doesn't seem to be easy to work around but which might be specifically fixable.

PR #2968, opened 2022-02-04
restore try catch for unpacking archive

KristofferC:
Allows us to continue after a failed download from the PkgServer (e.g. from https://github.com/JuliaPackaging/PkgServer.jl/issues/126). Was removed in https://github.com/JuliaLang/Pkg.jl/commit/67ed9cb3541368e3f98867ed809de7c6db32f4cc#diff-6dfb4e40705b22a427451362ea83b4e3a763028e3dcaa3b9b1846aa1ec84370fL501

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -511,7 +511,15 @@ function install_archive(
         url_success || continue
         dir = joinpath(tempdir(), randstring(12))
         push!(tmp_objects, dir) # for cleanup
-        unpack(path, dir; verbose=false)
+        # Might fail to extract an archive (https://github.com/JuliaPackaging/PkgServer.jl/issues/126)
+        try
+            unpack(path, dir; verbose=false)
+        catch e
+            e isa InterruptException && rethrow()
+            @warn "failed to extract archive downloaded from $(url)"
+            url_success = false
+        end
+        url_success || continue
         if top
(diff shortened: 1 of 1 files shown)


StefanKarpinski:
Ugh, we need to get rid of all these try-catch blocks all over the place...

KristofferC:
Well, this seems like it is in a reasonable place and it is also very local. We cannot ensure that we get correct data from the internet and the only way to figure out is to try to extract it which throws. Of course we could use something like `ignorestatus` and then look at the status but I'm not sure that is any clearer.

giordano:
For context, as it is now, because of the aforementioned issue, sometimes PkgServer servers corrupted (empty) files, `Tar.jl` would fail to extract them because they aren't valid tarballs and `Pkg.jl` immediately errors out, preventing us from going through the available fallback option of fetching a package from github.

Of course the long-term solution would be to fix the issue in PkgServer, but as Kristoffer said we're downloading files from the Internet and can't be 100% sure they'll be valid.  At the moment this is causing to users [confusing errors when installing packages](https://github.com/JuliaPackaging/Yggdrasil/issues/4358), pushing them to [open pull desperate requests](https://github.com/JuliaGraphics/Gtk.jl/pull/621#issuecomment-1028462541).

PR #2972, opened 2022-02-06
CI: make the CI output quiet

DilumAluthge:
This PR makes the CI output quiet.

If a user is trying to debug a PR, they can always change this value back to `"false"` on their PR branch, and then the PR CI will be verbose.


KristofferC:
:+1:

PR #2976, opened 2022-02-08
`Pkg.Operations.install_archive`: when unpacking the archive, always rethrow the caught exception unless it is a `ProcessFailedException`

DilumAluthge:
This is a follow-up to #2968.

The purpose of #2968 is only to catch the error that has the error message `Can not open the file as archive`. This error message comes from 7zip when the following 7zip command fails:[^1] https://github.com/JuliaLang/Pkg.jl/blob/c991ce0a31b784133e061816d21f7d2556a182e7/src/PlatformEngines.jl#L387-L387

This is illustrated in the stacktrace posted in https://github.com/JuliaPackaging/Yggdrasil/issues/4358.

So, we don't actually need to catch all exceptions; instead, we only need to catch the exception that is thrown when the above 7zip command fails. That exception is of the `ProcessFailedException` type. Therefore, we should always rethrow the caught exception unless it is a `ProcessFailedException`.

[^1]: While the command itself is defined in `Pkg.PlatformEngines` in the code snippet shown abve, the line on which that command is actually run is this line: https://github.com/JuliaIO/ArgTools.jl/blob/08b11b2707593d4d7f92e5f1b9dba7668285ff82/src/ArgTools.jl#L75. That is, the method that is used to run the command is `Base.open(f::Function, arg::Cmd)`.


DilumAluthge:
@StefanKarpinski @KristofferC Does this look correct? I figure it would be best if we only catch the narrowest necessary set of exceptions.

IanButterworth (approved):
Seems correct from the descriptions.

PR #2980, opened 2022-02-10
Inform user about Julia versions that support new manifest in warning msg.

tpapp:
(trivial change, see [discussion](https://discourse.julialang.org/t/julia-version-lower-bound-for-new-manifest-format/76152/))


fredrikekre:
Thanks. I am not sure if this message is in 1.6, but should be backportable to 1.7 at least.

Issue #2911, opened 2022-01-02
Suggest nearby (& popular?) package when name not found


IanButterworth:
Turning https://github.com/JuliaLang/Pkg.jl/pull/2493 into an issue given it was decided to be the wrong approach

> Suggesting similar package names based on what is available in the registry would probably work better.

> Edit distance with a low cost for case changes would be a good way to do this. Now that we have usage stats we can combine with a "language model" that uses package popularity to decide what's most likely.

PR #2985, opened 2022-02-13
add ranked suggestions to `add` error

IanButterworth:
When a package isn't found by name this adds a number of ranked matches (depending on the result scores and terminal width), via the same fuzzyscore search that `REPL` does for help mode.

Screen Shot 2022-02-13 at 3 51 00 PM

If there are no match scores above zero, no suggestions are shown.
It seems like the search could be improved in REPL because it's unfortunate that the diffeq spelling mistake isn't found.
Also the `fuzzysort` method tweak could be upstreamed.

It adds a bit of slowness, but I think that's ok for this kind of user feedback (this is with 3 registries including General)

master
```
julia> @time try Pkg.add("Image"); catch  end
  0.009007 seconds (23.61 k allocations: 2.039 MiB)

julia> @time try Pkg.add("Image"); catch  end
  0.008462 seconds (23.62 k allocations: 2.040 MiB)
```

this PR
```
julia> @time try Pkg.add("Image"); catch end
  0.027184 seconds (101.59 k allocations: 13.357 MiB)

julia> @time try Pkg.add("Image"); catch end
  0.044085 seconds (101.59 k allocations: 13.357 MiB, 22.40% gc time)

julia> @time try Pkg.add("Image"); catch end
  0.026322 seconds (101.59 k allocations: 13.356 MiB)
```

Closes https://github.com/JuliaLang/Pkg.jl/issues/2911

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -627,7 +627,7 @@ function set_repo_source_from_registry!(ctx, pkg)
         Pkg.Operations.update_registries(ctx; force=false)
         registry_resolve!(ctx.registries, pkg)
     end
-    ensure_resolved(ctx.env.manifest, [pkg]; registry=true)
+    ensure_resolved(ctx, ctx.env.manifest, [pkg]; registry=true)
     # We might have been given a name / uuid combo that does not have an entry in the registry
     for reg in ctx.registries
         regpkg = get(reg, pkg.uuid, nothing)
@@ -885,7 +885,7 @@ function stdlib_resolve!(pkgs::AbstractVector{PackageSpec})
 end
 
 # Ensure that all packages are fully resolved
-function ensure_resolved(manifest::Manifest,
+function ensure_resolved(ctx::Context, manifest::Manifest,
         pkgs::AbstractVector{PackageSpec};
         registry::Bool=false,)::Nothing
     unresolved_uuids = Dict{String,Vector{UUID}}()
@@ -901,7 +901,6 @@ function ensure_resolved(manifest::Manifest,
         push!(unresolved_names, pkg.uuid)
     end
     isempty(unresolved_uuids) && isempty(unresolved_names) && return
-    ctx = Context()
     msg = sprint(context = ctx.io) do io
         if !isempty(unresolved_uuids)
             print(io, "The following package names could not be resolved:")
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -901,21 +901,30 @@ function ensure_resolved(manifest::Manifest,
         push!(unresolved_names, pkg.uuid)
     end
     isempty(unresolved_uuids) && isempty(unresolved_names) && return
-    msg = sprint() do io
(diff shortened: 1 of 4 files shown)


test/new.jl
```diff
@@ -407,6 +407,19 @@ end
         @test_throws PkgError(
             "version specification invalid when tracking a repository: `0.5.0` specified for package `Example`"
             ) Pkg.add(name="Example", rev="master", version="0.5.0")
+        # Adding with a slight typo gives suggestions
+        try Pkg.add("Examplle")
+            @assert false # to fail if add doesn't error
```
KristofferC, line 413: `@test` instead of assert maybe?

IanButterworth:
Let's merge to get people trying this out in the first 1.8 cut

src/Types.jl
```diff
@@ -928,6 +936,27 @@ function ensure_resolved(manifest::Manifest,
     pkgerror(msg)
 end
 
+# copied from REPL to efficiently expose if any score is >0
+function fuzzysort(search::String, candidates::Vector{String})
+    scores = map(cand -> (REPL.fuzzyscore(search, cand), -Float64(REPL.levenshtein(search, cand))), candidates)
+    candidates[sortperm(scores)] |> reverse, any(s -> s[1] > 0, scores)
+end
```
ericphanson, line 943: from Slack- it would be better to use REPL's `levsort` instead (https://github.com/JuliaLang/julia/blob/f3e3554910a79fa38d5662d14abe04e137075e18/stdlib/REPL/src/docview.jl#L652-L660), which is what the REPL uses for corrections: https://github.com/JuliaLang/julia/blob/f3e3554910a79fa38d5662d14abe04e137075e18/stdlib/REPL/src/docview.jl#L700-L707.

This does a lot better on `DifferentielEquations`:
```julia
julia> levsort("DifferentielEquations", all)
3-element Vector{String}:
 "DifferentialEquations"
 "DifferenceEquations"
 "DifferentialEvolutionMCMC"
```
vs
```julia
julia> fuzzysort("DifferentielEquations", all)
(["DifferentialEvolutionMCMC", "DifferentialDynamicsModels", "DifferentiableStateSpaceModels", "DifferentialEquations", "SimpleDifferentialOperators", "DiscreteDifferentialGeometry", "DifferentialForms", "DifferentialDynamicProgramming", "DifferenceEquations", "AbstractDifferentiation"  …  "BenchmarkingEconomicEfficiency", "BayesianNonparametricStatistics", "ApproxFunOrthogonalPolynomials", "GAP_pkg_normalizinterface_jll", "ReinforcementLearningEnvironmentAtari", "MultivariateOrthogonalPolynomials", "BlackBoxOptimizationBenchmarking", "SemiclassicalOrthogonalPolynomials", "OpenTelemetryExporterOtlpProtoGrpc", "ReinforcementLearningEnvironmentClassicControl"], false)
```

PR #2986, opened 2022-02-14
`Pkg.Registry.uncompress`: if overlapping ranges are found, print detailed debugging information in the error message

### LINKED ISSUE #2899: Undefined Variable in `uncompress()`
There is undefined variable `path` on  [https://github.com/JuliaLang/Pkg.jl/blob/master/src/Registry/registry_instance.jl#L86](https://github.com/JuliaLang/Pkg.jl/blob/master/src/Registry/registry_instance.jl#L86).

DilumAluthge:
Fixes #2899

@andreasnoack and I ran into this error with a private registry, and it was very difficult to pinpoint the exact location of the problem, because the current error message does not provide very much information.

With this PR, the error message for overlapping ranges will include the following data:
1. Registry.toml -> `path`
2. Registry.toml -> `registry_path`
3. Registry.toml -> `name`
4. Registry.toml -> `uuid`
5. Package.toml -> `repo`
6. Package.toml -> `subdir`

This should make it much easier to figure out where the problem is.


andreasnoack:
This seems like a great improvement to me (and it fixes a clear bug). I'm not familiar with this code so I'll let others review.

timholy:
I think we just hit this too (not my machine and via zoom, so I can't do a detailed diagnosis).

DilumAluthge:
Needs a review from someone that knows the code.

DilumAluthge:
Bump @KristofferC

andreasnoack:
@KristofferC bumping this again

DilumAluthge:
Bump @KristofferC

andreasnoack:
There are some conflicts here now. Are the changes still relevant?

KristofferC:
Sorry for dropping the ball on this one. In my defense, I do have a note on my profile saying to ping me on Slack. I just have not found a good way to keep up with GitHub notifications without missing things. I think this is still good to have, yes.

src/Registry/registry_instance.jl
```diff
@@ -183,6 +193,45 @@ mutable struct PkgEntry
...
+        ")"
+    return summary
+end
+
+function getproperty_maybe(x::Union{PkgInfo, PkgEntry, Nothing}, field::Symbol)
+    default_value = ""
+    if x isa PkgInfo
+        return something(getproperty(x, field), default_value)
+    elseif x isa PkgEntry
+        if !LazilyInitializedFields.islazyfield(typeof(x), field)
```
KristofferC, line 224: I think this dependency (`LazilyInitializedFields`) is completely dropped now.

KristofferC:
Just FYI, this has been "fixed" with

https://github.com/JuliaLang/Pkg.jl/blob/88a46e68f9649ec3f5cadc2b8f82825683967240/src/Registry/registry_instance.jl#L321-L322

but the verification has been turned off because it turned out to be too expensive by default. Maybe it can be optimized..

PR #2987, opened 2022-02-14
doc: introduce temporary environment and shared environment

johnnychen94:
Pkg has so many secret yet convenient commands to improve the dev workflow.
This PR introduces two that I know: `pkg> activate --temp` and `pkg> activate @shared_pkg`.
It also includes a brief introduction on `LOAD_PATH` and caveats on it.


docs/src/environments.md
```diff
@@ -107,3 +112,146 @@ Otherwise, it will resolve the latest versions of the dependencies compatible wi
...
+```julia-repl
+(@v1.6) pkg> activate @mytoolbox # require Julia 1.4 or later
+  Activating new project at `~/.julia/environments/mytoolbox`
+```
+
+This provides a quick way to activate the project environment stored in `~/.julia/environments` folder.
+
+### Shared environment caveat: the reproducibility issue
+
+Using shared environment can cause reproducibility issues. For instance,
```
stillyslalom, line 217:
```suggestion
!!! warning Shared environments and reproducibility
    Because shared environments allow you to load packages without specifying them in a particular environment's
    `Project.toml` and `Manifest.toml`, you may encounter problems with reproducibility.
```

docs/src/environments.md
```diff
@@ -107,3 +112,146 @@ Otherwise, it will resolve the latest versions of the dependencies compatible wi
...
+
+using BenchmarkTools
+
+greet() = print("Hello World!")
+
+end # module
+```
+
+Now when you tries to load `MyPackage`, because `BenchmarkTools` is recorded in the global shared environment `v#.#`,
+Julia still succeeds loading it but will print a warning message:
```
stillyslalom, line 243:
```suggestion
When you try to load `MyPackage`, Julia searches for and finds `BenchmarkTools` in the global shared environment `v#.#`, so loading is successful, but a warning message is printed:
```

docs/src/environments.md
```diff
@@ -107,3 +112,146 @@ Otherwise, it will resolve the latest versions of the dependencies compatible wi
...
+│ - If you have MyPackage checked out for development and have
+│   added BenchmarkTools as a dependency but haven't updated your primary
+│   environment's manifest file, try `Pkg.resolve()`.
+│ - Otherwise you may need to report an issue with MyPackage
+└ Loading BenchmarkTools into MyPackage from project dependency, future warnings for MyPackage are suppressed.
+```
+
+Imaging that you want to ship your wonderful package to other people, because `BenchmarkTools` is not recoreded in `MyPackage/Project.toml`,
+you can't ensure that `BenchmarkTools` can be successfully loaded even if they initialize your package with `pkg> instantiate`.
+In such sense, this is a reproducibility warning; when you develop a package that meant to be shared by others, you should avoid doing this.
```
stillyslalom, line 257:
```suggestion
Imagine that you want to ship your wonderful package to other people. Because `BenchmarkTools` is not recorded in `MyPackage/Project.toml`,
you can't ensure that `BenchmarkTools` will be successfully loaded. For this reason, developers of shared packages should be especially mindful of reproducibility issues when using shared environments.
```

docs/src/environments.md
```diff
@@ -49,18 +50,15 @@ julia> print(read("Project.toml", String))
...
-version = "0.5.1"
-
-[[Test]]
-uuid = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
+version = "0.5.3"
 ```
 
-This new environment is completely separate from the one we used earlier.
-
+This new environment is completely separate from the one we used earlier. See [`Project.toml` and `Manifest.toml`](@ref Project-and-Manifest) for more detailed explaination on them.
```
stillyslalom, line 61:
```suggestion
This new environment is completely separate from the one we used earlier. See [`Project.toml` and `Manifest.toml`](@ref Project-and-Manifest) for a more detailed explanation.
```

stillyslalom (changes requested):
I've made a few typo corrections and grammar suggestions.

docs/src/environments.md
johnnychen94: I'm not sure if it's appropriate to use `!!! ` for a long thread like this, so I kept the `###` title here.

docs/src/environments.md
```diff
@@ -107,3 +112,145 @@ Otherwise, it will resolve the latest versions of the dependencies compatible wi
     ```bash
     $ julia --project=. myscript.jl
     ```
+
+
+## Temporary environments
+
+It is not uncommon to test some package without installing it into your usual environment.
```
IanButterworth, line 119:
```suggestion
Temporary environments make it easy to start an environment from a blank slate to test a package or set of
packages, and have Pkg automatically delete the environment when you're done.
```

jkrumbiegel:
I think it would be helpful if these REPL commands also had the `Pkg.func` version side-by-side. In scripts, if you do `pkg"activate @myenv"` you get a warning that you're not supposed to use this interactive version, but I wouldn't know what the non-interactive incantation is, as `Pkg.activate("@myenv")` doesn't work.

docs/src/environments.md
```diff
@@ -107,3 +112,146 @@ Otherwise, it will resolve the latest versions of the dependencies compatible wi
...
+    Updating `/tmp/juliatmp/environments/v1.6/Manifest.toml`
+  [7876af07] + Example v0.5.3
+Precompiling project...
+  1 dependency successfully precompiled in 0 seconds
+```
+
+This way, anything downloaded or modified by package operations will be stored in the temporary directory
+`/tmp/juliatmp`, with no need to worry that it might break your regular workflow.
+
+## [shared environment](@id shared-environment)
```
KristofferC, line 171: I don't think this description is correct for shared environments. A shared environment is just an environment that can be referred to by name and is put in `~/.julia/environments/`. For example:

```
(@v1.7) pkg> activate --shared Foo
  Activating project at `~/.julia/environments/Foo`
```

This is completely separate from `LOAD_PATH` and a shared environment created like above will not be in the default load path if you activate a different project.

KristofferC:
I removed some things so this can be merged straight away, but the old stuff is still in the original commit so it isn't lost.

docs/src/environments.md
```diff
@@ -112,3 +112,146 @@ Otherwise, it will resolve the latest versions of the dependencies compatible wi
...
+    Updating `/private/var/folders/34/km3mmt5930gc4pzq1d08jvjw0000gn/T/jl_a31egx/Project.toml`
+  [7876af07] + Example v0.5.3
+    Updating `/private/var/folders/34/km3mmt5930gc4pzq1d08jvjw0000gn/T/jl_a31egx/Manifest.toml`
+  [7876af07] + Example v0.5.3
+```
+
+The temporary directory will be removed when you exit Julia so you don't need to worry this operation
+breaks your common workflow.
+
+Note that this still reuses the contents in the global depot path in `~/.julia`. Thus you might still hit
```
KristofferC: I felt this started to get way too long. I have almost never needed to change the depot path because things in it are content tracked so it should never " break your common workflow.". We have to try keep the text a bit short so readers don't get fatigued.

PR #2990, opened 2022-02-15
Don't warn if showing manifest and outdated indirect deps are hidden

IanButterworth:
Fix #2989 

Closes #2991 

cc. @fingolfin

Issue #2989, opened 2022-02-15
Mysterious "Some packages have new versions but cannot be upgraded" message


fingolfin:
I really like the new `status --outdated` feature coming in Julia 1.8! It relieves a serious pain point that's been bothering some of us for years. Thanks!

There is a minor weirdness about it, though: in some cases it keeps telling me
> Info Some packages have new versions but cannot be upgraded. To see why use `status --outdated`

But entering that command shows no packages:
```
$ julia-master
               _
   _       _ _(_)_     |  Documentation: https://docs.julialang.org
  (_)     | (_) (_)    |
   _ _   _| |_  __ _   |  Type "?" for help, "]?" for Pkg help.
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.8.0-DEV.1558 (2022-02-15)
 _/ |\__'_|_|_|\__'_|  |  Commit 9542577c88* (0 days old master)
|__/                   |

(@v1.8) pkg> update
    Updating registry at `~/.julia/registries/General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
  No Changes to `~/.julia/environments/v1.8/Project.toml`
  No Changes to `~/.julia/environments/v1.8/Manifest.toml`
        Info Some packages have new versions but cannot be upgraded. To see why use `status --outdated`

(@v1.8) pkg> status --outdated
Status `~/.julia/environments/v1.8/Project.toml`

(@v1.8) pkg> status
Status `~/.julia/environments/v1.8/Project.toml`
  [6e4b80f9] BenchmarkTools v1.3.1
  [c3a54625] JET v0.5.6
  [f1435218] Oscar v0.8.1-DEV `~/.julia/dev/Oscar`
  [295af30f] Revise v3.3.1
  [aa65fe97] SnoopCompile v2.9.0
  [e2b509da] SnoopCompileCore v2.9.0

(@v1.8) pkg>
```

The answer, "of course" (well... as I realize while typing this, but after having been puzzled by it for several days) is that there *is* a package in my manifest that is outdated, but it is only a dependency of one of my packages:
```
(v1.8) pkg> status --outdated -m
Status `/tmpbig/horn/julia_depot/environments/v1.8/Manifest.toml`
⌅ [4fe8b98c] Mongoc v0.6.2 (<v0.7.1): Polymake
```

I see two solutions for this:

[...]

PR #2991, opened 2022-02-15
add `--manifest` to warn if showing manifest status

IanButterworth:
Alt to #2990 

Fix https://github.com/JuliaLang/Pkg.jl/issues/2989

Closes #2990

cc. @fingolfin @DilumAluthge

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -2075,7 +2075,8 @@ function print_status(env::EnvCache, old_env::Union{Nothing,EnvCache}, registrie
             printpkgstyle(io, :Info, "Packages marked with $upgradable_indicator and $heldback_indicator have new versions available, but those with $heldback_indicator cannot be upgraded. To see why use `status --outdated`", color=Base.info_color(), ignore_indent)
         end
         if hidden_upgrades_info && no_visible_packages_heldback && !no_packages_heldback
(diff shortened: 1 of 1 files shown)


DilumAluthge (approved):
LGTM!

KristofferC:
This doesn't make sense to me. Why would you then not show all packages (even those in the manifest) when you do `status`? The point about splitting up showing direct and indirect dependencies is that the direct dependencies are those you can directly change by changing the project file (e.g its compat). If all your packages are at the latest version but you still have some indirect dependency at the non latest version, that will not directly affect you since you are consuming the latest API of all your dependencies and indirect dependencies are more an implementation detail of your direct dependencies.

DilumAluthge:
If I understand correctly, this isn't about `] status`, right? It's about `] update`.

I think that there are plenty of people that will assume that if they do `] update`, it will update all packages in their environment (both direct dependencies and indirect dependencies). So, if a user does `] update`, and then, after the update, one or more indirect dependencies are not at the latest version, then I absolutely think that we should print a warning to the user.

KristofferC:
> then I absolutely think that we should print a warning to the user.

I don't think so because there is nothing as a user that you can do. If you are a power user / developer that wants to go and update the compat of your dependencies then you can just do `st --manifest` after. Showing warnings should be done very sparingly, people getting warnings when there is nothing they can really do is not user friendly in my opinion.

PR #2992, opened 2022-02-16
Add the `Pkg.__diagnostics__()` function

DilumAluthge:
See https://github.com/JuliaLang/julia/pull/43834#issuecomment-1034211488 for context.

---

1. Registry status
2. Project status
3. Manifest status
4. List of outdated packages in the project
5. List of outdated packages in the manifest
6. Information about the current Pkg server


IanButterworth:
Might just be simpler to call it `Pkg.diagnostics()`?

Also, might it be worth adding ping/download speed tests for the pkgserver & GitHub?

DilumAluthge:
> Might just be simpler to call it `Pkg.diagnostics()`?

So, the current plan is that the user-facing interface will be a single function called `InteractiveUtils.diagnostics()`. This will be the only function that users will need to call. When the user calls `InteractiveUtils.diagnostics()`, it will automatically call the `__diagnostics__()` function for any module in `Base.loaded_modules` that has a `__diagnostics__()` function.

So then we can add these `__diagnostics__` functions to Pkg, LinearAlgebra, NetworkOptions, etc., and the user only has to call the `InteractiveUtils.diagnostics()` to run all of them.

So that's why the function in this PR is named `__diagnostics__`.

DilumAluthge:
> Also, might it be worth adding ping/download speed tests for the pkgserver & GitHub?

Definitely! Any ideas on how we might go about implementing those? I'm sure it's doable with the Downloads and/or LibCURL_jll stdlibs, but I don't actually know how to do so.

KristofferC:
I think it is strange to start this with `__`. The only other thing that I can think of that does this is `__init__` and that one is "magic" so it makes you think that this method also has some magic. Also, I don't think there is need for a whole new source + test file.

StefanKarpinski:
Agree that the underscores are weird. In Python they mean that a name is reserved as having special meaing for the language itself, which is actually also how we use them in `__init__` which has special behavior. This doesn't, it's just a regular function.

Issue #2997, opened 2022-02-19
Uncompressible registry error should be more helpful


IanButterworth:
From slack.

The error here could be more informative/actionable
```
(@v1.7) pkg> activate .
 Activating new project at `~/Projects/model-selection`

(model-selection) pkg> add DataFrames
ERROR: Unexpected end of data : General.tar
ERROR: EOFError: read end of file
Stacktrace:
 [1] read_data(tar::Base.Process, file::IOBuffer; size::Int64, buf::Vector{UInt8}, tee::Base.DevNull)
  @ Tar ~/julia-1.7.2/share/julia/stdlib/v1.7/Tar/src/extract.jl:706
 [2] #12
  @ ~/julia-1.7.2/share/julia/stdlib/v1.7/Pkg/src/Registry/registry_instance.jl:201 [inlined]
 [3] read_tarball(callback::Pkg.Registry.var"#12#15"{Base.Process, IOBuffer, Vector{UInt8}, Dict{String, String}}, predicate::Pkg.Registry.var"#13#16", tar::Base.Process; buf::Vector{UInt8}, skeleton::Base.DevNull)
  @ Tar ~/julia-1.7.2/share/julia/stdlib/v1.7/Tar/src/extract.jl:399
 [4] #11
  @ ~/julia-1.7.2/share/julia/stdlib/v1.7/Pkg/src/Registry/registry_instance.jl:199 [inlined]
 [5] open(::Pkg.Registry.var"#11#14"{IOBuffer, Vector{UInt8}, Dict{String, String}}, ::Cmd; kwargs::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
  @ Base ./process.jl:395
 [6] open
  @ ./process.jl:393 [inlined]
 [7] uncompress_registry(tar_gz::String)
  @ Pkg.Registry ~/julia-1.7.2/share/julia/stdlib/v1.7/Pkg/src/Registry/registry_instance.jl:198
 [8] Pkg.Registry.RegistryInstance(path::String)
  @ Pkg.Registry ~/julia-1.7.2/share/julia/stdlib/v1.7/Pkg/src/Registry/registry_instance.jl:266
 [9] reachable_registries(; depots::Vector{String})
  @ Pkg.Registry ~/julia-1.7.2/share/julia/stdlib/v1.7/Pkg/src/Registry/registry_instance.jl:373
 [10] reachable_registries
[...]

PR #2998, opened 2022-02-19
Detect uncompressible registry tarballs during update. Add path to error

IanButterworth:
Tries to help https://github.com/JuliaLang/Pkg.jl/issues/2997

I wanted to add some guidance like "Try a `pkg> registry rm/add`" but `registry rm` tries to load the `RegistryInstance` so hits the same error.

Saying simply `rm regpath` also doesn't help because you also need to delete the toml, so the instruction gets a bit long.

Alternatively, could the file hash of the tarball be stored in the registry toml, and it checked in `verify_compressed_registry_toml`

### DIFF
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -369,7 +369,7 @@ function update(regs::Vector{RegistrySpec} = RegistrySpec[]; io::IO=stderr_f(), 
                                     @goto done_tarball_read
                                 end
                                 try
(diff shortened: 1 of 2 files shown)


IanButterworth:
In place of a faster file integrity check like a file hash, I just added an uncompress check that doesn't store the data

IanButterworth:
Fixed by https://github.com/JuliaLang/Pkg.jl/pull/3408

PR #2999, opened 2022-02-21
Use levsort for `add` suggestions

IanButterworth:
As @ericphanson [pointed out](https://github.com/JuliaLang/Pkg.jl/pull/2985#discussion_r810556268), levsort seems to work better

Screen Shot 2022-02-20 at 6 57 34 PM

However, some of these are quite distant, especially for short names.

@ericphanson any idea of a better threshold to use here?

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -913,12 +913,18 @@ function ensure_resolved(ctx::Context, manifest::Manifest,
                     join(io, what, ", ", " or ")
                     print(io, ")")
                     all_names = available_names(ctx; manifest, include_registries = registry)
-                    all_names_ranked, any_score_gt_zero = fuzzysort(name, all_names)
-                    if any_score_gt_zero
+                    closest_names_ranked = REPL.levsort(name, all_names)
+                    if !isempty(closest_names_ranked)
                         println(io)
                         prefix = "   Suggestions:"
                         printstyled(io, prefix, color = Base.info_color())
-                        REPL.printmatches(io, name, all_names_ranked; cols = REPL._displaysize(ctx.io)[2] - length(prefix))
+                        printed_length = length(prefix)
+                        disp_width = displaysize(ctx.io)[2]
+                        for name in closest_names_ranked
+                            printed_length += length(name) + 1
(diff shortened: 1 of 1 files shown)


ericphanson:
We could take top 3 results maybe? But that’s not ideal- if we have say a 3 or 4 letter package, there might be many packages one substitution away, so we are depending on fuzzy score for tie-breaking to decide the top 3 results. But fuzzy score seems to roughly measure how close the start of the names matches- not sure how useful that is.

In those examples though, at least the first result seems good, so maybe just limiting the # of results will help.

What would be better would be if we had a weighted Levenshtein distance that used typo frequencies on say qwerty keyboards to determine the weights. Then we wouldn’t have so many ties, since the weights would give us a better ranking. (I’d like this for RegistryCI too!). But probably that’s out of scope for this…

IanButterworth:
It could be a pkgserver query. Send a name, get back a shortlist of suggestions. Then we could throw any/all algorithms at it. Package popularity, seo ranking etc etc

I think that was @StefanKarpinski's suggestion somewhere. Obviously a fair bit of work, and wouldn't work for other registries or local packages

ericphanson:
I don't think we have a weighted Levenshtein (or even better, Damerau-Levenshtein) in Julia, so we'd still need to write that if we wanted that algorithm, but at least it could be a separate package, true. Seems a lot simpler to not need to make a network call though.

> Package popularity, seo ranking etc etc

That does sound cool... I think at this point though, we don't have so many packages / name clashes that we need that kind of indicator, and hopefully with automerge's nudge towards name dissimilarity, it can stay that way for awhile. I could see how this could be necessary at some point though, especially if typo squatting started to happen... but ideally that can be handled better at the registry maintenance-level.

> wouldn't work for other registries or local packages

I suppose you could setup a private registry with a private package server, and implement the API endpoint for name similarity, and that could work. But as someone who uses a private registry without a package server, I'd still prefer it to work without a package server set up :).

waldyrious:
Cross-referencing #1834, for historical record.

KristofferC:
Subsumed by https://github.com/JuliaLang/Pkg.jl/pull/4287 which does better than this on the examples given here.

PR #3000, opened 2022-02-21
Handle custom manifest name

IanButterworth:
Paired with https://github.com/JuliaLang/julia/pull/44286

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -180,12 +180,11 @@ end
 function manifestfile_path(env_path::String; strict=false)
     man_names = Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
     if length(man_names) == 1
-            man_path = joinpath(env_path, only(man_names))
-            if strict && !isfile(man_path)
-                return nothing
-            else
-                return man_path
-            end
+        man_path = joinpath(env_path, only(man_names))
+        if strict && !isfile(man_path)
+            return nothing
+        else
+            return man_path
         end
     end
     for name in man_names
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -178,7 +178,7 @@ function projectfile_path(env_path::String; strict=false)
 end
 
 function manifestfile_path(env_path::String; strict=false)
-    man_names = Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
+    man_names = @static Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
     for name in man_names
         maybe_file = joinpath(env_path, name)
         isfile(maybe_file) && return maybe_file
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -178,15 +178,7 @@ function projectfile_path(env_path::String; strict=false)
 end
 
 function manifestfile_path(env_path::String; strict=false)
-    man_names = Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
-    if first(man_names) == last(man_names) # indicative of a custom manifest name
-        man_path = joinpath(env_path, first(man_names))
-        if strict && !isfile(man_path)
-            return nothing
-        else
-            return man_path
-        end
-    end
+    man_names = Base.manifest_names isa Vector ? Base.manifest_names : Base.manifest_names()
     for name in man_names
         maybe_file = joinpath(env_path, name)
         isfile(maybe_file) && return maybe_file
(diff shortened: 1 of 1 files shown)


src/Types.jl
```diff
@@ -178,7 +178,17 @@ function projectfile_path(env_path::String; strict=false)
...
 
 function manifestfile_path(env_path::String; strict=false)
-    for name in Base.manifest_names
+    man_names = Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
+    if length(man_names) == 1
+            man_path = joinpath(env_path, only(man_names))
+            if strict && !isfile(man_path)
+                return nothing
+            else
+                return joinpath(env_path, only(man_names))
```
DilumAluthge, line 187:
```suggestion
            man_name = only(man_names)
            man_path = joinpath(env_path, man_name)
            if strict && !isfile(man_path)
                return nothing
            else
                return joinpath(env_path, man_name)
```

src/Types.jl
```diff
@@ -178,7 +178,17 @@ function projectfile_path(env_path::String; strict=false)
 end
 
 function manifestfile_path(env_path::String; strict=false)
-    for name in Base.manifest_names
+    man_names = Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
```
DilumAluthge, line 181:
```suggestion
    @static if Base.VERSION >= v"1.9.0-"
        man_names = Base.manifest_names()
    else
        man_names = Base.manifest_names
    end
```

IanButterworth: This will have to follow. We can't merge this kind of catch before it's merged in Base

IanButterworth:
Errm..
```
ERROR: LoadError: UndefVarError: man_names not defined
Stacktrace:
 [1] top-level scope
   @ ~/work/Pkg.jl/Pkg.jl/src/Types.jl:191
```

Is it me, or does this look like the purity PR deleting `man_names`.. I can't see how it's not defined on 191

DilumAluthge:
You could try building Julia from source with the commit right before the purity modeling PR was merged.

DilumAluthge:
I think you might have a syntax error somewhere.

DilumAluthge:
You have one too many `end` in that function.

DilumAluthge:
You have an `end` on 188, 189, and 190. One of those shouldn't be there.

src/Types.jl
DilumAluthge: Just as a short-term stop-gap, how about this?

```suggestion
    @static if Base.manifest_names isa Tuple
        man_names = Base.manifest_names()
    else
        man_names = Base.manifest_names
    end
```

Then we can replace it with the `Base.VERSION` check once the Base PR is merged and the nightly binaries catch up.

DilumAluthge: Or, if you want to keep the ternary:

```suggestion
    man_names = @static Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
```

DilumAluthge (approved):
I'd like that `@static` to be added, but other than that, this looks good to me.

DilumAluthge:
Once this PR is merged, we can bump the Pkg version on Julia master, and after that bump PR is merged, we can rebase https://github.com/JuliaLang/julia/pull/44286.

src/Types.jl
```diff
@@ -178,7 +178,16 @@ function projectfile_path(env_path::String; strict=false)
...
+    if first(man_names) == last(man_names) # indicative of a custom manifest name
+        man_path = joinpath(env_path, first(man_names))
+        if strict && !isfile(man_path)
+            return nothing
+        else
+            return man_path
+        end
+    end
```
DilumAluthge, line 189: @IanButterworth I may be missing something, but now that we guarantee that `man_names` is always the same length as `Base.project_names`, can we not just delete lines 182 to 189?

IanButterworth: Sure but it will check for the file twice if it doesn't exist.

DilumAluthge: Ah. Yeah, it's probably good to avoid doing unnecessary filesystem operations.

src/Types.jl
```diff
@@ -178,7 +178,16 @@ function projectfile_path(env_path::String; strict=false)
 end
 
 function manifestfile_path(env_path::String; strict=false)
-    for name in Base.manifest_names
+    man_names = Base.manifest_names isa Tuple ? Base.manifest_names : Base.manifest_names()
+    if first(man_names) == last(man_names) # indicative of a custom manifest name
+        man_path = joinpath(env_path, first(man_names))
```
DilumAluthge, line 183: We can get rid of one of these calls to `first(man_names)`.

```suggestion
    first_man_name = first(man_names)
    if first_man_name == last(man_names) # indicative of a custom manifest name
        man_path = joinpath(env_path, first_man_name)
```

PR #3010, opened 2022-03-01
improve error message for erroring when dev package that is in sysimage

KristofferC:


DilumAluthge (approved):
Looks like an improvement to me!

PR #3013, opened 2022-03-02
list outdated kwarg of Pkg.status

ranocha:
I really like this new feature available in Julia v1.8. I tried to improve the docstring of `Pkg.status` a bit to make the new keyword `outdated` more clear, at least for me.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -399,12 +399,13 @@ from packages that are tracking a path.
 const resolve = API.resolve
 
 """
-    Pkg.status([pkgs...]; mode::PackageMode=PKGMODE_PROJECT, diff::Bool=false, compat::Bool=false, io::IO=stdout)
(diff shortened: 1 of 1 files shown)


src/Pkg.jl
```diff
@@ -399,12 +399,13 @@ from packages that are tracking a path.
...
 
 """
-    Pkg.status([pkgs...]; mode::PackageMode=PKGMODE_PROJECT, diff::Bool=false, compat::Bool=false, io::IO=stdout)
+    Pkg.status([pkgs...]; outdated::Bool=false, mode::PackageMode=PKGMODE_PROJECT, diff::Bool=false, compat::Bool=false, io::IO=stdout)
 
 Print out the status of the project/manifest.
 
 Packages marked with `⌃` have new versions that can be installed, e.g. via [`Pkg.up`](@ref).
-Those marked with `⌅` have new versions available, but that cannot be installed. To see why use the `outdated` kwarg.
+Those marked with `⌅` have new versions available, but that cannot be installed. To see why, set the
```
KristofferC, line 407:
```suggestion
Those marked with `⌅` have new versions available, but cannot be installed due to compatibility conflicts with other packages. To see why, set the
```

PR #3017, opened 2022-03-06
Remove BinaryPlatforms_compat

timholy:
This no longer appears to be necessary, and having extra methods of
functions like Base.BinaryPlatforms.libgfortran_version has a cost: it
becomes impossible for abstract inference to succeed, since it pushes
the number of methods over `MAX_METHODS`.


timholy:
This is harder than I thought. I'm going to close this for now, but if someone who knows more about the JLL ecosystem can sketch a path forward I'd be happy to try again.

PR #3024, opened 2022-03-09
Warn when activating manifest that was resolved in a different julia version

### LINKED ISSUE #3019: Proposal: warn (at least) on activation of a project made with a different Julia version
Making an issue here as suggested on the [discourse thread](https://discourse.julialang.org/t/julia-shoudnt-silently-activate-a-project-associated-to-a-previous-julia-version/77480).

The idea is that now Manifest holds the information on the Julia version itself but "ignores" this information if activating the environment from another Julia version.
While it is true that it does issue a warning in instantiating it, sometimes you don't need to instantiate.
For example, I did start a new environment in Julia 1.7, add a DataFrames and then activated it from Julia 1.8 beta, I had no warning. Yes, the package I were using were the same, but Julia wasn't, and I think this breaks the promise of replicability of Manifest.toml. 

An option could always be added to `activate` NOT to warn/error in such cases (eg `strict = false`)...

IanButterworth:
Intended to fix #3019 

```
(@v1.9) pkg> activate .
  Activating project at `~/Documents/GitHub/Pkg.jl`
┌ Warning: The active manifest file has dependencies that were resolved with a different julia version (1.7.1). Unexpected behavior may occur.
└ @ ~/Documents/GitHub/Pkg.jl/Manifest.toml:0
```

I was surprised that Pkg wasn't already using the `Base.set_active_project` API. This PR leverages that so that it can also cover when julia is started with `--project`. But that is untested as I haven't built julia locally with this Pkg in the sysimage yet.

Also, I've been thinking that Pkg should use fewer `@error`'s and instead format the error messages using `printpkgstyle`? The clash of the two formatting modes is a little inelegant.


DilumAluthge:
> Also, I've been thinking that Pkg should use fewer `@error`'s and instead format the error messages using `printpkgstyle`? The clash of the two formatting modes is a little inelegant.

Agreed, we should be consistent within Pkg.

fredrikekre:
I predict this is going to be super annoying.

If something like this is included it should at the very least tell the user what to do.

DilumAluthge:
Can you print both Julia versions (both the current Julia version as well as the manifest's Julia version)? Just to make the contrast easier to see when you're quickly skimming the logs?

DilumAluthge:
> If something like this is included it should at the very least tell the user what to do.

It seems reasonable to give the user a possible next step? E.g. something like this:

> You may want to do `imporrt Pkg; Pkg.resolve()` to re-resolve the manifest with the active Julia version (1.y.z).

PR #3035, opened 2022-03-22
Add the `throw_on_failure` keyword argument to `Pkg.precompile`, which causes `Pkg.precompile` to throw an exception if precompilation encounters any errors

### LINKED ISSUE #3026: Feature request: `Pkg.precompile(; throw_error = true)`
Currently, if I do `julia -e 'import Pkg; Pkg.precompile()'`, if one or more of the packages in my manifest fails to precompile, the Julia process will still exit with exit code zero.

I'd like to have the ability to do something like `Pkg.precompile(; throw_error = true)`, in which case if one or more of the packages in my manifest fails to precompile, the `Pkg.precompile` function will throw an exception, and thus the Julia process will exit with nonzero exit code.

DilumAluthge:
Fixes #3026


DilumAluthge:
cc: @andreasnoack

IanButterworth:
I don't think this kwarg is needed. We just need to make the existing extended error message that shows when you directly call Pkg.precompile throw as a pkgerror.

PR #3040, opened 2022-03-23
Fix alignment with indicators present

### LINKED ISSUE #3018: Wrong output alignment when environment has outdated and packages missing the source
```
  [8e7c35d0] BlockArrays v0.16.11
→⌃ [e30172f5] Documenter v0.27.13
  [c061ca5d] Ferrite v0.3.4 `..`
```

barucden:
Adjusts the logic for printing the padding and indicators for packages in the Status operation.

I would like to add a test with a package that is downloaded but can be upgraded. Can you give me a hint how to do that?

Fixes #3018


IanButterworth:
> I would like to add a test with a package that is downloaded but can be upgraded. Can you give me a hint how to do that?

I'd just add a test project beside `test/test_packages/Status` with a checked-in manifest with an outdated version of Example, then duplicate the tests you've modified here

barucden:
I figured I could just instantiate and check again using the same Status test package. Let me know if I should improve it.

IanButterworth:
Yeah that works! Thanks!

PR #3045, opened 2022-03-30
Fix Artifacts test on aarch64-apple-darwin14

staticfloat:
When we test that the old version of GMP_jll gets installed properly, we
need to realize that the old version of GMP_jll may not have an artifact
for some of our newer platforms.

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -2284,6 +2284,7 @@ end
 # # Caching
 #
 @testset "Repo caching" begin
+    default_branch = LibGit2.getconfig("init.defaultBranch", "master")
     # Add by URL should not overwrite files.
     isolate(loaded_depot=true) do
         Pkg.add(url="https://github.com/JuliaLang/Example.jl")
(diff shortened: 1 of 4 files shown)


test/new.jl
```diff
@@ -2821,16 +2821,21 @@ end
...
         artifacts_toml = joinpath(gmp_jll_dir, "Artifacts.toml")
         @test isfile(artifacts_toml)
-        meta = artifact_meta("GMP", artifacts_toml)
-        @test meta !== nothing
 
-        gmp_artifact_path = artifact_path(Base.SHA1(meta["git-tree-sha1"]))
-        @test isdir(gmp_artifact_path)
+        # These tests require a matching platform artifact for this old version of GMP_jll,
+        # which is not the case on some of our newer platforms.
+        if !(Sys.isapple() && Sys.ARCH != :aarch64)
```
giordano, line 2827:
```suggestion
        if !(Sys.isapple() && Base.BinaryPlatforms.arch(Base.BinaryPlatforms.HostPlatform()) == "aarch64")
```
may be more robust because in some cases `Sys.ARCH` could be `:arm64`?  Also, this should be done for armv6l, too, right (in case someone is going to run the tests on that board...)?

staticfloat: Yeah, I decided to take a deeper approach to fixing these things.

staticfloat:
HSG failure fixed by https://github.com/JuliaLang/Pkg.jl/pull/3048 so I'm going to merge this.

PR #3046, opened 2022-03-31
Update Historical Stdlibs list

staticfloat:
Note that while it looks strange that v1.9 was dropped, it's actually
because we lost v1.8 due to not having a beta for so long, and v1.8 and
v1.9 share the same list of stdlibs (so far).


DilumAluthge:
How did you generate this pull request? It looks different from #3047, which was generated by the bot.

DilumAluthge:
Closing in favor of #3047 (but please re-open if I was mistaken).

PR #3050, opened 2022-03-31
Fix REPL tests lacking `stdin` on `Distributed` workers

staticfloat:
When running the `Pkg` tests as part of the Julia test suite (e.g. via
`Base.runtests(["triplequote", "Pkg"])`, note that you need multiple
tests in order to trigger the `Distributed` behavior) we get run inside
of a `Distributed` worker which does not have access to `stdin`.
Attempting to run `redirect_stdin()` then fails with an error from
`dup()` such as the following:

```
SystemError: dup: Bad file descriptor
Stacktrace:
  [1] systemerror(p::String, errno::Int32; extrainfo::Nothing)
    @ Base ./error.jl:176
  [2] #systemerror#80
    @ ./error.jl:175 [inlined]
  [3] systemerror
    @ ./error.jl:175 [inlined]
  [4] dup
    @ ./libc.jl:37 [inlined]
  [5] _redirect_io_libc
    @ ./stream.jl:1213 [inlined]
  [6] (::Base.RedirectStdStream)(handle::Base.PipeEndpoint)
    @ Base ./stream.jl:1223
```

To fix this we follow the same convention as used in the base tests [0,1]
That is, we schedule work items on worker 1.  The main difference
between this and the base tests is that we must import `Pkg` and its
test utils on worker 1.

[0] https://github.com/JuliaLang/julia/blob/ef835ac1337ccc1463934af092ab37d9903eb6e3/test/spawn.jl#L336-L341
[1] https://github.com/JuliaLang/julia/blob/ef835ac1337ccc1463934af092ab37d9903eb6e3/test/show.jl#L723-L731

### DIFF
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -684,31 +684,43 @@ end
     @inferred Pkg.REPLMode.CompoundSpecs(Pair{String,Vector{Pkg.REPLMode.CommandDeclaration}}[])
 end
 
-# To be used to reply to a prompt
-function withreply(f, ans)
-    p = Pipe()
-    try
-        redirect_stdin(p) do
-            @async println(p, ans)
-            f()
-        end
-    finally
-        close(p)
-    end
+
+# stdin not available on other workers, so if we're being run on a Distributed worker
+# make sure this stuff happens on worker 1:
+using Distributed
+@everywhere [1] quote
+    using Pkg, Test
+    include("utils.jl")
+    using .Utils
 end
 
-@testset "REPL missing package install hook" begin
-    isolate(loaded_depot=true) do
-        @test Pkg.REPLMode.try_prompt_pkg_add(Symbol[:notapackage]) == false
-
-        # don't offer to install the dummy "julia" entry that's in General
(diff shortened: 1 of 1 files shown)


staticfloat:
@vtjnash I'm not sure if this is truly the best way to do this.  I tried to copy the examples in Base's tests, but I'm a little worried that blithely running `@spawnat 1 :(using Pkg, Test)` is going to break some other test that just ran on worker 1.

DilumAluthge:
Alternatively, could we just add Pkg to the list of "move to node 1" test sets [here](https://github.com/JuliaLang/julia/blob/3888176736b3f12aff132c2adbe5fd68f3fd0358/test/runtests.jl#L74-L79)?

DilumAluthge:
Oh, now I see what you're going for.

Instead of running all of the Pkg tests on node 1, we only need to run a few of the REPL tests on node 1.

DilumAluthge (commented):
The idea sounds good to me.

PR #3051, opened 2022-04-06
Respect the IO for downloads in instantiate

Pangoraw:
Hi!
This was initially introduced in https://github.com/JuliaLang/Pkg.jl/pull/2391, but was overriden in https://github.com/JuliaLang/Pkg.jl/pull/2268.

For now, nothing gets printed in the context io if it is not a `Base.TTY`:

https://github.com/JuliaLang/Pkg.jl/blob/a0f67d7ccdb4ba3a78d085e273bdde9334fc0f5e/src/Operations.jl#L657-L658

But the custom io could benefit from having at least the prints with "Downloading/Downloaded" in with_show_download_info:

https://github.com/JuliaLang/Pkg.jl/blob/a0f67d7ccdb4ba3a78d085e273bdde9334fc0f5e/src/Artifacts.jl#L474-L489

If that makes sense, I can add it to the PR or make a follow-up PR.


Pangoraw:
Bumping for feedback on how we can get the "Downloading/Downloaded artifact" message to be printed to a custom IO. :wave: 

Related: https://github.com/fonsp/Pluto.jl/issues/1451

KristofferC:
Looks like this PR was missed. Seems clearly this should be merged. (Also, cases like this are why I don't like putting default arguments on internal functions, too easy to miss passing one along.)

fonsp:
@KristofferC 🏓

PR #3056, opened 2022-04-14
Before running the "test atomicity of write_env_usage" testset, make sure that the General registry has been installed

DilumAluthge:
I'm hoping that this will fix the non-deterministic (but frequent) `Pkg` failures that I'm seeing in the `rr-net` job.

### DIFF
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -370,7 +370,7 @@ temp_pkg_dir() do project_path
         # to precompile Pkg given we're in a different depot
         run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg"`)
         # make sure the General registry is installed
-        run(pipeline(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg; Pkg.Registry.add()"`; stdout=devnull, stderr=devnull))
+        Utils.show_output_if_command_errors(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg; Pkg.Registry.add()"`)
         flag_start_dir = tempdir() # once n=Sys.CPU_THREADS files are in here, the processes can proceed to the concurrent test
         flag_end_file = tempname() # use creating this file as a way to stop the processes early if an error happens
         for i in 1:Sys.CPU_THREADS
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -368,6 +368,7 @@ temp_pkg_dir() do project_path
         iobs = IOBuffer[]
         Sys.CPU_THREADS == 1 && error("Cannot test for atomic usage log file interaction effectively with only Sys.CPU_THREADS=1")
         run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg"`) # to precompile Pkg given we're in a different depot
+        run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg; Pkg.Registry.add()"`) # make sure the General registry is installed
         flag_start_dir = tempdir() # once n=Sys.CPU_THREADS files are in here, the processes can proceed to the concurrent test
         flag_end_file = tempname() # use creating this file as a way to stop the processes early if an error happens
         for i in 1:Sys.CPU_THREADS
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -367,8 +367,10 @@ temp_pkg_dir() do project_path
         tasks = Task[]
         iobs = IOBuffer[]
         Sys.CPU_THREADS == 1 && error("Cannot test for atomic usage log file interaction effectively with only Sys.CPU_THREADS=1")
-        run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg"`) # to precompile Pkg given we're in a different depot
-        run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg; Pkg.Registry.add()"`) # make sure the General registry is installed
+        # to precompile Pkg given we're in a different depot
+        run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg"`)
+        # make sure the General registry is installed
+        run(pipeline(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg; Pkg.Registry.add()"`; stdout=devnull, stderr=devnull))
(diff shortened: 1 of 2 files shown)


DilumAluthge:
cc: @IanButterworth (who wrote these tests)

test/pkg.jl
```diff
@@ -367,7 +367,10 @@ temp_pkg_dir() do project_path
         tasks = Task[]
         iobs = IOBuffer[]
         Sys.CPU_THREADS == 1 && error("Cannot test for atomic usage log file interaction effectively with only Sys.CPU_THREADS=1")
-        run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg"`) # to precompile Pkg given we're in a different depot
+        # to precompile Pkg given we're in a different depot
+        run(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg"`)
+        # make sure the General registry is installed
+        run(pipeline(`$(Base.julia_cmd()) --project="$(pkgdir(Pkg))" -e "import Pkg; Pkg.Registry.add()"`; stdout=devnull, stderr=devnull))
```
vtjnash, line 373: Piping stderr to null is usually a bad idea, since that is where fatal error messages go

DilumAluthge: I don't want to show the output unless there is an error.

How about I pipe `stderr` to an `IOBuffer`, and print the contents of the `IOBuffer` only if the command fails?

vtjnash: stderr is where errors go to, so that should be the same

DilumAluthge: > stderr is where errors go to, so that should be the same

Unfortunately, it looks like `Pkg` is violating that rule 😂 . It prints the `Installing known registries into...` message to `stderr` even if there are no errors.

To reproduce:

```
julia> env2 = copy(ENV);

julia> delete!(env2, "JULIA_PKG_SERVER");

julia> env2["JULIA_DEPOT_PATH"] = mktempdir();

julia> julia = Base.julia_cmd().exec[1];

julia> cmd = `$(julia) -e 'import Pkg; Pkg.Registry.add(Pkg.Registry.RegistrySpec[])'`;

julia> run(pipeline(setenv(cmd, env2); stdout=devnull));
  Installing known registries into `/var/folders/dg/7mtlrlqd6fz01jdb9djb33bm0000gp/T/jl_NitKkG`
```

I've redirected `stdout` to `devnull`, so I'm assuming that the `Installing known registries...` message is being printed to `stderr`.

DilumAluthge: Alright, here's what I just pushed:
1. We don't redirect `stderr`, so `stderr` will always be shown (regardless of the command's exit code).
2. `stdout` will be shown if and only if the command exits with nonzero exit code.

DilumAluthge:
Here's the full stacktrace for the `ERROR: IOError: open("D:\\a\\Pkg.jl\\Pkg.jl\\test\\tmp\\logs\\manifest_usage.toml.pid", 1282, 292): permission denied (EACCES)` error I'm seeing on this PR on Windows:

Link to CI log: https://github.com/JuliaLang/Pkg.jl/runs/6033892020?check_suite_focus=true

```
Activating new project at `C:\Users\runneradmin\AppData\Local\Temp\jl_ZhyUdq`
Errored after 2 iterations
ERROR: IOError: open("D:\\a\\Pkg.jl\\Pkg.jl\\test\\tmp\\logs\\manifest_usage.toml.pid", 1282, 292): permission denied (EACCES)
Stacktrace:
[1] uv_error
  @ .\libuv.jl:97 [inlined]
[2] open(path::String, flags::UInt16, mode::UInt16)
  @ Base.Filesystem .\filesystem.jl:105
[3] tryopen_exclusive(path::String, mode::UInt16)
  @ FileWatching.Pidfile C:\hostedtoolcache\windows\julia\nightly\x64\share\julia\stdlib\v1.9\FileWatching\src\pidfile.jl:188
[4] open_exclusive(path::String; mode::UInt16, poll_interval::Int64, wait::Bool, stale_age::Int64)
  @ FileWatching.Pidfile C:\hostedtoolcache\windows\julia\nightly\x64\share\julia\stdlib\v1.9\FileWatching\src\pidfile.jl:237
[5] mkpidlock(at::String, pid::Int32; stale_age::Int64, refresh::Float64, kwopts::Base.Pairs{Symbol, Union{}, Tuple{}, NamedTuple{(), Tuple{}}})
  @ FileWatching.Pidfile C:\hostedtoolcache\windows\julia\nightly\x64\share\julia\stdlib\v1.9\FileWatching\src\pidfile.jl:56
[6] mkpidlock(f::Pkg.Types.var"#43#46"{String, Dates.DateTime, String}, at::String, pid::Int32; kwopts::Base.Pairs{Symbol, Int64, Tuple{Symbol}, NamedTuple{(:stale_age,), Tuple{Int64}}})
  @ FileWatching.Pidfile C:\hostedtoolcache\windows\julia\nightly\x64\share\julia\stdlib\v1.9\FileWatching\src\pidfile.jl:80
[7] #mkpidlock#6
  @ C:\hostedtoolcache\windows\julia\nightly\x64\share\julia\stdlib\v1.9\FileWatching\src\pidfile.jl:77 [inlined]
[8] write_env_usage(source_file::String, usage_filepath::String)
  @ Pkg.Types D:\a\Pkg.jl\Pkg.jl\src\Types.jl:492
[9] Pkg.Types.EnvCache(env::Nothing)
  @ Pkg.Types D:\a\Pkg.jl\Pkg.jl\src\Types.jl:353
[10] EnvCache
  @ D:\a\Pkg.jl\Pkg.jl\src\Types.jl:333 [inlined]
[11] top-level scope
  @ .\none:18
test atomicity of write_env_usage with 2 parallel processes: Test Failed at D:\a\Pkg.jl\Pkg.jl\test\pkg.jl:416
Expression: any(istaskfailed, tasks) == false
 Evaluated: true == false
```

DilumAluthge:
I'm not sure if the error is occuring in `Pidfile` code, or in the callback function that `Pkg` is passing to `Pidfile`.

DilumAluthge:
It looks like it's coming from the `tryopen_exclusive` function here?

https://github.com/JuliaLang/julia/blob/master/stdlib/FileWatching/src/pidfile.jl#L186-L193

DilumAluthge:
@vtjnash When `EACCES` is thrown, instead of rethrowing the error, would it make sense for `Pidfiles` to sleep for 10 seconds and then try again to acquire the lockfile?

DilumAluthge:
I've moved this to an issue: https://github.com/JuliaLang/Pkg.jl/issues/3058

PR #3057, opened 2022-04-15
[DO NOT MERGE] Dummy PR to trigger CI

DilumAluthge:


DilumAluthge:
@IanButterworth @vtjnash I am seeing frequent failures in the "test atomicity of write_env_usage" test set on Windows, but not on any other operating systems. See the CI on this PR for an example?

Is it cool if I allow that test to fail on Windows?

vtjnash:
If you want to delete a file on windows, you must rename it first. I generally do something polite like name it "delete this.please"

DilumAluthge:
In both this PR and https://github.com/JuliaLang/Pkg.jl/pull/3056, when I get an error about Windows trying to delete a file, the code that's trying to delete the file is in `FileWatching.Pidfile`. So presumably the "rename the file on Windows before deleting it" logic needs to be implemented in `FileWatching.Pidfile`?

DilumAluthge:
Actually, in this PR, the error is when trying to delete the file, but in https://github.com/JuliaLang/Pkg.jl/pull/3056, the error is a `ERROR: IOError: open("D:\\a\\Pkg.jl\\Pkg.jl\\test\\tmp\\logs\\manifest_usage.toml.pid", 1282, 292): permission denied (EACCES)` when trying to open a file.

In both cases, however, the code in question is in `FileWatching.Pidfile`.

vtjnash:
Don't blame your closure callbacks on Pidfile. I mean, you can, but it won't help.

Pidfile does know this trick, but I will take a look at the other one.

DilumAluthge:
Ohhhh, I see, the actual `rm` is in a callback we passed to `Pidfile`, not in `Pidfile` itself.

DilumAluthge:
Yeah https://github.com/JuliaLang/Pkg.jl/pull/3056 is the main one I'm interested in.

It's non-deterministic. Here is a `test (windows-latest, x64, pkg.julialang.org, nightly)` job that failed with `ERROR: IOError: open("D:\\a\\Pkg.jl\\Pkg.jl\\test\\tmp\\logs\\manifest_usage.toml.pid", 1282, 292): permission denied (EACCES)`:
- https://github.com/JuliaLang/Pkg.jl/runs/6033405710?check_suite_focus=true

But when I retried the job, the same `test (windows-latest, x64, pkg.julialang.org, nightly) ` job now passes:
- https://github.com/JuliaLang/Pkg.jl/runs/6033548145?check_suite_focus=true

vtjnash:
I apparently didn't call the proper deletion code on cleanup here and just used rm, so maybe there is a risk here of a race condition causing EACCES? https://github.com/vtjnash/Pidfile.jl/commit/30709e47d440f887d34d7425ee3daf54c780e22d

DilumAluthge:
I'm closing this PR just because it accomplished its goal (demonstrate that there is a nondeterministic failure on `Pkg#master`), but we can continue the discussion of how to fix the underlying issue.

PR #3059, opened 2022-04-18
Flush stdin before showing package install prompt

### LINKED ISSUE #3054: Prompt to install package should purge input buffer before reading from it
This just happened to me.  Note the newlines and the following REPL prompts (that immediately appeared after installing):

```
julia> using Foo


 │ Package Foo not found, but a package named Foo is available from a registry. 
 │ Install package?
 │   (@v1.7) pkg> add Foo 
 └ (y/n) [y]: n


   Resolving package versions...
   Installed Foo ─ v0.3.2
    Updating `~/.julia/environments/v1.7/Project.toml`
  [a1216acd] + Foo v0.3.2
    Updating `~/.julia/environments/v1.7/Manifest.toml`
  [a1216acd] + Foo v0.3.2
  [4722fa14] + Bar v1.1.1
  [d3707c4a] + Baz_jll v1.55.1+0
Precompiling project...
  1 dependency successfully precompiled in 6 seconds (33 already precompiled)

julia> 

julia> n
ERROR: UndefVarError: n not defined
```

IanButterworth:
Fixes #3054

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -699,7 +699,7 @@ function try_prompt_pkg_add(pkgs::Vector{Symbol})
             println(ctx.io, line)
         end
         printstyled(ctx.io, " └ "; color=:green)
-        flush(stdin)
+        readavailable(stdin) # ignore any keypresses before the prompt shows
         Base.prompt(stdin, ctx.io, "(y/n/o)", default = "y")
     catch err
         if err isa InterruptException # if ^C is entered
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -699,6 +699,7 @@ function try_prompt_pkg_add(pkgs::Vector{Symbol})
             println(ctx.io, line)
         end
         printstyled(ctx.io, " └ "; color=:green)
+        flush(stdin)
         Base.prompt(stdin, ctx.io, "(y/n/o)", default = "y")
     catch err
         if err isa InterruptException # if ^C is entered
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -692,7 +692,10 @@ function withreply(f, ans)
     p = Pipe()
     try
         redirect_stdin(p) do
-            @async println(p, ans)
+            Timer(5) do t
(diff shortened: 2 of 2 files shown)


IanButterworth:
Actually, the flush doesn't seem to be enough. I'm throwing a bunch of newlines in here while the prompt compiles, and they're not being ignored
```
julia> using Flux



 │ Package Flux not found, but a package named Flux is available from a registry. 
 │ Install package?
 │   (jl_fvEGFB) pkg> add Flux 
 └ (y/n/o) [y]: n
...
```

IanButterworth:
@vtjnash any idea why a flush doesn't work here? I'm trying to ignore any keypresses from before the prompt shows

vtjnash:
What do you think flush does?

IanButterworth:
Empty any buffered data?

If that's not the case (which I'm inferring from your response) then I assume this needs a `read` instead?

IanButterworth:
This doesn't work locally.

Without any keypresses it hangs until a return keypress here, so I assume `readavailable` is waiting (which seems odd)
```
julia> using Flux
 │ Package Flux not found, but a package named Flux is available from a registry. 
 │ Install package?
 │   (@v1.9) pkg> add Flux 
 └ 
```

and with returns pressed before the prompt it interprets the returns as a confirm:
```
julia> using Flux


 │ Package Flux not found, but a package named Flux is available from a registry. 
 │ Install package?
 │   (@v1.9) pkg> add Flux 
 └ (y/n/o) [y]: n
   Resolving package versions...
...
```

If what should be done is clear to you @vtjnash I'd appreciate that being pointed out. Thanks

vtjnash:
Ah, you want TCIFLUSH instead of this, which does a tty drain (awkwardly, tty got the names swapped from everyone else 🤦)

IanButterworth:
Ah ok. Am I right in saying that's not implemented in Base?

I did find this though https://github.com/kdheepak/TERMIOS.jl/blob/c7e23eb42cf9edf0b2b053a79956e38a741cd2c0/src/TERMIOS.jl#L670

IanButterworth:
Closing to tidy up given there is an issue to track

PR #3061, opened 2022-04-20
Ensure that test preferences are properly stacked

staticfloat:
When calling `Pkg.test("Foo")`, some information from the calling
environment is incorporated into the tests of `Foo`; for example, if the
calling environment places restrictions on the package versions of
`Bar`, a dependency of `Foo`, those restrictions will be honored when
resolving the dependencies of the test environment for `Foo`.
Similarly, we should honor preferences set within the calling
environment.  The easiest way to do this is to layer the test
environment onto the end of the current `LOAD_PATH` instead of wholesale
replacing it, as we were doing when copying preferences over for the
test environment.

With this change it is now viable to test alternate configurations of a
package without dev'ing it out and modifying it directly.  If a package
`Foo` uses preference to alter its behavior, previously you would need
to `dev` it out and insert an override into its `LocalPreferneces.toml`
(or `test/LocalPreferences.toml`, if it had a separate test project) in
order for the preference to be read during test time.  Now, one can
simply set the preference in an environment that has `Foo` added, then
run `Pkg.test("Foo")`, and as long as the package itself doesn't define
the preference within its own testing environment, it will pick up the
default from the environment.

We allow the package to override values coming in from the environment,
as there may be rather sensitive preferences that it is best to not
allow easy overriding from.  For these cases, there is of course still
the `dev`-and-edit method available for tests.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -992,12 +992,12 @@ function build_versions(ctx::Context, uuids::Set{UUID}; verbose=false)
         local build_project_override, build_project_preferences
         if isfile(projectfile_path(builddir(source_path)))
             build_project_override = nothing
-            with_load_path([builddir(source_path)]) do
+            with_load_path([builddir(source_path), Base.LOAD_PATH...]) do
                 build_project_preferences = Base.get_preferences()
             end
         else
             build_project_override = gen_target_project(ctx, pkg, source_path, "build")
-            with_load_path([projectfile_path(source_path)]) do
+            with_load_path([projectfile_path(source_path), Base.LOAD_PATH...]) do
                 build_project_preferences = Base.get_preferences()
             end
         end
@@ -1725,12 +1725,12 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
         local test_project_preferences, test_project_override
         if isfile(projectfile_path(testdir(source_path)))
             test_project_override = nothing
-            with_load_path(testdir(source_path)) do
+            with_load_path([testdir(source_path), Base.LOAD_PATH...]) do
                 test_project_preferences = Base.get_preferences()
             end
         else
             test_project_override = gen_target_project(ctx, pkg, source_path, "test")
-            with_load_path(projectfile_path(source_path)) do
+            with_load_path([projectfile_path(source_path), Base.LOAD_PATH...]) do
                 test_project_preferences = Base.get_preferences()
             end
         end
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /tmp
+/test/tmp
 *.jl.mem
 *.jl.cov
 *.jl.*.cov
--- a/test/sandbox.jl
+++ b/test/sandbox.jl
@@ -64,6 +64,7 @@ end
     # Preferences should be copied over into sandbox
     temp_pkg_dir() do project_path; mktempdir() do tmp
         copy_test_package(tmp, "Sandbox_PreservePreferences")
+        spp_uuid = UUID("3872bf94-3adb-11e9-01dc-bf80c7641364")
         Pkg.activate(joinpath(tmp, "Sandbox_PreservePreferences"))
 
         # Create fake test/Project.toml and test/LocalPreferences.toml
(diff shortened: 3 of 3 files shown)


staticfloat:
I think this may require https://github.com/JuliaPackaging/Preferences.jl/pull/37

src/Operations.jl
```diff
@@ -992,12 +992,12 @@ function build_versions(ctx::Context, uuids::Set{UUID}; verbose=false)
         local build_project_override, build_project_preferences
         if isfile(projectfile_path(builddir(source_path)))
             build_project_override = nothing
-            with_load_path([builddir(source_path)]) do
+            with_load_path([Base.LOAD_PATH..., builddir(source_path)]) do
```
vchuravy, line 995: Is the order here right? I would have expected the LOAD_PATH to come after not before

staticfloat: Yeah, I think you're right.

staticfloat:
@vchuravy is it possible for you to test this with MPIPreferences to ensure that it's actually sufficient for your usecase?

vchuravy:
Will try to give it a go today!

giordano:
Doesn't seem to work for me:
```
% ZES_ENABLE_SYSMAN=1 julia-master --project=/tmp --startup-file=no -q
julia> using MPIPreferences; MPIPreferences.use_jll_binary()
┌ Warning: The underlying MPI implementation has changed. You will need to restart Julia for this change to take effect
│   binary = "MPICH_jll"
└ @ MPIPreferences ~/.julia/packages/MPIPreferences/uArzO/src/MPIPreferences.jl:65

(tmp) pkg> test MPI
[...]
     Testing Running tests...
┌ Info: Running MPI tests
│   ArrayType = Array
│   nprocs = 4
│   MPIPreferences.abi = "MPICH"
└   MPIPreferences.binary = "MPICH_jll"
[...]

julia> using MPIPreferences; MPIPreferences.use_system_binary()
┌ Info: MPI implementation
│   libmpi = "libmpi"
│   version_string = "Open MPI v4.1.2, package: Open MPI builduser@arojas Distribution, ident: 4.1.2, repo rev: v4.1.2, Nov 24, 2021\0"
│   impl = "OpenMPI"
│   version = v"4.1.2"
└   abi = "OpenMPI"
┌ Warning: The underlying MPI implementation has changed. You will need to restart Julia for this change to take effect
│   libmpi = "libmpi"
│   abi = "OpenMPI"
│   mpiexec = "mpiexec"
└ @ MPIPreferences ~/.julia/packages/MPIPreferences/uArzO/src/MPIPreferences.jl:119

(tmp) pkg> test MPI
[...]
     Testing Running tests...
┌ Info: Running MPI tests
│   ArrayType = Array
│   nprocs = 4
│   MPIPreferences.abi = "MPICH"
└   MPIPreferences.binary = "MPICH_jll"
```

vchuravy:
```
(mpitemp) pkg> add MPI#master
...
  [da04e1cc] + MPI v0.20.0-dev `https://github.com/JuliaParallel/MPI.jl.git#master`
  [3da0fdf6] + MPIPreferences v0.1.1
...
┌ Info: Running MPI tests
│   ArrayType = Array
│   nprocs = 4
│   MPIPreferences.abi = "MPICH"
└   MPIPreferences.binary = "MPICH_jll"
...
julia> MPI.use_jll_binary("OpenMPI_jll")
┌ Warning: The underlying MPI implementation has changed. You will need to restart Julia for this change to take effect
│   binary = "OpenMPI_jll"
└ @ MPIPreferences ~/.julia/packages/MPIPreferences/uArzO/src/MPIPreferences.jl:65
(mpitemp) pkg> test MPI
     Testing Running tests...
┌ Info: Running MPI tests
│   ArrayType = Array
│   nprocs = 4
│   MPIPreferences.abi = "OpenMPI"
└   MPIPreferences.binary = "OpenMPI_jll"
```

So that looks right to me! I built Julia with

```diff
diff --git a/stdlib/Pkg.version b/stdlib/Pkg.version
index a595ed0213..c1f14c1982 100644
--- a/stdlib/Pkg.version
+++ b/stdlib/Pkg.version
@@ -1,4 +1,4 @@
 PKG_BRANCH = master
-PKG_SHA1 = 7676431149332ae400a805632082b12263c20269
+PKG_SHA1 = 5e5189b84523a41c67e3ade44fe774e69b66cea5
 PKG_GIT_URL := https://github.com/JuliaLang/Pkg.jl.git
 PKG_TAR_URL = https://api.github.com/repos/JuliaLang/Pkg.jl/tarball/$1
```

staticfloat:
Looks good Valentin!  With that, I'm going to merge!

omlins:
@staticfloat , @vchuravy : how about having in the loadpath a default location where the default preferences are looked up - **analogue to `@stdlib`**? It would be nice if we could be 100% sure that preferences defined in the Julia installation by cluster administrators are taken **always** into account if not explicitly and intentionally overwritten by the user-  **even in cases where the Pkg manager wants to create a "clean" environment** etc.

staticfloat:
> It would be nice if we could be 100% sure that preferences defined in the Julia installation by cluster administrators are taken always into account if not explicitly and intentionally overwritten by the user- even in cases where the Pkg manager wants to create a "clean" environment etc.

This can be done by having users define a default `JULIA_DEPOT_PATH` that includes a system-wide depot.  I don't suggest using `@stdlib` as this, since `@stdlib` has kind of a special meaning; it doesn't act like a true "depot".  Let's open a new issue about the best way for system administrators to provide defaults (preferences, packages, etc...) for their users.

omlins:
> I don't suggest using @stdlib as this, since @stdlib has kind of a special meaning; it doesn't act like a true "depot".

My suggestion is not to abuse `@stdlib` for preferences, but to create in Pkg.jl an analogue to `@stdlib` for preferences, e.g. `@stdprefs`, which points to the standard location of preferences in the Julia installation.

omlins:
@vchuravy , @staticfloat, what do you think about the above comment?

PR #3065, opened 2022-04-21
Remove stdin forwarding from Pkg.test

### LINKED ISSUE #3062: Pkg depends on `stdin` being open
https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_date/2022-04/15/AAindex.primary.log

We could possibly substitute `devnull` here transparently when the stream was previously exhausted and therefore now closed, but I don't know if that is ideal. In general though, passing `stdin` to a concurrent process is a terrible idea. Either you might want `RawFD(0)`, or just don't.

IanButterworth:
Fixes #3062


IanButterworth:
I confirmed locally to still be able to interrupt tests. The stdin forwarding removed here was a misunderstanding

PR #3073, opened 2022-04-29
[release-1.6] Make 1.6 manifest error clearer

IanButterworth:
Currently 1.6.6 can error in CI with
```
┌ Warning: The active manifest file has dependencies that were resolved with a different julia version (1.7.2). Unexpected behavior may occur.
└ @ ~/actions-runner/_work/Foo.jl/Foo.jl/Manifest.toml:0
ERROR: The manifest file you are using was most likely generated by a different version of Julia and is not compatible with this Julia version
```

which makes it seem like julia is refusing to use the manifest because of a version mismatch.

It's actually trying but erroring, which I think is meaningfully different.

This tries to make that clearer.

Untested locally


IanButterworth:
Looks like cloning https://github.com/JuliaLang/Example.jl is being rate limited (http err 429)

IanButterworth:
Bump @DilumAluthge 🙏

Issue #2664, opened 2021-07-06
Don't download artifact when overwritten via Overrides.toml


crstnbr:
Came up over at MKL.jl [here](https://github.com/JuliaLinearAlgebra/MKL.jl/issues/82).

Currently, if you use `Overrides.toml` to point Julia to a custom artifact Pkg will still eagerly download the original artifact (as indicated in the corresponding jll wrapper package) and then just not use it afterwards.

**Example: `IntelOpenMP_jll`:**

`Overrides.toml`
```toml
[1d5cc7b8-4909-519e-a0f8-d0f5ad9712d0]
IntelOpenMP = "/path/to/my/custom/artifact/" # contains lib/libiomp5.so and  lib/libiompstubs5.so
```
Pkg still downloads the original artifact:
```Julia
(@v1.8) pkg> add IntelOpenMP_jll
   Resolving package versions...
  Downloaded artifact: IntelOpenMP
    Updating `~/.julia/environments/v1.8/Project.toml`
  [1d5cc7b8] + IntelOpenMP_jll v2018.0.3+2
    Updating `~/.julia/environments/v1.8/Manifest.toml`
  [1d5cc7b8] + IntelOpenMP_jll v2018.0.3+2
Precompiling project...
  1 dependency successfully precompiled in 1 seconds (9 already precompiled)
```

```
➜  bauerc@ln-0001 .julia/artifacts  ls
947793e42b663bacd09f00d96aa96a47095f3b1c  Overrides.toml
➜  bauerc@ln-0001 .julia/artifacts  cd 947793e42b663bacd09f00d96aa96a47095f3b1c/lib/; ls
libiomp5.so  libiompstubs5.so
```

But doesn't use it (that's desired of course 😄):
```
julia> IntelOpenMP_jll.libiomp_path
"/path/to/my/custom/artifact/lib/libiomp5.so"
```

**Doesn't happen for `lazy` artifacts**

If an artifact is marked as `lazy = true` the unnecessary download of the original artifact doesn't happen, as desired. For example, `MKL_jll.jl` is [lazy](https://github.com/JuliaBinaryWrappers/MKL_jll.jl/blob/19da215d915da456e2ac4995c5d3f077f040ebdb/Artifacts.toml#L4).

**Desired behaviour:**
IMO, if an artifact is overwritten in an `Overrides.toml`, the original artifact should never be downloaded.

mkitti:
[...]

PR #3076, opened 2022-05-03
prevent overridden artifacts from downloading

KristofferC:
fixes https://github.com/JuliaLang/Pkg.jl/issues/2664

I think this needs a bit of thinking though. For example, if you use PackageCompiler you want that `ensure_artifact_installed` to download the artifact even if it is overridden. So it probably needs an option.

cc @carstenbauer, @staticfloat (who is on vacation AFAIU).

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -415,8 +415,8 @@ function ensure_artifact_installed(name::String, meta::Dict, artifacts_toml::Str
                                    quiet_download::Bool = false,
                                    io::IO=stderr_f())
     hash = SHA1(meta["git-tree-sha1"])
-
-    if !artifact_exists(hash)
+    overridden = query_override(hash) !== nothing
+    if !overridden && !artifact_exists(hash)
         errors = Any[]
         # first try downloading from Pkg server
         # TODO: only do this if Pkg server knows about this package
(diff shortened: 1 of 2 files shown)


staticfloat:
I think we should use the same option naming as the rest of the `Artifacts` functions, which is to add an `honor_overrides::Bool` parameter.  Note that some methods, such as `artifact_paths()` [default to `honor_overrides = true`](https://github.com/JuliaLang/julia/blob/7f84d46ce01b9400a61dd5660c7607a4629abad6/stdlib/Artifacts/src/Artifacts.jl#L206-L216), while other methods such as `verify_artifact()` [default to `honor_overrides = false`](https://github.com/JuliaLang/Pkg.jl/blob/ab65c7c56b2c5f0b605aa0e31df2b39c12237c07/src/Artifacts.jl#L113-L128).

In this case, I think if `honor_overrides` is `true`, then any overridden artifact is considered already downloaded.  If `honor_overrides` is `false`, then overriding is ignored and `ensure_artifact_installed()` will download the "normal" artifact.  We default to `honor_overrides=true`.

smartalecH:
Could we resurrect this? Or has an alternate solution already been implemented?

amontoison:
When we locally precompile an artifact with BinaryBuilder.jl, the url fields in the `Artifacts.toml` of the `*_jll` package are wrong and we can't reinstall / test the `.tar.gz` archives that contain the share library because `Pkg.jl` try to download the artifact even if it is overridden.

Issue #2915, opened 2022-01-03
Removing registries should only remove registry in DEPOT_PATH[1]


mkitti:
`registry rm` should not remove registries from further down the DEPOT_PATH stack. Depots further down the stack should be read-only to Julia.

```julia
julia> DEPOT_PATH
3-element Vector{String}:
 "/home/mkitti/anaconda3/envs/julia171_test/share/julia"
 "/home/mkitti/.julia"
 "/home/mkitti/anaconda3/envs/julia171_test/local/share/julia"

(@julia171_test) pkg> registry rm General
    Removing registry `General` from ~/anaconda3/envs/julia171_test/share/julia/registries/General.toml
    Removing registry `General` from ~/.julia/registries/General
```

The action described by the last line, should not have happened.

PR #3078, opened 2022-05-03
only remove registries from first entry in `DEPOT_PATH`

KristofferC:
Fxes https://github.com/JuliaLang/Pkg.jl/issues/2915

cc @mkitti

### DIFF
--- a/test/registry.jl
+++ b/test/registry.jl
@@ -145,15 +145,6 @@ end
         @test is_pkg_available(Example1)
         @test is_pkg_available(Example2)
 
-        # Behaviour with conflicting registry names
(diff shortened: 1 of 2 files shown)


test/registry.jl
```diff
@@ -145,15 +145,6 @@ end
         @test is_pkg_available(Example1)
         @test is_pkg_available(Example2)
 
-        # Behaviour with conflicting registry names
-        @test_throws PkgError pkgstr("registry up RegistryFoo")
```
KristofferC, line 149: @fredrikekre I think these tests are redundant with this since you can't have to registries with the same name in the same depot, and `registry rm` now only looks at the first depot. Do you agree?

Issue #1861, opened 2020-06-10
Add Special case `add` syntax for GitHub branches


Keno:
I can add packages on branches using syntax like (using aa/gitlab as an example, which is the first branch in this repo
```
]add Pkg#aa/gitlab
```
or
```
]add https://github.com/JuliaLang/Pkg.jl#/aa/gitlab
```
I think it would be convenient to also allow
```
]add https://github.com/JuliaLang/Pkg.jl/tree/aa/gitlab
```
as a special case and just rewrite it to the second form.
This is the URL that GitHub displays when you browse to a branch
in the Web UI. I sometimes want to just add the branch I'm currently
looking at (e.g. from a pull request), and it's more convenient to just
copy the URL than to have to reformat it manually.

PR #3079, opened 2022-05-04
pattern match in Pkg REPL for github branch URL

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/1861

### DIFF
--- a/src/REPLMode/argument_parsers.jl
+++ b/src/REPLMode/argument_parsers.jl
@@ -7,7 +7,18 @@ import ..isdir_nothrow, ..Registry.RegistrySpec, ..isurl
 Parser for PackageSpec objects.
 """
 function parse_package(args::Vector{QString}, options; add_or_dev=false)::Vector{PackageSpec}
-    args = PackageToken[PackageToken(pkgword) for pkgword in package_lex(args)]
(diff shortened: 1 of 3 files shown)


src/REPLMode/argument_parsers.jl
```diff
@@ -7,7 +7,18 @@ import ..isdir_nothrow, ..Registry.RegistrySpec, ..isurl
 Parser for PackageSpec objects.
 """
 function parse_package(args::Vector{QString}, options; add_or_dev=false)::Vector{PackageSpec}
-    args = PackageToken[PackageToken(pkgword) for pkgword in package_lex(args)]
+    words′ = package_lex(args)
+    words = String[]
+    for word in words′
+        if (m = match(r"https://github.com/(.*?)/(.*?)/tree/(.*?)$", word)) !== nothing
```
fredrikekre, line 13: Maybe also allow
```suggestion
        if (m = match(r"https://github.com/(.*?)/(.*?)/(?:tree|commit)/(.*?)$", word)) !== nothing
```

KristofferC: Nice

PR #3081, opened 2022-05-09
Transition to pkg mode from shell mode

skleinbo:
Regarding https://github.com/JuliaLang/julia/issues/42687

Switch to Pkg mode from shell mode.

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -604,6 +604,11 @@ function repl_init(repl::REPL.AbstractREPL)
         end
     )
     main_mode.keymap_dict = LineEdit.keymap_merge(main_mode.keymap_dict, keymap)
+    shell_mode_index = findfirst(x->startswith(x.prompt, REPL.SHELL_PROMPT), repl.interface.modes)
+    if !isnothing(shell_mode_index)
+        shell_mode = repl.interface.modes[shell_mode_index]
+        shell_mode.keymap_dict = LineEdit.keymap_merge(shell_mode.keymap_dict, keymap)
+    end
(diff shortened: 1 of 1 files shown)


src/REPLMode/REPLMode.jl
```diff
@@ -604,6 +604,11 @@ function repl_init(repl::REPL.AbstractREPL)
         end
     )
     main_mode.keymap_dict = LineEdit.keymap_merge(main_mode.keymap_dict, keymap)
+    shell_mode_index = findfirst(x->startswith(x.prompt, REPL.SHELL_PROMPT), repl.interface.modes)
```
skleinbo, line 607: It's either detecting the shell mode by its prompt, or relying on it being second in the list `repl.interface.modes`. I opted for the former.

LilithHafner:
Are there any shell commands that start with `]`?
This is new functionality so it should get tests.
I don't know enough about the internals to comment on the implementation.

src/REPLMode/REPLMode.jl
albheim: Relying on the name would not work with a customized prompt, e.g. [OhMyREPL](https://kristofferc.github.io/OhMyREPL.jl/latest/features/prompt/#Prompt-changing) provides methods for doing this for julia mode, but similar things would be easy to do for other modes.

Though maybe if one goes to the trouble of changing the prompt, it wouldn't be too much more effort to also update a keymap if it was desired. And it is not like it breaks, just that the map wouldn't be there. 

But it does seem a little brittle to rely on the name, seems like it would be nice to have a proper identifier for these things.

PR #3086, opened 2022-05-18
In `] st --outdated`, add a hint explaining that `] add Example@1.2.3` is the correct syntax for adding a package by version

DilumAluthge:


IanButterworth:
For review could you elaborate what this looks like and why it's needed? Thanks!

IanButterworth:
I assume `] add GLMakie#...` failed? Perhaps that's the place for a tip?

DilumAluthge:
In this case, `] add GLMakie#...` did error, because the tag in question didn't exist on the repo in question. But in general, it may or may not error. E.g. `] add Example#v0.5.3` doesn't error.

IanButterworth:
Yeah it seems to me like it's cleaner to not add what this PR currently adds but give a tip during errors for a failed `add Foo#1.2.3`. 
```
Error...
Info: It looks like you tried to install by version number. Use the form `add Foo@1.2.3` to add by version number, not by tag.
```

PR #3088, opened 2022-05-19
Allow `Registry.{add,update}()` to target specific depots

staticfloat:
Matching many other `Pkg` functions, we allow `update()` here to operate
on a custom set of depots, and `add()` to operate on a single specificed
depot.  This is useful for manipulating registries inside of depots that
are not currently on the `DEPOT_PATH`.


staticfloat:
Review with the whitespace filter on; I indented a function so it looks like a giant change.

KristofferC:
The other way is of course to do the `DEPOT_PATH` juggling but this is probably nicer.

staticfloat:
Yeah, I thought about that, but I'm always wary of manipulating global variables, because I just know someone is going to try and manipulate two depots in two different tasks.

PR #3091, opened 2022-05-24
Switch release-1.8 CI to run on 1.8

IanButterworth:
Standard practice, but also to debug this issue https://buildkite.com/julialang/julia-release-1-dot-8/builds/65#dbe68b90-6238-4f40-af1e-f3e7a2b286b9/483-1349 from https://github.com/JuliaLang/julia/pull/44789

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -35,8 +35,7 @@ jobs:
           - ""
           - "pkg.julialang.org"
         julia-version:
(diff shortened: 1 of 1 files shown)


IanButterworth:
Looks like 32-bit linux binaries aren't available for that julia branch.

But otherwise, the issue above isn't presenting itself, meaning it's likely a stdout redirect issue

PR #3096, opened 2022-05-29
Remove `Pkg.Types.STDLIBS_BY_VERSION` from Pkg, thus removing the need to constantly keep `STDLIBS_BY_VERSION` up to date

DilumAluthge:
Actually, `Pkg.Types.STDLIBS_BY_VERSION` is still there, but it is now an empty vector. Consumers (e.g. BinaryBuilder) must populate it.


test/new.jl
```diff
@@ -2790,9 +2794,15 @@ using Pkg.Types: is_stdlib
     @test is_stdlib(pkg_uuid, v"999.999.999")
     @test is_stdlib(pkg_uuid, v"0.7")
     @test is_stdlib(pkg_uuid, nothing)
+
+    empty!(Pkg.Types.STDLIBS_BY_VERSION)
+    @test is_stdlib(networkoptions_uuid)
+    @test_throws Pkg.Types.PkgError is_stdlib(networkoptions_uuid, v"1.6")
```
staticfloat, line 2800:
```suggestion
    empty!(Pkg.Types.STDLIBS_BY_VERSION)
    
    # Test that we can probe for stdlibs for the current version with no STDLIBS_BY_VERSION,
    # but that we throw a PkgError if we ask for a particular julia version.
    @test is_stdlib(networkoptions_uuid)
    @test_throws Pkg.Types.PkgError is_stdlib(networkoptions_uuid, v"1.6")
```

staticfloat (approved):
This seems good to me!  We can have a separate package that gets auto-updated outside of the Pkg repository, and use that from BB.  Thanks for doing this Dilum!

DilumAluthge:
@KristofferC Does this look good to you? The idea here is that by removing this list from Pkg, we don't need to constantly make PRs to Pkg to update the list.

IanButterworth:
Bump

IanButterworth:
Given this is lingering, I'll update the stdlibs

DilumAluthge:
Bump @KristofferC

DilumAluthge:
(I can resolve the merge conflicts before merging.)

DilumAluthge:
Bump @KristofferC

IanButterworth:
I'm going to update this and merge it.
The historical stdlib check just failed 5 minutes after a stdlib update PR was merged. It's too much overhead.

AFAIK @KristofferC isn't a user of this interface, so if it's ok by @staticfloat that seems ok to me

giordano:
> Consumers (e.g. BinaryBuilder) must populate it.

How exactly?  Getting a notice about these breaking changes would also be nice.

giordano:
> We can have a separate package that gets auto-updated outside of the Pkg repository, and use that from BB.

Maybe let's merge these PRs ***after*** such packages actually exist instead of breaking downstream packages?

IanButterworth:
Yeah, sorry, did this in the wrong order. Here's the package https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl

Being registered https://github.com/JuliaRegistries/General/pull/71291

PR #3097, opened 2022-06-01
remove usage of hardcoded stdlib list when `julia_version` == VERSION

KristofferC:


KristofferC:
I only want to do the `is_unregistered_stdlib` check (which looks up stdlibs from a hardcoded list) when `julia_version` has been explicitly passed in. That way, inconsistencies in that list only affects users who set custom `julia_version`s.

IanButterworth:
Ok. Makes sense

staticfloat (approved):
I actually think I originally had it working this way; it must have changed due to some refactoring at some point.

PR #3104, opened 2022-06-09
Expand precompilation docs

IanButterworth:
I realized that the error skipping wasn't explained, and that the docs could be expanded.


docs/src/environments.md
```diff
@@ -62,29 +62,61 @@ This new environment is completely separate from the one we used earlier. See [`
...
+  ◐ CoordinateTransformations
+  ◑ ArnoldiMethod
+  ◑ IntegralArrays
+  ◒ RegionTrees
+  ◐ ChangesOfVariables
+  ◓ PaddedViews
+```
+
+The exception is the `develop` command, which neither builds nor precompiles the package. When
+that happens is left up to the user to manually build and precompile the package.
```
IanButterworth, line 90:
```suggestion
that happens is left up to the user decide.
```

docs/src/environments.md
```diff
@@ -62,29 +62,61 @@ This new environment is completely separate from the one we used earlier. See [`
...
 
-To disable this auto-precompilation, set `ENV["JULIA_PKG_PRECOMPILE_AUTO"]=0`, after which precompilation can be triggered
-manually either serially via code loading
+If a given package version errors during auto-precompilation, Pkg will remember for the next time it
```
IanButterworth, line 92:
```suggestion
If a given package version errors during auto-precompilation, Pkg will remember for the next times it
```

docs/src/environments.md
```diff
@@ -62,29 +62,61 @@ This new environment is completely separate from the one we used earlier. See [`
...
+```
+
+The exception is the `develop` command, which neither builds nor precompiles the package. When
 that happens is left up to the user to decide.
 
-If a package that has been updated is already loaded in the session, the precompilation process will go ahead and precompile
-the new version, and any packages that depend on it, but will note that the package cannot be used until session restart.
+If a given package version errors during auto-precompilation, Pkg will remember for the following times it
+automatically tries, and will skip that package with a brief warning. Note that manual precompilation will
+always retry all packages i.e. `pkg> precompile`
```
IanButterworth, line 94:
```suggestion
automatically tries, and will skip that package with a brief warning. Manual precompilation can be used to force these packages to be retried, as `pkg> precompile` will
always retry all packages.
```

PR #3107, opened 2022-06-13
Use add instead of free for unregistered packages

mzaffalon:


mzaffalon:
Trying to `free` an unregistered package gave the unhelpful error message `ERROR: unable to free unregistered package ...`, but this comment by `@fredrikekre` in https://discourse.julialang.org/t/free-dev-ed-path/29065/6 saved me.

mzaffalon:
What was the reason to close this PR?

Issue #2570, opened 2021-05-20
Error from Pkg if no active project and loading a package that does not exist


KristofferC:
```jl
julia> using Foo
ERROR: no active project
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:68
  [2] find_project_file(env::Nothing)
    @ Pkg.Types ~/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:186
  [3] Pkg.Types.EnvCache(env::Nothing) (repeats 2 times)
    @ Pkg.Types ~/julia/usr/share/julia/stdlib/v1.7/Pkg/src/Types.jl:285
  [4] Pkg.Types.Context()
    @ Pkg.Types ./util.jl:471
  [5] try_prompt_pkg_add(pkgs::Vector{Symbol})
    @ Pkg.REPLMode ~/julia/usr/share/julia/stdlib/v1.7/Pkg/src/REPLMode/REPLMode.jl:654
  [6] #invokelatest#2
    @ ./essentials.jl:726 [inlined]
  [7] invokelatest
    @ ./essentials.jl:724 [inlined]
  [8] check_for_missing_packages_and_run_hooks(ast::Any)
    @ REPL ~/julia/usr/share/julia/stdlib/v1.7/REPL/src/REPL.jl:174
  [9] eval_user_input(ast::Any, backend::REPL.REPLBackend)
    @ REPL ~/julia/usr/share/julia/stdlib/v1.7/REPL/src/REPL.jl:145
 [10] repl_backend_loop(backend::REPL.REPLBackend)
    @ REPL ~/julia/usr/share/julia/stdlib/v1.7/REPL/src/REPL.jl:240
 [11] start_repl_backend(backend::REPL.REPLBackend, consumer::Any)
    @ REPL ~/julia/usr/share/julia/stdlib/v1.7/REPL/src/REPL.jl:225
 [12] run_repl(repl::REPL.AbstractREPL, consumer::Any; backend_on_current_task::Bool)
    @ REPL ~/julia/usr/share/julia/stdlib/v1.7/REPL/src/REPL.jl:358
 [13] run_repl(repl::REPL.AbstractREPL, consumer::Any)
    @ REPL ~/julia/usr/share/julia/stdlib/v1.7/REPL/src/REPL.jl:345
[...]

PR #3117, opened 2022-06-25
Exit missing package prompt early if no active project

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/2570

I used a catch because if we can't form a `Context()` it might just be good to abort.

Or is there a narrower but reliable check we could make?


IanButterworth:
@KristofferC does this seem reasonable to you?

KristofferC:
Yeah, seems OK.

PR #3122, opened 2022-06-28
warn & strip on accidental leading `]`

t-bltg:
I find myself often trying to switch to the `Pkg REPL` while already in the `Pkg REPL`, resulting in the command error-ing out.

This PR trims accidental leading `]` so that the following examples don't error:
```julia
(@v1.7) pkg> ]?
(@v1.7) pkg> ] ?
(@v1.7) pkg> ]st
(@v1.7) pkg> ] st
(@v1.7) pkg> ]st -m
(@v1.7) pkg> ] st -m
```
A warning message is also printed:
```julia
(@v1.7) pkg> ]st
┌ Warning: Removing leading `]`, which should only be used once to switch to pkg> mode
└ @ Pkg.REPLMode [...]/Pkg.jl/src/REPLMode/REPLMode.jl:230
```


fredrikekre:
See https://github.com/JuliaLang/Pkg.jl/pull/880

t-bltg:
Thanks for pointing #880, I didn't find it when searching the existing issues.

It would be disappointing not to discuss the relevance of this PR, because of a closed issue 4 years ago.

My point is that the issue described in https://github.com/JuliaLang/Pkg.jl/pull/3122#issue-1287364983 happens all the time, and it's time consuming to hit `up` + `home` + `delete` + `enter` to fix the (long) command.

> Having your key disappear into the void is also surprising.

We could add a warning if necessary.

> Also, to be consistent with other REPL modes I don't think we will do this.

We could also make the other REPL modes (Shell mode ?) consistent with this behavior, if there are no counter examples.

> I wanted to bring up a possible alternative for what could happen when pressing ] in the Pkg REPL: namely, exiting back to the standard REPL.

I don't think the `toggle` behavior described here would improve user experience, since (on some keyboards), hitting the backspace is faster than the combination of `AltGr` + the `]` key.

src/REPLMode/REPLMode.jl
```diff
@@ -226,6 +226,10 @@ end
 
 function tokenize(cmd::String)
     cmd = replace(replace(cmd, "\r\n" => "; "), "\n" => "; ") # for multiline commands
+    if startswith(cmd, ']')
+        @warn "Removing leading `]`"
```
IanButterworth, line 230: Perhaps?
```suggestion
        @warn "Removing leading `]`, which should only be used once to switch to pkg> mode"
```

t-bltg: Better, thanks.

IanButterworth (approved):
Seems like a friendly thing to do for new users 👍

PR #3124, opened 2022-07-06
Add the `PlatformEngines.request` function, and refactor `PlatformEngines.download` to just call `PlatformEngines.request`

DilumAluthge:


fredrikekre:
What is the reason for this?

DilumAluthge:
When debugging, I often find myself wanting to call `Downloads.request`, but masquerading as Pkg (i.e. with the exact same auth headers and metadata headers that Pkg would provide). However, currently there is no way to do so. I can call `Pkg.PlatformEngines.download`, which is the Pkg equivalent of `Downloads.download`, but there is no Pkg equivalent of `Downloads.request`.

src/PlatformEngines.jl
```diff
@@ -279,8 +297,13 @@ function download(
     else
         (total, now) -> nothing
     end
-    try
-        Downloads.download(url, dest; headers, progress)
+    return try
+        if verbose && !do_fancy
+            response = Downloads.request(url; output = dest, headers, verbose)
+        else
+            response = Downloads.request(url; output = dest, headers, progress)
```
staticfloat, line 304: It seems strange to me that we might not pass `verbose` through here, despite the fact that it could be `true`.

DilumAluthge: Yeah. The problem is that I'm trying to preserve the current behavior of `PlatformEngines.download`. Currently, if you pass `verbose` to `PlatformEngines.download`, it only gets used to determine whether or not we show a progress bar; it never gets passed through to `Downloads.download.` The `Downloads.download` call is `Downloads.download(url, dest; headers, progress)`.

PR #3134, opened 2022-07-14
Add `import Pkg` to manifest warning

jishnub:
Since `Pkg` is not imported by default in the REPL, this might help new users

### DIFF
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -285,7 +285,7 @@ end
 function write_manifest(manifest::Manifest, manifest_file::AbstractString)
     if manifest.manifest_format.major == 1
         @warn """The active manifest file at `$(manifest_file)` has an old format that is being maintained.
-            To update to the new format, which is supported by Julia versions ≥ 1.6.2, run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.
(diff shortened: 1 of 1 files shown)


src/manifest.jl
```diff
@@ -285,7 +285,7 @@ end
 function write_manifest(manifest::Manifest, manifest_file::AbstractString)
     if manifest.manifest_format.major == 1
         @warn """The active manifest file at `$(manifest_file)` has an old format that is being maintained.
-            To update to the new format, which is supported by Julia versions ≥ 1.6.2, run `Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.
+            To update to the new format, which is supported by Julia versions ≥ 1.6.2, run `using Pkg; Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.
```
IanButterworth, line 288:
```suggestion
            To update to the new format, which is supported by Julia versions ≥ 1.6.2, run `import Pkg; Pkg.upgrade_manifest()` which will upgrade the format without re-resolving.
```

IanButterworth (changes requested):
Good point!
Pkg usually suggests `import` here.

IanButterworth (approved):

PR #3141, opened 2022-07-19
Allow local default registry (by `path`)

t-bltg:
This PR allows to use a local registry (from `file://` protocol) instead of cloning from `https`, and is useful for setting up multiple julia for multiple users where the bandwidth is limited and a local clone of https://github.com/JuliaRegistries/General is available.

The implementation adds `JULIA_PKG_DEFAULT_REGISTRY_PATH`, `JULIA_PKG_DEFAULT_REGISTRY_URL` and `JULIA_PKG_DEFAULT_REGISTRY_LINKED` environment variables, in order to dynamically parametrize the default registries from environment, instead of hard-coding the url.

As a side effect, this PR simplifies the tests by extending the `withenv` constructs for setting up different default registries.

The docs have been enhanced to document the usage of the added environment variables.

All tests are passing locally on nightly.

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -512,6 +512,27 @@ julia> ENV["JULIA_PKG_SERVER"] = ""
 ""
 ```
 
+A local registry path can be used by setting the `JULIA_PKG_DEFAULT_REGISTRY_PATH` environment variables, it is
+best used in combination with `JULIA_PKG_SERVER` and `JULIA_PKG_DEFAULT_REGISTRY_URL`:
+
+```julia
+# use a .git repo instead of data from the default pkg server
+julia> ENV["JULIA_PKG_SERVER"] = "";
+
+# do not use default url https://github.com/JuliaRegistries/General [1]
+julia> ENV["JULIA_PKG_DEFAULT_REGISTRY_URL"] = "";
+
+# given path is a local clone of [1]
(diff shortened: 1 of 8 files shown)


IanButterworth:
I believe what you want is already achieved by using `Pkg.UPDATED_REGISTRY_THIS_SESSION[] = true` ?

DilumAluthge:
Combined with having a system-wide depot that has the General registry in it.

t-bltg:
> I believe what you want is already achieved by using Pkg.UPDATED_REGISTRY_THIS_SESSION[] = true ?

AFAIK, this doesn't work when `first(DEPOT_PATH)` is empty (fresh install).

 > Combined with having a system-wide depot that has the General registry in it.

Do you mean using `export JULIA_DEPOT_PATH='/foo:'` ?
AFAIK, if this is centralized, and the user doesn't have write access to `/foo`, `Pkg` fails.

DilumAluthge:
What if you do `export JULIA_DEPOT_PATH=':/foo:`? The registry should still be found as long as it is in one of the depots. But the user's depot will still be first.

t-bltg:
> What if you do export JULIA_DEPOT_PATH=':/foo? The registry should still be found as long as it is in one of the depots. But the user's depot will still be first.

Nice trick thanks, this solve my registry issues for multiple users. Let me see if I can find something else ...

t-bltg:
Thanks again for the comments ;)

PR #3143, opened 2022-07-22
Support the `JULIA_TEST_VERBOSE_LOGS_DIR` environment variable, which gives the ability to send logs to a file

gbaraldi:
The Pkg tests print a lot of things that can be useful, but on main julia CI they tend to make the ci logs a bit messy. So adding an easy option to send those logs to a file seems useful

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -29,7 +29,7 @@ if islogging
     logfile = joinpath(logdir, "Pkg.log")
     Pkg.DEFAULT_IO[] = open(logfile, "a")
 elseif hide_logs
-    Pkg.DEFAULT_IO[] = devnull
+    Pkg.DEFAULT_IO[] = BufferStream()
 else
     Pkg.DEFAULT_IO[] = stdout
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -28,6 +28,7 @@ islogging = logdir !== nothing
 if islogging
     logfile = joinpath(logdir, "Pkg.log")
     Pkg.DEFAULT_IO[] = open(logfile, "a")
+    @info "Pkg test output is being logged to file" logfile
 elseif hide_logs
     Pkg.DEFAULT_IO[] = Base.BufferStream()
 else
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -20,7 +20,7 @@ end
 ### Disable logging output if true (default)
 hide_logs = Pkg.get_bool_env("JULIA_PKG_TEST_QUIET", default="true")
 
-logfile = get(ENV, "JULIA_PKG_LOG_PATH", nothing)
+logdir = get(ENV, "JULIA_TEST_VERBOSE_LOGS_DIR", nothing)
 ### Send all Pkg output to a BufferStream if true (default)
 hide_stdoutstderr = hide_logs
 
@@ -54,8 +54,9 @@ Logging.with_logger(hide_logs ? Logging.NullLogger() : Logging.current_logger())
     end
 end
 
-if logfile !== nothing
-    open(logfile, "w") do io
+if logdir !== nothing
+    logfile = joinpath(logdir, "Pkg.log")
+    open(logfile, "a") do io
         write(io, Pkg.DEFAULT_IO[])
     end
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -31,6 +31,7 @@ if islogging
     @info "Pkg test output is being logged to file" logfile
 elseif hide_logs
     Pkg.DEFAULT_IO[] = Base.BufferStream()
+    @info "Pkg test output is silenced"
 else
     Pkg.DEFAULT_IO[] = stdout
 end
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -38,7 +38,7 @@ Pkg.REPLMode.minirepl[] = Pkg.REPLMode.MiniREPL() # re-set this given DEFAULT_IO
 
 include("utils.jl")
 
-redirect_stdio(stdout=Pkg.DEFAULT_IO[], stderr=Pkg.DEFAULT_IO[]) do
+Utils.io_to_buffer(Pkg.DEFAULT_IO[]) do
     @testset "Pkg" begin
                 @testset "$f" for f in [
                     "new.jl",
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -20,7 +20,7 @@ end
 ### Disable logging output if true (default)
 hide_logs = Pkg.get_bool_env("JULIA_PKG_TEST_QUIET", default="true")
 
-logfile = haskey(ENV,"JULIA_PKG_LOG_PATH") ? ENV["JULIA_PKG_LOG_PATH"] : nothing
+logfile = get(ENV, "JULIA_PKG_LOG_PATH", nothing)
 ### Send all Pkg output to a BufferStream if true (default)
 hide_stdoutstderr = hide_logs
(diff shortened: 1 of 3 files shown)


test/runtests.jl
```diff
@@ -20,6 +20,7 @@ end
 ### Disable logging output if true (default)
 hide_logs = Pkg.get_bool_env("JULIA_PKG_TEST_QUIET", default="true")
 
+logfile = haskey(ENV,"JULIA_PKG_LOG_PATH") ? ENV["JULIA_PKG_LOG_PATH"] : nothing
```
IanButterworth, line 23:
```suggestion
logfile = get(ENV, "JULIA_PKG_LOG_PATH", nothing)
```

test/runtests.jl
```diff
@@ -53,4 +54,10 @@ Logging.with_logger(hide_logs ? Logging.NullLogger() : Logging.current_logger())
     end
 end
 
+if logfile !== nothing
+    open(logfile, "w") do io
+        write(io, Pkg.DEFAULT_IO[])
+    end
+end
```
IanButterworth, line 61: huh, I didn't know this kind of thing worked.. Doesn't `Pkg.DEFAULT_IO[]` get consumed before this?

gbaraldi: It doesn't look like it's consumed by anything :)

DilumAluthge (changes requested):
Instead of making this specific to Pkg, I'd like to create a more general interface.

So, for example, how about we have an environment variable of the form `JULIA_TEST_VERBOSE_LOGS_DIR` that points to a directory. And then any test set in the Base Julia test suite can choose to write a verbose log file to that directory. So in the Pkg example, the `Pkg` test set will write its verbose logs to `joinpath(ENV["JULIA_TEST_VERBOSE_LOGS_DIR"], "Pkg.log")`. But for example, the `LibGit2/online` test set might also write a verbose log file to e.g. `joinpath(ENV["JULIA_TEST_VERBOSE_LOGS_DIR"], "LibGit2_online.log")`. And so on.

And then at the end of the Buildkite job, we do an upload of the form `buildkite-agent artifact upload ${JULIA_TEST_VERBOSE_LOGS_DIR:?}/*.log`. And this upload will be done unconditionally, whether the tests pass or fail.

cc: @staticfloat @Keno

test/runtests.jl
```diff
@@ -20,6 +20,7 @@ end
 ### Disable logging output if true (default)
 hide_logs = Pkg.get_bool_env("JULIA_PKG_TEST_QUIET", default="true")
 
+logfile = get(ENV, "JULIA_PKG_LOG_PATH", nothing)
```
DilumAluthge, line 23:
```suggestion
logdir = get(ENV, "JULIA_TEST_VERBOSE_LOGS_DIR", nothing)
```

test/runtests.jl
```diff
@@ -53,4 +54,10 @@ Logging.with_logger(hide_logs ? Logging.NullLogger() : Logging.current_logger())
     end
 end
 
+if logfile !== nothing
```
DilumAluthge, line 57:
```suggestion
if logdir !== nothing
    logfile = joinpath(logdir, "Pkg.log")
```

test/runtests.jl
```diff
@@ -53,4 +54,10 @@ Logging.with_logger(hide_logs ? Logging.NullLogger() : Logging.current_logger())
...
+    open(logfile, "w") do io
```
DilumAluthge, line 58:
```suggestion
    open(logfile, "a") do io
```

DilumAluthge (changes requested):

DilumAluthge:
See https://github.com/JuliaCI/julia-buildkite/pull/181 for the companion PR on the `JuliaCI/julia-buildkite` repository.

gbaraldi:
That sounds like a great idea @DilumAluthge

DilumAluthge (approved):

DilumAluthge:
@gbaraldi Could you test this out locally and confirm that:
1. The output does not get printed to stdout/stderr.
2. The log file is created.
3. The log file contains all the correct output.

gbaraldi:
Interestingly it doesn't work if a test errors

fredrikekre:
Write to the file directly instead of the BufferStream indirection?

gbaraldi:
The issue is that it seems to stop the logger from capturing stdout

gbaraldi:
Logging wasn't robust as just `redirect_stdio` so I switched to it.

DilumAluthge:
@gbaraldi It looks like CI is now broken. Also, the CI logs don't seem to show anything, so it looks like `JULIA_TEST_VERBOSE_LOGS_DIR` is now being ignored?

DilumAluthge:
`BufferStream` is not exported.

DilumAluthge (approved):

DilumAluthge:
Hmmm CI is unhappy again.

gbaraldi:
Apparently `redirect_stdio` doesn't like when you give it a BufferStream or an IOStream, which is kind of annoying

test/runtests.jl
```diff
@@ -20,36 +20,48 @@ end
...
+                    "sandbox.jl",
+                    "resolve.jl",
+                    "misc.jl",
+                    "force_latest_compatible_version.jl",
+                    "manifests.jl",
+                    ]
+                    @info "==== Testing `test/$f`"
+                    flush(Pkg.DEFAULT_IO[])
+                    include(f)
+                end
```
IanButterworth, line 62: Why the extra indent?

gbaraldi: It's how it was before, Not exactly sure why :)

test/runtests.jl
```diff
@@ -50,6 +60,7 @@ Logging.with_logger(hide_logs ? Logging.NullLogger() : Logging.current_logger())
             flush(Pkg.DEFAULT_IO[])
             include(f)
         end
+    islogging && close(Pkg.DEFAULT_IO[])
```
IanButterworth, line 63:
```suggestion
        islogging && close(Pkg.DEFAULT_IO[])
```

IanButterworth: I think you'll also want this in a try finally

DilumAluthge:
@gbaraldi @IanButterworth Is this ready to merge?

test/runtests.jl
```diff
@@ -20,35 +20,49 @@ end
...
-### Send all Pkg output to a BufferStream if true (default)
-hide_stdoutstderr = hide_logs
+logdir = get(ENV, "JULIA_TEST_VERBOSE_LOGS_DIR", nothing)
+### Send all Pkg output to a file called Pkg.log
+
+islogging = logdir !== nothing
+
+if islogging
+    logfile = joinpath(logdir, "Pkg.log")
+    Pkg.DEFAULT_IO[] = open(logfile, "a")
```
IanButterworth, line 30:
```suggestion
    Pkg.DEFAULT_IO[] = open(logfile, "a")
    @info "Pkg test output is being logged to file" logfile
```

IanButterworth (commented):
Perhaps add a log to inform in the main stdout?

DilumAluthge:
In the case where the log output is being suppressed (and not written to file), could we print an info message there as well?

gbaraldi:
bump

DilumAluthge:
Any idea why `x86_64-apple-darwin` keeps failing on Julia nightly? The failure doesn't seem related to this PR, but you never know.

DilumAluthge:
Okay it's all green now.

PR #3145, opened 2022-07-24
Remove dead code in Graph constructor

IanButterworth:
I noticed this appears to be unused

### DIFF
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -252,8 +252,6 @@ mutable struct Graph
             compat[uuid_julia] = Dict{VersionNumber,Dict{UUID,VersionSpec}}()
         end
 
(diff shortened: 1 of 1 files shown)


LilithHafner (approved):

LilithHafner:
I don't see much benefit to backporting this. Seems like more of a performance/readability feature than a bugfix (also maybe this code wasn't dead in 1.6/1.8)

IanButterworth:
Given 1.8 isn't yet out, I think it's a reasonable backport

However, you're right, it's used in 1.6 https://github.com/JuliaLang/Pkg.jl/blob/292c1e637a0a8e15b8b5fa8e6fe14dcc0b0cd6e5/src/Resolve/graphtype.jl#L263-L264

Issue #3147, opened 2022-07-29
`add` removes `dev`ed `path` when package is pinned


t-bltg:
```julia
julia> using Pkg
julia> Pkg.develop(path="[...]/Example.jl")  # local clone
julia> Pkg.pin("Example")
julia> Pkg.add("Example")
`Example` is pinned at `v0.5.3`: maintaining pinned version  # fine but ...
[...]
~ Example v0.5.3 `[...]/Example.jl` ⚲ ⇒ v0.5.3 ⚲  # ... expected to keep `dev`ed path
```

`Pkg.pin` docs:
Pin a package to the current version (or the one given in the PackageSpec) or to a certain git revision. **A pinned package is never updated.**

I read this a `if a package is developed & pinned, then nothing should change`. This is annoying in an update workflow, since it re-triggers time costly precompilation (by un-pinning, and re-pinning again).

PR #3148, opened 2022-07-29
keep `path` & `repo` when `pinned`

t-bltg:
Fix https://github.com/JuliaLang/Pkg.jl/issues/3147.

- [x] add tests

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1211,7 +1211,8 @@ function update_package_add(ctx::Context, pkg::PackageSpec, entry::PackageEntry,
             println(ctx.io, "`$(pkg.name)` is pinned at `v$(entry.version)`: maintaining pinned version")
         end
         return PackageSpec(; uuid=pkg.uuid, name=pkg.name, pinned=true,
-                           version=entry.version, tree_hash=entry.tree_hash)
+                           version=entry.version, tree_hash=entry.tree_hash,
+                           path=entry.path, repo=entry.repo)
     end
     if entry.path !== nothing || entry.repo.source !== nothing || pkg.repo.source !== nothing
         return pkg # overwrite everything, nothing to copy over
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -296,7 +296,7 @@ const build = API.build
     Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}}; io::IO=stderr, all_pkgs::Bool=false)
 
 Pin a package to the current version (or the one given in the `PackageSpec`) or to a certain
-git revision. A pinned package is neverautomatically updated: if `pkg` is tracking a path, or a repository,
+git revision. A pinned package is never automatically updated: if `pkg` is tracking a path, or a repository,
 those remain tracked.
 
 !!! compat "Julia 1.7"
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -296,7 +296,7 @@ const build = API.build
     Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}}; io::IO=stderr, all_pkgs::Bool=false)
 
 Pin a package to the current version (or the one given in the `PackageSpec`) or to a certain
-git revision. A pinned package is never updated: if `pkg` is tracking a path, or a repository,
+git revision. A pinned package is neverautomatically updated: if `pkg` is tracking a path, or a repository,
 those remain tracked.
 
 !!! compat "Julia 1.7"
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -312,8 +312,9 @@ const build = API.build
     Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}}; io::IO=stderr, all_pkgs::Bool=false)
 
 Pin a package to the current version (or the one given in the `PackageSpec`) or to a certain
-git revision. A pinned package is never automatically updated: if `pkg` is tracking a path, or a repository,
-those remain tracked.
+git revision. A pinned package is never automatically updated: if `pkg` is tracking a path,
+or a repository, those remain tracked in a frozen state). For example, if one wishes to update
+a pinned package in the current manifest, one must explicitly call `Pkg.pin` again.
 
 !!! compat "Julia 1.7"
     The `all_pkgs` kwarg was introduced in julia 1.7.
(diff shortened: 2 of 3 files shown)


test/pkg.jl
```diff
@@ -926,4 +926,24 @@ end
     @test_throws Pkg.Types.PkgError("Package PackageSpec(\n  path = test_packages/Example\n  version = *\n) has neither name nor uuid") ensure_resolved(Pkg.Types.Context(), Pkg.Types.Manifest(), [p])
 end
 
+@testset "Issue #3147"
```
IanButterworth, line 929:
```suggestion
@testset "Issue #3147" begin
```

t-bltg: Yeah, sorry, draft for now ;)

IanButterworth: Yep, just meaning to help the test run :)

src/Operations.jl
```diff
@@ -1148,7 +1148,8 @@ function update_package_add(ctx::Context, pkg::PackageSpec, entry::PackageEntry,
             println(ctx.io, "`$(pkg.name)` is pinned at `v$(entry.version)`: maintaining pinned version")
         end
         return PackageSpec(; uuid=pkg.uuid, name=pkg.name, pinned=true,
-                           version=entry.version, tree_hash=entry.tree_hash)
+                           version=entry.version, tree_hash=entry.tree_hash,
```
fredrikekre, line 1151: Is it really possible to pin everything here? What if I update the version in the Project.toml file in my local path? Perhaps if you pin a package that tracks a path it should only be the path that is pinned?

t-bltg: If you change the version in the `dev`ed path (**and pinned pkg**), then `Pkg.add("Example")` results in:
```
`Example` is pinned at `v0.5.3`: maintaining pinned version
   Resolving package versions...
    Updating `~/.julia/environments/v1.7/Project.toml`
  [7876af07] ~ Example v0.5.3 `[...]/Example.jl` ⚲ ⇒ v0.5.8 `[...]/Example.jl` ⚲
    Updating `~/.julia/environments/v1.7/Manifest.toml`
  [7876af07] ~ Example v0.5.3 `[...]/Example.jl` ⚲ ⇒ v0.5.8 `[...]/Example.jl` ⚲
```
I guess it is just consistent with `develop`, although the version is not `pinned` per se.

t-bltg:
Bump, anyone.

t-bltg:
@fredrikekre or @IanButterworth, could your review this PR or explain why it cannot be finalized ?

t-bltg:
Hit this again, bump.

t-bltg:
What can I do to get this PR reviewed / approved / merged ?
Thanks.

src/Pkg.jl
```diff
@@ -296,7 +296,8 @@ const build = API.build
     Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}}; io::IO=stderr, all_pkgs::Bool=false)
 
 Pin a package to the current version (or the one given in the `PackageSpec`) or to a certain
-git revision. A pinned package is never updated. To pin all dependencies set `all_pkgs=true`.
+git revision. A pinned package is never updated: if `pkg` is tracking a path, or a repository,
+those remain tracked.
```
IanButterworth, line 300: Did you mean to remove the info about pinning all?

Also the new sentence seems confusing. "Never updated" and "remain tracked" seem like they mean opposite things?

t-bltg: What is meant is that an update is not automatically updated. (If you look at the previous sentence "A pinned package is never updated." it was also confusing before).

DilumAluthge:
@IanButterworth and @fredrikekre Can you review this?

IanButterworth (commented):
Its not clear to me what the behavior of `up Foo` if Foo is pinned and tracking a repo here. The language around being pinned but still tracking is a little confusing. It might be good to explain that in the docstring a little more explicitly and add tests for `up` behavior?

t-bltg:
@IanButterworth, I've clarified again and added more tests for `Pkg.update` behaviour under `Pkg.pin`.

Would you care to review again so that we can move forward on this PR ? Thanks.

src/Pkg.jl
```diff
@@ -312,7 +312,9 @@ const build = API.build
     Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}}; io::IO=stderr, all_pkgs::Bool=false)
 
 Pin a package to the current version (or the one given in the `PackageSpec`) or to a certain
-git revision. A pinned package is never updated. To pin all dependencies set `all_pkgs=true`.
+git revision. A pinned package is never automatically updated: if `pkg` is tracking a path,
+or a repository, those remain tracked in a frozen state). For example, if one wishes to update
+a pinned package in the current manifest, one must explicitly call `Pkg.pin` again.
```
IanButterworth, line 317: Does you have to call `Pkg.unpin` before `Pkg.update("Foo")`?

Sorry for the questions, I just don't use pin on tracked repos like this

t-bltg: Yes, in my experiments calling `Pkg.add` and `Pkg.update` on a pinned package is a noop.

One must either `Pkg.unpin` and update or one can also call `Pkg.pin` again, in order to update the manifest when `Project.toml` has changed (this is my use case: I track a path which is dev'ed, and pinned).

This is what I expect from pinning a package: freeze it until I explicitly want to work on it.

IanButterworth: Ok perhaps `one must unpin, update then pin the package again`

t-bltg: Thanks, (hopefully) done in https://github.com/JuliaLang/Pkg.jl/pull/3148/commits/414f9a0f40d098c13a7f1eb1ce70700feecddf2e.

src/Pkg.jl
```diff
@@ -312,7 +312,9 @@ const build = API.build
     Pkg.pin(pkgs::Union{PackageSpec, Vector{PackageSpec}}; io::IO=stderr, all_pkgs::Bool=false)
 
 Pin a package to the current version (or the one given in the `PackageSpec`) or to a certain
-git revision. A pinned package is never updated. To pin all dependencies set `all_pkgs=true`.
+git revision. A pinned package is never automatically updated: if `pkg` is tracking a path,
+or a repository, those remain tracked in a frozen state). For example, if one wishes to update
+a pinned package in the current manifest, one must unpin, update then pin the package again.
```
IanButterworth, line 317:
```suggestion
or a repository, those remain tracked but will not update. 
To get updates from the origin path or remote repository the package must first be freed.
```

Final suggestion.. :)

IanButterworth (approved):

PR #3151, opened 2022-07-30
Add some tips to resolve conflict handling docs

IanButterworth:


docs/src/managing-packages.md
```diff
@@ -351,6 +351,15 @@ This message means that a package named `D` has a version conflict.
 Even if you have never `add`ed `D` directly, this kind of error can arise
 if `D` is required by other packages that you are trying to use.
 
+!!! note
+    When tackling these conflicts, first consider that the bigger an project gets, the more likely this is to happen.
```
t-bltg, line 355:
```suggestion
    When tackling these conflicts, first consider that the bigger a project gets, the more likely this is to happen.
```

docs/src/environments.md
```diff
@@ -5,6 +5,8 @@ The following discusses Pkg's interaction with environments. For more on the rol
 ## Creating your own projects
 
 So far we have added packages to the default project at `~/.julia/environments/v1.6`. It is however easy to create other, independent, projects.
+This approach has the benefit of allowing you to check-in a Project.toml (and even a Manifest.toml if you wish) into version control (e.g. git)
+to ensure that the code you're version controlling also has a reproducible environment.
```
DilumAluthge, line 9: Maybe it's just the way I'm reading it, but to me, this "if you wish" language seems to imply that if you check the `Project.toml` file into source control, you get a reproducible environment, even if you choose to not check in the `Manifest.toml`. Perhaps this sentence should be reworded to make it clear that you only get reproducible environments if the `Manifest.toml` file is also checked in.

IanButterworth: Well, a Project.toml alone with the typical type of compat entries should give a reproducible environment in a semver sense. I think the concept of reproducibility is a spectrum and I'm not sure we want to promote checking in manifests too much?

DilumAluthge: You could call that a "semver-compatible" environment, but I definitely wouldn't call that a reproducible environment. If you don't get the exact same package versions, then I don't think that's a reproducible environment.

I agree that we don't want to encourage people to check in manifests for packages (only for applications). So I would prefer that we just remove this sentence entirely.

IanButterworth: Fair point

```suggestion
```

docs/src/managing-packages.md
```diff
@@ -351,6 +351,15 @@ This message means that a package named `D` has a version conflict.
 Even if you have never `add`ed `D` directly, this kind of error can arise
 if `D` is required by other packages that you are trying to use.
 
+!!! note
+    When tackling these conflicts, first consider that the bigger a project gets, the more likely this is to happen.
+    Using targeted projects for a given task is highly recommended, and removing unused dependencies is a good first
+    step when hitting these issues.
+    For instance, a common pitfall is having more than a few packages in your default (i.e. `(@1.7)`) environment,
```
DilumAluthge, line 358:
```suggestion
    For instance, a common pitfall is having more than a few packages in your default (i.e. `(@$(Base.VERSION.minor).$(Base.VERSION.minor))`) environment,
```

DilumAluthge (approved):

docs/src/managing-packages.md
IanButterworth: Huh, I didn't know this worked in docs

docs/src/environments.md
```diff
@@ -5,6 +5,7 @@ The following discusses Pkg's interaction with environments. For more on the rol
 ## Creating your own projects
 
 So far we have added packages to the default project at `~/.julia/environments/v1.6`. It is however easy to create other, independent, projects.
+This approach has the benefit of allowing you to check-in a Project.toml (and even a Manifest.toml if you wish) into version control (e.g. git)
```
IanButterworth, line 8:
```suggestion
This approach has the benefit of allowing you to check-in a Project.toml, and even a Manifest.toml if you wish, into version control (e.g. git) alongside your code.
```

docs/src/managing-packages.md
DilumAluthge: Oh, I just realized that this is a Markdown file; I thought this was a docstring.

It would definitely work inside a docstring, but I'm not sure how to get it to work inside a Markdown file. @mortenpi or @fredrikekre would know.

It definitely would be nice to avoid hardcoding Julia versions in the docs.

mortenpi: If this was a block (e.g. a whole paragraph) then you could use an at-eval block, but unfortunately we can't really interpolate into an inline context like this. The best you could do would be something like 

````
using Markdown
```@eval
md"""
!!! note
    The whole admonition with interpolation $(string(Base.VERSION.minor))...
"""
````

Note: it would need the `string()` since the fields are unsigned.

DilumAluthge: Damn. @IanButterworth Want to revert my suggestion, since it won't work correctly?

PR #3155, opened 2022-08-04
Clarify RegistrySpec docstring

LilithHafner:
It is possible to think that `Registry` may be a standalone Pkg command with the previous docstring (e.g. a command that constructs or displays registry info).


LilithHafner:
Bump

PR #3163, opened 2022-08-07
make tip about upgradability slightly more informative

KristofferC:


IanButterworth:
Isn't that symbol also shown for deps held back due to Project compat?

KristofferC:
True, I'll try reformulate it.

KristofferC:
Tweaked the wording a bit. Also added "may" to the green `^` symbol to make it clearer that it is not 100% the package can be upgraded. cc @IanButterworth

src/Operations.jl
```diff
@@ -2107,17 +2107,17 @@ function print_status(env::EnvCache, old_env::Union{Nothing,EnvCache}, registrie
...
         end
         if !no_visible_packages_heldback && no_packages_upgradable
-            printpkgstyle(io, :Info, "Packages marked with $heldback_indicator have new versions available but cannot be upgraded.$tip", color=Base.info_color(), ignore_indent)
+            printpkgstyle(io, :Info, "Packages marked with $heldback_indicator have new versions available but compatibility constraints restrict them from upgrading.$tip", color=Base.info_color(), ignore_indent)
         end
         if !no_visible_packages_heldback && !no_packages_upgradable
-            printpkgstyle(io, :Info, "Packages marked with $upgradable_indicator and $heldback_indicator have new versions available, but those with $heldback_indicator cannot be upgraded.$tip", color=Base.info_color(), ignore_indent)
+            printpkgstyle(io, :Info, "Packages marked with $upgradable_indicator and $heldback_indicator have new versions available, but those with $heldback_indicator are restricted by compatibility constraints from upgrading.$tip", color=Base.info_color(), ignore_indent)
```
IanButterworth, line 2116:
```suggestion
            printpkgstyle(io, :Info, "Packages marked with $upgradable_indicator and $heldback_indicator have new versions available, but those with $heldback_indicator are known to be restricted by compatibility constraints from upgrading.$tip", color=Base.info_color(), ignore_indent)
```

KristofferC: Seems like 3 extra words with not so much extra info. If it is restricted, it is also "known to be restricted"? :P

IanButterworth: I was hoping to help the user infer that the ^ might not be upgradable because of an unknown compat limit, which they might come across. How the language currently is seems to be definitive that only ⌅ is limited.

KristofferC: I'll go with this for now and we can tweak it further later if people still get confused.

PR #3168, opened 2022-08-10
Document that `.pkg/` hooks are only allowed to use `Base`, `Artifacts`, `Libdl`, and `TOML`

DilumAluthge:


DilumAluthge:
Obviously, it will be better if we can figure out a way to enforce this limitation, but for now I think documenting it is sufficient.

PR #3169, opened 2022-08-11
fixes for moving out DelimitedFiles.jl from Julia

KristofferC:
The `julia_version` functionality is incompatible with moving out stdlibs so those tests have been commented out here.


KristofferC:
It was discussed that for now we will disable the tests for custom `julia_version` and bring them back in a while so we are not blocked on moving DelimitedFiles out. I will open an issue about this.

PR #3173, opened 2022-08-19
update the Pkg.test docs

KristofferC:
I didn't really like the

> If a manifest file exists and the `allow_reresolve`
keyword argument is set to `false`, the versions in the manifest file are used.
Otherwise a feasible set of packages is resolved and installed.

part, since it is only in the case of a conflict that new versions will be resolved. This tries to clarify that a bit.

Also, remove the hint about inlining since I don't think that is true anymore.


src/Pkg.jl
```diff
@@ -231,17 +231,16 @@ const update = API.up
...
 
-The tests are run by generating a temporary environment with only the `pkg` package
-and its (recursive) dependencies in it. If a manifest file exists and the `allow_reresolve`
-keyword argument is set to `false`, the versions in the manifest file are used.
-Otherwise a feasible set of packages is resolved and installed.
+The tests are run in a temporary environment that also includes the test specific dependencies
+of the package. The versions of dependencies in the current project are used for the
+test environment unless there is a compatibility conflict between the version of the dependencies and
+the test-specific dependencies. In that case, if `allow_reresolve` is `false` an error is thrown and
+if `allow_reresolve` is `true` a feasible set of versions of the dependencies is resolved and used.
```
IanButterworth, line 241: It'd be good to add a note about yanked versions here. It's caught me out before when I've expected a manifest to be observed.
```suggestion
The tests are run in a temporary environment that also includes the test specific dependencies
of the package. The versions of dependencies in the current project are used for the
test environment unless there is a compatibility conflict between the version of the dependencies and
the test-specific dependencies (or if the version in the manifest has been "yanked" from the registry and 
is thus marked as unavailable). In that case, if `allow_reresolve` is `false` an error is thrown and if 
`allow_reresolve` is `true` a feasible set of versions of the dependencies is resolved and used.
```

IanButterworth: bump @KristofferC the change here is `(or if the version in the manifest has been "yanked" from the registry and 
is thus marked as unavailable)`

PR #3183, opened 2022-08-30
fix invalidations from Static.jl

ranocha:
This is based on the following code:

```julia
julia> using Pkg; Pkg.activate(temp=true); Pkg.add("Static")

julia> using SnoopCompileCore; invalidations = @snoopr(using Static); using SnoopCompile

inserting !(::False) in Static at ~/.julia/packages/Static/sVI3g/src/Static.jl:427 invalidated:
...
                 77: signature Tuple{typeof(!), Any} triggered MethodInstance for Pkg.Types.Manifest(::Dict{String, Any}, ::String) (97 children)
...
```

I suggest the labels `latency` and `backport-1.8`.

### DIFF
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -132,8 +132,8 @@ end
 
 function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
     julia_version = haskey(raw, "julia_version") ? VersionNumber(raw["julia_version"]) : nothing
-    manifest_format = VersionNumber(raw["manifest_format"])
-    if !(in(manifest_format.major, 1:2)::Bool)
+    manifest_format = VersionNumber(raw["manifest_format"]::String)
+    if !in(manifest_format.major, 1:2)
         if f_or_io isa IO
             @warn "Unknown Manifest.toml format version detected in streamed manifest. Unexpected behavior may occur" manifest_format
         else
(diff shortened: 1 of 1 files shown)


ranocha:
The CI failure of `Run tests / test (ubuntu-latest, x86, nightly) (pull_request) ` does not seem to be related to this PR:

```
Downloading: Error During Test at /home/runner/work/Pkg.jl/Pkg.jl/test/platformengines.jl:132
118
119
  RequestError: HTTP/2 302 (Operation too slow. Less than 1 bytes/sec transferred the last 20 seconds) while requesting https://github.com/staticfloat/small_bin/raw/f1a92f5eafbd30a0c6a8efb6947485b0f6d1bec3/socrates.tar.gz
120
  Stacktrace:
```

src/manifest.jl
```diff
@@ -133,7 +133,7 @@ end
 function Manifest(raw::Dict, f_or_io::Union{String, IO})::Manifest
     julia_version = haskey(raw, "julia_version") ? VersionNumber(raw["julia_version"]) : nothing
     manifest_format = VersionNumber(raw["manifest_format"])
```
KristofferC, line 135:
```suggestion
    manifest_format = VersionNumber(raw["manifest_format"]::String)
```

Might also do it?

ranocha: Sounds reasonable. I changed it in cd830cb

ranocha:
@KristofferC Can we backport this to 1.8, please? I would like to see these fixes in Julia 1.8.1

Issue #3185, opened 2022-08-30
`stdin` closed during testing


t-bltg:
https://github.com/JuliaLang/Pkg.jl/pull/3065 broke [`Sixel.jl`](https://github.com/JuliaIO/Sixel.jl) testing.

We need `stdin` opened in order to query terminal properties via control sequences.

@johnnychen94

oxinabox:
This also broke DataDeps.jl oxinabox/DataDeps.jl#104
where it is a feature that the user is prompted to confirm that they do infact want to download data during the test.
(which they can bypass by setting an enviroment variable).

oxinabox:
also https://github.com/JuliaCI/PkgTemplates.jl/issues/370

IanButterworth:
I don't understand what happened here.

https://github.com/JuliaLang/Pkg.jl/pull/3065 removed stdin forwarding that was added in https://github.com/JuliaLang/Pkg.jl/pull/2933. Both of which were during development of 1.8.0

If this used to work before 1.8, the change & revert above isn't the reason for the breakage

IanButterworth:
1.7.3
https://github.com/JuliaLang/Pkg.jl/blob/0fae7809dbaa400d99cbe3a5b39c82332a1381d1/src/Operations.jl#L1630

1.8.0
https://github.com/JuliaLang/Pkg.jl/blob/63f4405d17e11decd2e5eb786cc491322b68c58c/src/Operations.jl#L1754

IanButterworth:
Ok. 

From the `Base.run` docs
> When wait is true (deffault), I/O streams are shared with the parent process. Use [pipeline](https://docs.julialang.org/en/v1/base/base/#Base.pipeline-Tuple{Any,%20Any,%20Any,%20Vararg{Any}}) to control I/O redirection.

@vtjnash in https://github.com/JuliaLang/Pkg.jl/issues/3062 you said that one shouldn't forward stdin, but isn't that what `run` was doing by default before 1.8?

vtjnash:
You should not forward and set wait=false. It is application dependent though if you need stdin

IanButterworth:
Ok. So would this be better?
```
p = run(ignorestatus(cmd), stdin, sandbox_ctx.io, stderr_f())
```

vtjnash:
The Pkg tests are run on a worker which does not have access to `stdin`

vtjnash:
Tests that need a TTY, should be BYO

[...]

PR #3186, opened 2022-08-30
add `forward_stdin` test option

t-bltg:
Fix https://github.com/JuliaLang/Pkg.jl/issues/3185.


fredrikekre:
Wouldn't it be better to setup your own stdin that you control instead?

t-bltg:
> Wouldn't it be better to setup your own stdin that you control instead?

How can I do that ? AFAIK there is now way to recover the `stdin` fd from the parent process.

PR #3189, opened 2022-09-01
fix invalidations due to `vect(::Nothing, Vararg{String}}` call

thchr:
It seems `projectfile_path` can return either a `String` or `Nothing` but here, it really must return a `String` (at least, there are no methods of `with_load_path` that accept a vector with `Nothing` entries).

Previously, this was causing a specialization of `vect(::Nothing, ::Vararg(String))` which then gave rise to a bunch of invalidations in StaticArrays due to its overloading of `getindex`. I think this should fix about ~100 of them.


thchr:
Gentle bump.

PR #3191, opened 2022-09-03
Test suite: restore the original depot path and load path at the end of the test suite

DilumAluthge:


DilumAluthge:
Motivation: as seen in [this CI log](https://buildkite.com/julialang/julia-master/builds/15431#0183048d-73ec-47e7-ba99-361a006a1882), the Pkg test suite currently does not restore the original values of `Base.DEPOT_PATH` at the end of the test suite.

fredrikekre:
Seems like a band-aid, can't we identify the problematic tests and fix those?

DilumAluthge:
Yeah, that's the better approach. I didn't have time to diagnose which tests are the problem, so I made this PR because we need to fix Base CI. But once the problematic tests are fixed, we can merge those fixes and remove the workaround in this PR.

PR #3200, opened 2022-09-14
improve type stability of `Command(statement::Statement)`

ranocha:
I improved type stability to fix some invalidations. This is based on the following code:
```julia
julia> using Pkg; Pkg.activate(temp=true); Pkg.add("ArrayInterface")

julia> using SnoopCompileCore; invalidations = @snoopr(using ArrayInterface); using SnoopCompile

...
inserting !(::Static.False) in Static at ~/.julia/packages/Static/sVI3g/src/Static.jl:427 invalidated:
...
```
The problem here is that `arguments` is only inferred as `Any`. I do not know whether we can assert something better (which may also fix the invalidations).

I suggest the labels `latency` and `backport-1.8`.

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -365,7 +365,7 @@ function Command(statement::Statement)::Command
     # arguments
     arg_spec = statement.spec.argument_spec
     arguments = arg_spec.parser(statement.arguments, options)
-    if !(arg_spec.count.first <= length(arguments) <= arg_spec.count.second)
+    if !((arg_spec.count.first <= length(arguments) <= arg_spec.count.second)::Bool)
         pkgerror("Wrong number of arguments")
     end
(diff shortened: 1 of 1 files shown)


src/REPLMode/REPLMode.jl
```diff
@@ -365,7 +365,7 @@ function Command(statement::Statement)::Command
     # arguments
     arg_spec = statement.spec.argument_spec
     arguments = arg_spec.parser(statement.arguments, options)
```
timholy, line 367: `default_parser` looks like it always returns `Vector{String}` but I don't know enough to promise that's true for all. `Command` (which gets called with `arguments`) has that field typed as `::Vector` with no specified eltype. Typing it `::AbstractVector` might be safe in general? I'd defer to others more knowledgeable about Pkg internals than I.

WRT `Command`, it might be a good idea to modify that field type declaration to something concrete.

PR #3204, opened 2022-09-18
Allow optional dependencies in the resolver

### LINKED ISSUE #3125: Resolver feature request: compatibility on optional dependencies
@carlobaldassi, we are thinking about possible ways of adding conditional dependencies. In almost all cases it seems there is a need for a package to be able to add compatibility to a package that does not have to end up getting installed.

So as an example:

```toml
[deps]
A = "..."
B = "..."

[weakdeps]
C = "..."

[compat]
A = "0.1"
B = "0.2"
C = "0.3"
```

In this case, we want to require `C` to have the `"0.3"` compat *if the package C ends up getting installed*.  But if it happens that C does not get installed, then that is also fine.

How hard do you think something like that would be to implement. And do you see any immediate problems conceptually by such a thing?

I guess one question is how much score you give C being installed vs the score of upgrading other versions (that might be on lower versions if C is installed). But the existing resolver already have to deal with questions like that I guess.

carlobaldassi:
This closes #3125.
A couple of things:
1. The first commit adds a `weak` attribute to `VersionSpec`, like in the idea by @IanButterworth. Indeed, in the resolver code, that is definitely the most straightforward way to do it. It required a couple of design decisions though. This is what I did, which could be debatable:
    * When intersecting a weak and a non-weak `VersionSpec`, the result is non-weak
    * It is not allowed to take a union of a weak and non-weak `VersionSpec` (the resolver code does not use `union` and I wasn't sure what, if any, would be a possible use case)
    * In printing, weak dependencies ranges show up with a `ʷ` character appended. This is maybe ugly, but in this way the information is preserved when showing log traces etc. An alternative proposal: let the ranges show as normal, letting them be less informative, but add a `; weak=true` string when using `show`.
  Opinions or further proposals welcome.
2. I have revised all the code and I don't think that any other part of the code should need to be changed, but I performed only limited testing (see the two new schemes in `"test/resolve.jl"), mainly for lack of time but also because it's not exactly easy to come up with interesting artificial scenarios. Hopefully, someone can try this out on more realistic cases, and even add to the test files.

### DIFF
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -268,7 +268,7 @@ mutable struct Graph
                 p1 == p0 && error("Package $(pkgID(pkgs[p0], uuid_to_name)) version $vn has a dependency with itself")
                 # check conflicts instead of intersecting?
                 # (intersecting is used by fixed packages though...)
-                req_p1 = get!(VersionSpec, req, p1)
+                req_p1 = get!(()->VersionSpec(;weak=true), req, p1)
                 req[p1] = req_p1 ∩ vs
             end
             # Translate the requirements into bit masks
@@ -276,10 +276,11 @@ mutable struct Graph
             req_msk = Dict{Int,BitVector}()
             for (p1, vs) in req
                 pv = pvers[p1]
-                req_msk_p1 = BitVector(undef, spp[p1] - 1)
+                req_msk_p1 = BitVector(undef, spp[p1])
                 @inbounds for i in 1:spp[p1] - 1
                     req_msk_p1[i] = pv[i] ∈ vs
                 end
+                req_msk_p1[end] = vs.weak
                 req_msk[p1] = req_msk_p1
             end
             extended_deps[p0][v0] = req_msk
@@ -319,13 +320,11 @@ mutable struct Graph
                 bmt = gmsk[p1][j1]
             end
 
-            for v1 = 1:(spp[p1]-1)
+            for v1 = 1:spp[p1]
                 rmsk1[v1] && continue
                 bm[v1, v0] = false
                 bmt[v0, v1] = false
             end
-            bm[end,v0] = false
-            bmt[v0,end] = false
         end
 
         req_inds = Set{Int}()
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -581,7 +581,7 @@ Finds a minimal collection of ranges as a `VersionSpec`, that permits everything
 `subset`, but does not permit anything else from the `pool`.
 """
 function range_compressed_versionspec(pool, subset=pool)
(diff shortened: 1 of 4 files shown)


KristofferC:
> The first commit adds a weak attribute to VersionSpec, like in the idea by @IanButterworth. Indeed, in the resolver code, that is definitely the most straightforward way to do it.

I think this is probably fine in practice. However, it is slightly inelegant in the sense that the majority of `VersionSpec`s will be non-weak but we still need to pay the cost of adding a `Bool` field to all instances (this doubles the size of the object, 64 extra bits due to padding, to add one extra bit of information). Generally, in cases like this, you would probably do it more Struct-of-Array style and have two collections, one with weak dependencies and one with strong ones.

Looking at the code here, it looks kind of difficult to do the change because the `.weak` property of the `VersionSpec` becomes quite entangled with the code. But then again, maybe this is more of a theoretical concern than a practice one.

carlobaldassi:
Another relatively simple option would be changing the `compat` structure that gets passed to the graph constructor, currently it is a `Dict{UUID,Dict{VersionNumber,Dict{UUID,VersionSpec}}}` but we could use a `Tuple{VersionSpec,Bool}` for the leaves instead.

KristofferC:
> Or we could add a parallel Dict{UUID,Dict{VersionNumber,Set{UUID}}} to denote the weak dependencies,

That was my thought for how to best structure it. But it is totally fine by me to go with this first and then do the optimization later if the need arises and if people are currently constrained for time.

carlobaldassi:
Ok I did that in #3205. Depending on which you like best, we can close the other PR (I don't have a preference).
Also, in case anyone wants to explore further designs, it should be easy to do by looking at the diff; the only requirement is that the weakness information about the dependency reaches at [this line](https://github.com/carlobaldassi/Pkg3.jl/blob/56f5fdc9419b3a9ddc10cac5f73dc6da05d55734/src/Resolve/graphtype.jl#L287), then the following changes to the `"graphtype.jl"` file are independent from the interface.

BioTurboNick:
Considering a different sense of "optional" that may impact this decision.

"uninstalled" is considered to be higher than all versions, it sounds like? I think it would be useful to also have the opposite concept, where "uninstalled" can be the lower than all versions.

Consider a package like ArrayInterface, which split into several packages in v6.

In theory, a package could be entirely compatible with the monolithic earlier versions of ArrayInterface and also the newest split packages; but currently there is no way to express that. Packages were forced to move up to v6 to use the new structure, which led to a cascade of conflicting compat changes across the ecosystem.

Is it better to have a separate section to indicate potentially-uninstalled dependencies, or could the `compat` version list handle it some way?

KristofferC:
I'll close this since we decided to go with #3205

PR #3205, opened 2022-09-19
Allow optional dependencies in the resolver (without changing VersionSet)

### LINKED ISSUE #3125: Resolver feature request: compatibility on optional dependencies

carlobaldassi:
This closes #3125 and it's an alternative implementation to #3204.
It does not change `VersionSet` so it does not have any of the associated difficulties there. 
It adds a keyword argument to the `GraphType` constructor instead ([here](https://github.com/carlobaldassi/Pkg3.jl/blob/56f5fdc9419b3a9ddc10cac5f73dc6da05d55734/src/Resolve/graphtype.jl#L245)). The type is `Dict{UUID,Dict{VersionNumber,Set{UUID}}}`.
@KristofferC this is what was discussed [here](https://github.com/JuliaLang/Pkg.jl/pull/3204#issuecomment-1251017819).
The building of the structure is a little cumbersome, an example can be seen [here](https://github.com/carlobaldassi/Pkg3.jl/blob/56f5fdc9419b3a9ddc10cac5f73dc6da05d55734/test/resolve_utils.jl#L76-L79).

This still needs testing, just as much as #3204 did.

### DIFF
--- a/src/Resolve/Resolve.jl
+++ b/src/Resolve/Resolve.jl
@@ -19,15 +19,18 @@ const Requires = Dict{UUID,VersionSpec}
 struct Fixed
     version::VersionNumber
     requires::Requires
+    weak::Set{UUID}
 end
-Fixed(v::VersionNumber) = Fixed(v, Requires())
+Fixed(v::VersionNumber, requires::Requires = Requires()) = Fixed(v, requires, Set{UUID}())
 
-Base.:(==)(a::Fixed, b::Fixed) = a.version == b.version && a.requires == b.requires
-Base.hash(f::Fixed, h::UInt) = hash((f.version, f.requires), h + (0x68628b809fd417ca % UInt))
+Base.:(==)(a::Fixed, b::Fixed) = a.version == b.version && a.requires == b.requires && a.weak == b.weak
+Base.hash(f::Fixed, h::UInt) = hash((f.version, f.requires, f.weak), h + (0x68628b809fd417ca % UInt))
 
 Base.show(io::IO, f::Fixed) = isempty(f.requires) ?
     print(io, "Fixed(", repr(f.version), ")") :
-    print(io, "Fixed(", repr(f.version), ",", f.requires, ")")
+    isempty(f.weak) ?
+    print(io, "Fixed(", repr(f.version), ",", f.requires, ")") :
+    print(io, "Fixed(", repr(f.version), ",", f.requires, ", weak=", f.weak, ")")
 
 
 struct ResolverError <: Exception
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -377,7 +377,7 @@ function add_reqs!(graph::Graph, reqs::Requires)
     return graph
 end
 
-function _add_reqs!(graph::Graph, reqs::Requires, reason)
+function _add_reqs!(graph::Graph, reqs::Requires, reason; weak_reqs::Set{UUID} = Set{UUID}())
     gconstr = graph.gconstr
     spp = graph.spp
     req_inds = graph.req_inds
@@ -392,7 +392,8 @@ function _add_reqs!(graph::Graph, reqs::Requires, reason)
             rvn = pvers[rp0][rv0]
             rvn ∈ rvs || (new_constr[rv0] = false)
         end
-        new_constr[end] = false
+        weak = rp ∈ weak_reqs
+        new_constr[end] = weak
         old_constr = copy(gconstr[rp0])
         gconstr[rp0] .&= new_constr
         reason ≡ :explicit_requirement && push!(req_inds, rp0)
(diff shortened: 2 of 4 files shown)


KristofferC:
Thanks, this is great to have!

IanButterworth:
Yeah. Thanks for both! It does seem like this is the better route.

KristofferC:
Question, does the `Fixed` input to the resolver also need to distinguish between weak/strong dependencies? I would guess so..

KristofferC:
@carlobaldassi, when adding support for the weak stuff in `Fixed`, I am hitting an assertion. It can be reprod with this Pkg commit (https://github.com/JuliaLang/Pkg.jl/commit/167cd57f93eb0daa854ed19b1b6a5a68618db937) using the following script:

```julia
using Pkg
using LibGit2
Pkg.activate(; temp=true)
tmp = mktempdir()
cd(tmp)
LibGit2.clone("https://github.com/IanButterworth/WeakDepsExamples", "WeakDepsExamples")
Pkg.develop(path="WeakDepsExamples/HasWeakDeps.jl")
```

The assertion is:

```julia
julia> Pkg.develop(path="WeakDepsExamples/HasWeakDeps.jl")
   Resolving package versions...
ERROR: AssertionError: !((gconstr[rp0])[end])
Stacktrace:
  [1] log_event_req!(graph::Pkg.Resolve.Graph, rp::Base.UUID, rvs::Pkg.Versions.VersionSpec, reason::Tuple{Base.UUID, Pkg.Resolve.ResolveLogEntry})
    @ Pkg.Resolve ~/JuliaPkgs/Pkg.jl/src/Resolve/graphtype.jl:678
  [2] _add_reqs!(graph::Pkg.Resolve.Graph, reqs::Dict{Base.UUID, Pkg.Versions.VersionSpec}, reason::Tuple{Base.UUID, Pkg.Resolve.ResolveLogEntry}; weak_reqs::Set{Base.UUID})
    @ Pkg.Resolve ~/JuliaPkgs/Pkg.jl/src/Resolve/graphtype.jl:399
  [3] _add_fixed!(graph::Pkg.Resolve.Graph, fixed::Dict{Base.UUID, Pkg.Resolve.Fixed})
    @ Pkg.Resolve ~/JuliaPkgs/Pkg.jl/src/Resolve/graphtype.jl:428
  [4] Pkg.Resolve.Graph(compat::Dict{Base.UUID, Dict{VersionNumber, Dict{Base.UUID, Pkg.Versions.VersionSpec}}}, compat_weak::Dict{Base.UUID, Dict{VersionNumber, Set{Base.UUID}}}, uuid_to_name::Dict{Base.UUID, String}, reqs::Dict{Base.UUID, Pkg.Versions.VersionSpec}, fixed::Dict{Base.UUID, Pkg.Resolve.Fixed}, verbose::Bool, julia_version::VersionNumber)
    @ Pkg.Resolve ~/JuliaPkgs/Pkg.jl/src/Resolve/graphtype.jl:342
  [5] deps_graph(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, uuid_to_name::Dict{Base.UUID, String}, reqs::Dict{Base.UUID, Pkg.Versions.VersionSpec}, fixed::Dict{Base.UUID, Pkg.Resolve.Fixed}, julia_version::VersionNumber)
    @ Pkg.Operations ~/JuliaPkgs/Pkg.jl/src/Operations.jl:537
  [6] resolve_versions!(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, julia_version::VersionNumber)
```

If I disable the assertion, things work as I expect for my limited testing but not sure if something else will blow up at some point.

The only thing that commit does is that it passes a ` Set([UUID("6fe1bfb0-de20-5000-8ca7-80f57d26f881")]))` to the `Fixed` constructor for the package being developed (because that is a weak dep for the package: https://github.com/IanButterworth/WeakDepsExamples/blob/adb19bec21d763566c4419bf1d459fcf130ee9ae/HasWeakDeps.jl/Project.toml#L9).

Any ideas?

carlobaldassi:
@KristofferC you're right, that assert is no longer correct. The good news is, it can be safely disabled just like you did in the next commit, or one can keep it around writing `reason ≡ :explicit_requirement && @assert !gconstr[rp0][end]` as I did in this branch.

DilumAluthge:
Are the `@assert`s here being used for debugging purposes only, or are they necessary to ensure correctness? If the latter, they should be switched to explicit errors, right? Since we'll eventually disable `@assert` by default in Julia.

carlobaldassi:
They're purely for debugging.

carlobaldassi:
I have pushed a better fix now, that also fixes the logging (that assert was in fact signalling an assumption about the code below after all...)

PR #3207, opened 2022-09-20
misc fixes from JET report

thchr:
This is essentially a grab bag of fixes found by playing with JET.jl's `report_package` utility.

There's a lot more raised in the report, but I figured I might as well turn what I had into a PR.

Mostly, the fixes involve narrowing type information manually.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -60,8 +60,8 @@ end
 
 function package_info(env::EnvCache, pkg::PackageSpec, entry::PackageEntry)::PackageInfo
     git_source = pkg.repo.source === nothing ? nothing :
-        isurl(pkg.repo.source) ? pkg.repo.source :
-        Operations.project_rel_path(env, pkg.repo.source)
+        isurl(pkg.repo.source::String) ? pkg.repo.source::String :
+        Operations.project_rel_path(env, pkg.repo.source::String)
     info = PackageInfo(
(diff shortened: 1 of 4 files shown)


src/REPLMode/argument_parsers.jl
```diff
@@ -100,7 +100,7 @@ function parse_package_args(args::Vector{PackageToken}; add_or_dev=false)::Vecto
         elseif modifier isa Rev
             pkg.rev = modifier.rev
         else
-            pkgerror("Package name/uuid must precede subdir specifier `[$arg]`.")
+            pkgerror("Package name/uuid must precede subdir specifier `$args`.")
```
thchr, line 103: Here, the issue was that there isn't any `arg` variable. I wasn't completely sure of what the intended argument should have been here - it could've also been `args[1]`?

src/API.jl
```diff
@@ -1560,14 +1563,13 @@ function instantiate(ctx::Context; manifest::Union{Bool, Nothing}=nothing,
     # Handling packages tracking repos
     for pkg in pkgs
         pkg.repo.source !== nothing || continue
+        repo_source = pkg.repo.source::String
```
KristofferC, line 1566: Move this up and use in the `!== nothing` check?

thchr: Done 👍

PR #3211, opened 2022-09-26
Make CURLOPT_LOW_SPEED_TIME configurable

jgiannuzzi:
This PR exposes the `CURLOPT_LOW_SPEED_TIME` through `JULIA_PKG_LOW_SPEED_TIME` so that large packages that take a long time to download on a proxy can still be installed.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -11,6 +11,9 @@ using Base.BinaryPlatforms, p7zip_jll
 
 export verify, unpack, package, download_verify_unpack
 
+const DOWNLOADER_LOCK = ReentrantLock()
+const DOWNLOADER = Ref{Downloads.Downloader}()
+
 const EXE7Z_LOCK = ReentrantLock()
 const EXE7Z = Ref{String}()
 
@@ -282,7 +285,15 @@ function download(
         (total, now) -> nothing
     end
     try
-        Downloads.download(url, dest; headers, progress)
+        lock(DOWNLOADER_LOCK) do
+            if !isassigned(DOWNLOADER)
+                downloader = Downloads.Downloader()
+                timeout = parse(Int, get(ENV, "JULIA_PKG_LOW_SPEED_TIME", "20"))
+                downloader.easy_hook = (easy, info) -> Downloads.Curl.setopt(easy, Downloads.Curl.CURLOPT_LOW_SPEED_TIME, timeout)
+                DOWNLOADER[] = downloader
+            end
+        end
+        Downloads.download(url, dest; headers, progress, downloader=DOWNLOADER[])
     finally
         do_fancy && end_progress(io, bar)
     end


StefanKarpinski:
Thanks for the PR. A couple of comments. If you're going to change a setting like `CURLOPT_LOW_SPEED_TIME` for Pkg downloads, wouldn't you also want to change it for all downloads? The only situation where I've seen `CURLOPT_LOW_SPEED_TIME` be an issue is when there's a MITM firewall/proxy that waits to get an entire large download and scan it before serving any content back to the client. If that's the case, this would be a problem for large files downloaded with `Downloads.download` as well and not just for Pkg downloads.

It would also be better to figure out a way to avoid this being a problem in the first place. Sure, it's good to have a workaround, but if you've got this problem, then surely many others are also silently suffering from this and it would be best to not need a workaround in the first place. The intent of the timeout settings is to prevent downloads from just hanging when servers are non-responsive. You can't just set [`CURLOPT_TIMEOUT`](https://curl.se/libcurl/c/CURLOPT_TIMEOUT.html) because there's no fixed amount of time that is too much for a download. Instead, we put a timeout on establishing a connection with [`CURLOPT_CONNECTTIMEOUT`](https://curl.se/libcurl/c/CURLOPT_CONNECTTIMEOUT.html) and make sure that the server isn't just completely stalled out by making sure it returns *some* data every 30 seconds with [`CURLOPT_LOW_SPEED_TIME`](https://curl.se/libcurl/c/CURLOPT_LOW_SPEED_TIME.html). That works great when the client connects to a normal HTTP server since large downloads are served incrementally. It fails in the MITM firewall/proxy case bacause the proxy waits until the entire file is downloaded before sending any data to the client.

Maybe we can figure out a better alternative to `CURLOPT_LOW_SPEED_TIME` for preventing stuck servers from leaving connections hanging indefinitely? I have some ideas, but I'd really need to test with an actual MITM firewall/proxy setup in order to figure out something that works. Can you tell me any more about your setup? Do you know what kind of firewall you're going through? One possibility is detecting that we're going through a MITM proxy and automatically disabling the ``CURLOPT_LOW_SPEED_TIME` setting, but I'd need to figure out the logic for detecting that. Another possibility is using the `Content-Length` header to figure out a `CURLOPT_TIMEOUT` setting that would work. I don't know if download settings can be adjusted after a download has begun, so I'd have to test that out. If not, then it might be necessary to wait until a download fails and automatically retry if it meets certain criteria that suggest it might have failed for this reason. Yet another possibility is using [range requests](https://developer.mozilla.org/en-US/docs/Web/HTTP/Range_requests) to download large files in chunks. However, that depends on both the origin server and the proxy supporting range requests, which isn't always a given.

KristofferC:
Imo, there should either be support for providing your own `Downloader` object or this should be set on the global downloader. But adding an environment variable into Pkg that just replicates the options for libcurl will be too verbose / repetitive.

PR #3223, opened 2022-10-10
Allow users to set the max number of precompile tasks by setting `JULIA_MAX_NUM_PRECOMPILE_TASKS="<=n"`

DilumAluthge:
I have a hetereogeneous collection of machines. Across all of these machines, I want to enforce a limit of no more than `10` precompile tasks. However, some machines are small and only have e.g. four CPU threads; on those machines, instead of using `10` precompile tasks, I want to be using `5` tasks, which is what Pkg would decide to use if I didn't set any environment variables.

Unfortunately, with the existing `JULIA_NUM_PRECOMPILE_TASKS` environment variable, if I set `JULIA_NUM_PRECOMPILE_TASKS="10"`, then Pkg will use `10` precompile tasks even on the small machines.

Therefore, this PR adds the ability to set e.g. `JULIA_NUM_PRECOMPILE_TASKS="<=10"`.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1079,9 +1079,22 @@ function precompile(ctx::Context, pkgs::Vector{String}=String[]; internal_call::
     # Windows sometimes hits a ReadOnlyMemoryError, so we halve the default number of tasks. Issue #2323
     # TODO: Investigate why this happens in windows and restore the full task limit
     default_num_tasks = Sys.iswindows() ? div(Sys.CPU_THREADS::Int, 2) + 1 : Sys.CPU_THREADS::Int + 1
-    default_num_tasks = min(default_num_tasks, 16) # limit for better stability on shared resource systems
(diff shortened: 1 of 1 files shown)


IanButterworth:
Just so I follow, you're saying `Sys.CPU_THREADS` is wrong on those systems?

DilumAluthge:
No, `Sys.CPU_THREADS` is correct on all of the systems. It's just that on the larger systems, I don't want to launch 16 tasks; instead, I want the upper limit to be 12 or 10 or something.

DilumAluthge:
This PR looks like it's stale.

KristofferC:
It kind of feels like this should be a general thing among the different julia thread environment variables and not specific to Pkg? If e.g. Base adds support for it then it would be obvious to add it here as well. Without Base having support for it, it isn't super clear to me that having special syntax just for this case is worth it.

PR #3229, opened 2022-10-19
add the ability to add compat on "weak dependencies"

KristofferC:
This is the part of the Pkg changes of the weak deps / glue code saga that is independent of the actual mechanism of how the conditional dependency gets loaded. This is useful on its own, for example with Requires.jl.

To properly test this we need to either make our own registry that has a weak dependency, or add some packages to the General registry.

It would also be good to start working on Registrator support for this feature (cc @GunnarFarneback).

### DIFF
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -18,6 +18,27 @@ The format of the version specifier is described in detail below.
 !!! info
     Use the command `compat` to edit the compat entries in the Pkg REPL, or manually edit the project file.
 
+## Compatibility on "weak dependencies"
+
+!!! compat "Julia 1.9"
+    This feature requires Julia 1.9
+
+A weak dependency is a package that you do not depend on but still want to be able to add
+compatibility constraints on. This is for example useful when using Requires.jl to load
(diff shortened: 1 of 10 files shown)


docs/src/compatibility.md
```diff
@@ -18,6 +18,27 @@ The format of the version specifier is described in detail below.
...
+[weakdeps]
+Example = "7876af07-990d-54b4-ab0e-23690620f79a"
+
+[compat]
+Example = "0.5"
+```
+
+Possible use cases for this is when using Requires.jl and one wants to add a compat constrained
+on the "conditional dependency" that Requires.jl tends to be used for.
+
```
GunnarFarneback, line 41:
```suggestion

Weak dependencies are ignored by Julia before 1.9. A dependency can be listed in both the `deps` and `weakdeps` sections. In that case it acts as a weak dependency for Julia 1.9 and later and as a normal dependency for older Julia versions.
```

KristofferC:
This didn't make it into 1.9 so closing this since we will have the full thing in 1.10 anyway.

Issue #3212, opened 2022-09-30
Artifact hash errors on nightly 64-bit


KristofferC:
https://github.com/JuliaLang/Pkg.jl/actions/runs/3157873266/jobs/5139255214

```
Artifacts.toml Utilities: Test Failed at /home/runner/work/Pkg.jl/Pkg.jl/test/artifacts.jl:237
  Expression: artifact_hash("foo_txt", artifacts_toml) == hash2
   Evaluated: SHA1("5679c650921bf0f891224295dd95572a5e9db8da") == SHA1("09795a47b3e5d3bcc4b44beace0072e1c3eb8aab")
```

Not sure what changed.

KristofferC:
Also fails on https://github.com/JuliaLang/Pkg.jl/pull/3178 which I just updated.

DilumAluthge:
Interesting. They're still passing on Buildkite... what's different between Buildkite and GitHub Actions? The hardware, maybe?

KristofferC:
Tests passed on https://github.com/JuliaLang/Pkg.jl/pull/3205 now... Something intermittent with GitHub serving bad stuff maybe..?

vtjnash:
We got a SHA failure in rr https://buildkite.com/julialang/julia-master/builds/17283#0183faf8-d50b-41b7-bffb-10f237f7b6c0. Not sure how useful it is though

gbaraldi:
I'm able to reproduce this locally on x86 linux. I believe we might have a data race in this test.
```julia
using Pkg.Artifacts
using Test
using TOML
using Pkg
path = mktempdir()
function torture()
    while true
        hash = create_artifact() do path
        open(joinpath(path, "foo.txt"), "w") do io
            print(io, "hello, world!")
        end
    end
    artifacts_toml = joinpath(path, "Artifacts.toml")
    bind_artifact!(artifacts_toml, "foo_txt", hash;  force=true)
    hash2 = create_artifact() do path
        open(joinpath(path, "foo.txt"), "w") do io
            print(io, "goodbye, world!")
        end
    end
    @test_throws ErrorException bind_artifact!(artifacts_toml, "foo_txt", hash2)
    # sleep(0.1)
    @test artifact_hash("foo_txt", artifacts_toml) == hash
    # sleep(0.1)
    bind_artifact!(artifacts_toml, "foo_txt", hash2; force=true)
    @test artifact_hash("foo_txt", artifacts_toml) == hash2
end
end

torture()
[...]

PR #3234, opened 2022-10-24
Add sleeps when overwriting the TOML files

gbaraldi:
Checking the hash of a file that was just written to might lead to wrong results, this test has a data race that leads to weird results, adding a sleep will probably help it. It seems it was "broke" it https://github.com/JuliaLang/julia/pull/46944/files. 
@vtjnash 
Should hopefully fix https://github.com/JuliaLang/Pkg.jl/issues/3212


gbaraldi:
This breaks the other test. It seems overwriting manifest hashes is a bit weird.

gbaraldi:
Superceded by https://github.com/JuliaLang/Pkg.jl/pull/3235

PR #3235, opened 2022-10-24
Use atomic writes when binding artifacts

staticfloat:
**EDIT**: No longer doing this, Jameson has Shown Me The Way™

We need to clear this cache when modifying the TOML file.  This should fix https://github.com/JuliaLang/Pkg.jl/issues/3212.

It might be nice to expose a `Base.drop_parsed_toml(path::String)` API so that I don't have to manually reach into `TOML_CACHE.d` here like this.


src/Artifacts.jl
```diff
@@ -226,6 +226,8 @@ function bind_artifact!(artifacts_toml::String, name::String, hash::SHA1;
         open(artifacts_toml, "w") do io
             TOML.print(io, artifact_dict, sorted=true)
         end
```
vtjnash, line 228: Ah, these lines are buggy, causing this failure. Pkg must never open an existing file for writing, as it causes problems for the filesystem.

vtjnash: e.g. https://github.com/JuliaLang/julia/issues/36664

vtjnash: and https://github.com/JuliaLang/julia/issues/39892

KristofferC: I think https://github.com/JuliaLang/julia/pull/46944 removed the check that was intended to caught things like this.

vtjnash: It did, but that check was fundamentally flawed, and didn't actually check for things like this, it merely checked for attempts to *test for* this problem.

vtjnash (approved):
Excellent

IanButterworth:
Made an issue for the MacOS error which appears unrelated but has been sticking around https://github.com/JuliaLang/Pkg.jl/issues/3238

Also the historical stdlib check is now removed as of #3096

simeonschaub:
@staticfloat This broke support for relative `Artifact.toml` paths in https://github.com/simeonschaub/ArtifactUtils.jl/issues/19. Should this be fixed here or was this never a supported use case and ArtifactUtils should be updated?

staticfloat:
A patch like yours should be fine.

PR #3246, opened 2022-10-29
make `JULIA_PKG_IGNORE_HASHES` ignore file download hashes too

IanButterworth:
`JULIA_PKG_IGNORE_HASHES` was introduced to work around odd filesystem tree hashing issues, where the user just wants to make things work.

There is another usecase however, where the source tarball has changed e.g. been recreated and swapped out.
Currently the user is stuck until upstream hashes have been fixed.

This just makes `JULIA_PKG_IGNORE_HASHES` ignore file download hashes too, but with a visible verbose warning when it does.

Note that `JULIA_PKG_IGNORE_HASHES` is not currently in the manual, so is quite private.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -5,7 +5,7 @@
 module PlatformEngines
 
 using SHA, Downloads, Tar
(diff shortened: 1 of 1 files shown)


IanButterworth:
I think this is more of a pkg dev tool than something for roll out. Can be revived if needed

PR #3261, opened 2022-11-16
Remove weird constructors

### LINKED ISSUE #3215: Weird constructor for type alised `Dict`
Happened to see this:

https://github.com/JuliaLang/Pkg.jl/blob/ca9f4426568d85d2ed276b4ab3083dc8f2184478/src/REPLMode/REPLMode.jl#L296-L309

This is not great, it adds a constructor to a certain type of dictionary. This should likely just be its own type.

barucden:
Resolves #3215 

The individual commits are standalone, one addresses `APIOptions` and the other `PackageToken`, so they don't have to be squashed.

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -32,7 +32,7 @@ end
 
 # TODO assert names matching lex regex
 # assert now so that you don't fail at user time
-# see function `REPLMode.APIOptions`
+# see function `REPLMode.api_options`
 function OptionSpec(;name::String,
                     short_name::Union{Nothing,String}=nothing,
                     takes_arg::Bool=false,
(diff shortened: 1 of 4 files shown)


fredrikekre:
I think it would be nicer not to introduce all the new types, instead just use a function `api_options` or w/e that return the dictionary.

barucden:
I did as suggested.

barucden:
The failed test is due to a network error during a git clone.

barucden:
What do you think, @fredrikekre? Is it ready for merge?

(cc @KristofferC, who reported the issue)

KristofferC:
Yep looks good, thank you!

KristofferC:
Actually, could the `PackageToken` still be a union?

barucden:
> Actually, could the `PackageToken` still be a union?

I believe so. It just appeared to me like the perfect use-case for an abstract type with several subtypes. It seemed cleaner to me.

Is there a particular reason for Union?

KristofferC:
> Is there a particular reason for Union?

When there are only a few "subtypes" it means the compiler can branch on the different unions. If it is an abstract type the compiler doesn't know anything. It might not matter a lot here but there has been some issues earlier with "invalidations" (a bit of a technical thing) when Pkg has some type unstable code. To prevent that, I think maybe we should just put back the union so that things are the way they used to be.

barucden:
I was not aware of that; thanks for the explanation. `PackageToken` is a `Union` now. 

I still left `PackageIdentifier` as a struct, not just an alias for a `String`. I hope I don't miss another technical reason why it was a good idea to have it as an alias :)

PR #3264, opened 2022-11-24
add support for package extensions

KristofferC:
Requires https://github.com/JuliaLang/julia/pull/47695

PR #3265, opened 2022-11-25
Make `Pkg.precompile` work with glue packages

IanButterworth:
Builds on #3264 

(While I was here I added some comments to Pkg.precompile too, in a separate commit)

```
shell> rm -rf ~/.julia/compiled/v1.10

julia> import Pkg

(Pkg) pkg> activate /Users/ian/Documents/GitHub/GluePkgExamples.jl/HasDepWithGluePkgs.jl
  Activating project at `~/Documents/GitHub/GluePkgExamples.jl/HasDepWithGluePkgs.jl`

(HasDepWithGluePkgs) pkg> precompile
Precompiling environment...
  6 dependencies successfully precompiled in 3 seconds

julia> using HasDepWithGluePkgs

julia> HasDepWithGluePkgs.do_something()
[ Info: First do something with the basic array support in B
┌ Info: HasGluePkgs.jl: A custom logging method for AbstractArray
│   x =
│    2-element Vector{Float64}:
│     0.22960852293312484
└     0.5867772603987441
[ Info: Now do something with extended OffsetArray support in B
┌ Info: HasGluePkgs.jl: A custom logging method for AbstractArray
│   x =
│    2-element OffsetArray(::Vector{Float64}, 0:1) with eltype Float64 with indices 0:1:
│     0.09267832976119839
└     0.38042557254693776
```

Tested on https://github.com/KristofferC/GluePkgExamples.jl/pull/1

Todo
- [ ] Add tests

### DIFF
--- a/docs/make.jl
+++ b/docs/make.jl
@@ -35,6 +35,7 @@ makedocs(
         "managing-packages.md",
         "environments.md",
         "creating-packages.md",
+        "gluedeps.md",
         "compatibility.md",
         "registries.md",
         "artifacts.md",
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -245,6 +245,81 @@ using Test
 Every dependency should in general have a compatibility constraint on it.
 This is an important topic so there is a separate chapter about it: [Compatibility](@ref Compatibility).
 
+## Conditional loading of code in packages (Glue packages)
+
+!!! note
+    This is a somewhat advanced section which can be skipped for people new to Julia and Julia packages.
+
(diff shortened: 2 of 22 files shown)


KristofferC:
> Add tests

This should get tested by the existing Glue package tests (since the glue packages will get precompiled there in the `Pkg.develop`)

KristofferC:
image

KristofferC:
I think that there will be quite a lot of collisions with glue package names, so maybe it is worth to write it in the progress output as `Parent -> GluePackage` or something so you can keep track from which parent package the glue package comes from?

Issue #3260, opened 2022-11-16
Add option for pkg> compat to auto populate compat bounds


cossio:
It would be nice to have a command within `pkg > compat` to automatically add compat bounds to all dependencies, using the current major sem.version installed.

This is different from https://github.com/JuliaLang/Pkg.jl/issues/776, https://github.com/JuliaLang/Pkg.jl/issues/2463. I don't think it's so convenient to have compat bounds automatically added at `add`.

CC: @IanButterworth , @jakobnissen

PR #3266, opened 2022-11-27
add `pkg> compat --current` mode

IanButterworth:
Edit: Renamed to `compat --current`

---

Closes https://github.com/JuliaLang/Pkg.jl/issues/3260

Screenshot 2025-07-07 at 3 25 14 PM

cc. @cossio @ericphanson


DilumAluthge:
What happens if the active environment has not been instantiated, and there isn't a manifest file? I'm assuming it throws an error?

ericphanson:
What about naming this `compat current`, and then here could also be a `compat latest` to set the bounds to the latest version? And maybe it could work with individual package names like `compat current DataFrames` or `compat current *` which would be the default if a package isn't specified.

Moelf:
if we try to continue the semantics of:
```
(v1.10) pkg> ?compat
  compat [pkg] [compat_string]
```

it looks like it should be `compact * current` instead of `compat current *`

But I would argue having `*` is too weird -- do we need to support regex next? I'm happy with `compat auto` or `compat current`. It's likely only be used once at the beginning or when compat changes a lot, probably not getting in the way of user if they have to do `compact current` and maually edit a few.

IanButterworth:
@Moelf @ericphanson After thinking about it a bit I've updated it to this

Screenshot 2025-07-07 at 3 25 14 PM


What do you think?

PR #3273, opened 2022-12-07
Disable parallel LLVM image generation for now

IanButterworth:
This will likely be needed before https://github.com/JuliaLang/julia/pull/47797 can be merged, to avoid overloading the system.

This leaves speed improvements on the table, but figuring out the load balancing here shouldn't block that PR IMO.


pchintalapudi (approved):
Seems fine to me, assuming compilecache calls into `jl_dump_native` somewhere down the line.

PR #3276, opened 2022-12-13
try improve the merging of sandbox manifest and parent manifest

KristofferC:
I only have a rudimentary understanding of what this piece of code is currently trying to do. The `prune_manifest` call for example doesn't make sense to me.
So I rewrote it to something that I feel I would have written based on what I understand this code should do.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/1585 since this failure mode is no more.

This probably needs a PkgEval due to the "level of uncertainty" of this code.


src/Operations.jl
```diff
@@ -1679,15 +1678,15 @@ function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::S
...
-                        pkgerror("can not merge projects")
-                    end
+        abspath!(sandbox_env, sandbox_env.manifest)
+        for (uuid, entry) in sandbox_env.manifest.deps
+            entry_working = get(working_manifest, uuid, nothing)
+            if entry_working === nothing
+                working_manifest[uuid] = entry
+            else # Check for collision between the sandbox manifest and the "parent" manifest
+                if entry_working != entry && (ctx.env.pkg !== nothing && ctx.env.pkg.uuid != uuid)
+                    @warn "Entry in manifest at \"$sandbox_path\" for package \"$(entry_working.name)\" differs from that in \"$(ctx.env.manifest_file)\""
```
IanButterworth, line 1688: Might be worth letting the user see what the difference is
```suggestion
                    @warn "Entry in manifest at \"$sandbox_path\" for package \"$(entry_working.name)\" differs from that in \"$(ctx.env.manifest_file)\"" sandbox_entry=entry_working original_entry=entry
```

KristofferC: These print really ugly though.

IanButterworth: I can try to make a nice show method

IanButterworth: Maybe this could sprint a status with a diff and include it in the error?

PR #3281, opened 2022-12-15
Run Pkg.precompile before tests in same julia setup as tests for correct caching

IanButterworth:
Required to avoid double-precompilation when native code caching is enabled

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -154,7 +154,9 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why)
             pkgs = deepcopy(pkgs) # don't mutate input
             foreach(handle_package_input!, pkgs)
             ret = $f(ctx, pkgs; kwargs...)
-            $(f in (:add, :up, :pin, :free, :build)) && Pkg._auto_precompile(ctx)
+            $(f in (:add, :up, :pin, :free)) && Pkg._auto_precompile(ctx)
+            # don't precompile early in CI during build because the native cached code will be regenerated for the test setup
+            $(f == :build) && !(tryparse(Bool, get(ENV, "CI", "")) === true) && Pkg._auto_precompile(ctx)
             $(f in (:up, :pin, :free, :rm)) && Pkg._auto_gc(ctx)
             return ret
         end
--- a/src/API.jl
+++ b/src/API.jl
@@ -154,9 +154,7 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why)
             pkgs = deepcopy(pkgs) # don't mutate input
             foreach(handle_package_input!, pkgs)
             ret = $f(ctx, pkgs; kwargs...)
-            $(f in (:add, :up, :pin, :free)) && Pkg._auto_precompile(ctx)
-            # don't precompile early in CI during build because the native cached code will be regenerated for the test setup
-            $(f == :build) && !(tryparse(Bool, get(ENV, "CI", "")) === true) && Pkg._auto_precompile(ctx)
+            $(f in (:add, :up, :pin, :free, :build)) && Pkg._auto_precompile(ctx)
             $(f in (:up, :pin, :free, :rm)) && Pkg._auto_gc(ctx)
             return ret
         end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -14,7 +14,8 @@ import ..Artifacts: ensure_artifact_installed, artifact_names, extract_all_hashe
                     artifact_exists, select_downloadable_artifacts
 using Base.BinaryPlatforms
 import ...Pkg
-import ...Pkg: pkg_server, Registry, pathrepr, can_fancyprint, printpkgstyle, stderr_f, OFFLINE_MODE, UPDATED_REGISTRY_THIS_SESSION, RESPECT_SYSIMAGE_VERSIONS
+import ...Pkg: pkg_server, Registry, pathrepr, can_fancyprint, printpkgstyle, stderr_f, OFFLINE_MODE
+import ...Pkg: UPDATED_REGISTRY_THIS_SESSION, RESPECT_SYSIMAGE_VERSIONS, should_autoprecompile
 
 #########
 # Utils #
@@ -1573,10 +1574,7 @@ function free(ctx::Context, pkgs::Vector{PackageSpec}; err_if_free=true)
     end
 end
 
-function gen_test_code(source_path::String;
-        coverage=false,
-        julia_args::Cmd=``,
-        test_args::Cmd=``)
+function gen_test_code(source_path::String; coverage, julia_args::Cmd, test_args::Cmd)
     test_file = testfile(source_path)
     code = """
         $(Base.load_path_setup_code(false))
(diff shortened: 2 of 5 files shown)


DilumAluthge:
Hmmm. This means that if someone has auto-precompilation disabled, and then they call `Pkg.precompile`, and then they try to use their packages in a normal Julia session, e.g. `using Foo`, they'll hit precompilation, right?

Can we have a kwarg `test_environment` to the `Pkg.precompile` function? And then if `test_environment` is true, we do the precompilation with the same Julia flags as the tests will be run in, but if `test_environment` is false, we do the current behavior (i.e. we use the same Julia flags as the parent process)? And then we default to `test_environment = true`.

So then, in my above example of someone that has auto-precompilation disabled, they can do `Pkg.precompile(; test_environment = false)`, and then when they do `using Foo`, they won't hit precompilation.

DilumAluthge:
Or am I misreading the diff, and this change only applies to the `Pkg.precompile` call that happens at the beginning of the test suite?

IanButterworth:
> this change only applies to the Pkg.precompile call that happens at the beginning of the test suite?

Correct

staticfloat (approved):
looks reasonable to me!

src/Operations.jl
```diff
@@ -1573,17 +1573,28 @@ function free(ctx::Context, pkgs::Vector{PackageSpec}; err_if_free=true)
...
         cd($(repr(dirname(test_file))))
         append!(empty!(ARGS), $(repr(test_args.exec)))
         include($(repr(test_file)))
         """
+    return gen_subprocess_cmd(code; coverage, julia_args)
+end
+
+function gen_test_precompile_code(; coverage, julia_args::Cmd, test_args::Cmd)
+    code = """
+        $(Base.load_path_setup_code(false))
```
KristofferC, line 1589: I don't get why this is needed. If the tests start with the correct project, why isn't just a `Pkg.precompile` enough?

KristofferC (commented):
Didn't we say we needed to disable auto precompile on CI? Otherwise, you will precompile once when the deps are installed (which has the wrong flags) and then once before the tests (with different flags so will be a full re-precompile).

src/Operations.jl
IanButterworth: I assumed that importing `Pkg` into the test process might be breaking, so I left that untouched

KristofferC:
Oh,  I didn't know that. Nice :).

Then I think this should be good. Would be good to confirm this whole PR by setting `ENV["CI"] = "true`, and running a `Pkg.instantiate()` + `Pkg.test()` on some package with the code caching PR and see that things work as expected.

IanButterworth:
I think this shows it's working.

All 19 dependencies had to be re-precompiled before tests ran.


```
shell> rm -rf ~/.julia/compiled/v1.10/

(jl_zoE1I6) pkg> st
Status `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_zoE1I6/Project.toml`
  [336ed68f] CSV v0.10.8

(jl_zoE1I6) pkg> precompile
Precompiling environment...
  19 dependencies successfully precompiled in 53 seconds

(jl_zoE1I6) pkg> test CSV
     Testing CSV
      Status `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_AkwWE3/Project.toml`
  [336ed68f] CSV v0.10.8
  [944b1d66] CodecZlib v0.7.0
  [48062228] FilePathsBase v0.9.20
...
      Status `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_AkwWE3/Manifest.toml`
  [336ed68f] CSV v0.10.8
  [944b1d66] CodecZlib v0.7.0
  [34da2185] Compat v4.5.0
...
Precompiling environment...
  19 dependencies successfully precompiled in 64 seconds
     Testing Running tests...
## Test output starts with no serial precompilation hit
...
```

> by setting ENV["CI"] = "true"

@KristofferC I don't think `ENV["CI"]` is relevant here. Disabling autoprecompilation is specifically done in `Pkg.test` until we're ready to do it, and that hasn't changed, just that the `Pkg.precompile` is now called in the subprocess with the correct setup.

IanButterworth:
I did however just realise that we used to respect whether the user wanted auto-precompilation here, so I've reinstated that

KristofferC:
> Disabling autoprecompilation is specifically done in Pkg.test until we're ready to do it,

If you look at https://github.com/JuliaLang/JuliaSyntax.jl/actions/runs/3703309958/jobs/6274600432 in the `buildpkg` step it says:

```
Precompiling project...
[19](https://github.com/JuliaLang/JuliaSyntax.jl/actions/runs/3703309958/jobs/6274600432#step:5:22)
  ✓ JuliaSyntax
[20](https://github.com/JuliaLang/JuliaSyntax.jl/actions/runs/3703309958/jobs/6274600432#step:5:23)
  1 dependency successfully precompiled in 3 seconds
```

That precompile will now be useless. So that's why I am suggesting disabling it on CI so that doesn't run.

IanButterworth:
Ok, I follow now. Makes sense. Updated. I believe `ENV["CI"]` can be both "" and "true" on CI, so I tested for that

src/API.jl
```diff
@@ -154,7 +154,9 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why)
             pkgs = deepcopy(pkgs) # don't mutate input
             foreach(handle_package_input!, pkgs)
             ret = $f(ctx, pkgs; kwargs...)
-            $(f in (:add, :up, :pin, :free, :build)) && Pkg._auto_precompile(ctx)
+            $(f in (:add, :up, :pin, :free)) && Pkg._auto_precompile(ctx)
+            # don't precompile early in CI during build because the native cached code will be regenerated for the test setup
+            $(f in (:build)) && get(ENV, "CI", false) === false  && Pkg._auto_precompile(ctx)
```
IanButterworth, line 159:
```suggestion
            $(f == :build) && get(ENV, "CI", false) === false && Pkg._auto_precompile(ctx)
```

src/API.jl
```diff
@@ -154,7 +154,9 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why)
             pkgs = deepcopy(pkgs) # don't mutate input
             foreach(handle_package_input!, pkgs)
             ret = $f(ctx, pkgs; kwargs...)
-            $(f in (:add, :up, :pin, :free, :build)) && Pkg._auto_precompile(ctx)
+            $(f in (:add, :up, :pin, :free)) && Pkg._auto_precompile(ctx)
+            # don't precompile early in CI during build because the native cached code will be regenerated for the test setup
+            $(f == :build) && get(ENV, "CI", false) === false && Pkg._auto_precompile(ctx)
```
IanButterworth, line 159:
```suggestion
            $(f == :build) && !in(get(ENV, "CI", "false"), ("", "true")) && Pkg._auto_precompile(ctx)
```

DilumAluthge:
> I believe `ENV["CI"]` can be both "" and "true" on CI, so I tested for that

What CI platforms are there that set `CI` to an empty string?

I think we should only assume that we are running on CI if `tryparse(Bool, get(ENV, "CI", "")) === true`.

IanButterworth:
> What CI platforms are there that set CI to an empty string?

Only a vague recollection. I've switched it to your suggestion. thanks

IanButterworth:
On reflection, I'm a bit torn about whether Pkg should be so opinionated about when precompilation happens in CI. 

`buildpkg` isn't just used before `julia-runtest`, it could be before a script that does something else.

IanButterworth:
I propose I revert [`476b368` (#3281)](https://github.com/JuliaLang/Pkg.jl/pull/3281/commits/476b36842b8e4504ac1c7dd7bd8b566e8f373f5c) here and we do https://github.com/julia-actions/julia-buildpkg/pull/26 instead

staticfloat:
I think what we should do is disable automatic precompilation during `Pkg.add()` on CI (by having the CI scripts specify that, not something in Pkg) and then have `Pkg.test()` do a `precompile()` with the same options as it would use when actually running the tests.  If that's what your comment means, then I agree. :)

IanButterworth:
Almost what I'm saying. There's no standard way to `Pkg.add` in CI that's like `julia-buildpkg` `julia-runtest` so IMO that's up to the user to manage. 

I'm proposing not precompiling during `julia-buildpkg` via control in that directly, rather than making Pkg decide. 

The rest is accurate

staticfloat:
In buildkite land I don't think we even manually instantiate (that's what I meant more than `Pkg.add()`, sorry for imprecise terminology), we just rely on `Pkg.test()` to do that for us: https://github.com/JuliaCI/julia-test-buildkite-plugin/blob/6859a2c70d88fb5711f423b6bf3e76342b480af4/hooks/command#L28-L30

IanButterworth:
Cool. Ok, I'm going to go ahead with that plan. I guess we can adapt if needed

PR #3284, opened 2022-12-20
Fix broken source_path handling. Explicitly test for coverage files

IanButterworth:
~~It's not clear to me why this didn't fail in #3281 .~~ Urgh.. it was picking up a function and string-ing the name 🤦‍♂️
```
julia> Pkg.Operations.source_path
source_path (generic function with 2 methods)

julia> string("@", Pkg.Operations.source_path)
"@source_path"
```

----

This fixes a rather nasty missing local var bug introduced by #3281 that broke coverage generation in `Pkg.test`, and reinstates and expands testing for coverage files where they should and shouldn't be generated

### DIFF
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -232,9 +232,12 @@ temp_pkg_dir() do project_path
     end
 
     @testset "testing" begin
-        pkgdir = Base.locate_package(Base.PkgId(TEST_PKG.uuid, TEST_PKG.name))
-        recursive_rm_cov_files(pkgdir) # clean out cov files from previous test runs
         Pkg.add(TEST_PKG.name)
+
+        pkgdir = Base.locate_package(Base.PkgId(TEST_PKG.uuid, TEST_PKG.name))
+        @test !isnothing(pkgdir)
+        recursive_rm_cov_files(pkgdir) # clean out cov files from previous test runs
+
         @test !any(endswith(".cov"), readdir(pkgdir)) # should be no cov files to start with
         Pkg.test(TEST_PKG.name; coverage=true)
         @test any(endswith(".cov"), readdir(pkgdir))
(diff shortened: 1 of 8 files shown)


test/test_packages/ExtensionExamples/HasExtensions.jl/ext/OffsetArraysExt.jl
```diff
@@ -1,6 +1,7 @@
 module OffsetArraysExt
 
 using HasExtensions, OffsetArrays
+import HasExtensions: foo
```
KristofferC, line 4: Oops :p

fredrikekre (approved):

test/pkg.jl
```diff
@@ -233,10 +233,14 @@ temp_pkg_dir() do project_path
...
+
+        pkgdir = dirname(Base.locate_package(Base.PkgId(TEST_PKG.uuid, TEST_PKG.name)))
+        @test !isnothing(pkgdir)
+        recursive_rm_cov_files(pkgdir) # clean out cov files from previous test runs
+
+        @test !any(endswith(".cov"), readdir(pkgdir)) # should be no cov files to start with
         Pkg.test(TEST_PKG.name; coverage=true)
-        pkgdir = Base.locate_package(Base.PkgId(TEST_PKG.uuid, TEST_PKG.name))
-        # No coverage files being generated?
-        @test_broken TEST_PKG.name * ".cov" in readdir(pkgdir)
```
KristofferC, line 239: What fixed this?

fredrikekre: Coverage files now have a PID, so maybe just not looking for the correct files?

KristofferC: I see. Makes sense.

test/test_packages/ExtensionExamples/HasExtensions.jl/ext/OffsetArraysExt.jl
IanButterworth: I did lol :)

PR #3302, opened 2023-01-08
Clarify what "tiered" means

LilithHafner:
The tier that preserves the most version information is always `all`, but `tiered` doesn't always use that tier. This clarifies exactly when it doesn't.

And bold `tiered` because it is the default.


src/REPLMode/command_declarations.jl
```diff
@@ -123,13 +123,14 @@ If the package is not located at the top of the git repository, a subdirectory c
...
-| `semver` | Preserve semver-compatible versions of direct dependencies                          |
-| `none`   | Do not attempt to preserve any version information                                  |
-| `tiered` | Use the tier which will preserve the most version information (this is the default) |
+| Argument     | Description                                                                        |
+|:-------------|:-----------------------------------------------------------------------------------|
+| `all`        | Preserve the state of all existing dependencies (including recursive dependencies) |
+| `direct`     | Preserve the state of all existing direct dependencies                             |
+| `semver`     | Preserve semver-compatible versions of direct dependencies                         |
+| `none`       | Do not attempt to preserve any version information                                 |
+| **`tiered`** | Use the tier which will preserve the most version information                      |
```
LilithHafner, line 132:
```suggestion
| **`tiered`** | Use the tier that will preserve the most version information                       |
```
[Grammar](https://www.diffen.com/difference/That_vs_Which)

LilithHafner:
Bump. Right now, the description of tiered is literally incorrect because the tier which preserves the most information is `all` or 'installed'. This helps.

IanButterworth:
Seems like an improvement. Can you also update the docstrings where this is explained.

LilithHafner:
Yep, I didn't realize this explanation was duplicated. I think I got all two cases now (from `grep -riI "Use the tier"`)

IanButterworth (approved):
Great, thanks

PR #3303, opened 2023-01-10
Respect manifest if provided in Project.toml

quinnj:
I recently discovered that loading in Base supports specifying an alternative `Manifest.toml` in a `Project.toml`. While playing around with this though, I discovered that Pkg.jl doesn't play as nice with this. It supports the parsing/storing of it, but gets confused often because so many path calculations are made relative to the _project file_ instead of the now-different manifest file.

This is relevant when there are devv-ed packages in the specified Manifest and the relative path calculations should be made wrt to the _manifest_ not the _project_.

The changes here, hence, are mainly about auditing uses of `env.project_file` and changing them to `env.manifest_file` if necessary. The cases where `manifest_file` should be used, as far as I understand, are where the `PackageSpec` we're computing a path for came from a manifest entry instead of a project dep or elsewhere.

The other change in this PR is in `prune_manifest` where we actually _don't_ want to prune the manifest if the project doesn't "own" it. I'm calculating that by checking `dirname(env.manifest_file) != dirname(env.project_file)`, but we could also potentially just check `env.pkg.manifest` if it's populated, but it wasn't clear if that would be a more sure thing since we're trying to query if the Manifest isn't the default.

### DIFF
--- a/test/project_manifest.jl
+++ b/test/project_manifest.jl
@@ -9,6 +9,11 @@ temp_pkg_dir() do project_path
         mktempdir() do dir
             path = abspath(joinpath(dirname(pathof(Pkg)), "../test", "test_packages", "monorepo"))
             cp(path, joinpath(dir, "monorepo"))
+            cd(joinpath(dir, "monorepo")) do
+                with_current_env() do
+                    Pkg.develop(path="packages/B")
+                end
+            end
             # test subpackage instantiates/tests
             # the order of operations here is important
             # when we instantiate and dev a dependency to the C subpackage
@@ -27,7 +32,6 @@ temp_pkg_dir() do project_path
             cd(joinpath(dir, "monorepo")) do
                 with_current_env() do
                     Pkg.develop(path="packages/C")
-                    Pkg.develop(path="packages/B")
                     Pkg.add("Test")
                     Pkg.test()
                 end
--- a/test/project_manifest.jl
+++ b/test/project_manifest.jl
@@ -1,0 +1,46 @@
+module ProjectManifestTest
+
+import ..Pkg # ensure we are using the correct Pkg
+using Test, Pkg
+using ..Utils
+
+temp_pkg_dir() do project_path
+    @testset "test Project.toml manifest" begin
+        mktempdir() do dir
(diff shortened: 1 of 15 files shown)


KristofferC:
Makes sense that paths etc are relative to the manifest and not the project. That will also be true for subprojects when those exist.

> The other change in this PR is in prune_manifest where we actually don't want to prune the manifest if the project doesn't "own" it. I'm calculating that by checking dirname(env.manifest_file) != dirname(env.project_file),

When we do the subproject thing we do want to prune the manifest but consider the packages in **all** subprojects to be "roots". But this is fine to do for now as an incremental workflow improvement.

test/project_manifest.jl
```diff
@@ -0,0 +1,43 @@
...
+            end
+            # test subpackage instantiates/tests
+            # the order of operations here is important
+            # when we instantiate and dev a dependency to the C subpackage
+            # the changes are being written to the monorepo/Manifest.toml
+            # but are not "sticky" in that if we ran a resolve a the monorepo/
+            # level, the C & D packages would be pruned since there's no direct
+            # dependency at the monorepo project level yet, so we first build up
+            # C's dependencies, then at the monorepo level, we need to dev C *first*
+            # to make those Manifest changes "stick" before devving B or adding Test
```
nickrobinson251, line 25:
```suggestion
            # to make those Manifest changes "stick" before adding Test.
```

test/project_manifest.jl
```diff
@@ -0,0 +1,43 @@
...
+            cd(joinpath(dir, "monorepo")) do
+                with_current_env() do
+                    Pkg.develop(path="packages/B")
+                end
+            end
+            # test subpackage instantiates/tests
+            # the order of operations here is important
+            # when we instantiate and dev a dependency to the C subpackage
+            # the changes are being written to the monorepo/Manifest.toml
+            # but are not "sticky" in that if we ran a resolve a the monorepo/
```
nickrobinson251, line 21:
```suggestion
            # but are not "sticky" in that if we ran a resolve at the monorepo/
```

nickrobinson251 (commented):
Thanks for walking me through this -- LGTM

PR #3308, opened 2023-01-12
also check build ids when checking for staleness of precompile files

KristofferC:
I noticed a case (which I can't repro now) where `pkg> precompile` considered everything precompiled, but loading the package still had to recompile. When I debugged, I saw that the check at https://github.com/JuliaLang/julia/blob/fec8304a8c6e2157759b76d20339b426201a5f61/base/loading.jl#L2602-L2606 failed which caused Julia to recompile the package.

However, in Pkg, we only call the 2-arg version of `stale_cachefile` which disables this check https://github.com/JuliaLang/julia/blob/fec8304a8c6e2157759b76d20339b426201a5f61/base/loading.jl#L2554-L2556

This PR copies the recursive checks for staleness in https://github.com/JuliaLang/julia/blob/79ceb8dbeab1b5a47c6bd664214616c19607ffab/base/loading.jl#L1309-L1336 but it adds some caching to prevent checking this for the same package multiple times. This still leads to some slowdown. On a pretty big environment I measure before this change a "no-op" precompile to take 469ms while with this change it takes 834 ms. However, with some more caching in Base (https://github.com/JuliaLang/julia/pull/48247) it now takes 578 ms which is not so bad compared to before.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1032,19 +1032,48 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
     Operations.build(ctx, Set{UUID}(pkg.uuid for pkg in pkgs), verbose)
 end
 
-function _is_stale(paths::Vector{String}, sourcepath::String)
+
+function _is_stale!(stale_cache::Dict, paths::Vector{String}, sourcepath::String)
     for path_to_try in paths
         staledeps = Base.stale_cachefile(sourcepath, path_to_try, ignore_loaded = true)
-        if staledeps !== true
-            try
-                # update timestamp of precompilation file so that it is the first to be tried by code loading
-                touch(path_to_try)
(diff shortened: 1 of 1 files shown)


timholy:
Thank you! I don't know whether it's exactly this issue, but I've seen the phenomenon you describe quite frequently, and never tracked it down. If this fixes it, it's a significant QOL improvement.

baggepinnen:
Fantastic! This bug has been eating up a lot of time and patience for quite some time now!

KristofferC:
Since I can't repro my original issue (I was being stupid not to save my `.compiled` folder) I am not 100% sure that this will fix all cases, but it should IMO no longer be able to trigger the same failure mode that I originally saw.

IanButterworth (approved):
LGTM. Thanks for hunting this out!

PR #3309, opened 2023-01-12
add some type annotations to improve type stability

KristofferC:


src/manifest.jl
```diff
@@ -204,7 +211,7 @@ function read_manifest(f_or_io::Union{String, IO})
     catch e
         if e isa TOML.ParserError
             pkgerror("Could not parse manifest: ", sprint(showerror, e))
-        end
+        end
```
timholy, line 214: whitespace

timholy (approved):

src/manifest.jl
KristofferC, line 214:
```suggestion
        end
```

PR #3320, opened 2023-01-17
only auto update registries on `add` once per day

KristofferC:
I think it is fine to only do the auto registry update on `Pkg.add` ~ once per day.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -256,7 +256,7 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PR
     # repo + unpinned -> name, uuid, repo.rev, repo.source, tree_hash
     # repo + pinned -> name, uuid, tree_hash
 
-    Operations.update_registries(ctx; force=false)
+    Operations.update_registries(ctx; force=false, update_cooldown=Day(1))
 
     project_deps_resolve!(ctx.env, pkgs)
     registry_resolve!(ctx.registries, pkgs)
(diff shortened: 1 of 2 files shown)


IanButterworth (approved):

IanButterworth:
I tried this out and somehow `registry_update_log[string(reg.uuid)] = now()` isn't getting set for me in `Registry.update` meaning the dict is always empty on write.

IanButterworth:
Also, it seems popular to just remove auto registry update from `add`, but perhaps add a message if the registry is older than a certain period, say 1 week??

KristofferC:
> I tried this out and somehow registry_update_log[string(reg.uuid)] = now() isn't getting set for me in Registry.update meaning the dict is always empty on write.

Hm, I tried this and it seemed to write for me. Can you check where the code is going that doesn't hit the dict setting code?

IanButterworth:
Unfortunately it's working for me now. I also can't see a gap in the code.

For the record I have General via packed tarball and two other git registries

fonsp:
This is great!  🎉

Perhaps this could also replace `Pkg.UPDATED_REGISTRY_THIS_SESSION`? It seems like `UPDATED_REGISTRY_THIS_SESSION` used to be a heuristic for "has been updated recently", which is now implemented directly. Or is there a specific reason to track that it has been updated within the current session?

IanButterworth:
Yeah people that were setting `Pkg. UPDATED_REGISTRY_THIS_SESSION = true` should be able to remove that if they're only using julia versions that have this. Though, this is less heavy-handed. It still allows `up` to update the registry once a day.

IanButterworth:
@KristofferC and I discussed offline and agreed this is good to go

PR #3322, opened 2023-01-18
Make precompile use the API harness

### LINKED ISSUE #3321: Pkg.precompile stack overflows when pkgs is a vector of PackageSpec
to reproduce

```julia
julia> using Pkg

julia> Pkg.precompile(PackageSpec(name="Example"))
```

this hangs forever for me on v1.8.5 and v1.9.0-beta2, might be another issue for stack overflow, but I think this should be at least throwing an error on the Pkg side and get mentioned in docstring if `PackageSpec` is not supported for `precompile`.

IanButterworth:
Fixes #3321 by making it behave like the other api functions

I guess this should've been done in https://github.com/JuliaLang/Pkg.jl/pull/2724

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -138,15 +138,15 @@ function require_not_empty(pkgs, f::Symbol)
 end
 
 # Provide some convenience calls
-for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why)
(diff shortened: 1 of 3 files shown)


IanButterworth:
bump @KristofferC

PR #3329, opened 2023-01-20
add option for Pkg.build to allow_reresolve=false

IanButterworth:
Like `Pkg.test` has `allow_reresolve` https://github.com/JuliaLang/Pkg.jl/pull/2531 `Pkg.build` should also have the same option, to make the build process fixed by manifest


DilumAluthge:
Huh.... I didn't realize that, currently, `Pkg.build` is allowed to do a re-resolve.

Why is that the case? For `Pkg.test`, we have the option to re-resolve because we are adding new test-only dependencies to the environment, and we might need to re-resolve our environment in order to satisfy the `[compat]` restrictions of our test-only dependencies.

But `Pkg.build` doesn't need to add any new dependencies to the environment, does it? So why does `Pkg.build` ever need to re-resolve?

IanButterworth:
True. Maybe if somehow the checked in manifest isn't a valid resolve of the project. 

Maybe it should default to false but still be overdidable

IanButterworth:
Curiously the environment the issue was seen in does have a checked-in manifest but it's a valid resolve of the project.

It's unclear why it had to re-resolve.

Octogonapus:
It is still unclear to me why the re-resolve happened, but over the weekend, the problem has _resolved_ itself. I haven't changed the manifest, but it's now a valid resolve :shrug: Maybe something moved in a registry?

IanButterworth:
Oh, maybe a version was yanked and un-yanked

IanButterworth:
The sandbox resolve should probably allow yanked versions

src/Operations.jl
```diff
@@ -953,12 +953,12 @@ function any_package_not_installed(manifest::Manifest)
     return false
 end
 
-function build(ctx::Context, uuids::Set{UUID}, verbose::Bool)
+function build(ctx::Context, uuids::Set{UUID}, verbose::Bool; allow_reresolve::Bool=true)
```
DilumAluthge, line 956: Shall we default to `allow_reresolve = false`? Since this use case (build-only deps) seems pretty uncommon.

```suggestion
function build(ctx::Context, uuids::Set{UUID}, verbose::Bool; allow_reresolve::Bool=false)
```

IanButterworth:
@DilumAluthge I propose that we add this PR as-is as a conservative way to expose this option at least.

I know that I have a situation where I want it enabled, but I have no idea whether the ecosystem would want it on by default.

DilumAluthge:
Can we do:

1. Default to `false`, but allow it to be overrided
2. Don't make it public (don't document it in the manual). So then we can re-evaluate later and change the default if we want.

IanButterworth:
As for 2, let's just put it in the docs and make it symmetrical with Pkg.test

PR #3330, opened 2023-01-20
Change the default registry flavour from `conservative` to `eager`

KristofferC:
I really think the current default flavour is wrong. I would argue that 99.99% of users do not care if they download a package / artifact from the Pkg server or GitHub. They much rather have the latest version and there are quite frequent questions about why a user gets an old version installed, even though it was x hours since it got registered. As it is right now, if the package server processes a package slowly, all registered versions after that are blocked for everyone which imo is quite unacceptable. The recent reported blockage was ~18 hours from what I recently read which imo is not really acceptable.

In special cases, where you are e.g. behind a fire wall so you cannot talk to GitHub you want to use the conservative version since otherwise you might fail the download. But this is the special case so imo it makes more sense for these people to have to set some option.

cc @StefanKarpinski, @DilumAluthge .

### DIFF
--- a/docs/src/registries.md
+++ b/docs/src/registries.md
@@ -100,30 +100,12 @@ The default Pkg Server (`pkg.julialang.org`) offers two different "flavors" of r
 !!! compat "Julia 1.8"
     Registry flavors are only available starting with Julia 1.8.
 
-- `conservative`: suitable for most users; all packages and artifacts in this registry flavor are available from the Pkg Server, with no need to download from other sources
-- `eager`: this registry offers the latest versions of packages, even if the Pkg and Storage Servers have not finished processing them; thus, some packages and artifacts may not be available from the Pkg Server, and thus may need to be downloaded from other sources (such as GitHub)
+- `eager`: suitable for most users; all packages and artifacts in this registry flavor are available to download but it might not come from the Pkg Server
+- `conservative`: this registry only offers the versions of packages that have been proccesed by the Storage Servers; thus, the latest version of some packages and artifacts may not be availale.
 
-The default registry flavor is `conservative`. We recommend that most users stick to the `conservative` flavor unless they know that they need to use the `eager` flavor.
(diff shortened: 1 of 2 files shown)


StefanKarpinski:
As a technical matter, this is the wrong way to do this. The client should never set a default for server options—the default is to set no value and have the server pick a default. In fact, server options should never appear anywhere in the Pkg code. The right way to change this default is to change the server to serve eager when the client doesn't send a preference. That change would immediately be picked up by all clients (except those who have an explicit preference already). This is why you don't hard code these things into the client—then you have to wait for a new release and for people to start using it. When the server controls the defaults we can change them immediately for everyone. That's also why the `JULIA_PKG_SERVER_REGISTRY_PREFERENCE` doesn't appear anywhere in the Pkg code; instead we just send any environment variable that starts with `JULIA_PKG_SERVER_` to the server as a header. That way we can introduce new server options at any time and allow people with any recent Julia version to start using them immediately.

As a social matter, I vehemently disagree with changing the default to eager. The vast majority of users don't care if the registry they're using is a a few hours old. They'll never even realize it—they don't ever look at the upstream registry and they just get package updates when they get them. The only people who really care about getting package updates promptly are developers of registered packages, and they _have_ to be in contact with us because they have to register their packages, which gives them a mechanism to learn about eager mode if this affects them. We should not, in general, regularly be putting people in a position to need to get packages and artifacts from upstream sources, which may be down or broken, and which we have no insight into the state of. We should keep everything flowing through package servers as much as possible, which is exactly what the conservative default does. I think that's a good principle for everyone, not just people behind firewalls. But this is necessary for people behind restrictive firewalls not to have a terrible experience. That number of people is conservatively some 10% of users. And unlike registered package developers, who inherently need to be in some contact with us, people behind firewalls tend to be the people we're least likely to ever hear from, who will just give up in frustration because from their perspective the Julia package manager is just constantly broken.

DilumAluthge:
> the idea is to prepare all the source and artifact tarballs _before_ allowing package registration to succeed, as part of a CI step

Cross-ref: https://github.com/JuliaRegistries/General/issues/76297

PR #3332, opened 2023-01-23
Disallow re-resolution in `Pkg.build`

DilumAluthge:
Closes #3329 

In `Pkg.test`, we need to have the ability for the user to allow or disallow re-resolution. This is because `Pkg.test` will add the test-only dependencies, which may come with their own `[compat]` constraints, which may conflict with the existing manifest. Therefore, the `Pkg.test` function has the `allow_reresolve` kwarg.

However, I don't think we need to ever allow `Pkg.build` to re-resolve. If for some reason, the active manifest is not a valid resolution of the active project, then we should just print an error. This is fine, because it's really easy for the user to fix: just run `Pkg.resolve()`, and then run `Pkg.build()` again.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1101,7 +1101,7 @@ function build_versions(ctx::Context, uuids::Set{UUID}; verbose=false)
         fancyprint && show_progress(ctx.io, bar)
 
         let log_file=log_file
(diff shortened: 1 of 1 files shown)


IanButterworth:
I think https://github.com/JuliaLang/Pkg.jl/pull/3332 is a better safer direction.

Issue #3340, opened 2023-01-26
`Pkg.add()` should add the git submodule as the usual git repository


Fenyutanchan:
1. The output of `versioninfo()`
    ```log
    Julia Version 1.8.5
    Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
    Platform Info:
      OS: macOS (arm64-apple-darwin21.5.0)
      CPU: 8 × Apple M1
      WORD_SIZE: 64
      LIBM: libopenlibm
      LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
      Threads: 1 on 4 virtual cores
    ```
2. How you installed Julia
    ```shell
    brew install --cask julia
    ```
3. A minimal working example (MWE), also known as a minimum reproducible example
    Create a git repository with some git submodules. Then `Pkg.add("/path/to/git-submodule.jl")` doesn't work and the output is shown
    ```log
    ERROR: Did not find a git repository at `/path/to/git-submodule.jl`
    ```
    which is the same as `Pkg.add("/path/to/some.jl")` without the directory of `.git/`. However, there is indeed a file `.git` in the directory of the git submodule which archives some information about the git message as
    ```log
    gitdir: /path/to/the/parent/git/directory/.git/modules/git-submodule.jl
    ```
    This directory is the missing directory `.git/` of the git submodule. Therefore, I think `Pkg.add()` should process this case correctly.

Fenyutanchan:
Misplace issue: JuliaLang/julia#48411

fredrikekre:
Duplicate of #708

Fenyutanchan:
@fredrikekre
I think this is not the duplicate of #708.
#708 says that `Pkg.add()` cannot add one git repository with submodules, but I am saying that `Pkg.add()` cannot add the git repository which is the git submodule at one git repository!
It is huge different between them!

Fenyutanchan:
Right now, only `Pkg.develop()` could add the git submodules directly.
But I don't think it is logistical that `Pkg.add()` could not resolve the git submodules, because every git message can be found.

MilesCranmer:
Ran into this too which is why AirspeedVelocity.jl can't work on projects within submodules: https://github.com/MilesCranmer/AirspeedVelocity.jl/issues/46. Indeed it is a different issue than #708.

In principle there's no reason this should not work. It's simply because the `.git` repository is stored in a different directory, so Julia thinks it's not actually a git repo.

[...]

Issue #708, opened 2018-09-03
Pkg.add on git repository with a submodule raises GitError


KristofferC:
@non-Jedi commented on [Fri Aug 31 2018](https://github.com/JuliaLang/julia/issues/28996)

Just upgraded to 1.0 and tried to checkout some of my projects. One has a git submodule in it as a convenience since it's being developed against a spec contained in another git repo, but I am unable to add that one using `Pkg`:

```julia
julia> using Pkg

julia> Pkg.add(PackageSpec(url="https://github.com/non-Jedi/MatrixClientProtocol.jl", rev="master"))
  Updating registry at `~/.julia/registries/General`
  Updating git-repo `https://github.com/JuliaRegistries/General.git`
  Updating git-repo `https://github.com/non-Jedi/MatrixClientProtocol.jl`
ERROR: GitError(Code:ERROR, Class:Submodule, cannot get submodules without a working tree)
Stacktrace:
 [1] macro expansion at /builddir/julia/usr/share/julia/stdlib/v1.0/LibGit2/src/error.jl:101 [inlined]
 [2] #checkout_tree#46(::LibGit2.CheckoutOptions, ::Function, ::LibGit2.GitRepo, ::LibGit2.GitTree) at /builddir/julia/usr/share/julia/stdlib/v1.0/LibGit2/src/repository.jl:358
 [3] #add_or_develop at ./none:0 [inlined]
 [4] #add_or_develop#12 at /builddir/julia/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:29 [inlined]
 [5] #add_or_develop at ./none:0 [inlined]
 [6] #add_or_develop#10(::Base.Iterators.Pairs{Symbol,Symbol,Tuple{Symbol},NamedTuple{(:mode,),Tuple{Symbol}}}, ::Function, ::Pkg.Types.PackageSpec) at /builddir/julia/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:27
 [7] #add_or_develop at ./none:0 [inlined]
 [8] #add#18 at /builddir/julia/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:69 [inlined]
 [9] add(::Pkg.Types.PackageSpec) at /builddir/julia/usr/share/julia/stdlib/v1.0/Pkg/src/API.jl:69
 [10] top-level scope at none:0
```

[...]

PR #3344, opened 2023-01-27
add support for `Pkg.add(path="/path/to/git-submodule.jl")` directly.

Fenyutanchan:
For #3340.

It seems the `ERROR: Did not find a git repository at /path/to/git-submodule.jl` is generated from the code of
```julia
if !isdir(joinpath(pkg.repo.source, ".git"))
    ...
end
```
Then I suggest that the modification as
```julia
git_info_path   =   (path -> !isfile(path) ? path : joinpath(dirname(path), (last∘split∘readline)(path)))(joinpath(pkg.repo.source, ".git"))
if !isdir(git_info_path)
    ...
end
```
which can resolve the real path to the directory `.git/` (generally `/path-to-super-git-repository/.git/modules/name-of-git-submodule.jl`) of the git submodule.

I have tested this modification at 
```log
Julia Version 1.8.5
Commit 17cfb8e65ea (2023-01-08 06:45 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin21.5.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, apple-m1)
  Threads: 1 on 4 virtual cores
```
It works as my expectation.
Further tests may be required.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -874,7 +874,8 @@ function handle_repo_add!(ctx::Context, pkg::PackageSpec)
     repo_source = pkg.repo.source
     if !isurl(pkg.repo.source)
         if isdir(pkg.repo.source)
-            if !isdir(joinpath(pkg.repo.source, ".git"))
+            git_info_path   =   (path -> !isfile(path) ? path : joinpath(dirname(path), (last∘split∘readline)(path)))(joinpath(pkg.repo.source, ".git"))
+            if !isdir(git_info_path)
                 msg = "Did not find a git repository at `$(pkg.repo.source)`"
                 if isfile(joinpath(pkg.repo.source, "Project.toml")) || isfile(joinpath(pkg.repo.source, "JuliaProject.toml"))
                     msg *= ", perhaps you meant `Pkg.develop`?"
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -874,7 +874,15 @@ function handle_repo_add!(ctx::Context, pkg::PackageSpec)
     repo_source = pkg.repo.source
     if !isurl(pkg.repo.source)
         if isdir(pkg.repo.source)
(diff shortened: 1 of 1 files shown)


nlw0:
Bump this PR. I've just ran into this issue. When setting up CI, I would like to `add` my packages from submodules, but it doesn't work. Then we're forced to do `dev`, which I guess is not the end of the hello world. But I'd argue that it's just not very intuitive that a submodule is not considered to be a git repo, and I really just want to `add` when running my tests, and not `dev`. #semantics

MilesCranmer:
@KristofferC would you be willing to take a look at this? It addresses the scenario where you have a repository containing submodules, and you try to add **_one_** of the submodules. 

This is different than #708 as there's no ambiguity here in the case of that particular submodule containing a single package.

Here's a MWE of the bug:

```bash
git init
git submodule add https://github.com/timholy/Revise.jl
julia --project=. -e 'using Pkg; pkg"add ./Revise.jl/"'

# ERROR: Did not find a git repository at `Revise.jl/`, perhaps you meant `Pkg.develop`?
```

The error is because git submodules have their `.git` repo stored at the base repository they are a submodule of.

src/Types.jl
```diff
@@ -879,7 +879,8 @@ function handle_repo_add!(ctx::Context, pkg::PackageSpec)
     repo_source = pkg.repo.source
     if !isurl(pkg.repo.source)
         if isdir(pkg.repo.source)
-            if !isdir(joinpath(pkg.repo.source, ".git"))
+            git_info_path   =   (path -> !isfile(path) ? path : joinpath(dirname(path), (last∘split∘readline)(path)))(joinpath(pkg.repo.source, ".git"))
```
KristofferC: I think this has to be rewritten, my human brain cannot understand what is going on with an anonymous function into a double `∘` etc.

nlw0: It seems the AI did a good job understanding the preferences of your human brain! :)

PR #3349, opened 2023-01-31
Don't clear executable bits in `set_readonly()` on Windows

staticfloat:
On windows, `filemode()` doesn't include executable bits.  This is already worked around in `gitmode()` but we don't quite want to do something as intrusive as `gitmode()` in `set_readonly()`; let's just work around this one platform-specific foible.

This fixes problems like executable permissions disappearing when creating artifacts on Windows, packages that contain executable files losing those permissions on install, etc...


staticfloat:
For further discussion, see https://github.com/simeonschaub/ArtifactUtils.jl/pull/8#discussion_r767210865

KristofferC:
Bump :)

staticfloat:
*gulp* here we go

staticfloat:
FTR I am not proposing we backport this to 1.9; I consider this a somewhat high-risk change, so let's let it settle on `master` a bit.

PR #3350, opened 2023-01-31
force `--warn-overwrite=yes` in subprocess

sjkelly:
Normally precompilation will warn for this,
however it is possible there may be cases that
the user has not precompiled the package in question. By enforcing this as a default in test conditions, we ensure method overwrites will always produce a warning.


IanButterworth:
Seems reasonable. `gen_subprocess_cmd` is also used by `Pkg.build` so this enables that there, which I think is ok

sjkelly:
I am thinking there is a slightly better, but more complicated approach here. We can inspect the `julia_args` and respect the user's setting(s), but have preferred defaults.

IanButterworth:
bump. This seems good to do, especially now `Pkg.precompile` shows any warnings

sjkelly:
We can merge this one as-is. I will work on a followup with the more sophisticated functionality.

sjkelly:
@IanButterworth I toyed around with this a bit only to realize that julia's command line will prioritize the last specified argument. So this is okay, and anything specified in `julia_args` will take president. This is good to go!

IanButterworth:
Great. Seems good to me but just marking @timholy for a double-check

PR #3353, opened 2023-02-01
fix `stale_cache` calculation

aviatesk:
xref JuliaLang/julia#48397

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1053,39 +1053,29 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
     Operations.build(ctx, Set{UUID}(pkg.uuid for pkg in pkgs), verbose)
 end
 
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -1053,40 +1053,30 @@ function build(ctx::Context, pkgs::Vector{PackageSpec}; verbose=false, kwargs...
...
-                break
-            end
-            if !modfound
-                staledeps = true
-                break
+                stale_cache_key = (modkey, modbuild_id, modpath, modpath_to_try)::StaleCacheKey
+                is_stale = get!(() -> Base.stale_cachefile(stale_cache_key...) === true,
+                                stale_cache, stale_cache_key)
+                is_stale && continue
+                @goto check_next_path
```
KristofferC, line 1077: This type of simplification could probably be made to `loading.jl` (in Base) as well, where this code was pretty much taken from.

PR #3357, opened 2023-02-03
[release-1.9] Backport manifest file issue from #3356

quinnj:


KristofferC:
I also included this in https://github.com/JuliaLang/Pkg.jl/pull/3360 so this commit is empty. But doesn't really matter.

PR #3358, opened 2023-02-04
precompile: warn if waiting on trailing task

IanButterworth:
Proposal to help with https://github.com/JuliaLang/julia/issues/48506

Screenshot 2023-02-04 at 8 54 30 AM

In non-fancyprint mode, i.e. in CI, if the trailing task warning happens it will warn then start printing stderr in realtime to allow the user to see why it's hanging and cancel the job.

cc. @LilithHafner @timholy

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1266,15 +1266,37 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
 
     function handle_interrupt(err, in_printloop = false)
         notify(interrupted_or_done)
-        sleep(0.2) # yield for a period to let the print loop cease first
         in_printloop || wait(t_print) # wait to let the print loop cease first
         if err isa InterruptException
             lock(print_lock) do
                 println(io, " Interrupted: Exiting precompilation...")
             end
             interrupted = true
-        else
-            @error "Pkg.precompile error" exception=(err, catch_backtrace())
+        end
+    end
+
+    stderr_outputs = Dict{Base.PkgId,String}()
+    taskwaiting = Set{Base.PkgId}()
+
+    function monitor_stderr(pkg, iob)
+        try
+            while isopen(iob)
+                str = readline(iob)
+                stderr_outputs[pkg] = get(stderr_outputs, pkg, "") * str * "\n"
+                if !in(pkg, taskwaiting) && occursin("waiting for IO to finish", str)
+                    !fancyprint && lock(print_lock) do
+                        println(io, pkg.name, color_string(" Waiting for background task, IO, or timer to finish.", Base.warn_color()))
+                    end
+                    push!(taskwaiting, pkg)
+                end
+                if !fancyprint && in(pkg, taskwaiting)
+                    lock(print_lock) do
+                        println(io, str)
+                    end
+                end
+            end
+        catch err
+            err isa InterruptException || rethrow()
         end
     end
 
@@ -1332,7 +1354,12 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                                 # If not offset, on larger terminal fonts it looks odd that they all sync-up
                                 anim_char = anim_chars[(i + Int(dep.name[1])) % length(anim_chars) + 1]
                                 anim_char_colored = dep in direct_deps ? anim_char : color_string(anim_char, :light_black)
-                                print(iostr, "  $anim_char_colored $name\n")
+                                waiting = if dep in taskwaiting
+                                    color_string(" Waiting for background task, IO, or timer to finish. Use ctrl-c to interrupt and inspect warnings", Base.warn_color())
+                                else
+                                    ""
+                                end
+                                print(iostr, "  $anim_char_colored $name$waiting\n")
                             else
                                 print(iostr, "    $name\n")
                             end
@@ -1353,8 +1380,6 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
             fancyprint && print(io, ansi_enablecursor)
         end
     end
-    stderr_buffers = Dict{Base.PkgId,IOBuffer}()
-    stderr_outputs = Dict{Base.PkgId,String}()
     tasks = Task[]
     Base.LOADING_CACHE[] = Base.LoadingCache()
     ## precompilation loop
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -1275,6 +1275,29 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
         end
     end
 
+    stderr_buffers = Dict{Base.PkgId,IOBuffer}()
```
vtjnash, line 1278: IOBuffer is actually not a valid type to use for output (or often with Tasks in general) since it is lossy, I just keep forgetting to implement the error for it to instruct people toward the correct one. The correct type is Base.BufferStream and should mostly be a drop-in replacement (since they are both buffered IO objects).

IanButterworth: Ok, but I really don't understand the nuance between the two, and the IOBuffer docstring examples seem to be quite similar to how I use it https://docs.julialang.org/en/v1/base/io-network/#Base.IOBuffer

But, I've switched it out here.

src/API.jl
```diff
@@ -1329,7 +1352,12 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                                 # If not offset, on larger terminal fonts it looks odd that they all sync-up
                                 anim_char = anim_chars[(i + Int(dep.name[1])) % length(anim_chars) + 1]
                                 anim_char_colored = dep in direct_deps ? anim_char : color_string(anim_char, :light_black)
-                                print(iostr, "  $anim_char_colored $name\n")
+                                waiting = if dep in taskwaiting
+                                    color_string(" Waiting for trailing task to finish. Use ctrl-c to interrupt and inspect warnings", Base.warn_color())
```
vtjnash, line 1356: The word `trailing` does not convey a clear meaning to me. Maybe 'background' or 'async'? Perhaps saying 'background task, IO, or timer to finish` in full would also be even clearer?

src/API.jl
IanButterworth: > should mostly be a drop-in replacement

Unfortunately I hit the gap
```
MethodError: no method matching take!(::Base.BufferStream)
```

vtjnash: Yeah, `take!` is not something defined for most `IO`

vtjnash: Normally the method to call is `readline(io)` or `read(io, String)`

IanButterworth: Ok. `read` on an empty BufferStream blocks. What's the equivalent to `take!` which doesn't block on an empty IOBuffer?

src/API.jl
IanButterworth: `background task, IO, or timer to finish` sounds good

src/API.jl
IanButterworth: `readavailable` also blocks on an empty BufferStream. But perhaps I can query `bytesavailable` first.

```
julia> bytesavailable(iob)
0

julia> println(iob, "hello")

julia> bytesavailable(iob)
6
```

IanButterworth: updated

vtjnash: readline would be the normal function to use. If you are worried about blocking, you might be doing the work in the wrong task

IanButterworth: Ok, I think I have it in better shape now. thanks

IanButterworth:
(I thought https://github.com/JuliaLang/julia/pull/46571 was being backported. If it does, this should be too)

timholy:
LGTM, thanks!

IanButterworth:
Great. I think it's worth trialling this in the wild. Wording/formatting tweaks etc. could be in a future PR

oscardssmith:
This PR is a really nice improvement, but it does currently cause the progress bar refreshing to derp out a little (and write new lines instead of over-writing the same line). e.g.
```
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [======================================>  ]  291/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  292/307
  Progress [=======================================> ]  294/307
  Progress [=======================================> ]  294/307
```

IanButterworth:
It hasn't done in my testing. Can you provide any more detail?

Related to terminal width, or platform, perhaps?

oscardssmith:
I don't think it's width related although I'm not sure. I'm on Linux.

IanButterworth:
I suspect it's happening when the line with the message wraps around. It should be truncated to terminal width

oscardssmith:
Can confirm that the cause is line wrapping.

IanButterworth:
Thanks. Should be alleviated by #3370

Issue #1568, opened 2019-12-29
Bug: there is no way to `Pkg.add` a specific version of a JLL package


DilumAluthge:
Suppose I want to `Pkg.add` a specific version of a package. For example, suppose that I want to install version `0.5.3` of the `Example` package. I can do this as follows:
```julia
julia> import Pkg

julia> Pkg.add(Pkg.PackageSpec(name = "Example", version = "0.5.3"))
   Cloning default registries into `~/.julia`
   Cloning registry from "https://github.com/JuliaRegistries/General.git"
     Added registry `General` to `~/.julia/registries/General`
 Resolving package versions...
 Installed Example ─ v0.5.3
  Updating `~/.julia/environments/v1.4/Project.toml`
  [7876af07] + Example v0.5.3
  Updating `~/.julia/environments/v1.4/Manifest.toml`
  [7876af07] + Example v0.5.3
```

I can also accomplish the same thing using the REPL mode:
```julia
(@v1.4) pkg> add Example@0.5.3
   Cloning default registries into `~/.julia`
   Cloning registry from "https://github.com/JuliaRegistries/General.git"
     Added registry `General` to `~/.julia/registries/General`
 Resolving package versions...
 Installed Example ─ v0.5.3
  Updating `~/.julia/environments/v1.4/Project.toml`
  [7876af07] + Example v0.5.3
  Updating `~/.julia/environments/v1.4/Manifest.toml`
  [7876af07] + Example v0.5.3
```

But this does not work with JLL packages. For example, suppose that I want to install version `2.23.0+1` of the `Git_jll` package. I tried this, but it gives me an error:
```julia
julia> import Pkg

julia> Pkg.add(Pkg.PackageSpec(name = "Git_jll", version = "2.23.0+1"))
ERROR: ArgumentError: invalid version range: "2.23.0+1"
Stacktrace:
 [1] Pkg.Types.VersionRange(::String) at /Users/dilum/dev/forks-DilumAluthge/julia/usr/share/julia/stdlib/v1.4/Pkg/src/versions.jl:114
 [2] VersionSpec at /Users/dilum/dev/forks-DilumAluthge/julia/usr/share/julia/stdlib/v1.4/Pkg/src/versions.jl:197 [inlined]
 [3] #Package#129 at /Users/dilum/dev/forks-DilumAluthge/julia/usr/share/julia/stdlib/v1.4/Pkg/src/API.jl:996 [inlined]
[...]

PR #3368, opened 2023-02-11
Add build number support

IanButterworth:
Update: This seems to be in working shape now. See https://github.com/JuliaLang/Pkg.jl/pull/3368#issuecomment-1429151030

----

Opening this as an early WIP to discuss strategy here.

The initial proposal is to support only the addition of a numerical build number, indicated by a `+`, so how JLLs do it.
Otherwise, mixing prerelease strings in seems too tricky?

Implements https://github.com/JuliaLang/Pkg.jl/issues/1568


DilumAluthge:
1. Should we support multiple build numbers, or should we restrict to only a single build number?
2. Should we support non-integer build values, or should we restrict to only integer build values?

```julia
julia> v"1.2.3+4".build
(0x0000000000000004,)

julia> v"1.2.3+4.5.6".build
(0x0000000000000004, 0x0000000000000005, 0x0000000000000006)

julia> v"1.2.3+foo".build
("foo",)

julia> v"1.2.3+foo.bar.baz".build
("foo", "bar", "baz")
```

src/Versions.jl
```diff
@@ -8,34 +8,37 @@ export VersionBound, VersionRange, VersionSpec, semver_spec, isjoinable
...
+        n == 1 && return new((tin[1],      0,      0,      0), n)
+        n == 2 && return new((tin[1], tin[2],      0,      0), n)
+        n == 3 && return new((tin[1], tin[2], tin[3],      0), n)
+        n == 4 && return new((tin[1], tin[2], tin[3], tin[4]), n)
         error("invalid $n")
     end
 end
 VersionBound(t::Integer...) = VersionBound(t)
-VersionBound(v::VersionNumber) = VersionBound(v.major, v.minor, v.patch)
+VersionBound(v::VersionNumber) = VersionBound(v.major, v.minor, v.patch, v.build)
```
DilumAluthge, line 24: Is this correct? Won't `v.build` be a tuple (not an integer)?

IanButterworth: true

DilumAluthge: Also, what about the case where no build number is passed, and thus `v.build` is the empty tuple?

IanButterworth: I guess build should be a `Union{Nothing,Int}` to handle that

DilumAluthge: Well, `VersionBound` currently uses a `t::NTuple{4,UInt32}`. It would be annoying to try to stick a `nothing` into that. How about this?

```suggestion
function VersionBound(v::VersionNumber)
    if isempty(v.build)
        return VersionBound(v.major, v.minor, v.patch)
    else
        return VersionBound(v.major, v.minor, v.patch, only(v.build))
    end
end
```

IanButterworth: I'm not sure that avoids the compatability with `t::NTuple{4,UInt32}`

Also, instead of `only` it should detect `length > 1` and give a friendly error, I think

DilumAluthge, line 24: Or, to get even fancier:

```suggestion
function VersionBound(v::VersionNumber)
    if isempty(v.build)
        return VersionBound(v.major, v.minor, v.patch)
    elseif length(v.build) > 1
            pkgerror("We only support a single build number; multiple build numbers are not supported")
    else
        if !(v.build isa Integer)
            pkgerror("We only support integer values for the build data field")
        end
        return VersionBound(v.major, v.minor, v.patch, only(v.build))
    end
end
```

IanButterworth: Perhaps make VersionBound parametric `t::NTuple{V,UInt32}` where V is 3 or 4

DilumAluthge: Isn't that basically already what `VersionBound` does? They have a `t::NTuple{4,UInt32}`, but the "true length" is stored in the `n::Int` field. So if e.g. `n == 2`, then Pkg knows that `t[1]` and `t[2]` have meaningful data, but that `t[3]` should be ignored.

IanButterworth: There is an argument that they should silently ignore it all, which is what Pkg currently does
```
julia> Pkg.Types.VersionBound(v"1.2.3+4")
Pkg.Versions.VersionBound((0x00000001, 0x00000002, 0x00000003), 3)

julia> Pkg.Types.VersionBound(v"1.2.3+4.5")
Pkg.Versions.VersionBound((0x00000001, 0x00000002, 0x00000003), 3)

julia> Pkg.Types.VersionBound(v"1.2.3+4.5.6")
Pkg.Versions.VersionBound((0x00000001, 0x00000002, 0x00000003), 3)

julia> Pkg.Types.VersionBound(v"1.2.3+foo")
Pkg.Versions.VersionBound((0x00000001, 0x00000002, 0x00000003), 3)
```

DilumAluthge: Maybe print a `@debug`?

DilumAluthge, line 24:
```suggestion
function VersionBound(v::VersionNumber)
    if isempty(v.build)
        return VersionBound(v.major, v.minor, v.patch)
    end
    if length(v.build) > 1
        @debug "We only use the first build number; any additional build numbers are ignored" v v.build
    end
    if v.build[1] isa Integer
        return VersionBound(v.major, v.minor, v.patch, v.build[1])
    else
        @debug "We only support integer values for the build data field" v v.build v.build[1]
        return VersionBound(v.major, v.minor, v.patch)
    end
end
```

src/Versions.jl
```diff
@@ -8,34 +8,37 @@ export VersionBound, VersionRange, VersionSpec, semver_spec, isjoinable
...
-    t::NTuple{3,UInt32}
+    t::NTuple{4,UInt32}
     n::Int
     function VersionBound(tin::NTuple{n,Integer}) where n
-        n <= 3 || throw(ArgumentError("VersionBound: you can only specify major, minor and patch versions"))
-        n == 0 && return new((0,           0,      0), n)
-        n == 1 && return new((tin[1],      0,      0), n)
-        n == 2 && return new((tin[1], tin[2],      0), n)
-        n == 3 && return new((tin[1], tin[2], tin[3]), n)
+        n <= 4 || throw(ArgumentError("VersionBound: you can only specify major, minor, patch versions, and a build number"))
```
DilumAluthge, line 14:
```suggestion
        n <= 4 || throw(ArgumentError("VersionBound: you can only specify major, minor, patch versions, and a single positive-integer-valued build number"))
```

src/Versions.jl
```diff
@@ -8,34 +8,42 @@ export VersionBound, VersionRange, VersionSpec, semver_spec, isjoinable
...
+        n == 4 && return new((tin[1], tin[2], tin[3], tin[4]), n)
         error("invalid $n")
     end
 end
 VersionBound(t::Integer...) = VersionBound(t)
-VersionBound(v::VersionNumber) = VersionBound(v.major, v.minor, v.patch)
+function VersionBound(v::VersionNumber)
+    isempty(v.build) && return VersionBound(v.major, v.minor, v.patch)
+    length(v.build) > 1 && @debug "Pkg does not recognize build numbers beyond the first number. These will be ignored" v
+    !(v.build[1] isa Integer) && @debug "Pkg does not recognize string build fields. This will be ignored" v
```
DilumAluthge: You need to return early here. If you try to call the next line, the `VersionBound` constructor will throw an error when you try to convert a string to an `UInt32`.

src/Versions.jl
```diff
@@ -8,34 +8,46 @@ export VersionBound, VersionRange, VersionSpec, semver_spec, isjoinable
...
-    t::NTuple{3,UInt32}
+    t::NTuple{4,UInt32}
     n::Int
     function VersionBound(tin::NTuple{n,Integer}) where n
-        n <= 3 || throw(ArgumentError("VersionBound: you can only specify major, minor and patch versions"))
-        n == 0 && return new((0,           0,      0), n)
-        n == 1 && return new((tin[1],      0,      0), n)
-        n == 2 && return new((tin[1], tin[2],      0), n)
-        n == 3 && return new((tin[1], tin[2], tin[3]), n)
+        n <= 4 || throw(ArgumentError("VersionBound: you can only specify major, minor, patch versions, and a single positive-integer-valued build number"))
```
DilumAluthge, line 14: I made an error - of course we support zero as a build number.

```suggestion
        n <= 4 || throw(ArgumentError("VersionBound: you can only specify major, minor, patch versions, and a single non-negative-integer-valued build number"))
```

IanButterworth:
This appears to be working now

```
shell> cd ~/.julia/registries/General
/Users/ian/.julia/registries/General

shell> git checkout ae8ec3b695efb04ddec4371b97477779b6c62549
HEAD is now at ae8ec3b695 New version: OpenAPI v0.1.7 (#77093)

(Pkg) pkg> activate --temp
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF`

(jl_jFdSNF) pkg> add ImageMagick_jll@6.9.11
┌ Error: Some registries failed to update:
│     — `~/.julia/registries/General` — registry detached
└ @ Pkg.Registry ~/Documents/GitHub/Pkg.jl/src/Registry/Registry.jl:476
   Resolving package versions...
   Installed Ghostscript_jll ─ v9.55.0+3
   Installed OpenJpeg_jll ──── v2.4.0+0
   Installed ImageMagick_jll ─ v6.9.11+3
   Installed LittleCMS_jll ─── v2.12.0+0
  Downloaded artifact: OpenJpeg
  Downloaded artifact: ImageMagick
  Downloaded artifact: LittleCMS
  Downloaded artifact: Ghostscript
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Project.toml`
⌃ [c73af94c] + ImageMagick_jll v6.9.11+3
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Manifest.toml`
  [692b3bcd] + JLLWrappers v1.4.1
  [21216c6a] + Preferences v1.3.0
  [61579ee1] + Ghostscript_jll v9.55.0+3
⌃ [c73af94c] + ImageMagick_jll v6.9.11+3
  [aacddb02] + JpegTurbo_jll v2.1.2+0
  [88015f11] + LERC_jll v3.0.0+1
  [89763e89] + Libtiff_jll v4.4.0+0
  [d3a379c0] + LittleCMS_jll v2.12.0+0
  [643b3616] + OpenJpeg_jll v2.4.0+0
  [3161d3a3] + Zstd_jll v1.5.2+0
  [b53b4c65] + libpng_jll v1.6.38+0
  [0dad84c5] + ArgTools v1.1.1
  [56f22d72] + Artifacts
  [2a0f44e3] + Base64
  [ade2ca70] + Dates
  [f43a241f] + Downloads v1.6.0
  [7b1f6079] + FileWatching
  [b77e0a4c] + InteractiveUtils
  [b27032c2] + LibCURL v0.6.3
  [76f85450] + LibGit2
  [8f399da3] + Libdl
  [56ddb016] + Logging
  [d6f4376e] + Markdown
  [ca575930] + NetworkOptions v1.2.0
  [44cfe95a] + Pkg v1.8.0
  [de0858da] + Printf
  [3fa0cd96] + REPL
  [9a3f8284] + Random
  [ea8e919c] + SHA v0.7.0
  [9e88b42a] + Serialization
  [6462fe0b] + Sockets
  [fa267f1f] + TOML v1.0.3
  [a4e569a6] + Tar v1.10.0
  [cf7118a7] + UUIDs
  [4ec0a83e] + Unicode
  [deac9b47] + LibCURL_jll v7.84.0+0
  [29816b5a] + LibSSH2_jll v1.10.2+0
  [c8ffd9c3] + MbedTLS_jll v2.28.0+0
  [14a3606d] + MozillaCACerts_jll v2022.10.11
  [83775a58] + Zlib_jll v1.2.13+0
  [8e850ede] + nghttp2_jll v1.48.0+0
  [3f19e933] + p7zip_jll v17.4.0+0
Precompiling environment...
  9 dependencies successfully precompiled in 4 seconds. 5 already precompiled.

(jl_jFdSNF) pkg> add Example
   Resolving package versions...
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Project.toml`
  [7876af07] + Example v0.5.3
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Manifest.toml`
  [7876af07] + Example v0.5.3
Precompiling environment...
  1 dependency successfully precompiled in 1 seconds. 14 already precompiled.

(jl_jFdSNF) pkg> rm Example
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Project.toml`
  [7876af07] - Example v0.5.3
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Manifest.toml`
  [7876af07] - Example v0.5.3

shell> git checkout 3dd229c080a140cf67146ceb3ed6dba8add22ddf
Previous HEAD position was ae8ec3b695 New version: OpenAPI v0.1.7 (#77093)
HEAD is now at 3dd229c080 New version: ImageMagick_jll v6.9.11+4 (#77094)

(jl_jFdSNF) pkg> add Example
   Resolving package versions...
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Project.toml`
  [7876af07] + Example v0.5.3
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Manifest.toml`
  [7876af07] + Example v0.5.3

(jl_jFdSNF) pkg> add ImageMagick_jll@6.9.11+2
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package ImageMagick_jll [c73af94c]:
 ImageMagick_jll [c73af94c] log:
 ├─possible versions are: 6.9.10+0-7.1.0+0 or uninstalled
 └─restricted to versions 6.9.11+2 by an explicit requirement — no versions left

(jl_jFdSNF) pkg> add ImageMagick_jll@6.9.11+3
   Resolving package versions...
  No Changes to `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Project.toml`
  No Changes to `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Manifest.toml`

(jl_jFdSNF) pkg> add ImageMagick_jll@6.9.11+4
   Resolving package versions...
   Installed ImageMagick_jll ─ v6.9.11+4
  Downloaded artifact: ImageMagick
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Project.toml`
⌃ [c73af94c] ↑ ImageMagick_jll v6.9.11+3 ⇒ v6.9.11+4
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_jFdSNF/Manifest.toml`
⌃ [c73af94c] ↑ ImageMagick_jll v6.9.11+3 ⇒ v6.9.11+4
Precompiling environment...
  1 dependency successfully precompiled in 1 seconds. 14 already precompiled.
```

Contrast that to what currently happens https://github.com/JuliaLang/Pkg.jl/issues/1568#issuecomment-1425948204

KristofferC:
This has to be some "limited" support since e.g. the resolver is not updated here:

https://github.com/JuliaLang/Pkg.jl/blob/22c05b00445de98254300e7d39081610d51d2afa/src/Resolve/versionweights.jl#L3-L9

If the goal is just to make jlls not update on `add`, then something a bit more minimal as in https://github.com/JuliaLang/Pkg.jl/pull/3371 might be to prefer?

IanButterworth:
#3371 is clearly a better fix for this.

PR #3371, opened 2023-02-14
fixup jll versions that are supposed to stay fixed during e.g Pkg.add

KristofferC:
Alternative to https://github.com/JuliaLang/Pkg.jl/pull/3368. This is a bit of a more "surgical" fix to the issue of jlls getting updated from e.g. `Pkg.add` than trying to fully support build numbers in the resolver etc.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -402,7 +402,8 @@ function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryIn
     vers_fix = copy(vers)
     for (uuid, vers) in vers
         old_v = get(jll_fix, uuid, nothing)
-        if old_v !== nothing
+        # We only fixup a JLL if the old major/minor/patch matches the new major/minor/patch
+        if old_v !== nothing && Base.thispatch(old_v) == Base.thispatch(vers_fix[uuid])
             vers_fix[uuid] = old_v
         end
     end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -347,6 +347,14 @@ function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryIn
             @warn "julia version requirement for project not satisfied" _module=nothing _file=nothing
         end
     end
+
+    jll_fix = Dict{UUID, VersionNumber}()
+    for pkg in pkgs
+        if !is_stdlib(pkg.uuid) && endswith(pkg.name, "_jll") && pkg.version isa VersionNumber
+            jll_fix[pkg.uuid] = pkg.version
+        end
+    end
+
     names = Dict{UUID, String}(uuid => name for (uuid, (name, version)) in stdlibs())
     # recursive search for packages which are tracking a path
     developed = collect_developed(env, pkgs)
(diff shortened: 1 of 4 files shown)


IanButterworth:
You call that a diff?

LGTM!

DilumAluthge:
Can we also backport this to earlier Julia versions, such as 1.6?

KristofferC:
Probably yes

DilumAluthge (approved):

staticfloat (approved):

src/Operations.jl
```diff
@@ -390,6 +398,16 @@ function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryIn
     Resolve.simplify_graph!(graph)
     vers = Resolve.resolve(graph)
 
+    # Fixup jlls that got their build numbers stripped
+    vers_fix = copy(vers)
+    for (uuid, vers) in vers
+        old_v = get(jll_fix, uuid, nothing)
+        if old_v !== nothing
```
DilumAluthge, line 405:
```suggestion
        # We only fixup a JLL if the old major/minor/patch matches the new major/minor/patch
        if old_v !== nothing && Base.thispatch(old_v) == Base.thispatch(vers_fix[uuid])
```

DilumAluthge (commented):
It occurs to me that if the resolver changes the major/minor/patch of a JLL, then we shouldn't do the fixup, right?

E.g. if the resolver changes `1.2.3+0` to `1.2.3+1`, then we should do the fixup - i.e. we should change it back to `1.2.3+0`.

But if the resolver changes `1.2.3+0` to `1.2.4+1`, then we shouldn't do the fixup - i.e. we should keep it as `1.2.4+1`.

Does that sound right?

KristofferC:
> It occurs to me that if the resolver changes the major/minor/patch of a JLL, then we shouldn't do the fixup, right?

This should never happen though since we require that the package has the given version. So the only reason this code does anything is due to the build number stripping that happens when passing things into the resolver.

But we can be super defensive and check that anyway, yes.

IanButterworth (changes requested):
Looks good, just needs tests

test/new.jl
```diff
@@ -680,56 +680,85 @@ end
...
     # Preserve syntax
+
+    # helper for installing a specific JLL build, then moving the registry to a position where a new build exists
+    imjll_uuid = UUID("c73af94c-d91f-53ed-93a7-00f77d67a9d7")
+    function add_specific_jll_build()
+        cd(dirname(only(Pkg.Types.Context().registries).path)) do
+            Pkg.add(name="ImageMagick_jll", version=v"6.9.11+3")
+            @test Pkg.dependencies()[imjll_uuid].version == v"6.9.11+3"
+        end
+    end
```
IanButterworth, line 693: This isnt needed now. I'd just add and check in the test directly

test/new.jl
```diff
@@ -726,10 +744,13 @@ end
     isolate(loaded_depot=true) do
         Pkg.add(name="Example", version="0.3.0")
         @test Pkg.dependencies()[exuuid].version == v"0.3.0"
+        @test Pkg.dependencies()[imjll_uuid].version == v"6.9.11+3"
         Pkg.add(Pkg.PackageSpec(;name="JSON", version="0.18.0"); preserve=Pkg.PRESERVE_NONE)
         @test Pkg.dependencies()[exuuid].version == v"0.5.3"
         @test Pkg.dependencies()[json_uuid].version == v"0.18.0"
     end
+    Pkg.add(name="ImageMagick_jll", version=v"6.9.11+4")
+    @test Pkg.dependencies()[imjll_uuid].version == v"6.9.11+4"
```
IanButterworth, line 753: I think this should be inside an isolate?

IanButterworth (approved):

PR #3373, opened 2023-02-14
no longer consider DelimitedFiles an stdlib even though it is stored in stdlib folder

KristofferC:


DilumAluthge (commented):
This means that every time we move a new stdlib out, we need to update this file in Pkg, and then update Pkg in Julia master, right?

Could we instead have Pkg look for a file somewhere in Julia itself, e.g. `stdlib/not-stdlibs.txt`, and read the information from that? So then we don't have to update Pkg ever time we move a new stdlib out?

DilumAluthge (approved):
I guess we can just hard-code it for now, and then come up with a more sustainable solution later.

Issue #3369, opened 2023-02-11
`Pkg.add` should not update registry


jakobnissen:
Running `Pkg.add` should not automatically update the registry. Optionally add a warning when running `Pkg.add` if the registry is more than, say, 3 months old and Pkg is not in offline mode.

## Background
[A recent Discourse thread](https://discourse.julialang.org/t/first-pluto-notebook-launches-are-slower-on-julia-1-9-beta-3/93429) on the increased precompilation time with Julia 1.9 led to [some discussion on Slack](https://julialang.slack.com/archives/C03LD7K8HLN/p1676028682517259) about what can be done to reduce the amount of precompilation.

I don't think the Slack discussion reached a consensus, but several good suggestions were brought up that deserve an issue here on Pkg. Notably, some of the proposed solutions are good ideas _regardless_ of whether Julia 1.9's precompilation woes will be solved elsewhere.

## Motivation
The motivation is twofold: First, from a theoretical standpoint, adding a package and updating are conceptually different and orthogonal operations. You should be able to add a package without updating anything. Hence, the two operation should be separated.

From a practical point of view, developers and experienced users of Julia often create environments and modify their environments many times per day, sometimes tens of times per day. With the current state where every `Pkg.add` operation updates automatically, these users are going to be served new releases of various packages (dependencies of dependencies) very often, each time triggering waiting time for fetching the registry and/or package precompilation. For most users, getting the very newest version of every package is completely unnecessary - certainly getting up-to-the-hour updates is not needed.

## Possible objections
Q: Wouldn't this mean lots of people would get stuck on months old packages?
A: The registry would only get stale if they never choose to run `Pkg.update` for _any_ project. And if they never do so, it's probably safe to assume they are fine running an old environment. Furthermore, we should in general expect many users explicitly want to run years old environments that they choose not to upgrade or modify.

IanButterworth:
The summary that 

> the current state where every Pkg.add operation updates automatically, these users are going to be served new releases of various packages (dependencies of dependencies) very often

Isn't quite fair. Even if the registry is regularly updated, by default `Pkg.add` uses a tiered preserve strategy, where it tries to add the requested package while updating no existing package in the environment first, but if it cannot will go onto try the more permissive strategies (see https://pkgdocs.julialang.org/v1/api/#Pkg.add)

If `Pkg.add` has updated any existing dep, it should mean it couldn't add the package without doing that.

[...]

PR #3378, opened 2023-02-17
Add `PRESERVE_ALL_INSTALLED` and `PRESERVE_TIERED_INSTALLED` strategies

IanButterworth:
Updated:

New `PRESERVE_ALL_INSTALLED` preserve strategy that is like `PRESERVE_ALL` except it also only selects installed versions of the new packages. 

New `PRESERVE_TIERED_INSTALLED` strategy that tries `PRESERVE_ALL_INSTALLED` first.

To opt-in to `PRESERVE_TIERED_INSTALLED` being the default set the env var `JULIA_PKG_PRESERVE_TIERED_INSTALLED=true` 

Discussion in:
- https://discourse.julialang.org/t/first-pluto-notebook-launches-are-slower-on-julia-1-9-beta-3/93429/158
- https://github.com/JuliaLang/Pkg.jl/issues/3369

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,11 @@ Pkg v1.9 Release Notes
   or `pkg> add https://github.com/Org/Package.jl/commit/bb9eb77e6dc`.
 - Timing of the precompilation of dependencies can now be reported via `Pkg.precompile(timing=true)` ([#3334])
 - Bug fix on `pin/free --all` which now correctly applies to all dependencies, not just direct dependencies ([#3346]).
+- To reduce the amount of time spent downloading and precompiling new package versions when working with multiple
+  environments, a new preserve strategy `PRESERVE_ALL_INSTALLED` has been added which will preserve all existing
+  dependencies and only add versions of the new packages that are already installed. i.e. `pkg> add --preserve=installed Foo`.
+  Also a new tiered resolve strategy `PRESERVE_TIERED_INSTALLED` that tries this first, which can be set to the default
+  strategy by setting the env var `JULIA_PKG_PRESERVE_TIERED_INSTALLED` to `true` ([#3378]).
 
 Pkg v1.8 Release Notes
 ======================
--- a/src/API.jl
+++ b/src/API.jl
@@ -180,7 +180,7 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why, :
 end
 
 function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
-                 preserve::PreserveLevel=PRESERVE_TIERED, platform::AbstractPlatform=HostPlatform(), kwargs...)
+                 preserve::PreserveLevel=Operations.default_preserve(), platform::AbstractPlatform=HostPlatform(), kwargs...)
     require_not_empty(pkgs, :develop)
     Context!(ctx; kwargs...)
 
@@ -223,7 +223,7 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
     return
 end
 
-function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PRESERVE_TIERED,
+function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=Operations.default_preserve(),
              platform::AbstractPlatform=HostPlatform(), kwargs...)
     require_not_empty(pkgs, :add)
     Context!(ctx; kwargs...)
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -21,6 +21,14 @@ import ...Pkg: UPDATED_REGISTRY_THIS_SESSION, RESPECT_SYSIMAGE_VERSIONS, should_
 # Utils #
 #########
 
+function default_preserve()
+    if Base.get_bool_env("JULIA_PKG_PRESERVE_TIERED_INSTALLED", false)
+        PRESERVE_TIERED_INSTALLED
+    else
+        PRESERVE_TIERED
+    end
+end
+
 function find_installed(name::String, uuid::UUID, sha1::SHA1)
     slug_default = Base.version_slug(uuid, sha1)
     # 4 used to be the default so look there first
@@ -52,7 +60,7 @@ function load_version(version, fixed, preserve::PreserveLevel)
         return VersionSpec() # some stdlibs dont have a version
     elseif fixed
         return version # dont change state if a package is fixed
-    elseif preserve == PRESERVE_ALL || preserve == PRESERVE_DIRECT
+    elseif preserve == PRESERVE_ALL || preserve == PRESERVE_ALL_INSTALLED || preserve == PRESERVE_DIRECT
         return something(version, VersionSpec())
     elseif preserve == PRESERVE_SEMVER && version != VersionSpec()
         return Types.semver_spec("$(version.major).$(version.minor).$(version.patch)")
@@ -337,7 +345,9 @@ dropbuild(v::VersionNumber) = VersionNumber(v.major, v.minor, v.patch, isempty(v
 # sets version to a VersionNumber
 # adds any other packages which may be in the dependency graph
 # all versioned packages should have a `tree_hash`
-function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryInstance}, pkgs::Vector{PackageSpec}, julia_version)
+function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryInstance}, pkgs::Vector{PackageSpec}, julia_version,
+                           installed_only::Bool)
+    installed_only = installed_only || OFFLINE_MODE[]
     # compatibility
     if julia_version !== nothing
         # only set the manifest julia_version if ctx.julia_version is not nothing
@@ -394,7 +404,7 @@ function resolve_versions!(env::EnvCache, registries::Vector{Registry.RegistryIn
     # happened on a different julia version / commit and the stdlib version in the manifest is not the current stdlib version
     unbind_stdlibs = julia_version === VERSION
     reqs = Resolve.Requires(pkg.uuid => is_stdlib(pkg.uuid) && unbind_stdlibs ? VersionSpec("*") : VersionSpec(pkg.version) for pkg in pkgs)
-    graph, compat_map = deps_graph(env, registries, names, reqs, fixed, julia_version)
+    graph, compat_map = deps_graph(env, registries, names, reqs, fixed, julia_version, installed_only)
     Resolve.simplify_graph!(graph)
     vers = Resolve.resolve(graph)
 
@@ -451,7 +461,8 @@ get_or_make!(d::Dict{K,V}, k::K) where {K,V} = get!(d, k) do; V() end
 const JULIA_UUID = UUID("1222c4b2-2114-5bfd-aeef-88e4692bbb3e")
 const PKGORIGIN_HAVE_VERSION = :version in fieldnames(Base.PkgOrigin)
 function deps_graph(env::EnvCache, registries::Vector{Registry.RegistryInstance}, uuid_to_name::Dict{UUID,String},
-                    reqs::Resolve.Requires, fixed::Dict{UUID,Resolve.Fixed}, julia_version)
+                    reqs::Resolve.Requires, fixed::Dict{UUID,Resolve.Fixed}, julia_version,
+                    installed_only::Bool)
     uuids = Set{UUID}()
     union!(uuids, keys(reqs))
     union!(uuids, keys(fixed))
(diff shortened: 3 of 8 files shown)


visr:
Thanks for working on this! For context for readers, see #3369 and [this long Discourse thread](https://discourse.julialang.org/t/first-pluto-notebook-launches-are-slower-on-julia-1-9-beta-3/93429).

If I understand this correctly, `add --installed Example` will avoid a registry update and attempt to add Example in offline mode, after which it restores the offline status.

This does improve on the UI of going offline for an add, but I think the driving use case here is wanting to add packages as quickly as possible. That would suggest falling back to an online add instead of an error. That also avoids having to guess if the package is already installed in the depot.

When I `add --installed` multiple packages at once, ideally it adds any already installed packages and installs the others. With this implementation that doesn't seem possible, as you either do an online or offline add for all packages. With that in mind perhaps implementing a new top tier in the resolver for installed packages like in https://github.com/JuliaLang/Pkg.jl/issues/3369#issuecomment-1429008826 seems preferable to me. If including this tier by default is considered undesirable or too soon we could easily add an environment variable for users to opt in to this behavior.

KristofferC:
> With that in mind perhaps implementing a new top tier in the resolver for installed packages like in https://github.com/JuliaLang/Pkg.jl/issues/3369#issuecomment-1429008826 seems preferable to me. If including this tier by default is considered undesirable or too soon we could easily add an environment variable for users to opt in to this behavior.

Using the tier structure for this makes sense to me.

IanButterworth:
I agree, it's an elegant suggestion, but I think that requires changing the default behavior, as the default of tiered would naturally start at the top. 

Or is there a way to not change the default?

IanButterworth:
I've pushed updated docs for an approach where a new `max` preserve strategy can be enabled. The implementation isn't done yet, but feedback is welcome before that's done

CHANGELOG.md
```diff
@@ -12,6 +12,8 @@ Pkg v1.9 Release Notes
   or `pkg> add https://github.com/Org/Package.jl/commit/bb9eb77e6dc`.
 - Timing of the precompilation of dependencies can now be reported via `Pkg.precompile(timing=true)` ([#3334])
 - Bug fix on `pin/free --all` which now correctly applies to all dependencies, not just direct dependencies ([#3346]).
+- New `add --installed Foo` or `Pkg.add("Foo", allow_max=true)` flag to allow the tiered resolver to first try to
```
timholy, line 15: I get why these aren't named the same, but it's a predictable source of confusion.

src/Pkg.jl
```diff
@@ -117,18 +117,34 @@ a package, also inside that package.
...
-| Value             | Description                                                                         |
-|:------------------|:------------------------------------------------------------------------------------|
-| `PRESERVE_ALL`    | Preserve the state of all existing dependencies (including recursive dependencies)  |
-| `PRESERVE_DIRECT` | Preserve the state of all existing direct dependencies                              |
-| `PRESERVE_SEMVER` | Preserve semver-compatible versions of direct dependencies                          |
-| `PRESERVE_NONE`   | Do not attempt to preserve any version information                                  |
-| `PRESERVE_TIERED` | Use the tier which will preserve the most version information (this is the default) |
+| Value             | Description                                                                          |
+|:------------------|:-------------------------------------------------------------------------------------|
+| `PRESERVE_MAX`    | Preserve the state of all existing dependencies and only add those already installed |
```
timholy, line 122: Maybe `PRESERVE_INSTALLED`? `PRESERVE_LOCAL`? `PRESERVE_AVAILABLE`? `MAX` seems a bit non-specific.

src/Pkg.jl
```diff
@@ -117,18 +117,34 @@ a package, also inside that package.
...
+
+!!! compat "Julia 1.9"
+    The `PRESERVE_MAX` strategy and `allow_max` option requires at least Julia 1.9.
 
 # Examples
 ```julia
 Pkg.add("Example") # Add a package from registry
-Pkg.add("Example"; preserve=Pkg.PRESERVE_ALL) # Add the `Example` package and preserve existing dependencies
+Pkg.add("Example"; preserve=Pkg.PRESERVE_ALL) # Add the `Example` package and strictly preserve existing dependencies
+Pkg.add("Example"; allow_max=true) # First try to add only the installed version of `Example` and its dependencies
```
timholy, line 147: This sounds more permissive (more tolerant) than `PRESERVE_ALL`. I'm guessing it's actually less tolerant?

timholy (commented):
This is mostly about clarity of terms, no comments on the actual implementation.

src/Pkg.jl
IanButterworth: Yeah the tricky thing is conveying it includes `all` AND the installed thing

IanButterworth:
This is working now:
```
julia> ENV["JULIA_DEBUG"]="Pkg"
"Pkg"

(Pkg) pkg> activate --temp
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2Zz11A`

(jl_2Zz11A) pkg> add CSV@v0.10.8
   Resolving package versions...
┌ Debug: tiered_resolve: trying PRESERVE_ALL
└ @ Pkg.Operations ~/Documents/GitHub/Pkg.jl/src/Operations.jl:1297
   Installed CSV ─ v0.10.8
  No Changes to `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2Zz11A/Project.toml`
  No Changes to `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_2Zz11A/Manifest.toml`
Precompiling environment...
  1 dependency successfully precompiled in 17 seconds. 20 already precompiled.

(jl_2Zz11A) pkg> activate --temp
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_uuXGfM`

(jl_uuXGfM) pkg> add --installed CSV
   Resolving package versions...
┌ Debug: tiered_resolve: trying PRESERVE_ALL_INSTALLED
└ @ Pkg.Operations ~/Documents/GitHub/Pkg.jl/src/Operations.jl:1290
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_uuXGfM/Project.toml`
⌃ [336ed68f] + CSV v0.10.8
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_uuXGfM/Manifest.toml`
...

(jl_uuXGfM) pkg> activate --temp
  Activating new project at `/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_qIkpLK`

(jl_qIkpLK) pkg> add CSV
   Resolving package versions...
┌ Debug: tiered_resolve: trying PRESERVE_ALL
└ @ Pkg.Operations ~/Documents/GitHub/Pkg.jl/src/Operations.jl:1297
   Installed CSV ─ v0.10.9
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_qIkpLK/Project.toml`
  [336ed68f] + CSV v0.10.9
    Updating `/private/var/folders/_6/1yf6sj0950vcg4t91m9ltb5w0000gn/T/jl_qIkpLK/Manifest.toml`
...
Precompiling environment...
  1 dependency successfully precompiled in 17 seconds. 20 already precompiled.
```

src/Pkg.jl
IanButterworth: Updated to `PRESERVE_ALL_INSTALLED`

src/Pkg.jl
IanButterworth: `allow_max` updated to simply `installed`,to match the `add --installed Example` form

CHANGELOG.md
IanButterworth: Yeah. I've made them both `installed` now

src/API.jl
```diff
@@ -224,7 +233,13 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
 end
 
 function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=PRESERVE_TIERED,
-             platform::AbstractPlatform=HostPlatform(), kwargs...)
+             platform::AbstractPlatform=HostPlatform(), installed::Union{Bool,Nothing}=nothing, kwargs...)
+
+    if installed === true && !in(preserve, (PRESERVE_TIERED, PRESERVE_ALL_INSTALLED))
```
KristofferC, line 238: Could this be moved to `tiered_resolve` so it doesn't have to be repeated in two places (both `develop` and `add`).?

src/API.jl
```diff
@@ -179,8 +179,17 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why, :
...
+
 function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
-                 preserve::PreserveLevel=PRESERVE_TIERED, platform::AbstractPlatform=HostPlatform(), kwargs...)
+                 preserve::PreserveLevel=PRESERVE_TIERED, platform::AbstractPlatform=HostPlatform(),
+                 installed::Union{Bool,Nothing}=nothing, kwargs...)
+
+    if installed === true && !in(preserve, (PRESERVE_TIERED, PRESERVE_ALL_INSTALLED))
+        pkgerror("`installed` mode requires a preserve strategy that is either `PRESERVE_TIERED` or `PRESERVE_ALL_INSTALLED`")
+    end
+    installed = @something installed Base.get_bool_env("JULIA_PKG_ADD_INSTALLED", false)
```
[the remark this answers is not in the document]
IanButterworth: It's in the `add` docstring as a note

src/Operations.jl
```diff
@@ -1546,7 +1574,9 @@ function pin(ctx::Context, pkgs::Vector{PackageSpec})
     foreach(pkg -> update_package_pin!(ctx.registries, pkg, manifest_info(ctx.env.manifest, pkg.uuid)), pkgs)
     pkgs = load_direct_deps(ctx.env, pkgs)
 
-    pkgs, deps_map = _resolve(ctx.io, ctx.env, ctx.registries, pkgs, PRESERVE_TIERED, ctx.julia_version)
+    # TODO: maybe this needs to allow PRESERVE_ALL_INSTALLED (disabled with false)
+    pkgs, deps_map = _resolve(ctx.io, ctx.env, ctx.registries, pkgs, PRESERVE_TIERED, ctx.julia_version, false)
```
KristofferC, line 1578: Imo we should change `pin` to not take a version and just have it pin on the current version. Then there is no need to resolve after a `pin`.

src/Pkg.jl
```diff
@@ -105,8 +105,8 @@ const PreserveLevel = Types.PreserveLevel
 
 # Define new variables so tab comleting Pkg. works.
 """
-    Pkg.add(pkg::Union{String, Vector{String}}; preserve=PRESERVE_TIERED)
-    Pkg.add(pkg::Union{PackageSpec, Vector{PackageSpec}}; preserve=PRESERVE_TIERED)
+    Pkg.add(pkg::Union{String, Vector{String}}; preserve=PRESERVE_TIERED, installed=false)
+    Pkg.add(pkg::Union{PackageSpec, Vector{PackageSpec}}; preserve=PRESERVE_TIERED, installed=false)
```
KristofferC, line 109: I don't understand why this extra keyword argument is added. Shouldn't this new feature just be a new `preserve`, `PRESERVE_TIERED_INSTALLED` or something that tells Pkg to first try the installed option and then go for the normal tiered resolve?

IanButterworth: It's added because the opinion seemed to be that it would be better to not change the default behavior of the tiered resolver. This arg is a specific opt-in, vs. The env var which is general

src/API.jl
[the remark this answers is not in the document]
IanButterworth: Ah, it was a typo

src/Operations.jl
```diff
@@ -1302,27 +1302,51 @@ function assert_can_add(ctx::Context, pkgs::Vector{PackageSpec})
...
+    @debug "tiered_resolve: trying PRESERVE_NONE"
     return targeted_resolve(env, registries, pkgs, PRESERVE_NONE, julia_version)
 end
 
 function targeted_resolve(env::EnvCache, registries::Vector{Registry.RegistryInstance}, pkgs::Vector{PackageSpec}, preserve::PreserveLevel, julia_version)
-    if preserve == PRESERVE_ALL
+
+    if preserve == PRESERVE_ALL_INSTALLED
+        was_offline = OFFLINE_MODE[]
+        OFFLINE_MODE[] = true
```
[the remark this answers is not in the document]
IanButterworth: Yeah that sounds good. I'll give it a try

src/Pkg.jl
IanButterworth: Oh, and map `--installed` to set preserve to `PRESERVE_TIERED_INSTALLED`? That makes sense

IanButterworth: Updated. I think `pkg> add --preserve=installed Foo` is short enough for local opt-in

src/Operations.jl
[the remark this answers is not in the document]
IanButterworth: done

src/Operations.jl
IanButterworth: Sounds good. Another PR?

KristofferC (approved):
LGTM

PR #3381, opened 2023-02-17
add test for JLL build no. fix PR

IanButterworth:
@KristofferC a test proposal for https://github.com/JuliaLang/Pkg.jl/pull/3371

I think the correct answer is that in some of these later strategies, the jll version should be updated, so I think this needs changes

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -680,55 +680,90 @@ end
             @test pkg.version == v"0.3.0"
         end
     end end
+
     # Preserve syntax
+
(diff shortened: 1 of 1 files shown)


KristofferC:
> so I think this needs changes

You mean that https://github.com/JuliaLang/Pkg.jl/pull/3371 needs changes?

IanButterworth:
No I mean i suspect some of these later tests should have their test values changed. I haven't run this, and I guess CI doesn't run on PRs to non master branches.

PR #3382, opened 2023-02-17
Some improvements to documentation about creating packages

### LINKED ISSUE #3379: Small  documentation fix for Pkg.jl documentation to aid understanding of new users.
I've been struggling for a while with how to do "multiple file projects" in the way I do in Python/Fortran/whatever in Julia - i.e. by having local packages - and due to some experimentation I think I've worked out what the problem is.

In the `Pkg.jl` docmentation on how to create a package (https://github.com/JuliaLang/Pkg.jl/blob/master/docs/src/creating-packages.md) there is an important step which looks like a "not important" step, namely:

```
julia> cd("HelloWorld")

shell> tree .
.
├── Project.toml
└── src
    └── HelloWorld.jl

1 directory, 2 files
```

To a naive reader of the documentation, this looks like a box they can skip over, because it's just showing the layout of the file-system, however as you all (and now I!) know it's actually important to execute the `cd` command or else nothing that follows will work.

My suggestion would be to make it clear that the `cd` is an important step by having a separate box for it, distinct from the `tree .` command and making clear its necessity.

giordano:
Clarify confusion about directories, make the tutorial more easy to follow by automatically generating the files that are otherwise supposed to be manually created by the user.

Fix #3379.  CC: @owainkenwayucl

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -14,18 +14,16 @@ To generate the bare minimum files for a new package, use `pkg> generate`.
 (@v1.8) pkg> generate HelloWorld
 ```
 
-This creates a new project `HelloWorld` with the following files (visualized with the external [`tree` command](https://linux.die.net/man/1/tree)):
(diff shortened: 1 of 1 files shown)


owainkenwayucl:
That change definitely clears it up for me.

Issue #3380, opened 2023-02-17
Autoname extensions


aplavin:
Basically everyone would follow the recommended naming scheme for extensions: extension triggered by `SomePkg` should be named `SomePkgExt`. Can this naming be done automatically to reduce repetition?

Currently, there are lots of unnecessary repetitions in the common case:
- `Project.toml` `[extensions]`
  - extension name `SomePkgExt`
  - package name `= SomePkg`
- Code
  - file name `SomePkgExt.jl`
  - module name `module SomePkgExt`

Maybe, at least some of these could be removed?
For example, instead of `SomePkgExt = SomePkg` allow just `SomePkg` in `[extensions`]?
And/or implicitly define `module SomePkgExt` in the corresponding file?

fredrikekre:
I like that things are explicit, it also isn't more repetitive than the package itself. I think it is also nice if there is just one way to specify the extension since it is enough to learn one way.

In addition, this is something you set it up once and then don't have to bother with it again, and not something you have to type everytime you use an extension.

KristofferC:
> For one, arbitrary names without a convention would prevent building sysimages

Due to some existing deficiencies. There is nothing inherently problematic about it.

PR #3386, opened 2023-02-22
Create autonames.jl

GOVINDFROMINDIA:
Implements https://github.com/JuliaLang/Pkg.jl/issues/3380

Currently, there are lots of unnecessary repetitions in the common case:

Project.toml [extensions]
extension name SomePkgExt
package name = SomePkg
Code
file name SomePkgExt.jl
module name module SomePkgExt

Automatic naming schemes might not always work for all packages, especially those with non-standard names or naming conventions. Define a function that takes the package name as input and returns the corresponding extension name

### DIFF
--- a/autonames.jl
+++ b/autonames.jl
@@ -1,0 +1,8 @@
+function generate_extension_name()
+    # Extract the package name from the current environment
+    pkgname = split.(keys(Base.loaded_modules))[end][1]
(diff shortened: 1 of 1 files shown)


fredrikekre:
I don't understand what this is supposed to do, and it doesn't look very useful towards potentially solveing the linked issue so I am closing.

PR #3393, opened 2023-02-27
Fix usage instructions of weak deps with Requires.jl

maleadt:
Doing both a top level in a single block yields a `LoadError: UndefVarError: @require not defined` (`@static` doesn't help there).


IanButterworth:
It's surprising to me that the way it was doesn't work. Perhaps a comment in the code example would help people understand the issue?

IanButterworth:
Makes sense. Thanks

PR #3401, opened 2023-03-06
rejigger some code paths that handles untyped data to avoid invalidations

KristofferC:


timholy (commented):
LGTM but you know the types involved far better than I.

PR #3408, opened 2023-03-14
Verify git-tree-sha1 of registry downloads

fredrikekre:
This patch adds verification of registries downloaded from Pkg server when using the default method of keeping registries as compressed tarballs. This verification is done before the registry is moved in place to ensure that a Pkg server that serves bad content does not also poison the users local machine.

### DIFF
--- a/src/PlatformEngines.jl
+++ b/src/PlatformEngines.jl
@@ -654,4 +654,22 @@ function verify(path::AbstractString, hash::AbstractString; verbose::Bool = fals
     end
 end
 
+# Verify the git-tree-sha1 hash of a compressed archive.
+function verify_archive_tree_hash(tar_gz::AbstractString, expected_hash::Base.SHA1)
+    # This can fail because unlike sha256 verification of the downloaded
+    # tarball, tree hash verification requires that the file can i) be
+    # decompressed and ii) is a proper archive.
+    calc_hash = try
+        Base.SHA1(open(Tar.tree_hash, `$(exe7z()) x $tar_gz -so`))
+    catch err
+        @warn "unable to decompress and read archive" exception=err
+        return false
+    end
+    if calc_hash != expected_hash
+        @warn "tarball content does not match expected git-tree-sha1"
+        return false
+    end
+    return true
+end
+
 end # module PlatformEngines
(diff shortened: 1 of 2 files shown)


staticfloat (approved):

IanButterworth:
How expensive is the uncompress? Is it worth passing that forward out of the verify function for later use?

fredrikekre:
I suppose if there is a way to `tee` the stream it migh improve a bit. Here are some timings (naively would have thought the hash computation would be faster):
```
julia> @time Downloads.download("https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/9226550d425f68a52fb15981e5601a6fef34ffcd", "reg.tar.gz");
  0.554557 seconds (5.65 k allocations: 5.920 MiB)

julia> @time Pkg.PlatformEngines.verify_archive_tree_hash("reg.tar.gz", Base.SHA1("9226550d425f68a52fb15981e5601a6fef34ffcd"));
  0.479317 seconds (3.41 M allocations: 200.620 MiB, 17.95% gc time)
```

IanButterworth:
Timing on my i7 2018 MBP
```
julia> @time Downloads.download("https://pkg.julialang.org/registry/23338594-aafe-5451-b93e-139f81909106/9226550d425f68a52fb15981e5601a6fef34ffcd", "reg.tar.gz");
  2.590834 seconds (17.59 k allocations: 6.416 MiB, 2.76% compilation time)

julia> @time Pkg.PlatformEngines.verify_archive_tree_hash("reg.tar.gz", Base.SHA1("9226550d425f68a52fb15981e5601a6fef34ffcd"));
  1.070156 seconds (3.80 M allocations: 224.332 MiB, 9.52% gc time, 48.54% compilation time)
```

fredrikekre:
Thanks. I don't think the verification will be noticeable in general then. In practice this will only happen if you run `Pkg.update` and there is a new registry. That operation likely will result in running the resolver, downloading new packages and artifacts, decompress and unpack, precompilation, etc.

fredrikekre:
Well, this came back to bite me (https://github.com/fredrikekre/jlpkg runs with `--compile=min`):
```julia
$ time JULIA_DEPOT_PATH=$(mktemp -d) julia --compile=min -e 'import Pkg; Pkg.Registry.add()'
real    5m56,043s
user    5m53,843s
sys     0m0,754s
```

IanButterworth:
What does a profile look like? Any low hanging fruit?

IanButterworth:
@fredrikekre should fixing this this hold up updating Pkg? https://github.com/JuliaLang/julia/pull/49295

KristofferC:
No, `compile=min` is not something g holding anythi g back. We could precompile this function presumably.

PR #3416, opened 2023-03-20
ENV variable to set number of concurrent requests

vdayanand:


src/Types.jl
```diff
@@ -382,13 +382,20 @@ end
 include("project.jl")
 include("manifest.jl")
 
+function num_concurrent_downloads()
+    num = parse(Int, get(ENV, "JULIA_PKG_CONCURRENT_DOWNLOADS", "8"))
```
KristofferC, line 386: Could also use `tryparse` here and give a good error message if it doesn't parse as an int.

vdayanand: Updated

vdayanand:
Thanks, I have added docs as well

PR #3429, opened 2023-03-31
try_prompt_pkg_add: on "other" default to 2nd env

StefanKarpinski:
When someone picks "o" from the y/n/o choices, they are explicitly skipping the first environment in the load path, so we should start at the second writable environment in the load path since that's usually the one someone wants to add the package to.


IanButterworth:
Seems like a bugfix so marking for backport

Issue #3256, opened 2022-11-11
Feature request: shortcuts for commands activate, instantiate, etc...


simonmandlik:
We have `up` for `update`.

Would it be possible to add `act` for `activate`, `res` for `resolve` and possibly others?

When making use of the great Julia package manager to quickly switch between environments, I find that typing out these rather long commands slows me down, not to mention that I often make typos.

fredrikekre:
`ac<TAB>`, `res<TAB>`.

What is your workflow if you use these commands so often that you feel they slow you down? I find that I very rarely use these when programming.

simonmandlik:
Yeah, that would work too, thanks, though it's one more keystroke.

Most often I find myself typing out these commands when quickly testing or benchmarking same code in different versions of julia. I usually run `activate --temp` and then have to `resolve` and possibly `instantiate` (or even delete the `Manifest.toml` for earlier julia versions). I guess for proper extensive benchmarking it makes sense to setup these envs separately, but this became my go-to solution for quick testing like this over time.

IanButterworth:
I don't think this is needed now we have tab completion hints. It's pretty self documenting.

PR #3440, opened 2023-04-16
Allow `act` in place of `activate` at the REPL

pbouffard:
The name says it all. I find I actually `activate` more than some of the other commands that are either already short like `add` or have short names like `st`.


IanButterworth:
Discussion: https://github.com/JuliaLang/Pkg.jl/issues/3256

IanButterworth:
I don't think this is needed now we have tab completion hints.

PR #3445, opened 2023-04-21
no longer specialize on the default IO used

KristofferC:
This changes time to first add (with Pkg outside the sysimage) from:

```
2.791889 seconds (4.83 M allocations: 322.060 MiB, 2.44% gc time, 51.78% compilation time)
```

to

```
1.430343 seconds (3.69 M allocations: 241.599 MiB, 3.67% gc time, 62.49% compilation time)
```

Pkg doesn't really print a huge number of things but I am a little bit worried that we are making invalidations a lot worse now with possibility of a lot of type instabilities in the printing code.


src/Pkg.jl
```diff
@@ -40,11 +40,17 @@ const OFFLINE_MODE = Ref(false)
...
-stderr_f() = something(DEFAULT_IO[], stderr)
-stdout_f() = something(DEFAULT_IO[], stdout)
+
+struct UnstableIO <: IO
+    io::IO
+end
+Base.write(io::UnstableIO, b::UInt8) = write(io.io, b)::Int
+Base.get(io::UnstableIO, val, default) = get(io.io, val, default)
+stderr_f() = something(DEFAULT_IO[], UnstableIO(stderr))
+stdout_f() = something(DEFAULT_IO[], UnstableIO(stderr))
```
IanButterworth, line 50:
```suggestion
stdout_f() = something(DEFAULT_IO[], UnstableIO(stdout))
```

vtjnash (commented):
You could slap an `invokelatest` in those to make sure, but I assume inference would give up  already on inferring them.

PR #3452, opened 2023-04-27
Add more info to deprecation warning of `installed`

KronosTheLate:
It is not clear from the current warning what should be used instead. This PR adds that information.
An example is also added to show how to get a list of direct dependency names. This example is 
a little longer than what all users can be expected to figure out themselves.


KronosTheLate:
Bump

src/Pkg.jl
```diff
@@ -724,7 +724,10 @@ end
 ################
 
 function installed()
-    @warn "Pkg.installed() is deprecated"
+    @warn "Pkg.installed() is deprecated. Use Pkg.dependencies() instead.
+    To get a list of package names that have been explicitly added to the 
+    current project (direct dependencies), you can do:
+    `[dep.second.name for dep in Pkg.dependencies() if dep.second.is_direct_dep]`."
```
IanButterworth:
```suggestion
    @warn "`Pkg.installed()` is deprecated. Use `Pkg.dependencies()` instead." maxlog=1
```

IanButterworth: I think the latter part is for `Pkg.dependencies()` docs, not here.

Issue #3453, opened 2023-04-28
Simultaneous writing to manifest_usage on NFS crashes Julia


lhupe:
We are having issues with Julia processes randomly crashing when simultaneously activating environments from different machines that share a NFS file system, due to IO issues when writing to the manifest usage log.

The error messages  generally look something like this (this particular example is on julia 1.9.0-rc2, but we've been seeing similar issues on different versions):
```
  Activating project at `/path/to/project`
┌ Warning: attempting to remove probably stale pidfile
│   path = "/path/to/.julia/logs/manifest_usage.toml.pid"
└ @ FileWatching.Pidfile /path/to/.julia/juliaup/julia-1.9.0-rc2+0.x64.linux.gnu/share/julia/stdlib/v1.9/FileWatching/src/pidfile.jl:244
ERROR: LoadError: IOError: stat(RawFD(16)): Unknown system error -116 (Unknown system error -116)
Stacktrace:
  [1] uv_error
    @ ./libuv.jl:100 [inlined]
  [2] stat(fd::RawFD)
    @ Base.Filesystem ./stat.jl:152
  [3] stat
    @ ./filesystem.jl:280 [inlined]
  [4] close(lock::FileWatching.Pidfile.LockMonitor)
    @ FileWatching.Pidfile /path/to/.julia/juliaup/julia-1.9.0-rc2+0.x64.linux.gnu/share/julia/stdlib/v1.9/FileWatching/src/pidfile.jl:307
  [5] mkpidlock(f::Pkg.Types.var"#51#54"{String, Dates.DateTime, String}, at::String, pid::Int32; kwopts::Base.Pairs{Symbol, Int64, Tuple{Symbol}, NamedTuple{(:stale_age,), Tuple{Int64}}})
    @ FileWatching.Pidfile /path/to/.julia/juliaup/julia-1.9.0-rc2+0.x64.linux.gnu/share/julia/stdlib/v1.9/FileWatching/src/pidfile.jl:84
  [6] mkpidlock
    @ /path/to/.julia/juliaup/julia-1.9.0-rc2+0.x64.linux.gnu/share/julia/stdlib/v1.9/FileWatching/src/pidfile.jl:79 [inlined]
  [7] #mkpidlock#6
    @ /path/to/.julia/juliaup/julia-1.9.0-rc2+0.x64.linux.gnu/share/julia/stdlib/v1.9/FileWatching/src/pidfile.jl:77 [inlined]
  [8] mkpidlock
[...]

PR #3454, opened 2023-04-28
Make usage file creation and parsing errors warnings

IanButterworth:
Suggested fix/mitigation for #3453


vtjnash (changes requested):
This should not probably not be just hacked around in Pkg, but actually fixed

Issue #3443, opened 2023-04-20
precompile not handling extensions correctly


vtjnash:
I see this issue pretty often during Pkg.precompile. My assumption is that Pkg fails to handle extensions explicitly when forming the topological sorting of the packages for the precompile call, so they end up getting simultaneously compiled by multiple downstream triggers, getting itself into a bad intermediate state because they are getting loaded during incremental compile in multiple subprocesses, but not tracked during Pkg.precompile to perform that work in advance:
```
┌ SimpleChains [de6bee2f-e2f4-4ec7-b6ed-219cc6f6e9e5]
│  ┌ Warning: Module SpecialFunctionsExt with build ID fafbfcfd-dfa3-1ca9-0000-f3387dd333bf is missing from the cache.
│  │ This may mean SpecialFunctionsExt [499d2952-c3ce-5339-96ee-b197ee6daaa6] does not support precompilation but is imported by a module that does.
│  └ @ Base loading.jl:1791
│  ┌ Error: Error during loading of extension SpecialFunctionsExt of LoopVectorization, use `Base.retry_load_extensions()` to retry.
│  │   exception =
│  │    1-element ExceptionStack:
│  │    Declaring __precompile__(false) is not allowed in files that are being precompiled.
│  │    Stacktrace:
│  │      [1] _require(pkg::Base.PkgId, env::Nothing)
│  │        @ Base ./loading.jl:1795
│  │      [2] _require_prelocked(uuidkey::Base.PkgId, env::Nothing)
│  │        @ Base ./loading.jl:1655
│  │      [3] _require_prelocked(uuidkey::Base.PkgId)
│  │        @ Base ./loading.jl:1653
│  │      [4] run_extension_callbacks(extid::Base.ExtensionId)
│  │        @ Base ./loading.jl:1195
│  │      [5] run_extension_callbacks(pkgid::Base.PkgId)
│  │        @ Base ./loading.jl:1250
│  │      [6] run_package_callbacks(modkey::Base.PkgId)
│  │        @ Base ./loading.jl:1083
│  │      [7] _require_prelocked(uuidkey::Base.PkgId, env::String)
│  │        @ Base ./loading.jl:1662
│  │      [8] macro expansion
│  │        @ ./loading.jl:1641 [inlined]
│  │      [9] macro expansion
│  │        @ ./lock.jl:267 [inlined]
│  │     [10] require(into::Module, mod::Symbol)
│  │        @ Base ./loading.jl:1604
[...]

PR #3456, opened 2023-05-03
precompile: fix race in precompiling exts

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3443


IanButterworth:
This should be good to go @KristofferC but it would be good if you could review

IanButterworth:
Review still appreciated but I'll get this onto master to get it used before backport for 1.9.0

Issue #3448, opened 2023-04-24
Printing during precompilation errors


gzagatti:
When precompiling a package. I keep getting the following error whenever an Error or Warning message is meant to be printed.

```
┌ Error: Pkg.precompile error
│   exception =
│    MethodError: no method matching (::Pkg.API.var"#color_string#264")(::String, ::Int64)
│    
│    Closest candidates are:
│      (::Pkg.API.var"#color_string#264")(::String, ::Symbol)
│       @ Pkg /data/dockerdata/guilherme/.julia/juliaup/julia-1.9.0-rc2+0.x64.linux.gnu/share/julia/stdlib/v1.9/Pkg/src/API.jl:1252
````

This is due to the function defined [here](https://github.com/JuliaLang/Pkg.jl/blob/master/src/API.jl#L1279), reproduced below for convenience:

```julia
function color_string(cstr::String, col::Symbol)
        enable_ansi  = get(Base.text_colors, col, Base.text_colors[:default])
        disable_ansi = get(Base.disable_text_style, col, Base.text_colors[:default])
        return string(enable_ansi, cstr, disable_ansi)
end
```

Now, the problem is that `typeof(Base.text_colors) == Dict{Union{Int64, Symbol}, String}`.

In my `startup.jl` I customize some colors:

```julia
ENV["JULIA_ERROR_COLOR"] = 1
ENV["JULIA_WARN_COLOR"] = 3
```

Which leads to exception. I think the solution is to modify the signature of `color_string` to accept `col::Union{Int64, Symbol}`. After all, this is the correct type for the keys of `Base.text_colors`.

IanButterworth:
Thanks for the report. A PR to fix this would be appreciated

PR #3457, opened 2023-05-03
fix color_string

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3448

@gzagatti


gzagatti:
Pefect! It fixes the problem, thanks for the PR.

Issue #3458, opened 2023-05-04
Disallow certain Pkg operations during precompilation


vchuravy:
` PDFIO@0.1.10 ` caused a fun hang/issue where the precompilation of the package, caused precompilation, and voila we had a fork-bomb. This can still happen on current Julia versions.

This is obviously bad user code and was fixed in https://github.com/sambitdash/PDFIO.jl/commit/57e2c068e009c163c119f39b28a22787cd78b85f

but we should guard in Pkg.jl and when we are currently running precompilation either error/or decline to auto-precompile.

PR #3459, opened 2023-05-05
Disallow Pkg.precompile during precompilation

IanButterworth:
Proposed fix for #3458 

Basically make it noop during precompilation, and noisy if it's specifically called, silent if it's autoprecompilation.

But I'm not sure if this would prevent `Pkg.precompile` being baked into the julia sysimage, and I'm pretty sure it will stop it from being baked into a Pkg pkgimage.

@vchuravy

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1115,6 +1115,14 @@ end
 function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool=false,
                     strict::Bool=false, warn_loaded = true, already_instantiated = false, timing::Bool = false,
                     _from_loading::Bool=false, kwargs...)
+
+    if ccall(:jl_generating_output, Cint, ()) == 1
+        if !internal_call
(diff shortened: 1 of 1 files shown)


vchuravy:
Having `precompile` statements should still work fine.

vchuravy (approved):

KristofferC:
We all this while creating the sysimage and also want to call it during precompilation when Pkg is not in sysimage.

IanButterworth:
I thought @vchuravy's comment meant that when julia is in the mode to generate `precompile` statements `jl_generating_output == 0`? So the sysimage mechanism should still work?

As for pkgimage precompilation, that does seem like it would need to circumvent this, which could be put in place as a global that could be set, perhaps?

vchuravy:
How do we prevent the fork-bomb from occuring if we were to execute this during the precompilation of Pkg? Probably because we changed to a temporary project? 

For PKG, the old method of collecting precompilation statements and then including those and regenerating the cache file might be better?

Issue #3461, opened 2023-05-08
rather unhelpful `Circular dependency detected` message


vtjnash:
this warning seems remarkably unhelpful due to its length, and ending with a StackOverflow in a function that does not seem recursive?
 ```
$ time ./julia --project="/Users/jameson/.julia/packages/OmniPackage/i70lB/" -q -e 'using Pkg; Pkg.precompile()'
┌ Warning: Circular dependency detected. Precompilation will be skipped for:
│   EllipsisNotation [da5c29d0-fa7d-589e-88eb-ea29b0a81949]
│   SimpleBatchedNonlinearSolveExt [33126265-fb3a-5c3f-b833-5fb22a06ca54]
│   DiffResults [163ba53b-c6d8-5494-b064-1a9d43ac40c5]
│   Interpolations [a98d9a8b-a2ab-59e6-89dd-64a1c18fca59]
│   Catalyst [479239e8-5488-4da2-87a7-35f2df7eef83]
│   GridLayoutBase [3955a311-db13-416c-9275-1d80ed98e5e9]
│   PreallocationTools [d236fae5-4411-538c-8e31-a6e3d9e00b46]
│   ArrayInterfaceStaticArraysCore [dd5226c6-a4d4-4bc7-8575-46859f9c95b9]
│   LeastSquaresOptim [0fc2ff8b-aaa3-5acd-a817-1944a5e08891]
│   StaticArrays [90137ffa-7385-5640-81b9-e52037218182]
│   Sundials [c3572dad-4567-51f8-b174-8c6c989267f4]
│   JpegTurbo [b835a17e-a41a-41e7-81f0-2f016b05efe0]
│   Transducers [28d57a85-8fef-5791-bfe6-a80928e7c999]
│   NonlinearSolve [8913a72c-1f9b-4ce2-8d82-65094dcecaec]
│   OrdinaryDiffEq [1dea7af3-3e70-54e6-95c3-0bf5283fa5ed]
│   AdaptStaticArraysExt [e1699a77-9e31-5da8-bb3e-0a796f95f0a0]
│   ImageIO [82e4d734-157c-48bb-816b-45c225c6df19]
│   ArrayInterfaceBandedMatricesExt [26e938bc-0cd5-5679-9003-44616cbf91d3]
│   StructArraysExt [7174a82b-1e85-55a8-8a3d-2691d660c3ac]
│   StaticArrayInterface [0d7ed370-da01-4f52-bd93-41d350b8b718]
[...]

PR #3462, opened 2023-05-08
Revert "precompile: fix race in precompiling exts (#3456)"

IanButterworth:
This reverts commit c8249204b52708c916ff9a4752b04386c28de46b.

Given https://github.com/JuliaLang/Pkg.jl/issues/3461

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1138,7 +1138,7 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     ]
     stale_cache = Dict{StaleCacheKey, Bool}()
     exts = Dict{Base.PkgId, String}() # ext -> parent
-    # make a flat map of each dep and its direct deps
+    # make a flat map of each dep and its deps
     depsmap = Dict{Base.PkgId, Vector{Base.PkgId}}()
     pkg_specs = PackageSpec[]
     for dep in ctx.env.manifest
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -1138,7 +1138,7 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     ]
     stale_cache = Dict{StaleCacheKey, Bool}()
     exts = Dict{Base.PkgId, String}() # ext -> parent
-    # make a flat map of each dep and its direct deps
+    # make a flat map of each dep and its deps
     depsmap = Dict{Base.PkgId, Vector{Base.PkgId}}()
```
vtjnash, line 1142: I think the algorithm you are looking for here is Tarjan's SCC, since you already have the list of pkg->deps. You could also do a post-order DFS traversal, but the slight advantage of Tarjan is that it not only detects a cycle, but also computes the minimum list of entities in it. It is not quite as strong as doing a full dominator analysis (since it cannot sort within a cycle).

I have seemed to find myself in need of this quite often lately haha, such as
https://github.com/JuliaLang/julia/pull/49664. That is my favorite one, because of all the extra processing that has to be done there to compute the edges, which shows the flexibility of the Tarjan SCC algorithm. Though my cleanest implementation might be https://github.com/JuliaLang/julia/blob/master/src/staticdata_utils.c#L985-L1050, since it has the least extra state.

IanButterworth: Thanks! I think it will take me time to figure that out, so I'll get this reverted on master

PR #3465, opened 2023-05-10
Reland "precompile: fix race in precompiling exts (#3456)"

IanButterworth:
Try again.
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3443

With this `OrdinaryDiffEq` precompiles. It was reported not to by @gbaraldi before revert.

No exts
```
Foo => []
Bar => [Foo]
```
With Exts
```
Foo => []
Foo.Ext1 => [Foo]
Foo.Ext2 => [Foo]
Bar => [Foo.Ext1, Foo.Ext2]
```

Previously we had a racy situation
```
Foo => []
Foo.Ext1 => [Foo]
Foo.Ext2 => [Foo]
Bar => [Foo]
```
where Bar would start up and race to compile the exts with the jobs for Foo.Ext1 & 2


IanButterworth:
Also OmniPackage with https://github.com/FluxML/Zygote.jl/pull/1420 and updated precompiles.

KristofferC (approved):

Issue #3395, opened 2023-03-01
Test environment doesn't inherit number of interactive threads


carstenbauer:
Only the number of threads in the default thread pool seems to propagate to the test environment but not the number of interactive threads (irrespective of whether set via `-t N,M` or `JULIA_NUM_THREADS=N,M`). AFAICS, this makes it impossible to test package features related to interactive threads.

**MWE:**
```julia
# test/runtests.jl
using Base.Threads: nthreads
@show nthreads(:default)
@show nthreads(:interactive)
```

`julia --project -t 3,4` followed by `] test` gives
```
nthreads(:default) = 3
nthreads(:interactive) = 0
```
both on 1.9 as well as on nightly.

kpamnany:
We ran into this too. [Here](https://github.com/JuliaLang/Pkg.jl/blob/master/src/Operations.jl#L1682) is the offfending code. Will post a PR to fix shortly.

KristofferC:
FWIW, I think it is better to pass stuff like this explicitly via the `julia_args` argument to `Pkg.test`.

kpamnany:
Agree, but this should work.

PR #3467, opened 2023-05-10
Use correct threads specification for subprocess

kpamnany:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3395.


NHDaly (commented):
Thanks @kpamnany. LGTM! But i'll let one of the Pkg devs give final approval.

IanButterworth:
Seems reasonable to me given that the previous threads setup was forwarded

PR #3468, opened 2023-05-11
remove entirely unnecessary LazilyInitializedFields encumberance

vtjnash:
The macros in here have problems with globals and hygiene, as well as not being particularly well-integrated into the Julia ecosystem, both because they need to overload getproperty and introduce magic constants like `uninit`.

Just look how much code is deleted (300 lines) without needing any replacement for them, because Julia already does this better, particularly now that we have the `const` attribute.


KristofferC:
How did you even find this was being used? 😆

KristofferC (approved):

vtjnash:
I was hijacking some macro hygiene tests and this failed.

vtjnash:
Did you notice you also got a PR over there to fix the hygiene issues, and other improvements?

KristofferC:
Yes, and thanks for that but how did you notice it at all :P. I guess it bugged somewhere?

PR #3474, opened 2023-05-14
extensions docs typo

cobac:
The original "suggest" was probably a typo instead of "suggestion".

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -354,8 +354,7 @@ end
 end # module
 ```
 
-The name of the extension (here `PlottingContourExt`) is not very important but using something similar to the suggest here is likely a good
-idea.
+Extensions can have any arbitrary name (here `PlottingContourExt`), but using something similar to this example is likely a good idea.
 
 A user that depends only on `Plotting` will not pay the cost of the "extension" inside the `PlottingContourExt` module.
 It is only when the `Contour` package actually gets loaded that the `PlottingContourExt` extension is loaded


docs/src/creating-packages.md
```diff
@@ -354,8 +354,7 @@ end
 end # module
 ```
 
-The name of the extension (here `PlottingContourExt`) is not very important but using something similar to the suggest here is likely a good
-idea.
+Extensions can have any arbitrary name (here `PlottingContourExt`), but using something similar to this example is likely a good idea.
```
IanButterworth, line 357:
```suggestion
Extensions can have any arbitrary name (here `PlottingContourExt`), but using something similar to the format of this example that makes the extended functionality and dependency of the extension clear is likely a good idea for clarity.
```

docs/src/creating-packages.md
```diff
@@ -354,7 +354,7 @@ end
 end # module
 ```
 
-Extensions can have any arbitrary name (here `PlottingContourExt`), but using something similar to this example is likely a good idea.
+Extensions can have any arbitrary name (here `PlottingContourExt`), but using something similar to the format of this example that makes the extended functionality and dependency of the extension clear is likely a good idea.
```
IanButterworth, line 357: Just a new line
```suggestion
Extensions can have any arbitrary name (here `PlottingContourExt`), but using something similar to the format of
this example that makes the extended functionality and dependency of the extension clear is likely a good idea.
```

IanButterworth (approved):

PR #3485, opened 2023-05-22
Revert "Reland "precompile: fix race in precompiling exts (#3456)""

KristofferC:
Reverts JuliaLang/Pkg.jl#3465

There are 600 packages that encounter this on PkgEval which is a bit too many. It suggests something is a bit off here.


IanButterworth:
Yeah seems much more frequent than the ext cache generation clash it's trying to solve

IanButterworth (approved):

PR #3486, opened 2023-05-22
release-1.9: Revert "Reland "precompile: fix race in precompiling exts (#3456)" (#3465)" (#3485)

KristofferC:


IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/3490 is the proper fix. 

The commit this claims to revert wasn't actually on `release-1.9` so I guess this was a manual backport as it's quite different to what the actual revert would look like

PR #3490, opened 2023-05-22
[release-1.9] Reland "precompile: fix race in precompiling exts (#3456)" (#3465)

IanButterworth:
Closes https://github.com/JuliaLang/Pkg.jl/pull/3486


IanButterworth:
I double-checked by building julia/backports-releasse-1.9 at [JuliaLang/julia@`5d43397` (#49680)](https://github.com/JuliaLang/julia/pull/49680/commits/5d43397ee52323f1c015513b2be3909078b646ef) with this Pkg commit and confirm that it doesn't hit the circular dep issue that it did before when running
```
julia> using Pkg; Pkg.activate(; temp=true); Pkg.add("HarmonicBalance");
```

PR #3504, opened 2023-06-01
Use `joinpath` to obtain extension file path in docs

jishnub:
This seems more general than using `/` as the path separator


KristofferC:
Just curious about what you mean by more general. Does this not work on e.g. windows?

jishnub:
After some reading, it seems that this should work on recent versions of Windows (I had not tested it on Windows). In that case, perhaps this isn't necessary. What's the use case for `joinpath`, if this does work?

jishnub:
Closing, as this doesn't seem necessary, and makes the paths harder to read

PR #3511, opened 2023-06-08
Document how public API is defined (depends on scoped export)

LilithHafner:
This PR is an extension of https://github.com/JuliaLang/julia/pull/50105 and should be merged or closed simultaneously.

The key sentence is 

> Unless you note otherwise, the public API of your package is defined as all the behavior you
describe about exported symbols

And the reason this definition is viable is that scoped exports allow you to export symbols without the namespace implications.

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -120,7 +120,7 @@ version no longer conforms to the specifications provided in the old version, yo
 adjust your package version number according to [Julia's variant on SymVer](#Version-specifier-format).
 If you would like to include a symbol in your public API without fully exporting it into the
 global namespace of folks who call `using YourPackage`, you can export that symbol with
-`export scoped=true that_symbol`. Symbols exported with `scoped=true` are equally public
+`export scoped=true that_symbol`. Symbols exported with `scoped=true` are just as public
 as those exported without it, but when folks call `using YourPackage`, they will still have
 to qualify access to those symbols with `YourPackage.that_symbol`.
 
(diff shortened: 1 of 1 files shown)


docs/src/creating-packages.md
```diff
@@ -107,6 +107,94 @@ julia> HelloWorld.greet_alien()
...
+
+If you want your package to be useful to other packages and you want folks to be able to
+easily update to newer version of your package when they come out, it is important to
+document what behavior will stay consistent across updates.
+
+Unless you note otherwise, the public API of your package is defined as all the behavior you
+describe about [exported symbols](https://docs.julialang.org/en/v1/manual/modules/#Export-lists).
+When you change the behavior of something that was previously exported so that the new
+version no longer conforms to the specifications provided in the old version, you should
+adjust your package version number according to [Julia's variant on SymVer](#Version-specifier-format).
```
ericphanson, line 120:
```suggestion
adjust your package version number according to [Julia's variant on SemVer](#Version-specifier-format).
```

KristofferC:
In general I think things like this are better off in the Julia manual. In some sense, Pkg is "just" a package manager and someone could in theory write another better one. And we should IMO then try to keep the documentation to package management and less about general julia package information. I realize that things are pretty muddy now (for example the extension documentation is in Pkg, which it probably shouldn't be), but I think it is something to strive for.

c42f:
Agree @KristofferC that this stuff would be better in the Julia manual. The issue as I see it, is that the manual has no place for tutorial-style docs.

So how about we just merge this here for now alongside the extensions stuff? Because it's strictly better to have this documented somewhere rather than sitting in a stale PR :)

KristofferC:
> So how about we just merge this here for now alongside the extensions stuff? Because it's strictly better to have this documented somewhere rather than sitting in a stale PR :)

I mean, we don't even have `public` in Base so it feels kind of premature to merge the doc change before the feature itself even exist.

PR #3513, opened 2023-06-09
compute paths relative manifest more consistently

KristofferC:
with `manifest = ...` usage in Project.toml (actually, unconditionally but it is only noticed when this feature is used) we want to compute paths relative the current manifest, not the current project.

cc @quinnj


quinnj:
Sorry for being so slow on this, but this (retroactively) LGTM; I tested this out on the RelationalAI monorepo setup we have.

PR #3519, opened 2023-06-18
precompile: respect and report cache pidlocking

IanButterworth:
Follow-up to https://github.com/JuliaLang/julia/pull/49052

Screenshot 2023-06-19 at 11 24 56 AM

Depends on
- https://github.com/JuliaLang/julia/pull/50214
- https://github.com/JuliaLang/julia/pull/50218
- https://github.com/JuliaLang/julia/pull/50254

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -10,6 +10,9 @@ import LibGit2
 import Logging
 using Serialization
 using REPL.TerminalMenus
+import FileWatching
+
+import Base: StaleCacheKey
 
 import ..depots, ..depots1, ..logdir, ..devdir, ..printpkgstyle
 import ..Operations, ..GitTools, ..Pkg, ..Registry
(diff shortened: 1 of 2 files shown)


KristofferC:
Is it really the package version that is locked on? It's the filename no (which doesn't include the version) but has a mix of other things like the project, sysimage, etc.

IanButterworth:
Specifically it's the sourcepath.

KristofferC:
Okay so that means that two julias with e.g. different sysimage that are precompiling will wait on each other even if they would produce a different precompile file in the end and there wouldn't really be a need for a lock, or?

With preferences, I don't think you even know what the end resulting file name will be until the process is done :/

IanButterworth:
Right. I don't think we can be specific to the cachefile because of all that

IanButterworth:
xref https://github.com/JuliaLang/julia/issues/50206

vchuravy:
I will also raise my earlier point again. We need to make this robust against user-interuption. Or at least make it obvious to the user how to cleanup after interruptions.

pacman as an example tells me which file it tried to acquire and if I think the other process was interrupted that I should remove it

IanButterworth:
Do you mean in addition to the stale detection stuff that mkpidlock does automatically?

IanButterworth:
This has addressed feedback and has tests, and I want to work on Pkg.precompile stuff after this so I'm going to merge and bump the stdlib, and then add the news to base NEWS

Issue #3074, opened 2022-05-02
Allow testing projects without a UUID


cossio:
Currently if you try to `Pkg.test` a project without an UUID, it will throw this error:

```
ERROR: trying to test unnamed project
```

Even if the project has a name.

Is there any reason why the Project.toml needs to have an UUID if I want to test it?

cossio:
Related: https://github.com/JuliaLang/Pkg.jl/issues/1982

AnHeuermann:
For me it would already be an improvement to make the error message more specific. I needed to search for a while until I noticed that the missing UUID was the reason for the error message.
I created a Julia project that is not a package using DrWatson, see https://github.com/JuliaDynamics/DrWatson.jl/issues/67.

So for the missing UUID something like

```
ERROR: trying to test unnamed project (missing UUID in Project.toml)
```

would improve my experience.

----------------

Also related to https://github.com/JuliaLang/Pkg.jl/issues/1253.

PR #3528, opened 2023-06-27
Clarify error with project missing attributes

alecloudenback:
I ran into this issue today and was confused because my project had a name but no UUID.

ref #3074


src/API.jl
```diff
@@ -430,7 +430,7 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     Context!(ctx; kwargs...)
 
     if isempty(pkgs)
-        ctx.env.pkg === nothing && pkgerror("trying to test unnamed project") #TODO Allow this?
+        ctx.env.pkg === nothing && pkgerror("Project.toml must have project name and UUID") #TODO Allow this?
```
IanButterworth, line 433:
```suggestion
        ctx.env.pkg === nothing && pkgerror("The Project.toml of the package being tested must have a name and a UUID entry") #TODO Allow this?
```

PR #3533, opened 2023-07-03
Clarify test dependencies paragraph by moving note about the "old" version

gdalle:
In the current version of the documentation, the `targets`-based method appears before the `test/Project.toml` one. Therefore the associated note about the "old" method needed to be moved.


IanButterworth:
I'm not sure about this proposal.

I've made an alternative proposal here, which makes a bit more sense to me https://github.com/JuliaLang/Pkg.jl/pull/3574

gdalle:
Fine by me!

PR #3537, opened 2023-07-04
recommend adding Requires to weakdeps

aplavin:
As I understand, installing Requires on new versions isn't needed, so it would make sense to `weakdep` it.

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -418,7 +418,8 @@ This is done by making the following changes (using the example above):
   ```julia
   isdefined(Base, :get_extension) ? (using Contour) : (using ..Contour)
   ```
-- Add `Requires` to `[weakdeps]` in your `Project.toml` file, so that it is only installed on Julia versions that do not support extensions.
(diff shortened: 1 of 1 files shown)


docs/src/creating-packages.md
```diff
@@ -418,6 +418,7 @@ This is done by making the following changes (using the example above):
   ```julia
   isdefined(Base, :get_extension) ? (using Contour) : (using ..Contour)
   ```
+- Add `Requires` to `[weakdeps]` in your `Project.toml` file, so that it is only installed on Julia versions that do not support extensions.
```
IanButterworth, line 421: It might be good to mention that it would be listed in both deps and weakdeps, and that 1.9+ knows to make that weak

IanButterworth:
```suggestion
- Add `Requires` to `[weakdeps]` in your `Project.toml` file, so that it is listed in both `[deps]` and `[weakdeps]`.
  Julia 1.9+ knows to not install it as a regular dependency, whereas earlier versions will consider it a dependency.
```

IanButterworth: bump

PR #3547, opened 2023-07-12
Remove mentions of “home project” in documentation and comments

### LINKED ISSUE #3127: `Pkg.activate()` doesn't seem to activate the home package
In a project directory at
```console
abcd/ $ pwd
/home/jishnu/Dropbox/JuliaPackages/abcd
```
I launch julia using
```console
abcd/ $ julia-1.8 --project=.
```
In julia
```julia
julia> LOAD_PATH
4-element Vector{String}:
 "@"
 "/tmp/jl_ztlVMr"
 "@v#.#"
 "@stdlib"

julia> Base.active_project()
"/home/jishnu/Dropbox/JuliaPackages/abcd/Project.toml"

julia> using OffsetArrays
 │ Package OffsetArrays not found, but a package named OffsetArrays is available from a registry. 
 │ Install package?
 │   (abcd) pkg> add OffsetArrays 
 └ Select environment:
   1: `~/Dropbox/JuliaPackages/abcd/Project.toml` (@)
 > 2: `/tmp/jl_ztlVMr` (/tmp/jl_ztlVMr)
   3: `~/.julia/environments/v1.8/Project.toml` (@v#.#)
   Resolving package versions...
    Updating `/tmp/jl_ztlVMr/Project.toml`
  [6fe1bfb0] + OffsetArrays v1.12.7
    Updating `/tmp/jl_ztlVMr/Manifest.toml`
  [79e6a3ab] + Adapt v3.3.3
  [6fe1bfb0] + OffsetArrays v1.12.7
  [56f22d72] + Artifacts
  [8f399da3] + Libdl
  [37e2e46d] + LinearAlgebra
  [e66e0078] + CompilerSupportLibraries_jll v0.5.2+0
  [4536629a] + OpenBLAS_jll v0.3.20+0
  [8e850b90] + libblastrampoline_jll v5.1.0+0

julia> Base.active_project()
"/home/jishnu/Dropbox/JuliaPackages/abcd/Project.toml"

julia> Base.current_project()
"/home/jishnu/Dropbox/JuliaPackages/abcd/Project.toml"

julia> using Pkg; Pkg.activate()
  Activating project at `/tmp/jl_ztlVMr`

julia> Base.active_project()
"/tmp/jl_ztlVMr/Project.toml"

julia> versioninfo()
Julia Version 1.8.0-rc1
Commit 6368fdc656 (2022-05-27 18:33 UTC)
Platform Info:
  OS: Linux (x86_64-pc-linux-gnu)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i5-1135G7 @ 2.40GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-13.0.1 (ORCJIT, tigerlake)
  Threads: 1 on 8 virtual cores
Environment:
  LD_LIBRARY_PATH = :/usr/lib/x86_64-linux-gnu/gtk-3.0/modules
  JULIA_EDITOR = vim
```
The docstring says "If no argument is given to activate, then activate the home project. The home project is specified by either the --project command line option to the julia executable, or the `JULIA_PROJECT` environment variable.". In this case, the specified `--project` flag is not being activated anymore, and `JULIA_PROJECT` is not set. In this case, `Pkg.activate()` should not change the path at all, as the home project is already activated.

Socob:
The concept of “home project” was removed in JuliaLang/julia#36434. Also see #1891.

Fixes #3127.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -539,9 +539,8 @@ The logic for what path is activated is as follows:
     activate the environment at the tracked path.
   * Otherwise, `s` is interpreted as a non-existing path, which is then activated.
 
-If no argument is given to `activate`, then activate the home project.
-The home project is specified by either the `--project` command line option to
-the julia executable, or the `JULIA_PROJECT` environment variable.
+If no argument is given to `activate`, then use `LOAD_PATH` to determine the
+project to activate.
 
 # Examples
 ```
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -539,8 +539,7 @@ The logic for what path is activated is as follows:
     activate the environment at the tracked path.
   * Otherwise, `s` is interpreted as a non-existing path, which is then activated.
 
-If no argument is given to `activate`, then use `LOAD_PATH` to determine the
-project to activate.
+If no argument is given to `activate`, then use the first project found in `LOAD_PATH`.
 
 # Examples
 ```
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -304,7 +304,7 @@ PSA[:name => "activate",
     activate [--shared] path
     activate --temp
 
-Activate the environment at the given `path`, or use `LOAD_PATH` to determine the project to activate if no `path` is specified.
+Activate the environment at the given `path`, or use the first project found in `LOAD_PATH` if no `path` is specified.
 The active environment is the environment that is modified by executing package commands.
 When the option `--shared` is given, `path` will be assumed to be a directory name and searched for in the
 `environments` folders of the depots in the depot stack. In case no such environment exists in any of the depots,
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -304,7 +304,7 @@ PSA[:name => "activate",
     activate [--shared] path
     activate --temp
 
-Activate the environment at the given `path`, or the home project environment if no `path` is specified.
+Activate the environment at the given `path`, or use `LOAD_PATH` to determine the project to activate if no `path` is specified.
 The active environment is the environment that is modified by executing package commands.
 When the option `--shared` is given, `path` will be assumed to be a directory name and searched for in the
 `environments` folders of the depots in the depot stack. In case no such environment exists in any of the depots,
--- a/test/api.jl
+++ b/test/api.jl
@@ -38,7 +38,7 @@ using ..Utils
         cd(mkdir("tests"))
         Pkg.activate("Foo") # activate developed Foo from another directory
         @test Base.active_project() == joinpath(path, "modules", "Foo", "Project.toml")
-        Pkg.activate() # activate home project
+        Pkg.activate() # activate LOAD_PATH project
         @test Base.ACTIVE_PROJECT[] === nothing
     end end
 end
--- a/test/new.jl
+++ b/test/new.jl
@@ -53,7 +53,7 @@ Pkg._auto_gc_enabled[] = false
             source053 = pkg.source
             source053_time = mtime(pkg.source)
         end
-        # - The home project was automatically created.
+        # - The active project was automatically created.
         @test haskey(Pkg.project().dependencies, "Example")
         @test length(Pkg.project().dependencies) == 1
         # Now we install the same package at a different version:
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -260,7 +260,7 @@ temp_pkg_dir() do project_path
         pkg"activate --shared Foo" # activate existing shared Foo
         @test Base.active_project() == joinpath(tmpdir, "Project.toml")
         pop!(Base.DEPOT_PATH)
-        pkg"activate" # activate home project
+        pkg"activate" # activate LOAD_PATH project
         @test Base.ACTIVE_PROJECT[] === nothing
         # expansion of ~
         if !Sys.iswindows()


src/Pkg.jl
```diff
@@ -539,9 +539,8 @@ The logic for what path is activated is as follows:
     activate the environment at the tracked path.
   * Otherwise, `s` is interpreted as a non-existing path, which is then activated.
 
-If no argument is given to `activate`, then activate the home project.
-The home project is specified by either the `--project` command line option to
-the julia executable, or the `JULIA_PROJECT` environment variable.
+If no argument is given to `activate`, then use `LOAD_PATH` to determine the
+project to activate.
```
IanButterworth, line 543: > then use `LOAD_PATH` to determine the project to activate

This might be a little to vague. Can we be more specific? 
Perhaps this?

```suggestion
If no argument is given to `activate`, then use the first usable project in `LOAD_PATH`.
```

Socob: Sure, although I don’t see how this could be understood to mean anything else. The “use […] usable” doesn’t sound great to me, so maybe like this?

> If no argument is given to `activate`, then use the first project found in `LOAD_PATH`.

IanButterworth: Yeah that sounds good

src/REPLMode/command_declarations.jl
```diff
@@ -304,7 +304,7 @@ PSA[:name => "activate",
     activate [--shared] path
     activate --temp
 
-Activate the environment at the given `path`, or the home project environment if no `path` is specified.
+Activate the environment at the given `path`, or use `LOAD_PATH` to determine the project to activate if no `path` is specified.
```
IanButterworth, line 307:
```suggestion
Activate the environment at the given `path`, or use the first project found in `LOAD_PATH` if no `path` is specified.
```

IanButterworth (approved):

danielwe:
> If no argument is given to `activate`, then use the first project found in `LOAD_PATH`.

This is still inaccurate. The first entry in the default `LOAD_PATH` is `"@"`, and according to the documentation,

> `"@"` refers to the "current active environment", the initial value of which is initially determined by the [`JULIA_PROJECT`](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_PROJECT) environment variable or the `--project` command-line option.

However, what happens when calling `activate` without arguments is that the project set through `JULIA_PROJECT` or `--project` is deliberately forgotten (by setting `Base.ACTIVE_PROJECT[] = nothing`). This turns `"@"` into a simple redirect to the next `LOAD_PATH` entry. In other words, `activate` without arguments isn't simply looking at `LOAD_PATH`, it's changing the meaning of the entries in `LOAD_PATH`.

How about:

> If no argument is given to `activate`, then discard any active project set via the [`JULIA_PROJECT`](https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_PROJECT) environment variable, the `--project` command-line option, or any previous calls to `activate`. Unless `LOAD_PATH` has been modified, the result is to activate the default `@v#.#` environment.

IanButterworth:
Note that https://github.com/JuliaLang/Pkg.jl/pull/1891 remains open, so it may be better to spend effort getting that merged?

Socob:
@danielwe Come to think of it, the somewhat unclear meaning of `"@"` is probably why I worded things the way I did in the original version of the PR (which I wouldn’t say was inaccurate!). In fact, unless I’m missing something, both Julia and Pkg.jl always simply ignore `"@"` in `LOAD_PATH`! Or are you aware of any situation where having `"@"` in `LOAD_PATH` makes any difference compared to not having it? Based on the code for `Base.active_project()` and `Pkg.activate()`, I don’t see any case where it’d do anything:

[`julia/base/initdefs.jl`](https://github.com/JuliaLang/julia/blob/750df9fb5bede16f321f5d5405943d12aec7b83e/base/initdefs.jl#L290-L312)

https://github.com/JuliaLang/Pkg.jl/blob/46b7eef2a07a69b6401c362e87be8f3e17201b94/src/API.jl#L1818-L1836

It seems like this is more of a historical leftover without any actual remaining functionality and should probably be removed?

IanButterworth:
@StefanKarpinski perhaps you could weigh in here about the design and perhaps a post merge review of these changes?

danielwe:
> Or are you aware of any situation where having `"@"` in `LOAD_PATH` makes any difference compared to not having it?

Not the expert here, I had never looked at this code before I wrote the previous comment, but it looks like it makes a difference in `load_path_expand`: https://github.com/JuliaLang/julia/blob/fd38d50cefd17c2e31df34c6b6c4a5b7c721c57d/base/initdefs.jl#L250

Without it, code loading won't know about the active project:

```julia
(@v1.9) pkg> activate --temp
  Activating new project at `/tmp/jl_ujp5um`

(jl_ujp5um) pkg> add EnumX
   Resolving package versions...
    Updating `/tmp/jl_ujp5um/Project.toml`
  [4e289a0a] + EnumX v1.0.4
    Updating `/tmp/jl_ujp5um/Manifest.toml`
  [4e289a0a] + EnumX v1.0.4

julia> popfirst!(LOAD_PATH)
"@"

julia> using EnumX
 │ Package EnumX not found, but a package named EnumX is available from a
 │ registry.
 │ Install package?
 │   (jl_ujp5um) pkg> add EnumX
 └ (y/n/o) [y]:
```

Socob:
Ah right, sorry. I think what’s confusing here is that `LOAD_PATH` is used for two different purposes: Its main purpose of giving the places to look for packages in the context of `using`/`import` (via `Base.load_path()`), and secondarily, also to determine what the “active project” is with `Base.active_project()`. But apart from being the project that Pkg.jl operates on, i. e. when it comes to Base, the only meaning of the “active project” is, in turn, to define what the placeholder `"@"` in `LOAD_PATH` means. So in that sense, `Pkg.activate()`/`Base.active_project()` *must* skip over `"@"` in `LOAD_PATH`, because the entire point of setting an active project is to define what `"@"` means in the first place!

Given all this, I think the clearest option would be to simply say

> If no argument is given to `activate`, then use the first project found in `LOAD_PATH` (ignoring `"@"`).

(By the way, the comment line

[``# `@` in JULIA_LOAD_PATH is expanded early (at startup time)``](https://github.com/JuliaLang/julia/blob/fd38d50cefd17c2e31df34c6b6c4a5b7c721c57d/base/initdefs.jl#L250)

seems to be inaccurate?)

danielwe:
> > If no argument is given to `activate`, then use the first project found in `LOAD_PATH` (ignoring `"@"`).

Yes, that's simple and clear. Though I do think it's worth pointing out what this implies in the usual case, i.e., that it activates `@v1.x`. You shouldn't have to read the docs for `LOAD_PATH` to understand what `activate` does when you haven't touched `LOAD_PATH`.



> (By the way, the comment line
> 
> > [`` # `@` in JULIA_LOAD_PATH is expanded early (at startup time) ``](https://github.com/JuliaLang/julia/blob/fd38d50cefd17c2e31df34c6b6c4a5b7c721c57d/base/initdefs.jl#L250)
> 
> seems to be inaccurate?)

No, this appears to be correct:

```bash
user@host:~/tmp$ ls
Manifest.toml  Project.toml
user@host:~/tmp$ JULIA_LOAD_PATH="@." julia -q --startup-file=no
julia> LOAD_PATH
1-element Vector{String}:
 "/home/daniel/tmp/Project.toml"
```

I suppose you'd have to look at the code that processes environment variables at startup to see how this is implemented.

Socob:
I don’t have any objections against mentioning what happens for the default `LOAD_PATH`. It’s a bit unfortunate that this discussion only happened after the PR was merged.

@IanButterworth What’s the best way forward here? Should I make another PR incorporating the things that were discussed here?

IanButterworth:
Sounds good to me. I think this PR is an improvement, at least.

danielwe:
> It’s a bit unfortunate that this discussion only happened after the PR was merged.

Yeah, sorry about that, I'm just a passer-by and didn't know about the PR until I got the email notification when the linked issue was closed by the PR merge (since I had contributed an MWE to the issue way back).

Socob:
Oh no worries, not blaming anyone for anything. That was just a general remark about the situation as a whole :slightly_smiling_face:

Issue #3499, opened 2023-05-27
Improving the docs for extensions


gdalle:
There have been several discussions on Slack and Discourse about extensions.
A major focal point seems to be that we cannot / should not define structs in extensions, only extend existing methods.
Let's try and formulate a clarification for the docs here, before opening a PR?

gdalle:
Here's where it could go: https://pkgdocs.julialang.org/v1/creating-packages/#Conditional-loading-of-code-in-packages-(Extensions)

gdalle:
Some of these conversations:
- https://discourse.julialang.org/t/are-extension-packages-importable/92527

Seelengrab:
Here are some more (perhaps not so obvious?) questions a user may have, that it ought to be good to answer:

 * Pkg extensions have me define a module. Can I define & use a struct/function defined exclusively in the extension in my main package?
 * Can I `@eval`/`eval` into my main package from an extension?
 * Can I precompile an extension?
 * How can I query whether an extension is loaded in my main package?
 * Should I query an extension for being loaded? What consequences does that have for precompilation of my package?
 * What happens when I have two mutually exclusive extensions that shouldn't be loaded at the same time? How can I handle this?
 * How can I teach my users about additional functionality being available when a given package is loaded?

PR #3552, opened 2023-07-21
Clarify extension docs about defining new symbols

gdalle:
Partial fix for #3499 

It's a very minimal addition and I would welcome more contributions


docs/src/creating-packages.md
```diff
@@ -299,6 +299,11 @@ loaded into the Julia session. This is very similar to functionality that the ex
 [Requires.jl](https://github.com/JuliaPackaging/Requires.jl) provides, but which is now available directly through Julia,
 and provides added benefits such as being able to precompile the extension.
 
+!!! warning
+    The recommended way to use extensions is to implement new methods for _existing functions_.
+    This can include zero-method functions defined in the main package by `function f end`.
+    On the other hand, if you create _new objects_ like functions or types in the extension, it is possible to access them (see below) but it is not considered good practice.
+
```
ffevotte, line 306: I'd perhaps have put this paragraph a little later in the documentation, perhaps even as far down as just before the second newly added paragraph about new functions/types.

But in any case, I feel this warning is more easily understood once the reader has seen an example of extensions in action. This could also allow referring to the code example inside the warning text. For example:

> The recommended way to use extensions is to implement new methods for _existing functions_ (as is the case of `Plotting.plot` in the example above). This can include...

docs/src/creating-packages.md
```diff
@@ -366,6 +371,14 @@ If one considers `PlottingContourExt` as a completely separate package, it could
...
 In fact, this form of type piracy is one of the most standard use cases for extensions.
 
+Conversely, although it is possible, defining new functions or types in the extension is not recommended.
+If your use case requires it, you can call `Base.get_extension(Plotting, :PlottingContourExt)` to retrieve the extension module.
+From there, any name such as `MyFancyContourPlot` can be retrieved with the standard module prefixing syntax.
+If you use it often in a third party package, you may want to shorten it as follows:
+```julia
+const MyFancyContourPlot = Base.get_extension(Plotting, :PlottingContourExt).MyFancyContourPlot
+```
+
```
ffevotte, line 381: Should we also mention here the "recipe" where the parent package provides a way to access things within its extension?

Like
```julia
# within the Plotting module:
function get_ext()
    ext = Base.get_extension(@__MODULE__, :PlottingContourExt)
    if ext === nothing
        error("extension not loaded...")
    else
        return ext  # or ext.something_defined_in_ext
    end
end
```

ffevotte (approved):
Thanks for the nice write-up! Seeing the frequency with which these questions come up on discourse, I think it's important to add this kind of details to the documentation.

timholy:
It's worth noting there's another pattern that was initially described in #3282 but which I then stripped out:

```
module MainPackage

export stub
function stub end   # no methods

# probably a good idea to `register_error_hint` in `__init__` describing what other package(s) must be loaded to make `stub` active

end

########################

module MainPackagePlotting  # extension package

MainPackage.stub(obj::SomeObject) = ...

end
```

KristofferC:
I personally don't think that this clarifies very much. And what is considered good practice is a bit hard to say at this point since the feature is so new. The first step should imo be to describe how everything works and do that well. Once that is done, there can be a discussion of best practices.

So the first question I would have is, are there currently lack of documentation how things work on a technical level (like how the name spacing of symbols defined in the extension work etc)?

ffevotte:
> are there currently lack of documentation how things work on a technical level (like how the name spacing of symbols defined in the extension work etc)?

One thing that I think is definitely lacking is a documentation for `get_extension` (which is currently only mentioned as the way to check whether the Julia version supports extensions via `!isdefined(Base, :get_extension)`.

Apart from that, I think all information is already there for users to understand how things work on a technical level. But judging by the questions asked on discourse, what appears to be missing is some form of "How to (do what I want)?"

AFAIU, this PR tries to provide answers to questions asked on discourse that all look like: "I used to use Requires.jl to define new functions/types/variables in my package when some optional dependency is also loaded. When switching to weak dependencies, these functions/types/variables are all namespaced into the extension module: how do I access them?"
Looking at the current documentation, it seems (to me at least) relatively clear why users are having such issues: because the extension is a new *module*, that comes with its own namespace. What isn't clear (IMO), is: what to do about it? I think this is what this PR tries to address.

Another way to look at this is the discussion in [Backwards compatibility](https://pkgdocs.julialang.org/v1/creating-packages/#Backwards-compatibility). This section gives a nice How-To guide about how to transition from (`Requires.jl` or strong dependencies) to weak dependencies. However, the "recipes" given there are only valid under some assumptions, (e.g. that the conditionally loaded code only adds new methods to functions already defined elsewhere). So the techniques discussed in this paragraph should be complemented with ways to solve issues appearing when these assumptions are not valid (or warn the user to not use extensions/weak deps in such cases).

<br/>

Writing all of this made me think of an alternate way of incorporating this in the documentation:
- in the main paragraph, perhaps simply note more explictly that since the extension has its own module, everything defined in it is namespaced (and therefore "insulated" from the main package namespace)
- in the discussion about backward compatibility & transition from Requires.jl, add some advice about how to make things defined in the extension available from the outside:
    - (a) add methods to functions already defined (and accessible from) elsewhere
    - (b) possibly define empty function stubs to make method (a) relevant
    - (c) use `get_extension` in some way to get direct access to the extension module (either from the main package or the client code)

docs/src/creating-packages.md
gdalle: I wasn't aware of this! Is it widely used?

gdalle:
> It's worth noting there's another pattern that was initially described in https://github.com/JuliaLang/Pkg.jl/pull/3282 but which I then stripped out:

That's what I meant when I mentioned "zero-method functions", perhaps I should be more explicit?

> And what is considered good practice is a bit hard to say at this point since the feature is so new. The first step should imo be to describe how everything works and do that well. Once that is done, there can be a discussion of best practices.

I based my PR mostly on comments by you that I had read on Slack or Discourse, like [this one](https://discourse.julialang.org/t/are-extension-packages-importable/92527/8?u=gdalle):

_Well yes, but I do think that best practices should be to make the functionality exposed by overloading functions in the parent. It can even be a zero method function that is overloaded. So you would not need to access the namespace of the extension._

But that was shortly after the 1.9 release. Now that we have a little more hindsight, would you say that accessing the extension namespace directly is a good idea?

> Are there currently lack of documentation how things work on a technical level (like how the name spacing of symbols defined in the extension work etc)?

I think the use of `get_extension` to actually access the namespace, and not just check whether it exists, is an important addition

gdalle:
> Writing all of this made me think of an alternate way of incorporating this in the documentation:

I understand where this comes from but I'm not sure it's a good idea to wait that long in the structure of the docs. Basically I'm afraid that the "backwards compatibility" section sounds too technical, and that people might just skip it.

Besides, there are many developers who never touched extensions before since they were not an official part of the language, but who will try them out now. Their packages may never have used Requires, and they may not even care about compatibility < 1.9. Yet these developers also need the info.

gdalle:
So where do we want to go with this? Should I just remove the notion of good practice? I personally don't think so, even in the extensions talk at JuliaCon it was rather clear that extending functions was the safe way to go

gdalle:
Gentlest bump on this, I'm more than willing to amend the PR if we agree on the direction

KristofferC:
To me, the core information to convey is that you cannot create new symbols in the parent package from the extension. Once that is established, everything else follows from that.

It is however totally fine to define new structs and functions in an extension and access them from the parent package with `Base.get_extension`.

docs/src/creating-packages.md
```diff
@@ -376,6 +369,34 @@ In fact, this form of type piracy is one of the most standard use cases for exte
...
+    However, for extensions, it is ok to assume that the extension owns the functions in its parent package.
+
+In other situations, one may need to define new symbols in the extension (types, structs, functions, etc.) instead of reusing those from the parent.
+Such symbols are created in a separate module corresponding to the extension (`PlottingContourExt`), and not in the parent itself (`Plotting`).
+To retrieve this module and the symbols it contains for outside use, all it takes is a call to [`Base.get_extension`](@ref).
+Here is an example with a custom type defined in the extension:
+
+```julia
+const ContourPlotType = Base.get_extension(Plotting, :PlottingContourExt).ContourPlotType
+```
```
ffevotte, line 398: Two things make me uneasy in this code example:
1. it does not account for the possibility of `get_extension` returning `nothing` if the extension is not loaded, and
2. I'm not sure whether the most common use case would be for a package to get access to its own extension, or for client code to access the extension of a package it depends on.

If the former is common enough, it might be useful to showcase an example along the lines of:
```julia
ext = Base.get_extension(@__MODULE__, :PlottingContourExt)
if !isnothing(ext)
   ContourPlotType = ext.ContourPlotType
end
```

As it stands, the example provided here hints at the latter use case, where code in a third-party package tries to access the extension to another package. I'm not sure this would be recommended practice:
- how can the developer of the 3rd party package know what is actually "exposed" by the extension as its "API"? In other words, is it safe to access `ContourPlotType` from the extension, or will its name change in the next match release?
- in any case, this "leaks" the extension name, which needs to be known for the 3rd party package to access it. I think I have seen somewhere that extension names should generally be considered internal implementation details and not be part of an API of the main package.

All in all, I'd be more comfortable if we advertised more prominently a use case in which the main package tries to gain access to its own extension.

ffevotte (approved):
This looks good to me; many thanks @gdalle!

I'm sure we'll have much more to say on this topic in a few months/years when more feedback on extensions will be available. But in the meantime, it looks to me like this proposal gives the two most useful (IMO) pieces of information:
1. extending (possibly empty) functions is the most common use case for extensions
2. `Base.get_extension` can be used to retrieve the extension module from its "outside".

StefanKarpinski:
It would be nice to get some kind of guidance along these lines in the docs. @KristofferC, are you happier with this now? Or could you suggest something that would make more sense to you?

jordancluts:
I am not remotely informed enough to express an opinion on the details discussed in this thread. I did, however, want to emphasize that something should be done to address this in the docs. There are near daily questions on Slack about confusion when defining functions/classes in an extension. It is obvious that the docs need to guide new users better (whether that is telling them not to do it, or how to go about doing so). I'm hoping that this anecdote will spur something getting over the finish line to help all the folks that are running headlong into this problem.

gdalle:
Whatever is needed to get this PR merged, I will gladly do ^^

jacobusmmsmit:
I think it's better to have something in the docs that might change in the future or is possibly imperfect or incomplete rather than nothing. The ideas described by the PR might not be the best, nor the way that extensions will be used in the future, but they do provide a consistent way to work with extensions and give an idea of best practices.

gdalle:
@KristofferC gentle ping

gdalle:
Is there anything that needs changing here?

gdalle:
Bump in case anyone is watching, this is a 15s reviez

Issue #3548, opened 2023-07-13
stdout messages lost by precompile [regression]


vtjnash:
Calling `using` now fails to print messages from stdout:

```
julia> write("MWE2/src/MWE2.jl", 
           """
           module MWE2
             println(stderr, "hello")
             println(stdout, "world")
           end""");

julia> using MWE2
Precompiling MWE2
  1 dependency successfully precompiled in 1 seconds. 1 already precompiled.
  1 dependency had warnings during precompilation:
┌ MWE2 [49dfab96-4e56-45b4-ab4d-cc1efaf838ee]
│  hello
└  
```

What happened to the "world"??

PR #3573, opened 2023-08-09
precompilation: capture stdout as well as stderr

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3548

@vtjnash in particular is the use of a single BufferStream `iob` ok here or should it use two buffers?

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1483,6 +1483,8 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                             rethrow()
                         end
                     finally
+                        isopen(std_pipe.in) && close(std_pipe.in) # close pipe to end the std output monitor
+                        wait(t_monitor)
                         Base.release(parallel_limiter)
                     end
                 else
--- a/src/API.jl
+++ b/src/API.jl
@@ -1277,37 +1277,26 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     taskwaiting = Set{Base.PkgId}()
     pkgspidlocked = Dict{Base.PkgId,String}()
 
-    function monitor_stderr_stdout(pkg, _stderr, _stdout)
-        @sync begin
-            @async try
-                while !eof(_stderr)
-                    str = readline(_stderr)
-                    std_outputs[pkg] = string(get(std_outputs, pkg, ""), str, "\n")
-                    if !in(pkg, taskwaiting) && occursin("waiting for IO to finish", str)
-                        !fancyprint && lock(print_lock) do
-                            println(io, pkg.name, color_string(" Waiting for background task / IO / timer.", Base.warn_color()))
-                        end
-                        push!(taskwaiting, pkg)
+    function monitor_std(pkg, pipe)
+        try
+            while !eof(pipe)
+                str = readline(pipe, keep=true)
+                std_outputs[pkg] = string(get(std_outputs, pkg, ""), str)
+                if !in(pkg, taskwaiting) && occursin("waiting for IO to finish", str)
+                    !fancyprint && lock(print_lock) do
+                        println(io, pkg.name, color_string(" Waiting for background task / IO / timer.", Base.warn_color()))
                     end
-                    if !fancyprint && in(pkg, taskwaiting)
-                        lock(print_lock) do
-                            println(io, str)
-                        end
+                    push!(taskwaiting, pkg)
+                end
+                if !fancyprint && in(pkg, taskwaiting)
+                    lock(print_lock) do
+                        print(io, str)
                     end
                 end
-            catch err
-                err isa InterruptException || rethrow()
             end
-            @async try
-                while !eof(_stdout)
-                    str = readline(_stdout)
-                    std_outputs[pkg] = string(get(std_outputs, pkg, ""), str, "\n")
-                end
-            catch err
-                err isa InterruptException || rethrow()
-            end
+        catch err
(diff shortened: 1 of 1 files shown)


IanButterworth:
Another question is whether we should just do this always and not just code load

vtjnash (commented):
It is strongly forbidden for the same object to be used asynchronously in two difference places.

vtjnash (commented):
This should be using a `Pipe()`. There also should not be a `close(iob)` call, ever, since that corrupts the stream.

src/API.jl
```diff
@@ -1469,11 +1481,13 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                         end
                         loaded && (n_loaded += 1)
                     catch err
-                        close(iob)
+                        # close pipes to end the std output monitor
+                        close(stderr_pipe)
+                        close(stdout_pipe)
                         wait(t_monitor)
```
IanButterworth, line 1487: > There also should not be a close(iob) call, ever, since that corrupts the stream.

@vtjnash the closes here are to stop the async `monitor_stderr_stdout`. Is that acceptable?

src/API.jl
```diff
@@ -1273,30 +1273,41 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
-            while isopen(iob)
-                str = readline(iob)
-                stderr_outputs[pkg] = get(stderr_outputs, pkg, "") * str * "\n"
-                if !in(pkg, taskwaiting) && occursin("waiting for IO to finish", str)
-                    !fancyprint && lock(print_lock) do
-                        println(io, pkg.name, color_string(" Waiting for background task / IO / timer.", Base.warn_color()))
+    function monitor_stderr_stdout(pkg, _stderr, _stdout)
+        @sync begin
+            @async try
+                while isopen(_stderr)
```
vtjnash, line 1283: `isopen` is the wrong predicate to use here, as stated in the docstring for it, use `eof`.

src/API.jl
```diff
@@ -1469,11 +1481,13 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                         end
                         loaded && (n_loaded += 1)
                     catch err
-                        close(iob)
+                        # close pipes to end the std output monitor
+                        close(stderr_pipe)
+                        close(stdout_pipe)
```
vtjnash, line 1486:
```suggestion                        
```
This introduces a data-race into your design, and suggests that the flow of events needs to be better managed so that they aren't happening from conflicting different places.

src/API.jl
```diff
@@ -1428,8 +1439,9 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                     is_direct_dep = pkg in direct_deps
 
                     # stderr monitoring
-                    iob = Base.BufferStream()
-                    t_monitor = @async monitor_stderr(pkg, iob)
+                    stderr_pipe = Base.Pipe()
+                    stdout_pipe = Base.Pipe()
```
vtjnash, line 1443:
```julia
                    stderr_pipe = Base.PipeEndpoint()
                    stdout_pipe = Base.PipeEndpoint()
```
If you only need to use them once, a PipeEndpoint is a bit simpler for that.

Otherwise, you must call
```julia
close(stderr_pipe.in)
```
later, once you are done with it.

But since you want to set up the async monitor call early, you are required to use the Pipe call, but you also must initialize it. This will give you much better performance IO anyways, so it is highly recommended.
```
    stderr_pipe =  Base.link_pipe!(Pipe(); reader_supports_async=true, writer_supports_async=true)
    ...
    close(stderr_pipe.in)
```

src/API.jl
```diff
@@ -1273,30 +1273,41 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
+                        end
                     end
                 end
+            catch err
+                err isa InterruptException || rethrow()
+            end
+            @async try
+                while isopen(_stdout)
+                    str = readline(_stdout)
+                    std_outputs[pkg] = string(get(std_outputs, pkg, ""), str, "\n")
```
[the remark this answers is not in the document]
IanButterworth: Do you mean why am I using `readline` rather than `read`? I was using `readline` on stderr so that it's easier to detect `"waiting for IO to finish"`. For stdout read can be used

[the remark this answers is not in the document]
vtjnash: readline is preferable, but any particular reason you are separating the handling now?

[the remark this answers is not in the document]
IanButterworth: I think I misunderstood when you said this

> It is strongly forbidden for the same object to be used asynchronously in two difference places.

I took it to mean a single buffer couldn't be used for stderr and stdout. But you meant that it can't be read from async in two places?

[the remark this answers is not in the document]
vtjnash: A Pipe is special, since it is cloned by the kernel instead of userspace

src/API.jl
```diff
@@ -1469,11 +1469,11 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                         end
                         loaded && (n_loaded += 1)
                     catch err
-                        close(iob)
+                        close(std_pipe.in) # close pipe to end the std output monitor
                         wait(t_monitor)
```
vtjnash, line 1473: Why do these only happen on the error path?

IanButterworth: Good point.

Issue #3535, opened 2023-07-03
"Test-specific dependencies" documentation needs proofreading


mgkuhn:
The section [Test-specific dependencies](https://pkgdocs.julialang.org/dev/creating-packages/#Test-specific-dependencies) in the documentation needs a bit of proof-reading, e.g.

* a paragraph starts with the sentence fragment `is given by test/Project.toml.`
* there is mention of an `old method of adding test-specific dependencies, described in the next section`, but there is no next section on the subject.

https://github.com/JuliaLang/Pkg.jl/blob/2cd30ad9203e1b331b84387580304dd86492e936/docs/src/creating-packages.md?plain=1#L207C1-L218

Issue #1788, opened 2020-04-20
Clarify current fragility of `test/Project.toml` approach


cscherrer:
@fredrikekre 's comment [here](https://github.com/JuliaLang/Pkg.jl/issues/1585#issuecomment-578187367) says,
> Don't use the functionality of `test/Project.toml`s which is pretty beta (and documented as such) 

But this is not clear at all from current documentation [here](https://julialang.github.io/Pkg.jl/v1/creating-packages/#Test-specific-dependencies-in-Julia-1.2-and-above-1), which only says

> The exact interaction between `Project.toml`, `test/Project.toml` and their corresponding `Manifest.toml`s are not fully worked out, and may be subject to change in future versions. The old method of adding test-specific dependencies, described in the next section, will therefore be supported throughout all Julia 1.X releases.

The section title here is "Test-specific dependencies in Julia 1.2 and above", and the qualification seems to me to only indicate that the current approach may change in the future, not that it's "beta" in any way.

I have set up a few repositories using this approach, only to find out today that it's extremely brittle. Can the current documentation be updated to clarify this?

phipsgabler:
There's something else that's not entirely clear here -- the docs say that "Pkg will add the tested package itself implictly".  This works fine when running `] test`, but I also expected the following to behave the same (that's on Julia 1.3.1):

```
(AutoGibbs) pkg> activate test
Activating environment at `~/.julia/dev/AutoGibbs/test/Project.toml`

julia> using AutoGibbs
ERROR: ArgumentError: Package AutoGibbs not found in current path:
- Run `import Pkg; Pkg.add("AutoGibbs")` to install the AutoGibbs package.
```

I guess this is related to the fact that the sub-project thing is still very much under discussion?

KristofferC:
> I guess this is related to the fact that the sub-project thing is still very much under discussion?

Yes, this would work with subprojects but with the current design, `Pkg.test` does "magic" to put these things into a new project that includes everything.

FelixBenning:
I just tried this approach and I can not get it to work:

`test/Project.toml`
```toml
[deps]
Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40
```

if I `activate .` :

```julia
(MyProject) pkg> test
....
ERROR: LoadError: ArgumentError: Package Tests not found in current path:
- Run `import Pkg; Pkg.add("Tests")` to install the Tests package.
```

if I `activate ./test` I can neither do

```julia
(test) pkg> test
[...]

PR #3574, opened 2023-08-09
Tweak test dep docs

IanButterworth:
Alternative proposal to https://github.com/JuliaLang/Pkg.jl/pull/3533
Closes https://github.com/JuliaLang/Pkg.jl/issues/3535

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -191,7 +191,7 @@ load when the package is tested).
 #### `target` based test specific dependencies
 
 Using this method of adding test-specific dependencies, the packages are added under an `[extras]` section and to a test target,
-e.g. to add `Markdown` and `Test` as test dependencies, add the following:
+e.g. to add `Markdown` and `Test` as test dependencies, add the following to the `Project.toml` file:
 
 ```toml
 [extras]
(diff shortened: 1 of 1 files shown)


IanButterworth:
It'd be good to get a 2nd opinion from a Pkg dev here, particularly marking the `test/Project.toml` as an "alternative", which is a bit of a demotion based on discussions like https://github.com/JuliaLang/Pkg.jl/issues/1788

gdalle:
I think it's fair but I understand the hesitation

lgoettgens:
Should this change imply a deprecation of the `test/Project.toml` approach? If yes, I would like to see a PSA about this, and maybe some motivation behind this change. I would like to understand, why the reasons that this approach was introduced in julia 1.2 are no longer valid to keep it around.
Furthermore, I wish for a *documented* way of integrating other sub-projects, most notably `docs/Project.toml` into the main `Project.toml` file.

IanButterworth:
Both approaches are valid. Just that the docs used to be worded in a way that made it seem like `test/Project.toml` was the new better way.

It's better to characterize it as an experimental way with caveats, caveats that were originally documented.

There's nothing wrong with using either, just that the docs now make it clearer that `test/Project.toml` has caveats and isn't a replacement.

PR #3579, opened 2023-08-11
Update manifest entry of project for non-local manifests

quinnj:
Instead of just skipping manifest pruning all together (which is tricky because we're often _not_ operating in a root-level context which would _correctly_ prune everything), we can at least update the dependency list of the currently activated project in the root-level manifest.

### DIFF
--- a/test/project_manifest.jl
+++ b/test/project_manifest.jl
@@ -52,6 +52,7 @@ temp_pkg_dir() do project_path
             m = Pkg.Types.read_manifest(joinpath(dir, "monorepo", "Manifest.toml"))
             # currently, we don't prune dependencies from the root manifest since when rm-ing a dep
             # in a subpackage, we don't also do a full resolve at the root level
+            # https://github.com/JuliaLang/Pkg.jl/issues/3590
             @test_broken !haskey(m, UUID("bf733257-898a-45a0-b2f2-c1c188bdd870")) # D subpackage, but no direct dependency
             pkgC = m[UUID("4ee78ca3-4e78-462f-a078-747ed543fa86")]
             @test !haskey(pkgC.deps, "D")
--- a/test/project_manifest.jl
+++ b/test/project_manifest.jl
@@ -1,7 +1,7 @@
 module ProjectManifestTest
 
 import ..Pkg # ensure we are using the correct Pkg
-using Test, Pkg
+using Test, Pkg, UUIDs
 using ..Utils
 
 temp_pkg_dir() do project_path
@@ -29,6 +29,12 @@ temp_pkg_dir() do project_path
                     Pkg.test()
                 end
             end
+            m = Pkg.Types.read_manifest(joinpath(dir, "monorepo", "Manifest.toml"))
+            @test haskey(m, UUID("dd0d8fba-d7c4-4f8e-a2bb-3a090b3e34f2")) # B subpackage
+            @test haskey(m, UUID("4ee78ca3-4e78-462f-a078-747ed543fa86")) # C subpackage
+            @test haskey(m, UUID("bf733257-898a-45a0-b2f2-c1c188bdd870")) # D subpackage, but no direct dependency
+            pkgC = m[UUID("4ee78ca3-4e78-462f-a078-747ed543fa86")]
+            @test haskey(pkgC.deps, "D")
             cd(joinpath(dir, "monorepo")) do
                 with_current_env() do
                     Pkg.develop(path="packages/C")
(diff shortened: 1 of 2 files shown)


quinnj:
Will try to think of a way to add a test here

nickrobinson251:
thanks for fixing this!

The problem we're seeing is that, if you have a package `Foo` with a `Project.toml` like (note manifest = `"../../Manifest.toml"`)
```toml
name = "Foo"
uuid = "801270c1-6d38-4c07-85b6-74a2f576d89a"
manifest = "../../Manifest.toml"
version = "0.1.0"

[deps]
Compat = "34da2185-b29b-5c13-b0c7-acf172513d20"
ContextVariablesX = "6add18c4-b38d-439d-96f6-d6bc489c04c5"
ExceptionUnwrapping = "460bff9d-24e4-43bc-9d9f-a8973cb893f4"
JSON = "682c06a0-de6a-54ab-a142-c8b1cf79cde6"
Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"

[compat]
ReTestItems = "1"

[extras]
DeepDiffs = "ab62b9b5-e342-54a8-a765-a90f495de1a6"
JSON3 = "0f8b85d8-7281-11e9-16c2-39a750bddbf1"
ReTestItems = "817f1d60-ba6b-4fd5-9520-3cf149f6a823"

[targets]
test = ["DeepDiffs", "JSON3", "ReTestItems", "Test"]
```
and then you add and then remove a package
```
] activate Foo
] add Example
] rm Example
```
The `Manifest.toml` would be left with a diff like
```diff
diff --git a/Manifest.toml b/Manifest.toml
index 1311df8842..2042e8dd9d 100644
--- a/Manifest.toml
+++ b/Manifest.toml
@@ -2,7 +2,7 @@
 
 julia_version = "1.9.2"
 manifest_format = "2.0"
-project_hash = "53c4ec67dcb0841eab26478fa1ef6c1836f9ee8b"
+project_hash = "c42550b13eba1529c8ab6377082cae8993859c62"
 
 [[deps.ANSIColoredPrinters]]
 git-tree-sha1 = "574baf8110975760d391c710b6341da1afa48d8c"
@@ -497,6 +497,11 @@ git-tree-sha1 = "bdb1942cd4c45e3c678fd11569d5cccd80976237"
 uuid = "4e289a0a-7415-4d19-859d-a7e5c4648b56"
 version = "1.0.4"
 
+[[deps.Example]]
+git-tree-sha1 = "46e44e869b4d90b96bd8ed1fdcf32244fddfb6cc"
+uuid = "7876af07-990d-54b4-ab0e-23690620f79a"
+version = "0.5.3"
+
 [[deps.ExceptionUnwrapping]]
 deps = ["Test"]
 git-tree-sha1 = "e90caa41f5a86296e014e148ee061bd6c3edec96"
@@ -1632,7 +1637,7 @@ uuid = "66d59315-31dc-41c8-b200-e39bcbf712f1"
 version = "0.1.0"
 
 [[deps.Foo]]
-deps = ["Compat", "ContextVariablesX", "Dates", "ExceptionUnwrapping", "JSON", "Logging"]
+deps = ["Compat", "ContextVariablesX", "Dates", "Example", "ExceptionUnwrapping", "JSON", "Logging"]
 path = "packages/TransactionLogging"
 uuid = "701270c1-6d38-4c07-85b6-74a2f576d89a"
 version = "0.1.0"
```
Which has two issues:
1. The `deps.Foo` entry is left with `Example` in its `deps`
2. There is a `deps.Example` entry, even though nothing depends on `Example` (given that `Foo` doesn't depend on `Example` after the `rm`)

This PR is to at least fix the first of those issues (cleaning up `deps.Foo`)

test/project_manifest.jl
```diff
@@ -29,13 +29,32 @@ temp_pkg_dir() do project_path
...
+            # now test removing a dependency from subpackage correctly updates root manifest
+            cd(joinpath(dir, "monorepo", "packages", "C")) do
+                with_current_env() do
+                    Pkg.rm("D")
+                    Pkg.test()
+                end
+            end
+            m = Pkg.Types.read_manifest(joinpath(dir, "monorepo", "Manifest.toml"))
+            # currently, we don't prune dependencies from the root manifest since when rm-ing a dep
+            # in a subpackage, we don't also do a full resolve at the root level
```
nickrobinson251, line 54:
```suggestion
            # in a subpackage, we don't also do a full resolve at the root level
            # https://github.com/JuliaLang/Pkg.jl/issues/3590
```

nickrobinson251 (approved):
LGTM!

Opened an issue for the second part of this https://github.com/JuliaLang/Pkg.jl/issues/3590

PR #3580, opened 2023-08-11
Ensure that `string(::VersionSpec)` is compatible with `semver_spec()`

staticfloat:
It is an unfortunate coincidence that if you try to print out a `VersionSpec` to a project's `compat` field, it can be in a format that `semver_spec()` rejects due to the ambiguity of `1.2.0-1.4.0` being either a prerelease with a very strange prerelease tag, or a range of versions.  We reduce the ambiguity by changing `VersionSpec` to always print its bounds with a separating space.

### DIFF
--- a/src/Versions.jl
+++ b/src/Versions.jl
@@ -161,15 +161,15 @@ function Base.print(io::IO, r::VersionRange)
     if (m, n) == (0, 0)
         print(io, '*')
     elseif m == 0
-        print(io, "0-")
+        print(io, "0 -")
         join(io, r.upper.t, '.')
     elseif n == 0
         join(io, r.lower.t, '.')
-        print(io, "-*")
+        print(io, " - *")
     else
         join(io, r.lower.t[1:m], '.')
         if r.lower != r.upper
-            print(io, '-')
+            print(io, " - ")
             join(io, r.upper.t[1:n], '.')
         end
     end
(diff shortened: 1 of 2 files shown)


KristofferC:
@staticfloat, looking at this again, it makes no sense to me. How does `string(::VersionSpec)` have anything to do with the input of `semver_spec`?

staticfloat:
I think this was for BB2, where I want to flow a dependency constraint (specified via a `VersionSpec`, [like this](https://github.com/JuliaPackaging/BinaryBuilder2.jl/blob/f582e2d8fb8fd27a25f40c6e070bff9f6bbf2da6/BinaryBuilderToolchains.jl/src/toolchains/HostToolsToolchain.jl#L43-L47)) down into a `compat` field in a generated `Project.toml`, [like this](https://github.com/JuliaPackaging/BinaryBuilder2.jl/blob/f582e2d8fb8fd27a25f40c6e070bff9f6bbf2da6/JLLGenerator.jl/src/JLLGenerator.jl#L832).  Looking back on this now, I don't think there's a strong reason to believe that `semver_spec(string(::VersionSpec))` should work for all `VersionSpec`, but I think _something_ should.  I notice that I have defensively added a `semver_spec()` call at write-out time to catch this error again, beacuse presumably it causes annoying errors downstream when the Project.toml is not parsed properly.

KristofferC:
Okay, AFAIK, there is virtually no relation between the text output of a VersionSpec and the input to the semver parser so if that is happening somewhere it should probably be looked at.

The extra spaces here makes the registry formatting inconsistent now after the registrator bot updated to Julia 1.11 where this is present.

staticfloat:
How are you supposed to write out a compat entry from a package spec?

KristofferC:
There is no such functionality right now (from what I know).

Issue #3487, opened 2023-05-22
show output for explicitly precompiled packages immediately


vtjnash:
When a user is developing a package, it is useful to be able to include debugging output. But `Pkg.precompile` suppresses that. This causes a regression in usability because of https://github.com/JuliaLang/julia/pull/49242. But when some explicit name passed to `] precompile Name`, the output could be shown immediately, and avoid this regression.

PR #3582, opened 2023-08-12
precompile: if a single package is requested, show output live

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3487

@vtjnash something like this, perhaps?

Screenshot 2023-08-11 at 9 04 01 PM
Screenshot 2023-08-11 at 9 04 32 PM

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1149,12 +1149,15 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
 
     # if the active environment is a package, add that
     ctx_env_pkg = ctx.env.pkg
-    if ctx_env_pkg !== nothing && isfile( joinpath( dirname(ctx.env.project_file), "src", "$(ctx_env_pkg.name).jl") )
-        depsmap[Base.PkgId(ctx_env_pkg.uuid, ctx_env_pkg.name)] = [
+    if ctx_env_pkg !== nothing && isfile(joinpath(dirname(ctx.env.project_file), "src", "$(ctx_env_pkg.name).jl"))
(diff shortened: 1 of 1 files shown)


IanButterworth:
An alternative would be for code loading precompilation to parallel precomp all of the depdendencies but leave off the requested package, then exit and leave the old serial precompilation to finish the main package like it used to. 
Though it would look a bit like a bug if it did that, I think.

src/API.jl
```diff
@@ -1277,10 +1278,21 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     taskwaiting = Set{Base.PkgId}()
     pkgspidlocked = Dict{Base.PkgId,String}()
 
-    function monitor_stderr(pkg, iob)
+    function monitor_stderr(pkg, iob; single_requested_pkg=false)
         try
+            liveprinting = false
             while isopen(iob)
                 str = readline(iob)
```
vtjnash, line 1285:
```suggestion
                str = readline(iob; keep=true)
```
(will give better performance)

IanButterworth:
Updated to
Screenshot 2023-08-14 at 10 09 16 PM

Note that now if you call `precompile` in a package project, it will follow this behavior too.

IanButterworth:
> I like the concept of having output appear as each task finishes for each process. 

It's an interesting idea, but would be a bigger change with some tricky UX because of the way packages disappear when done. This PR is a bit more manageable in UX because it only happens to the last singular package being precompiled.

IanButterworth:
Let's get this on master to get it tried out

PR #3583, opened 2023-08-14
precompile: add pid to waiting message

IanButterworth:
See https://github.com/JuliaLang/julia/pull/50914

Seems like the pid might useful for profiling what's lingering

Screenshot 2023-08-14 at 2 59 59 PM

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1274,7 +1274,7 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     end
 
     stderr_outputs = Dict{Base.PkgId,String}()
-    taskwaiting = Set{Base.PkgId}()
+    taskwaiting = Dict{Base.PkgId,String}()
     pkgspidlocked = Dict{Base.PkgId,String}()
 
     function monitor_stderr(pkg, iob)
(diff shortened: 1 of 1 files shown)


timholy:
If we decide it's actually useful to know the PID, then this would indeed be useful. Thanks!

I don't have time to play with this more today, but presumably this is easy to test. PyPlot 2.11.1 would make an easy test case, see https://github.com/JuliaPy/PyPlot.jl/issues/573.

IanButterworth:
Hmm, yeah. This doesn't look too helpful in this case, which was just a `Timer(30)` added to the module
```
(PNGFiles) pkg> precompile
Precompiling project...
  1 dependency successfully precompiled in 32 seconds. 28 already precompiled.
  1 dependency had output during precompilation:
┌ PNGFiles [f57f5aa1-a3ce-4bc8-8ab9-96f992907883]
│  [pid 52880] waiting for IO to finish:
│   TYPE[FD/PID]       @UV_HANDLE_T->DATA
│   timer              @0x6000027cd480->0x12bf29ea0
│  
│  ======================================================================================
│  Information request received. A stacktrace will print followed by a 1.0 second profile
│  ======================================================================================
│  
│  signal (29): Information request: 29
│  kevent at /usr/lib/system/libsystem_kernel.dylib (unknown line)
│  unknown function (ip: 0x0)
│  
│  ==============================================================
│  Profile collected. A report will print at the next yield point
│  ==============================================================
│  
│  ┌ Warning: There were no samples collected in one or more groups.
│  │ This may be due to idle threads, or you may need to run your
│  │ program longer (perhaps by running it multiple times),
│  │ or adjust the delay between samples with `Profile.init()`.
│  └ @ Profile ~/Documents/GitHub/julia/usr/share/julia/stdlib/v1.11/Profile/src/Profile.jl:1201
│  Overhead ╎ [+additional indent] Count File:Line; Function
│  =========================================================
│  Thread 1 Task 0x0000000102e18010 Total snapshots: 780. Utilization: 0%
│  
│  
│  [pid 52880] waiting for IO to finish:
│   TYPE[FD/PID]       @UV_HANDLE_T->DATA
│   timer              @0x6000027cd480->0x12bf29ea0
└  
```

IanButterworth:
There doesn't seem to be benefit to doing this. Also the output will be shown live in common dev usage patterns now thanks to https://github.com/JuliaLang/Pkg.jl/pull/3582

PR #3594, opened 2023-08-17
Update manifest entry of project for non-local manifests (#3579)

quinnj:
* Update manifest entry of project for non-local manifests

* Add a test

* Update test/project_manifest.jl

---------


IanButterworth:
This is now in https://github.com/JuliaLang/Pkg.jl/pull/3596

PR #3600, opened 2023-08-21
Clarify rationale for package extension naming

sloede:
Especially in light of issues such as https://github.com/JuliaLang/julia/issues/50873, where users are faced with warnings such as
```
Precompiling project...
  194 dependencies successfully precompiled in 92 seconds. 10 already precompiled.
  3 dependencies had warnings during precompilation:
┌ SummationByPartsOperatorsDiffEqCallbacksExt [82d7d246-636b-5c42-a4f7-7a8a453087a7]
│  [pid 720185] waiting for IO to finish:
│   TYPE[FD/PID]       @UV_HANDLE_T->DATA
│   timer              @0x107c040->0x7f8e2e0e9060
└
┌ LinearMapsStatisticsExt [364b1850-867c-5a65-9f2f-642d093ba2f6]
│  [pid 704422] waiting for IO to finish:
│   TYPE[FD/PID]       @UV_HANDLE_T->DATA
│   timer              @0x261b550->0x7f6d92ae8f10
└
┌ ExponentialUtilities [d4d017d3-3776-5f7e-afef-a10c40355c18]
│  [pid 712737] waiting for IO to finish:
│   TYPE[FD/PID]       @UV_HANDLE_T->DATA
│   timer              @0x955b000->0x7f9e61c56230
└
```
it would be very helpful to have both the main package name and the extension trigger package encoded in the extension name (in this example here, it is already done so for `SummationByPartsOperatorsDiffEqCallbacksExt` and `LinearMapsStatisticsExt`). This PR adds a sentence to the docs to clarify the rationale for the recommendation to further nudge pkg devs to stick to this convention.

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -356,6 +356,8 @@ end # module
 
 Extensions can have any arbitrary name (here `PlottingContourExt`), but using something similar to the format of
 this example that makes the extended functionality and dependency of the extension clear is likely a good idea.
+Encoding in the extension name both the name of the main package and of the package that triggers loading the
+extension can be very helpful especially when encountering and debugging errors.
 
 A user that depends only on `Plotting` will not pay the cost of the "extension" inside the `PlottingContourExt` module.
 It is only when the `Contour` package actually gets loaded that the `PlottingContourExt` extension is loaded


IanButterworth:
I think the length of that extension name is actually a reason not to and perhaps we should make Pkg show the parent and the ext name in the output log, like it does during the precompilation process

sloede:
I agree, it would be even better if the extension name would not show up anywhere without its parent. But is that even feasible?

IanButterworth:
See https://github.com/JuliaLang/Pkg.jl/pull/3603

sloede:
With #3603 incoming, are there any other prominent places left, where an extension's name is shown without its parent? Because if not, I would consider this PR here superseded by #3603 and withdraw it.

sloede:
> Maybe, it's better to ensure that extension packages are always shown with their parents (solved by #3603?), and actually exemplify and recommend naming `ContourExt` instead of `PlottingContourExt` in the docs?

I completely agree. However, as we just saw with this PR, there _was_ yet another case where a pkg extension was shown without reference to its parent. Once somebody with enough knowledge about Pkg.jl is reasonably sure that all such "orphaned" references to pkg extensions have been fixed, this would be a good step though!

IanButterworth:
It does seem better for extension names to not include the parent, otherwise they get huge, and it feels like too much burden for the developer to get right.

Screenshot 2023-09-15 at 9 14 47 AM

Can we make a list of all the places we need to fix where just the extension name is shown.

I think stack traces are ok because they show the path.

(perhaps this should be an issue)

IanButterworth:
We went in the other direction https://github.com/JuliaLang/Pkg.jl/pull/4000

sloede:
OK - thanks for letting me know!

Issue #3516, opened 2023-06-12
`compat` repl prints in wacky order.


oscardssmith:
Using the repl to update compat prints the dependencies in no apparent order despite the fact that the compat entries in the file are in alphabetical order.
![image](https://github.com/JuliaLang/Pkg.jl/assets/11729272/f700ad3e-a18b-4d74-83af-a91bbe53b31e)

PR #3605, opened 2023-08-28
sort compat entries in `pkg> compat`

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3516


IanButterworth (approved):

oscardssmith:
thanks! technically this probably should have used `sort!` to avoid an extra copy but that's a very minor detail.

KristofferC:
> thanks! technically this probably should have used sort! to avoid an extra copy but that's a very minor detail.

The extra character is not worth it over the negligible allocation change ;)

PR #3613, opened 2023-09-07
Note disabling autoprecompilation in Pkg.instantiate docstring

mortenpi:
Just to make this information a bit easier to find (especially in the REPL).


src/Pkg.jl
```diff
@@ -465,6 +465,8 @@ If no `Project.toml` exist in the current active project, create one with all th
 dependencies in the manifest and instantiate the resulting project.
 
 After packages have been installed the project will be precompiled.
+The automatic precompilation can be disabled by setting the environment variable
+`JULIA_PKG_PRECOMPILE_AUTO=0`.
 See more at [Environment Precompilation](@ref).
```
IanButterworth:
```suggestion
See more and how to disable auto-precompilation at [Environment Precompilation](@ref).
```

IanButterworth (approved):
I changed it to mention disabling it with the docs link, because we've got other ways of controlling it incoming https://github.com/JuliaLang/Pkg.jl/pull/4262

PR #3615, opened 2023-09-09
[CompatHelper] Debugging the behavior of `force_latest_compatible_version` when `test/Project.toml` exists

DilumAluthge:


KristofferC:
I'm guessing this is not used anymore

PR #3606, opened 2023-08-28
tweak how Pkg is loaded for precompiling when testing

KristofferC:
This is to prevent https://github.com/JuliaLang/Pkg.jl/issues/3442 when moving Pkg out of the sysimage.

PR #3620, opened 2023-09-13
fix active project of test precompilation

IanButterworth:
Fixes bug reported at https://github.com/JuliaLang/Pkg.jl/pull/3606#issuecomment-1716727635

Verified with a julia build pointing at this commit


KristofferC:
Thanks!

PR #3631, opened 2023-09-26
make precompile unwrap UnstableIO

KristofferC:
The new `UnstableIO` is good for latency but it seems to be a bit too big runtime penalty for precompile which outputs a lot of text to the terminal.

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -46,6 +46,7 @@ struct UnstableIO <: IO
 end
 Base.write(io::UnstableIO, b::UInt8) = write(io.io, b)::Int
 Base.get(io::UnstableIO, val, default) = get(io.io, val, default)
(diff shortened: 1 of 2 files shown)


IanButterworth (approved):
LGTM. Confirmed with a full julia build with this

src/Pkg.jl
```diff
@@ -46,6 +46,7 @@ struct UnstableIO <: IO
 end
 Base.write(io::UnstableIO, b::UInt8) = write(io.io, b)::Int
 Base.get(io::UnstableIO, val, default) = get(io.io, val, default)
+Base.print(io::UnstableIO, arg::Union{SubString{String}, String}) = print(io.io, arg)
```
vtjnash, line 49: This seems like a method you are not supposed to define as it is not compatible with the usual API for this. It seems like you should make UnstableIO be a subtype of `AbstractPipe`, and get all of the required API functions automatically instead.

Issue #3632, opened 2023-09-26
extraneous misleading output during precompile statements


vtjnash:
Precompile cache files are not supposed to depend on random files on the users' system. And generally is not supposed to go around deleting files either. Seems like Pkg.precompile is doing both of those bad things?
```
    JULIA build-x86/stdlib/Pkg.release.image
[ Info: We haven't cleaned this depot up for a bit, running Pkg.gc()...
```

PR #3633, opened 2023-09-27
Make auto GC message use printpkgstyle

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3632

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -797,7 +797,9 @@ function _auto_gc(ctx::Types.Context; collect_delay::Period = Day(7))
     end
 
     if curr_time - DEPOT_ORPHANAGE_TIMESTAMPS[depots1()] > delay_secs
-        @info("We haven't cleaned this depot up for a bit, running Pkg.gc()...")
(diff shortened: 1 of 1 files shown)


IanButterworth:
Maybe the better fix is to turn this into a `printpkgstyle`? It's one of the places where Pkg logs where I think it should use the Pkg print format

src/Pkg.jl
```diff
@@ -797,7 +797,9 @@ function _auto_gc(ctx::Types.Context; collect_delay::Period = Day(7))
     end
 
     if curr_time - DEPOT_ORPHANAGE_TIMESTAMPS[depots1()] > delay_secs
-        @info("We haven't cleaned this depot up for a bit, running Pkg.gc()...")
+        if ccall(:jl_generating_output, Cint, ()) != 1
+            @info("We haven't cleaned this depot up for a bit, running Pkg.gc()...")
+        end
```
IanButterworth, line 802:
```suggestion
        printpkgstyle(ctx.io, :Info, "We haven't cleaned this depot up for a bit, running Pkg.gc()...", color = Base.info_color())
```

PR #3642, opened 2023-10-04
feat: add JULIA_PKG_GC_AUTO env var

pfitzseb:
This PR adds a `JULIA_PKG_DISABLE_AUTO_GC` environment variable to allow users to disable automatic garbage collection without reaching into Pkg's internals and setting `_auto_gc_enabled[] = false` manually.


ericphanson:
This should get documented here too: https://docs.julialang.org/en/v1/manual/environment-variables/#Pkg.jl

src/Pkg.jl
```diff
@@ -311,6 +311,9 @@ finding artifacts and packages that are thereafter not used by any other project
 marking them as "orphaned".  This method will only remove orphaned objects (package
 versions, artifacts, and scratch spaces) that have been continually un-used for a period
 of `collect_delay`; which defaults to seven days.
+
+To disable automatic garbage collection, you can set the environment variable 
+`JULIA_PKG_DISABLE_AUTO_GC` to `"true"`.
```
ericphanson, line 316:
```suggestion
To disable automatic garbage collection, you can set the environment variable 
`JULIA_PKG_DISABLE_AUTO_GC` to `"true"` before starting Julia.
```

pfitzseb:
-> https://github.com/JuliaLang/julia/pull/51583

ericphanson (approved):
looks good to me! it could be nice to dynamically check before doing auto GC's rather than setting it in `__init__` so one can set the ENV variable later. But this already seems an improvement so IMO that shouldn't block merging.

staticfloat (approved):

staticfloat:
> so one can set the ENV variable later.

I think the general pattern of `ENV["JULIA_PKG_xxx"] = "false"` in scripts should be discouraged, and `Pkg._auto_gc_enabled[] = false` should be encouraged for these reasons:

* [Environment variables are not thread safe](https://github.com/JuliaLang/julia/issues/34726) and so should generally be thought of as a communication mechanism from outside of Julia to inside of Julia, not communication for things within Julia itself.
* Priority can be difficult to determine; if `ENV["JULIA_PKG_DISABLE_AUTO_GC"]` is set to `1`, but then the user later sets `_auto_gc_enabled[] = true`, which one wins?  In the current setup, this is clear (`_auto_gc_enabled[]` is the ground truth, it copies the value of `ENV["JULIA_PKG_DISABLE_AUTO_GC"]` at `__init__()` and then the value is thereafter ignored).

staticfloat (approved):

pfitzseb:
Yeah, that second point is also what I was mostly concerned about. 
Might be worth adding a function to set `_auto_gc_enabled` without touching that very internal looking ref, e.g. `Pkg.API.auto_gc(::Bool)`. Thoughts?

fingolfin:
This was approved almost a year ago, but not merged. What gives?

DilumAluthge:
@staticfloat Is this good to merge?

DilumAluthge:
Also bump @pfitzseb

fredrikekre:
Should it be `JULIA_PKG_AUTO_GC=true/false` instead?

IanButterworth:
Yeah like `JULIA_PKG_PRECOMPILE_AUTO=false`

IanButterworth:
Or `JULIA_PKG_GC_AUTO` to match the order

pfitzseb:
Renamed the env var and added `API.auto_gc` function for getting and setting the `_auto_gc_enabled` state.

src/API.jl
```diff
@@ -1582,4 +1582,18 @@ function upgrade_manifest(ctx::Context = Context())
     return nothing
 end
 
+"""
+    auto_gc()
+
+Query whether Pkg will automatically [`gc`](@ref) packages and artifacts.
+"""
+auto_gc() = _auto_gc_enabled[]
```
KristofferC, line 1590: Can we use the same API as `GC.enable` (where it returns the previous state) to avoid having two methods?

pfitzseb: Sure, done.

PR #3651, opened 2023-10-12
Fix algorithmic complexity of cycle detection

Keno:
The circular dependency detection in `precompile` has poor algorithmic complexity. In a project with deep and broad dependency trees of hundreds of packages, this introduces a significant delay before the start of precompilation work. Fix that by switching the cycle detection to a simple open-coded DFS walk.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1208,13 +1208,35 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
 
     # find and guard against circular deps
     circular_deps = Base.PkgId[]
-    function in_deps(_pkgs, deps, dmap)
-        isempty(deps) && return false
-        !isempty(intersect(_pkgs, deps)) && return true
-        return any(dep->in_deps(vcat(_pkgs, dep), dmap[dep], dmap), deps)
+    # Three states
+    # !haskey -> never visited
+    # true -> could be part of a cycle
+    # false -> not part of a cycle
+    could_be_cycle = Dict{Base.PkgId, Bool}()
+    dfs_num = 0
+    function scan_pkg!(pkg, dmap)
+        did_visit_dep = true
+        inpath = get!(could_be_cycle, pkg) do
+            did_visit_dep = false
+            return true
+        end
+        if did_visit_dep ? inpath : scan_deps!(pkg, dmap)
+            # Found a cycle. Delete this and all parents
+            # to let this cycle be found again by another toplevel dep
+            delete!(could_be_cycle, pkg)
+            return true
+        end
+        return false
     end
-    for (pkg, deps) in depsmap
-        if in_deps([pkg], deps, depsmap)
+    function scan_deps!(pkg, dmap)
+        for dep in dmap[pkg]
+            scan_pkg!(dep, dmap) && return true
+        end
+        could_be_cycle[pkg] = false
+        return false
+    end
+    for pkg in keys(depsmap)
+        if scan_pkg!(pkg, depsmap)
             push!(circular_deps, pkg)
             notify(was_processed[pkg])
         end
--- a/src/API.jl
+++ b/src/API.jl
@@ -1210,10 +1210,9 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     circular_deps = Base.PkgId[]
     # Three states
     # !haskey -> never visited
-    # true -> could be part of a cycle
-    # false -> not part of a cycle
+    # true -> cannot be compiled due to a cycle (or not yet determined)
+    # false -> not depending on a cycle
     could_be_cycle = Dict{Base.PkgId, Bool}()
-    dfs_num = 0
     function scan_pkg!(pkg, dmap)
         did_visit_dep = true
         inpath = get!(could_be_cycle, pkg) do
@@ -1223,7 +1222,6 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
         if did_visit_dep ? inpath : scan_deps!(pkg, dmap)
             # Found a cycle. Delete this and all parents
             # to let this cycle be found again by another toplevel dep
-            delete!(could_be_cycle, pkg)
             return true
         end
         return false
--- a/src/API.jl
+++ b/src/API.jl
@@ -1221,7 +1221,6 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
         end
         if did_visit_dep ? inpath : scan_deps!(pkg, dmap)
             # Found a cycle. Delete this and all parents
-            # to let this cycle be found again by another toplevel dep
             return true
         end
         return false


src/API.jl
```diff
@@ -1208,13 +1208,27 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
+            # Will be visited at top level
+            depdepth == 0 && continue
+            if depdepth < depth+1
+                # Found cycle
+                return true
+            elseif !already_visited
+                dfs_scan!(dep, dmap, depth+1) && return true
+            end
+        end
+        return false
```
vtjnash, line 1228:
```suggestion
        depths[dep] = 0
        return false
```
(and delete the already_visited code)

src/API.jl
```diff
@@ -1208,13 +1208,27 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
+        for dep in dmap[pkg]
+            already_visited = true
+            depdepth = get!(depths, dep) do
+                already_visited = false
+                depth + 1
+            end
+            # Will be visited at top level
+            depdepth == 0 && continue
+            if depdepth < depth+1
+                # Found cycle
```
vtjnash, line 1222:
```suggestion
                # Found cycle
                depths[dep] = typemax(Int)
```

vtjnash: Actually, a state enum might be easier for this (not-visited, visiting, cycle, not-cycle) since it does not need the depth itself (as it would if this was the full Tarjan's SCC for example instead)

IanButterworth:
Appreciate the help with this. Thanks. CI seems to have timed out, indicating it may not be catching every circular dep?

Keno:
Yes, this has a bug. Will try to fix.

src/API.jl
```diff
@@ -1208,13 +1208,35 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
+    function scan_pkg!(pkg, dmap)
+        did_visit_dep = true
+        inpath = get!(could_be_cycle, pkg) do
+            did_visit_dep = false
+            return true
+        end
+        if did_visit_dep ? inpath : scan_deps!(pkg, dmap)
+            # Found a cycle. Delete this and all parents
+            # to let this cycle be found again by another toplevel dep
+            delete!(could_be_cycle, pkg)
```
vtjnash, line 1226:
```suggestion
```
This does not seem needed. The dep is already in the is-cycle state.

src/API.jl
```diff
@@ -1208,13 +1208,35 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
+            return true
+        end
+        return false
```
vtjnash, line 1229: no reason to visit this again later, since it is proven here to be not in a cycle
```suggestion
        could_be_cycle[pkg] = false
        return false
```

src/API.jl
```diff
@@ -1208,13 +1208,35 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
     end
-    for (pkg, deps) in depsmap
-        if in_deps([pkg], deps, depsmap)
+    function scan_deps!(pkg, dmap)
+        for dep in dmap[pkg]
+            scan_pkg!(dep, dmap) && return true
+        end
+        could_be_cycle[pkg] = false
```
vtjnash, line 1235:
```suggestion
```

src/API.jl
```diff
@@ -1208,13 +1208,35 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
-    function in_deps(_pkgs, deps, dmap)
-        isempty(deps) && return false
-        !isempty(intersect(_pkgs, deps)) && return true
-        return any(dep->in_deps(vcat(_pkgs, dep), dmap[dep], dmap), deps)
+    # Three states
+    # !haskey -> never visited
+    # true -> could be part of a cycle
+    # false -> not part of a cycle
+    could_be_cycle = Dict{Base.PkgId, Bool}()
+    dfs_num = 0
```
vtjnash, line 1216:
```suggestion
```

src/API.jl
```diff
@@ -1208,13 +1208,35 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
 
     # find and guard against circular deps
     circular_deps = Base.PkgId[]
-    function in_deps(_pkgs, deps, dmap)
-        isempty(deps) && return false
-        !isempty(intersect(_pkgs, deps)) && return true
-        return any(dep->in_deps(vcat(_pkgs, dep), dmap[dep], dmap), deps)
+    # Three states
+    # !haskey -> never visited
+    # true -> could be part of a cycle
```
vtjnash, line 1213:
```suggestion
    # true -> cannot be compiled due to a cycle (or not yet determined)
```

src/API.jl
```diff
@@ -1208,13 +1208,35 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
+    # false -> not part of a cycle
```
vtjnash, line 1214:
```suggestion
    # false -> not depending on a cycle
```

src/API.jl
Keno: I don't see why this is necessarily better here

Keno: It just does a useless extra dict lookup in the hot path.

vtjnash: Ah, right they are equivalent but I misread it and thought all of the dict manipulation needed to be from the same function--it doesn't.

IanButterworth:
CI failures are:
- Windows timing out (for some reason Windows tests have just got slower recently - needs investigating)
- Intermittent MacOS git registry bug

I tested this locally and it looks good, but I have to say that things didn't used to be as bad as they appear on master currently.. calling `precompile` on OmniPackage on my new mac takes > 2 minutes to start 😱 where in the past it started in < 10-20 seconds. (otherwise I would've found it too painful and prioritized this kind of thing)

However this brings things back to a few seconds, which is great. Thanks

vtjnash:
FWIW, this is also a potential great start on a DFS algorithm for deciding the order to compile packages: a post-order listing of the result of `scan_deps!` is a valid linear traversal of the tree as well as an optimal implementation of the `collect_all_deps` functions that follows.

src/API.jl
```diff
@@ -1208,13 +1208,32 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
-        if in_deps([pkg], deps, depsmap)
+    function scan_deps!(pkg, dmap)
+        for dep in dmap[pkg]
+            scan_pkg!(dep, dmap) && return true
+        end
+        could_be_cycle[pkg] = false
+        return false
+    end
+    for pkg in keys(depsmap)
+        if scan_pkg!(pkg, depsmap)
```
vtjnash, line 1236: Doesn't this have to be `scan_deps!`, since otherwise it is missing the `could_be_cycle[pkg]` determination at the end?

PR #3652, opened 2023-10-12
Switch datastructure Vector -> Set for algorithmic complexity

Keno:
In the same project as #3651, this code was also having trouble because of the rapidly growing alldeps, which would end up with many repeated entries. Switch the data structure to fix the algorithmic complexity here also.


IanButterworth (approved):
Makes sense. Thanks

PR #3653, opened 2023-10-14
Make stdlib upgradeability opt-out

vchuravy:
Upgradeability is determined by having a registry entry for the stdlib in the registry.
There is a set of stdlibs that have compatibility entries in the registry, but those are not
yet updated with the new "1.11" and above versions.

Unsure what to do about the "UNREGISTERED_STDLIBS" dictionary, it currently contains "Statistics",
but "Statistics" is upgradeable. So do we need to duplicate that data by versions?

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -44,12 +44,12 @@ end
 # more accurate name is `should_be_tracking_registered_version`
 # the only way to know for sure is to key into the registries
 tracking_registered_version(pkg::Union{PackageSpec, PackageEntry}, julia_version=VERSION) =
-    !is_stdlib(pkg.uuid, julia_version) && pkg.path === nothing && pkg.repo.source === nothing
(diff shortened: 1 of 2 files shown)


vchuravy:
I am a bit uncertain about HistoricalStdlibVersions.jl

It currently has:

```
        UUID("0dad84c5-d112-42e6-8d28-ef12dabb789f") => ("ArgTools", v"1.1.1"),
        UUID("4af54fe1-eca0-43a8-85a7-787d91b784e3") => ("LazyArtifacts", nothing),
```

Both are registered. The version we have for ArgTools on julia master is https://github.com/JuliaIO/ArgTools.jl/commit/4eccde4 which is different from `v"1.1.1"` https://github.com/JuliaIO/ArgTools.jl/releases/tag/v1.1.1

PR #3654, opened 2023-10-15
cache pidlock tweaks

IanButterworth:
See https://github.com/JuliaLang/julia/pull/51714

Needs to wait for https://github.com/JuliaLang/julia/pull/51714


src/API.jl
```diff
@@ -1656,20 +1656,21 @@ function color_string(cstr::String, col::Union{Int64, Symbol})
 end
 
 function maybe_cachefile_lock(f, io::IO, print_lock::ReentrantLock, fancyprint::Bool, pkg::Base.PkgId, pkgspidlocked::Dict{Base.PkgId,String})
+    stale_age = 10  # match stale_age in loading.jl
```
vchuravy, line 1659: Should we have loading export a global with this default?

IanButterworth: Yeah. I guess even when this moves to base it will be in 2 different locations still.

PR #3656, opened 2023-10-15
Remove test that depends on Random being in the sysimg

vchuravy:
x-ref: https://github.com/JuliaLang/julia/pull/51432#issuecomment-1758460235

### DIFF
--- a/test/api.jl
+++ b/test/api.jl
@@ -271,7 +271,6 @@ import .FakeTerminals.FakeTerminal
         @test_throws Pkg.Types.PkgError Pkg.precompile("DoesNotExist") # fail to find a nonexistant dep in an empty env
 
         Pkg.add("Random")
-        @test_throws Pkg.Types.PkgError Pkg.precompile("Random") # Random is a dep but in the sysimage
         @test_throws Pkg.Types.PkgError Pkg.precompile("DoesNotExist")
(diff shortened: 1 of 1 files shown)


test/api.jl
```diff
@@ -271,7 +271,6 @@ import .FakeTerminals.FakeTerminal
         @test_throws Pkg.Types.PkgError Pkg.precompile("DoesNotExist") # fail to find a nonexistant dep in an empty env
 
         Pkg.add("Random")
-        @test_throws Pkg.Types.PkgError Pkg.precompile("Random") # Random is a dep but in the sysimage
```
IanButterworth, line 274: Replace with something in the sysimage? Also the load above needs changing

vchuravy: The issue is that if all my plans are going to work there will be nothing in the sysimg.

The add there still is useful since it tests that adding a stdlib works.

IanButterworth: Ahem.. https://github.com/JuliaLang/julia/pull/51463#issuecomment-1739754305

True, but Pkg still needs to serve people that make sysimages with modules in them.
Building a whole other sysimage would be too much, so perhaps this just has to be untested then..

vchuravy: We might need an "EmptyStdlib" to make sure that the loading infrastructure is properly precompiled.

PR #3657, opened 2023-10-15
Fix scan_pkg issue with cycle over-detection

vtjnash:
The main intent here is to ensure that `could_be_cycle[pkg] = false` happens even when called from the root package, and not just recursively from `scan_deps`

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1219,15 +1219,11 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
             did_visit_dep = false
             return true
         end
-        if did_visit_dep ? inpath : scan_deps!(pkg, dmap)
-            # Found a cycle. Delete this and all parents
-            return true
+        if did_visit_dep 
+            return inpath
         end
-        return false
-    end
-    function scan_deps!(pkg, dmap)
         for dep in dmap[pkg]
-            scan_pkg!(dep, dmap) && return true
+            scan_pkg!(dep, dmap) && return true # Encountered a cycle. Delete this and all parents.
         end
         could_be_cycle[pkg] = false
         return false


Keno:
I don't understand why this patch isn't a no-op. On what path does this change behavior?

vtjnash (commented):
For the root, I didn't see where it would get set to false eventually in could_be_cycle, but I could have just missed it

Keno:
It should get set on return from scan_deps!. I mean, I may have messed it up, but just from what I can tell, this PR is literally equivalent.

vtjnash:
Doesn't the root of the search call scan_pkg and not call scan_deps? I also might not be reading that bit right, this change is purely by visual inspection

Keno:
> Doesn't the root of the search call scan_pkg and not call scan_deps? I also might not be reading that bit right, this change is purely by visual inspection

scan_deps! always gets called on the first call to scan_pkg. Again, as far as I can tell, this PR is a no-op.

vtjnash:
Ah, okay, right. I was confused by having a mutating helper method in a conditional if statement, and that wasn't processing in my mind as being executed always. This PR might remove the need for Box-ing of the two functions for the mutual recursion pairing, but they are functionally equivalent.

IanButterworth:
It'd be tidy to resolve this before moving this code to Base.

On a fully precompiled env with DiffEq

This PR
```
julia> @time Pkg.precompile()
  0.333714 seconds (579.76 k allocations: 77.397 MiB, 1.35% gc time)

julia> @time Pkg.precompile()
  0.340852 seconds (579.76 k allocations: 77.397 MiB, 2.02% gc time)
```

master
```
julia> @time Pkg.precompile()
  0.416346 seconds (579.03 k allocations: 77.364 MiB, 23.80% gc time)

julia> @time Pkg.precompile()
  0.332330 seconds (579.03 k allocations: 77.364 MiB)
```

KristofferC:
> they are functionally equivalent.

So I don't think this really needs to be "resolved".

PR #3661, opened 2023-10-16
Let the user disable Pkg's precompilation workload by setting the `JULIA_PKG_PRECOMPILATION_WORKLOAD` environment variable to `false` or `0`

DilumAluthge:
I need some way of disabling this precompilation workload so that I can fix the RegistryCI.jl test suite on Julia nightly.

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -108,5 +108,8 @@ function pkg_precompile()
     end
 end
 
-pkg_precompile()
+if Base.get_bool_env("JULIA_PKG_PRECOMPILATION_WORKLOAD", true)
+    pkg_precompile()
+end
+
 end


vchuravy:
This should be fine, but why is PKG precompilation failing/triggering?

DilumAluthge:
> why is PKG precompilation failing/triggering?

Example log: https://github.com/JuliaRegistries/RegistryCI.jl/actions/runs/6528846119/job/17725625648?pr=514

Snippet:

```
ERROR: LoadError: The following package names could not be resolved:
 * TestPkg (not found in project, manifest or registry)
   Suggestions: IsoPkg TestEnv TestImages
```

Pkg's precompilation workload creates a fake registry and adds a package called `TestPkg` to it. And Pkg makes this an uncompressed tarball registry, i.e.:
1. `Registry.toml` exists.
2. `Registry.tar.gz` exists.
3. The tarball has not been extracted.
4. The registry isn't a Git repo.

But in RegistryCI.jl, for a variety of reasons, we set `JULIA_PKG_SERVER` to `""`. So then, when running the workload, maybe Pkg isn't using the fake registry and is instead trying to get a Git registry?

DilumAluthge:
Nope, that can't be it. The workload overrides any pre-existing value of `JULIA_PKG_SERVER`:

https://github.com/JuliaLang/Pkg.jl/blob/a8648f7c83bf561e0c3c6d33d23476067f382d0f/src/precompile.jl#L83-L83

vchuravy:
@IanButterworth pointed out that PrecompileTools has a feature for disabling workloads per package.

My recommendation for now is to add the global depot to the subprocess depot such that we don't need to recompile PKG in the first place.

DilumAluthge:
Ah. RegistryCI also sets `JULIA_PKG_UNPACK_REGISTRY` - maybe that's related?

DilumAluthge:
We could try https://github.com/JuliaLang/Pkg.jl/pull/3662

KristofferC:
Can be closed?

PR #3668, opened 2023-10-17
Precompilation workload: restore the original depot and load paths

DilumAluthge:


src/precompile.jl
```diff
@@ -107,6 +110,11 @@ function pkg_precompile()
             # end
         # end
     end
+    empty!(DEPOT_PATH)
```
KristofferC, line 113: Is this the same as `copy!(DEPOT_PATH, original_depot_path)`?

src/precompile.jl
```diff
@@ -107,6 +110,11 @@ function pkg_precompile()
...
+    empty!(LOAD_PATH)
+    append!(DEPOT_PATH, original_depot_path)
```
fatteneder, line 115: Would it make sense to wrap the code into a `try ... finally ... end` block so that the paths are restored upon failure?

KristofferC: If the precompilation process fails, there won't be any outputted precompile files, so does it matter at that point?

KristofferC: I don't really see how that is relevant since the precompilation process runs in a separate process so any modifications to variables are only local for that process. So if this errors then we will immediately exit the process. 

In fact, I am not sure how this change does anything useful...

KristofferC:
Since this is the last thing that the process does I don't see how this does anything. But maybe it is nice aesthetically.

vchuravy:
> Since this is the last thing that the process does I don't see how this does anything.

Afterwards we do the work of serializing the actual data and https://github.com/JuliaLang/julia/pull/49866 is going to look at the depot path to rewrite the source paths we store in the cache file

DilumAluthge (commented):
The various code review suggestions LGTM.

PR #3676, opened 2023-10-25
add Pkg to test env as a dep rather than via LOAD_PATH

IanButterworth:
Fixes https://github.com/JuliaLang/julia/issues/51820

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1650,9 +1650,7 @@ function gen_test_code(source_path::String; coverage, julia_args::Cmd, test_args
 end
 
 function gen_test_precompile_code(source_path::String; coverage, julia_args::Cmd, test_args::Cmd)
-    pkgdir = joinpath(@__DIR__, "..")
     code = """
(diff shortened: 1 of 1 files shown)


src/Operations.jl
```diff
@@ -1853,6 +1851,7 @@ function gen_target_project(ctx::Context, pkg::PackageSpec, source_path::String,
     if projectfile_path(source_path; strict=true) === nothing
         # no project file, assuming this is an old REQUIRE package
         test_project.deps = copy(env.manifest[pkg.uuid].deps)
+        get!(test_project.deps, "Pkg", PkgUUID) # so that the test environment can be instantiated & precompiled
```
IanButterworth, line 1854: Perhaps to catch the edge case where Pkg is another package..
```suggestion
        # add Pkg so that the test environment sandbox subprocesses can be precompiled
        if get!(test_project.deps, "Pkg", PkgUUID) != PkgUUID
            @warn """
            A package called Pkg is declared with a UUID that doesn't match the Pkg stdlib.
            This may cause unexpected behavior"""
        end
```

PR #3683, opened 2023-11-03
Don't clean test temp files

IanButterworth:
Both GitHub CI here and buildkite CI on the julia repo restore the fs to a clean state, likely faster than doing it within the julia process, so we don't need to spend time tidying up.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -28,9 +28,9 @@ p7zip_jll = "3f19e933-33d8-53b3-aaab-bd5110c3b7a0"
 HistoricalStdlibVersions = "1.2"
 
 [extras]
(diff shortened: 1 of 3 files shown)


IanButterworth:
Oof. I guess there's no `mktempdir(f; cleanup = false)` method. Makes sense, but makes this require lot more code change

KristofferC:
I thought we would just empty the `Base.TEMP_CLEANUP` variable or something like that in case `ENV["CI"]` exists.

IanButterworth:
Seems like a decent reduction but `windows-latest, x86, pkg.julialang.org, nightly` is an outlier and remains > 1hr. Odd

PR #3694, opened 2023-11-10
docs: add `manifest` field documentation to Project.toml guide in Pkg.jl manual

sjkelly:
This should be mentioned here.
Added in: https://github.com/JuliaLang/Pkg.jl/pull/3579

### DIFF
--- a/docs/src/toml-files.md
+++ b/docs/src/toml-files.md
@@ -77,6 +77,17 @@ Note that Pkg.jl deviates from the SemVer specification when it comes to version
 the section on [pre-1.0 behavior](@ref compat-pre-1.0) for more details.
 
 
+### The `manifest` field
+
+`manifest` specifies an external manifest file (`Manifest.toml`) for the project. This is particularly useful for managing package dependencies in a more granular or shared manner across multiple projects, for example:
+```toml
+manifest = "../Manifest.toml"
+```
+
+This specified `Manifest.toml` does not need to be a subset of the project's own manifest. 
+The project inherits any version constraints or package specifications from the referenced manifest and is integrated into the dependency resolution process.
+
+
 ### The `[deps]` section
 
 All dependencies of the package/project are listed in the `[deps]` section. Each dependency


docs/src/toml-files.md
```diff
@@ -77,6 +77,17 @@ Note that Pkg.jl deviates from the SemVer specification when it comes to version
...
+### The `manifest` field
+
+`manifest` specifies an external manifest file (`Manifest.toml`) for the project. This is particularly useful for managing package dependencies in a more granular or shared manner across multiple projects, for example:
+```toml
+manifest = "../Manifest.toml"
+```
+
+This specified `Manifest.toml` does not need to be a subset of the project's own manifest. 
+The project inherits any version constraints or package specifications from the referenced manifest and is integrated into the dependency resolution process.
+
```
mortenpi, line 89: Needs a compat note I think. I hope this is accurate:

```suggestion

!!! compat "Julia 1.11"

    This feature requires at least Julia 1.11. On earlier Julia version, the `manifest` entry in the `Project.toml` file gets ignored.
```

docs/src/toml-files.md
```diff
@@ -77,6 +77,17 @@ Note that Pkg.jl deviates from the SemVer specification when it comes to version
...
 
 
+### The `manifest` field
+
+`manifest` specifies an external manifest file (`Manifest.toml`) for the project. This is particularly useful for managing package dependencies in a more granular or shared manner across multiple projects, for example:
+```toml
+manifest = "../Manifest.toml"
+```
+
+This specified `Manifest.toml` does not need to be a subset of the project's own manifest.
```
lassepe, line 87: I think the wording "subset" here is unclear.

sjkelly: Yes, I'm not sure this is correct behavior. I'm somewhat exercising Cunningham's law here to get critique on what is expected.

IanButterworth: It's not clear to me what this sentence actually means. Can we just drop it?

KristofferC:
I don't think this should be added, the `manifest` field has many pitfalls if you alias many projects to the same manifest. The workspace feature should be the thing to use for "shared manner across multiple projects".

IanButterworth:
Ah that's what the subset thing means. Yeah I agree

PR #3695, opened 2023-11-10
Fix indentation in precompile.jl

mkitti:
This deletes the two commented `end` statements and then reindents the `withenv` block.

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -47,41 +47,41 @@ let
         write(
             "registries/Registry/Registry.toml",
             """
-name = "Registry"
-uuid = "37c07fec-e54c-4851-934c-2e3885e4053e"
-repo = "https://github.com/JuliaRegistries/Registry.git"
-[packages]
-$uuid = { name = "TestPkg", path = "T/TestPkg" }
-""",
+            name = "Registry"
+            uuid = "37c07fec-e54c-4851-934c-2e3885e4053e"
+            repo = "https://github.com/JuliaRegistries/Registry.git"
+            [packages]
+            $uuid = { name = "TestPkg", path = "T/TestPkg" }
+            """,
         )
         write(
             "registries/Registry/T/TestPkg/Compat.toml",
             """
-["0"]
-julia = "1"
-""",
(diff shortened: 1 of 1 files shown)


IanButterworth:
Might as well also run JuliaFormatter on the file. Note the outer let block isn't indented.

mkitti:
I ran JuliaFormatter.jl on precompile.jl with default settings.

src/precompile.jl
```diff
@@ -4,141 +4,195 @@ using Downloads
...
+        mkpath("registries/Registry/T/TestPkg")
+        write(
+            "registries/Registry/Registry.toml",
+            """
+name = "Registry"
+uuid = "37c07fec-e54c-4851-934c-2e3885e4053e"
+repo = "https://github.com/JuliaRegistries/Registry.git"
+[packages]
+$uuid = { name = "TestPkg", path = "T/TestPkg" }
+""",
```
IanButterworth, line 55: These should be indented

src/precompile.jl
```diff
@@ -4,141 +4,195 @@ using Downloads
...
+            Base.precompile(Tuple{typeof(Pkg.REPLMode.promptf)})
+            Base.precompile(Tuple{typeof(Pkg.REPLMode.repl_init),REPL.LineEditREPL})
+            Base.precompile(Tuple{typeof(Pkg.API.status)})
+            Base.precompile(
+                Tuple{
+                    typeof(Pkg.Types.read_project_compat),
+                    Base.Dict{String,Any},
+                    Pkg.Types.Project,
+                },
+            )
```
IanButterworth, line 159: Lets leave these as one-liners.

mkitti: Would you prefer that I do these manually or that I figure out a house style for Pkg.jl?

IanButterworth: Just manually. I don't think we should go beyond this file

mkitti:
Does this work or you do you want me to revert some of the altered lines?

IanButterworth:
Thanks

PR #3696, opened 2023-11-10
remove recompilation dependency check in Pkg.precompile

KristofferC:
This check forced packages to recompile if triggers for extensions of their dependencies were added.

This could be reproduced the following way:

- Add `ColorSchemes`
- Precompile
- Add `SpecialFunctions`
- Notice that `ColorSchemes` now required precompilation again because the extension of SpecialFunctions in ColorVectorSpace is now "active".

The last step doesn't make sense because the extension SpecialFunctions doesn't influence ColorSchemes in any way.

This check should in theory just be removable since `Base.isprecompiled` should handle checking that all dependencies are valid for the package being precompiled.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1079,11 +1079,11 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
                     strict::Bool=false, warn_loaded = true, already_instantiated = false, timing::Bool = false,
                     _from_loading::Bool=false, kwargs...)
     Context!(ctx; kwargs...)
-    if !already_instantiated 
+    if !already_instantiated
         instantiate(ctx; allow_autoprecomp=false, kwargs...)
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -1498,9 +1498,8 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
 
                 circular = pkg in circular_deps
                 # skip stale checking and force compilation if any dep was recompiled in this session
-                any_dep_recompiled = any(map(dep->was_recompiled[dep], deps))
```
IanButterworth, line 1501: Delete comment too

IanButterworth (approved):
Just to explain this. Our ability to check the cache state was limited in the past and I forgot to remove this failsafe when `Base.isprecompiled` was added.

Issue #3687, opened 2023-11-05
`Pkg.build`  is seeing packages from startup.jl


MasonProtter:
This is occuring for me both on `master` and on `v1.10-rc1`. Adding packages that have build scripts functions fine, but if I explicitly call `Pkg.build`, then packages (such as LinearAlgebra.jl) which are loaded in my startup.jl appear to contaminate the environment:

```julia
> julia-1.10 -q --startup=no                                                                  
julia> using Pkg; Pkg.build("ModernGL")
    Building ModernGL → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/b76ea40b5c0f45790ae09492712dd326208c28b2/build.log`
Precompiling project...
  2 dependencies successfully precompiled in 27 seconds. 405 already precompiled.
```
vs
```julia
>  julia-1.10 -q                                                                                                 
julia> using Pkg; Pkg.build("ModernGL")
    Building ModernGL → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/b76ea40b5c0f45790ae09492712dd326208c28b2/build.log`
ERROR: Error building `ModernGL`: 
ERROR: LoadError: ArgumentError: Package LinearAlgebra not found in current path.
- Run `import Pkg; Pkg.add("LinearAlgebra")` to install the LinearAlgebra package.
Stacktrace:
 [1] macro expansion
   @ Base ./loading.jl:1766 [inlined]
 [2] macro expansion
   @ Base ./lock.jl:267 [inlined]
 [3] __require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1747
 [4] #invoke_in_world#3
   @ Base ./essentials.jl:921 [inlined]
 [5] invoke_in_world
   @ Base ./essentials.jl:918 [inlined]
 [6] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1740
in expression starting at /home/mason/.julia/config/startup.jl:1
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/julia-1.10/usr/share/julia/stdlib/v1.10/Pkg/src/Types.jl:70
  [2] (::Pkg.Operations.var"#67#74"{Bool, Pkg.Types.Context, String, Pkg.Types.PackageSpec, String})()
[...]

PR #3699, opened 2023-11-16
Use `startup=no` during `build` unless julia session explicitly set `startup=yes`

MasonProtter:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3687 (identified by @KristofferC on Slack)

Before:
```julia
(@v1.10) pkg> build ModernGL
    Building ModernGL → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/b76ea40b5c0f45790ae09492712dd326208c28b2/build.log`
ERROR: Error building `ModernGL`: 
ERROR: LoadError: ArgumentError: Package LinearAlgebra not found in current path.
- Run `import Pkg; Pkg.add("LinearAlgebra")` to install the LinearAlgebra package.
Stacktrace:
 [1] macro expansion
   @ Base ./loading.jl:1766 [inlined]
 [2] macro expansion
   @ Base ./lock.jl:267 [inlined]
 [3] __require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1747
 [4] #invoke_in_world#3
   @ Base ./essentials.jl:921 [inlined]
 [5] invoke_in_world
   @ Base ./essentials.jl:918 [inlined]
 [6] require(into::Module, mod::Symbol)
   @ Base ./loading.jl:1740
in expression starting at /home/mason/.julia/config/startup.jl:1
```
Attempted fix:
```julia
julia> @eval Pkg.Operations begin
           function gen_build_code(build_file::String; inherit_project::Bool = false)
           code = """
               cd($(repr(dirname(build_file))))
               include($(repr(build_file)))
               """
           return ```
               $(Base.julia_cmd()) -O0 --color=no --history-file=no --startup=no
               $(inherit_project ? `--project=$(Base.active_project())` : ``)
               --eval $code
              ```
           end
       end

(@v1.10) pkg> build ModernGL
    Building ModernGL → `~/.julia/scratchspaces/44cfe95a-1eb2-52ea-b672-e2afdf69b78f/b76ea40b5c0f45790ae09492712dd326208c28b2/build.log`
Precompiling project...
  2 dependencies successfully precompiled in 28 seconds. 461 already precompiled.
```

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1028,7 +1028,7 @@ function gen_build_code(build_file::String; inherit_project::Bool = false)
         include($(repr(build_file)))
         """
     return ```
-        $(Base.julia_cmd()) -O0 --color=no --history-file=no
+        $(Base.julia_cmd()) -O0 --color=no --history-file=no --startup=no
         $(inherit_project ? `--project=$(Base.active_project())` : ``)
         --eval $code
         ```
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1027,9 +1027,12 @@ function gen_build_code(build_file::String; inherit_project::Bool = false)
         cd($(repr(dirname(build_file))))
         include($(repr(build_file)))
         """
+    # This will make it so that running Pkg.build runs the build in a session with --startup=no
+    # *unless* the parent julia session is started with --startup=yes explicitly.
+    start_flag = Base.JLOptions().startupfile == 1 ? "yes" : "no"
     return ```
         $(Base.julia_cmd()) -O0 --color=no --history-file=no
-        --startup-file=$(Base.JLOptions().startupfile == 1 ? "yes" : "no")
+        --startup-file=$(startup_flag)
         $(inherit_project ? `--project=$(Base.active_project())` : ``)
         --eval $code
         ```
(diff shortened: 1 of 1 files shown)


KristofferC:
For the tests this is used

https://github.com/JuliaLang/Pkg.jl/blob/fd00e3f326700dd6d800382c56fadb78ee4c035a/src/Operations.jl#L1690

so that if you run with `--startup-file=yes` it is also used in testing. I am not sure if that is a good idea or not but perhaps worth it to be consistent?

MasonProtter:
Hm, I'd think that using the `startup=yes` would be universally problematic, but I don't really have enough experience to know. I'll defer to your judgement.

KristofferC:
No one has complained about the test behavior (but no one has said it was useful either...). I just changed to that, don't think it matters in practice. 🤷

MasonProtter:
I could imagine that being able to force it on could be useful if you need to include some localized pirated fixes or something in your `startup.jl`, so I guess it makes sense.

IanButterworth:
It'd be good to update the PR title and add a comment to explain that it only sets to yes if explicitly on. It's not clear from the code alone

PallHaraldsson:
I never actually heard of anyone using --startup-file=yes because why would you, it's the default in all cases!

I don't see it better here special-casing for it (nor for tests, though haven't thought about either carefully).

I think we shouldn't do the check for this for neither, since it doesn't seems too useful, and I would want --startup-file=no the default for non-interactive at least... That was not well received, but if it were to happen, then people might start using --startup-file=yes in some cases.

PR #3701, opened 2023-11-20
Revise the API of is_manifest_current.

GunnarFarneback:
The API function `is_manifest_current` was not working as intended and was mostly useless for its intended purpose of verifying in tests that the manifest and project files of a package were synchronized. See 
https://github.com/JuliaLang/Pkg.jl/pull/2815#issuecomment-1819237619 for details.

This PR removes the zero-argument method of `is_manifest_current` and adds a new `is_manifest_current(::AbstractString)` method that can be used in tests in, e.g., the form
```
using Pkg, Test, Package
@test Pkg.is_manifest_current(pkgdir(Package))
```

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -455,7 +455,15 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     return
 end
 
-is_manifest_current(ctx::Context = Context()) = Operations.is_manifest_current(ctx.env)
+is_manifest_current(ctx::Context) = Operations.is_manifest_current(ctx.env)
+function is_manifest_current(path::AbstractString)
+    project_file = projectfile_path(path, strict = true)
+    if project_file === nothing
+        pkgerror("could not find project file at `$path`")
+    end
+    env = EnvCache(project_file)
+    return Operations.is_manifest_current(env)
+end
 
 const UsageDict = Dict{String,DateTime}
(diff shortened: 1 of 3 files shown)


IanButterworth (approved):
Seems good to me. There don't seem to be [any public users of the function](https://juliahub.com/ui/Search?q=is_manifest_current&type=code), so this shouldn't be too bad of a breaking bugfix, I think.

src/API.jl
```diff
@@ -455,7 +455,15 @@ function test(ctx::Context, pkgs::Vector{PackageSpec};
     return
 end
 
-is_manifest_current(ctx::Context = Context()) = Operations.is_manifest_current(ctx.env)
+is_manifest_current(ctx::Context) = Operations.is_manifest_current(ctx.env)
```
devmotion, line 458: It broke support of Julia 1.11 in Quarto (which is fixed by now) 😄 Generally, shouldn't documented functionality in Pkg.API be stable and not be broken in semver-nonbreaking releases? Maybe this should be changed to
```suggestion
is_manifest_current(ctx::Context = Context()) = Operations.is_manifest_current(ctx.env)
```
again?

GunnarFarneback: I don't have a strong opinion about this but if you managed to find a working use case for it I guess it has some value. But I'd suggest opening a new PR rather than adding a change suggestion to a merged one.

PR #3709, opened 2023-11-27
optionally add deved package to vscode workspace

IanButterworth:
This might be too magical? But it's quite nice.


IanButterworth:
@pfitzseb Just wondering what you think of this

KristofferC:
Feels a bit fancy to me tbh.

PR #3710, opened 2023-11-27
allow authors to be a TOML array

### LINKED ISSUE #3705: Richer author information
I would like to propose that we allow for richer `authors` information in the Project.toml. Instead of a string of `"name <email>"`, I propose that each author be a TOML table. I would propose the following keys be supported initially:
- `name`
- `email`
- `github` for github username
- [`orcid`](https://orcid.org/)

Of these, only `name` would be required.

Some others we may want to consider:
- greater name granularity (e.g. `givenname`/`familyname`?), to make it easier to parse different reference styles.
- `affiliation`, similar to a paper?

Benefits:
- name-email pairs are not great as identifiers: emails can become stale (people move institutions), some people dislike having their emails public
- it would make it easier to parse and process for other uses, e.g. zenodo submissions or including a list of authors + github links in the documentation.

This would mean that authors could be represented in TOML as an [array of tables](https://toml.io/en/v1.0.0#array-of-tables) (though personally I find that syntax a bit clunky).

simonbyrne:
Fixes #3705.

### DIFF
--- a/docs/src/toml-files.md
+++ b/docs/src/toml-files.md
@@ -34,7 +34,7 @@ For example:
 authors = [
   "Some One <someone@email.com>",
   "Foo Bar <foo@bar.com>",
-  {name = "Baz Qux", email = "bazqux@example.com", orcid = "0000-0001-8048-6810"},
+  {name = "Baz Qux", email = "bazqux@example.com", orcid = "0000-0000-0000-0000"},
 ]
 ```
 
--- a/docs/src/toml-files.md
+++ b/docs/src/toml-files.md
@@ -23,21 +23,20 @@ are described below.
 ### The `authors` field
 
 For a package, the optional `authors` field is a TOML array describing the package authors. 
-Entries in the array can either be a string in the form `NAME <EMAIL>`, or arrays with the following keys:
- - `name` (required): the name of the author
- - `email`: email address
- - `orcid`: 16-digit [ORCID](https://orcid.org/) identifier as a string
- - `github`: [GitHub](https://github.com/) username
+Entries in the array can either be a string in the form of `NAME` or `NAME <EMAIL>`, or arrays with the following keys:
+ - `name` (string, required): the name of the author
+ - `email` (string, optional): email address 
+ - `orcid` (string, optional): 16-digit [ORCID](https://orcid.org/) identifier
+ - `url` (string, optional): a URL associated with the author (e.g. personal website, GitHub page)
 
 For example:
 ```toml
 authors = [
   "Some One <someone@email.com>",
   "Foo Bar <foo@bar.com>",
-  {name = "Baz Qux", email = "bazqux@example.com", orcid = "0000-0000-0000-0000"},
+  {name = "Baz Qux", email = "bazqux@example.com", orcid = "0000-0000-0000-0000", url = "https://github.com/bazqux"},
 ]
 ```
-
 
 ### The `name` field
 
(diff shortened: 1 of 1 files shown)


fredrikekre:
Are there any consumer of this information?

simonbyrne:
https://github.com/KlausC/PkgVersion.jl is the only package I am aware of (and that just returns the value)

docs/src/toml-files.md
```diff
@@ -22,11 +22,20 @@ are described below.
...
 ### The `authors` field
 
-For a package, the optional `authors` field is a list of strings describing the
-package authors, in the form `NAME <EMAIL>`. For example:
+For a package, the optional `authors` field is a TOML array describing the package authors. 
+Entries in the array can either be a string in the form `NAME <EMAIL>`, or arrays with the following keys:
+ - `name` (required): the name of the author
+ - `email`: email address
+ - `orcid`: 16-digit [ORCID](https://orcid.org/) identifier as a string
+ - `github`: [GitHub](https://github.com/) username
```
StefanKarpinski, line 30: Baking GitHub in here feels weird. How about link and it can be any site. If it happens to be a github  user url then you can infer that its their github handle.

simonbyrne: That's a fair point, which I somewhat agree with. The main reason to include it is that GitHub is the primary platform/network most Julia package developers use.

The downside of a generic URL is that there can then only be 1: I guess we can have "urls" be a list?

StefanKarpinski: Don't you think that each person can pick one link that's identifying for them?

simonbyrne: True.

simonbyrne:
For reference, GitHub suggests people use the CITATION.cff format, which also provides a richer schema:
[`authors`](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#authors) can be either a [`person`](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsperson) or [`entity`](https://github.com/citation-file-format/citation-file-format/blob/main/schema-guide.md#definitionsentity)

The main differences are:
- `orcid` is the full URL, not just the 16-digit id (which probably makes sense, as it is easier to copy-paste)
- `website` instead of `url` (I don't have strong opinions)
- `name` is only for entities, persons instead use `family-names`,`given-names`, etc. This is the only one which is a bit sticky, but we could probably just allow either (or both)?

Thoughts?

simonbyrne:
One minor advantage of using `given-names`/`family-names` is that it matches the Zenodo fields.

simonbyrne:
Bump. Any thoughts?

StefanKarpinski:
Sure 🤷🏻‍♂️. We don't actually use this data anywhere, do we?

simonbyrne:
Not that I know of.

docs/src/toml-files.md
```diff
@@ -22,13 +22,38 @@ are described below.
...
+given-names = "Foo"
+family-names = "Bar"
+email = "foo@bar.com"
+
+[[authors]]
+given-names = "Baz"
+family-names = "Qux"
+email = "bazqux@example.com"
+orcid = "https://orcid.org/0000-0000-0000-0000"
+website = "https://github.com/bazqux"}
```
simonbyrne, line 55:
```suggestion
website = "https://github.com/bazqux"
```

PR #3715, opened 2023-11-30
docs/src/environments.md: update Julia version

jcsahnwaldt:
1.8 -> 1.9, 1.8.2 -> 1.9.4 (current version)

### DIFF
--- a/docs/src/environments.md
+++ b/docs/src/environments.md
@@ -4,13 +4,13 @@ The following discusses Pkg's interaction with environments. For more on the rol
 
 ## Creating your own environments
 
-So far we have added packages to the default environment at `~/.julia/environments/v1.8`. It is however easy to create other, independent, projects.
+So far we have added packages to the default environment at `~/.julia/environments/v1.9`. It is however easy to create other, independent, projects.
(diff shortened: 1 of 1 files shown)


jcsahnwaldt:
I guess the current Julia version occurs in other files as well. It might be nice if it could be replaced by the build process, but that would probably be more work. Is there already a build step that replaces variables in MarkDown or other text files?

IanButterworth:
The julia version on master is actually 1.11
There's unlikely to be another 1.9 release, but we could backport another PR onto the `release-1.10` branch to update 1.10 though

jcsahnwaldt:
I could update the PR and set the version to 1.11 instead. But a better long-term solution would be to replace the hard-coded version by a variable that gets replaced during the build process.

I had a quick look at how the docs are built. Maybe an [@eval-block](https://documenter.juliadocs.org/stable/man/syntax/#@eval-block) would work? Is the target version available as a Julia variable during the build? If not, can we add it somewhere, e.g. in [docs/make.jl](https://github.com/JuliaLang/Pkg.jl/blob/master/docs/make.jl)? Then I'd change the PR and add something like `@eval SomeModule.targetVersion()` instead of the hard-coded version strings.

PR #3720, opened 2023-12-06
extensions: fixup entire manifest

### LINKED ISSUE #3719: bad/incorrect/missing handling of stdlibs in Manifest
While investigating https://github.com/JuliaStats/Statistics.jl/pull/134, I noticed that calling `add Statistics` fails to update the global Manifest correctly (does not add the extensions section or update the version) and if then delete the global Manifest and run `resolve` (which updates the Manifest correctly) then `rm Statistics` it again fails to update the global Manifest correctly and becomes wrong again (keeping the information for the non-stdlib copy) until I again manually delete the global Manifest

vtjnash:
Not just explicitly installed packages.

Do I understand this logic? No.

Do I understand if this fixes anything? No.

Fixes #3719


vtjnash:
Bump?

PR #3711, opened 2023-11-28
Precompile REPLMode.repl_init, add force_compile

mkitti:
Adds some precompilation for Pkg.REPLMode.repl_init

Removes direct reference to Base.active_repl

Add Base.Experimental.@force_compile

PR #3724, opened 2023-12-09
Split REPLMode into an package extension called REPLExt

mkitti:
This PR separates Pkg.REPLMode into an package extension called PkgREPLMode.

* REPL.jl is now a weak dependency of Pkg.jl
* Some code duplication was necessary in DocView.jl. Perhaps this could be a separate package used by Pkg.jl and REPL.jl.
* This pull request includes precompilation code for the REPL Mode as in https://github.com/JuliaLang/Pkg.jl/pull/3711


ext/PkgREPLMode/PkgREPLMode.jl
```diff
@@ -1,16 +1,22 @@
 # This file is a part of Julia. License is MIT: https://julialang.org/license
 
-module REPLMode
+module PkgREPLMode
```
IanButterworth, line 3:
```suggestion
module REPLExt
```
I really don't like the trend to put the parent package name first and this should bear the Ext name

IanButterworth:
Also, nit but it's a "package extension" not an "extension package"

ext/PkgREPLMode/PkgREPLMode.jl
mkitti: How about `REPLModeExt`?

mkitti:
This requires coordination with https://github.com/JuliaLang/julia/pull/52467

IanButterworth:
Would this break usage like this?

```
julia -e 'import Pkg; pkg"add Foo"'
```

mkitti:
> Would this break usage like this?\n\njulia -e 'import Pkg; pkg\"add Foo\"'

Yes. Although I think it has been quite clear that this is not the intended use.

```julia
julia -e 'using Pkg; pkg"status"'
┌ Warning: The Pkg REPL mode is intended for interactive use only, and should not be used from scripts. It is recommended to use the functional API instead.
```

To make it work, they would need to load it as follows.

```
julia -e 'using Pkg, REPL; pkg"status"'
```

mkitti:
Since it is a macro, could we just add `@eval using REPL` to the top of the macro?

IanButterworth:
I think that breakage is very reasonable tbh just might come up in pkgeval

fredrikekre:
Does anything in that code path even use REPL though? Why hinge this functionality on whether REPL is loaded?

mkitti:
> Does anything in that code path even use REPL though? Why hinge this functionality on whether REPL is loaded?

Yes, it is very dependent on the REPL.

From master,
https://github.com/JuliaLang/Pkg.jl/blob/f7f222f28e929dda69e4f5b1d660fc51a133fe09/src/REPLMode/REPLMode.jl#L459-L470

This branch,
https://github.com/JuliaLang/Pkg.jl/blob/4a784ae943a3dda391479a1dc439ee1616ae28a8/ext/REPLExt/REPLExt.jl#L458-L492

fredrikekre:
Thats surprising, looks like it is only for the help view though, perhaps it could be refactored.

On another note, typically the entry point to the extension code is some type that you pass to the function, but in this case it will just magically work if REPL happen to be loaded, right?

mkitti:
The `pkg""` commands go through the Pkg REPL mode interface and are thus meant to take advantage of the "interactive" code. I personally think the the macro should be `@pkg_cmd` rather than `@pkg_str` but there might be historical reasons for this.

> On another note, typically the entry point to the extension code is some type that you pass to the function, but in this case it will just magically work if REPL happen to be loaded, right?

Yes. There are actually hooks on both sides. I just did some refactoring in [03a7da0](https://github.com/JuliaLang/Pkg.jl/pull/3724/commits/03a7da054bfada9517804573fd183dadc69ec3cf) to make the initialization a bit more robust.

Also there is a hook on the REPL side that needs some modification for this pull request to work.
https://github.com/JuliaLang/julia/pull/52467

I kept the name PkgREPLMode in REPL because referring to `REPLExt` inside REPL seems confusing and silly.
https://github.com/JuliaLang/julia/blob/f115ef7d66aa8261d199e642281bf6f4e70cda0b/stdlib/REPL/src/REPL.jl#L1138-L1144

mkitti:
Maintaining this branch as other pull requests get merged is going to be a challenge since so many files are affected.

src/Types.jl
```diff
@@ -6,12 +6,10 @@ using UUIDs
 using Random
 using Dates
 import LibGit2
-import REPL
 import Base.string
-using REPL.TerminalMenus
```
KristofferC, line 11: This probably breaks the `RadioMenu` stuff in this file. However, I think that is more or less useless and should be removed.

KristofferC:
Regarding the `pkg""` stuff, it should be possible I think with some tweaks to make that work without a REPL. I don't think there is anything that is really needs from the REPL module. Some packages (unfortunately) use the `pkg""` way to inform users how to install their package, e.g. https://github.com/bat/BAT.jl/blob/v2.0.5/docs/src/installation.md#installing-batjl-and-related-julia-packages.

Avoiding having to load the REPL for Pkg will be a very nice thing.

src/Types.jl
mkitti: Is this the only section that is affected?

https://github.com/JuliaLang/Pkg.jl/blob/989b507d26b4f6fd7f28043a772c22e3d5bcfa4b/src/Types.jl#L1076C18-L1084

We could probably create some kind of hook to move that into the extension if we still need it. Perhaps the more pressing thing is developing a test to cover that part of the code.

mkitti: Addressed in [d9138bd](https://github.com/JuliaLang/Pkg.jl/pull/3724/commits/d9138bd07abf110eb61bd807339c84208a30d12a)

KristofferC:
Subsumed by https://github.com/JuliaLang/Pkg.jl/pull/3777

PR #3725, opened 2023-12-10
Make tests almost entirely quiet

IanButterworth:
Best I can do is only this leaking.
```
Username for 'https://github.com/': Username for 'https://github.com/':
```

To do better than that we would need to reduce test coverage, or fix git cli https auth on CI, which seems difficult to figure out.

Otherwise everything is now piped into the Pkg.log which is uploaded as matrix-specific artifacts here and similar on buildkite.

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -2714,17 +2714,14 @@ for v in (nothing, "true")
                         @test_throws SystemError open(pathof(eval(Symbol(TEST_PKG.name))), "w") do io end  # check read-only
                         Pkg.rm(TEST_PKG.name)
                     end
-                    if Base.get_bool_env("JULIA_PKG_USE_CLI_GIT", false) == false
-                        # TODO: fix. on GH windows runners cli git will prompt for credentials here
-                        # on other runners it noisily prompts but continues
-                        println("here $v")
-                        @testset "via url" begin
-                            # git cli can be noisy on CI where user auth isn't set up, so ignore stderr
-                            Pkg.add(url="https://github.com/JuliaLang/Example.jl", use_git_for_all_downloads=true)
-                            @test haskey(Pkg.dependencies(), TEST_PKG.uuid)
-                            Pkg.rm(TEST_PKG.name)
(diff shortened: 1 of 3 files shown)


test/new.jl
```diff
@@ -2706,26 +2706,22 @@ for v in (nothing, "true")
...
-                        Pkg.add(TEST_PKG.name; use_git_for_all_downloads=true)
+                @testset "via name" begin
+                    Pkg.add(TEST_PKG.name; use_git_for_all_downloads=true)
+                    @test haskey(Pkg.dependencies(), TEST_PKG.uuid)
+                    @eval import $(Symbol(TEST_PKG.name))
+                    @test_throws SystemError open(pathof(eval(Symbol(TEST_PKG.name))), "w") do io end  # check read-only
+                    Pkg.rm(TEST_PKG.name)
+                end
+                if (Base.get_bool_env("JULIA_PKG_USE_CLI_GIT", false) == false) && !Sys.iswindows()
+                    # TODO: fix. On GH windows runners cli git will prompt for credentials and hang.
```
DilumAluthge, line 2717: Should we do a `@test_skip` or something? Or a `@test ... skip=Sys.iswindows()`?

DilumAluthge (approved):

DilumAluthge:
Any idea why macOS? CI is failing

IanButterworth:
One solid failure on MacOS is https://github.com/JuliaLang/julia/issues/52458

The other is flaky git for unknown reasons. Maybe a libgit2 issue.. I don't think anyone has reported it in the wild

test/new.jl
IanButterworth: Can we skip testsets like that? That's what would be needed here

IanButterworth: bump

DilumAluthge: Oh, hmmm, skip the whole testset? I don't think that's possible. So we're probably stuck with the existing approach.

PR #3731, opened 2023-12-14
fix preferential Manifest.toml naming

IanButterworth:
Reverts https://github.com/JuliaLang/Pkg.jl/pull/3000
and adapts for the latest https://github.com/JuliaLang/julia/pull/43845

Because the lengths of the `Base.project_names` and `Base.manifest_names` tuples have diverged, we can no longer rely on idx lookups, so this just does a bit more hardcoding.


IanButterworth:
Backporting because of https://github.com/JuliaLang/julia/pull/56600

PR #3734, opened 2023-12-16
Use realpaths for temp dirs during tests. Fix SparseArrays `why` breakage

IanButterworth:
- Avoids https://github.com/JuliaLang/julia/issues/52458
- Fixes new failure due to SparseArrays becoming a weak dep of Statistics

### DIFF
--- a/test/new.jl
+++ b/test/new.jl
@@ -1556,7 +1556,6 @@ end
         @test str ==
         """  StaticArrays → LinearAlgebra
           StaticArrays → Statistics → LinearAlgebra
-          StaticArrays → Statistics → SparseArrays → LinearAlgebra
         """
     end
 end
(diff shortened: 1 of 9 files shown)


IanButterworth:
Something's going wrong in the path-specific coverage setup. I need to see what the code info source paths for the files are to compare to the coverage path, but I'm not sure how to get that

test/new.jl
```diff
@@ -1556,7 +1556,6 @@ end
         @test str ==
         """  StaticArrays → LinearAlgebra
           StaticArrays → Statistics → LinearAlgebra
-          StaticArrays → Statistics → SparseArrays → LinearAlgebra
```
KristofferC, line 1559: :tada:

Issue #3702, opened 2023-11-21
Pkg.BinaryPlatforms invalidates Base.BinaryPlatforms


mkitti:
```julia
using SnoopCompileCore
invalidations = @snoopr begin
    using Pkg
    Pkg.activate()
end

using SnoopCompile
trees = invalidation_trees(invalidations)
```

```julia
julia> trees
3-element Vector{SnoopCompile.MethodInvalidations}:
 inserting ==(a::Union{Pkg.BinaryPlatforms.FreeBSD, Pkg.BinaryPlatforms.Linux, Pkg.BinaryPlatforms.MacOS, Pkg.BinaryPlatforms.Windows}, b::Base.BinaryPlatforms.AbstractPlatform) @ Pkg.BinaryPlatforms ~/src/julia/usr/share/julia/stdlib/v1.11/Pkg/src/BinaryPlatforms_compat.jl:89 invalidated:
   backedges: 1: superseding ==(x, y) @ Base Base.jl:177 with MethodInstance for ==(::Base.BinaryPlatforms.AbstractPlatform, ::Base.BinaryPlatforms.AbstractPlatform) (4 children)

 inserting tags(::Pkg.BinaryPlatforms.UnknownPlatform) @ Pkg.BinaryPlatforms ~/src/julia/usr/share/julia/stdlib/v1.11/Pkg/src/BinaryPlatforms_compat.jl:17 invalidated:
   mt_backedges:  1: signature Tuple{typeof(Base.BinaryPlatforms.tags), Base.BinaryPlatforms.AbstractPlatform} triggered MethodInstance for Base.BinaryPlatforms.triplet(::Base.BinaryPlatforms.AbstractPlatform) (0 children)
                  2: signature Tuple{typeof(Base.BinaryPlatforms.tags), Base.BinaryPlatforms.AbstractPlatform} triggered MethodInstance for Base.BinaryPlatforms.os_version(::Base.BinaryPlatforms.AbstractPlatform) (1 children)
                  3: signature Tuple{typeof(Base.BinaryPlatforms.tags), Base.BinaryPlatforms.AbstractPlatform} triggered MethodInstance for Base.BinaryPlatforms.libgfortran_version(::Base.BinaryPlatforms.AbstractPlatform) (1 children)
[...]

PR #3735, opened 2023-12-17
Replace UnstableIO with IOContext{IO}

### LINKED ISSUE #3722: missing API method closewrite for Pkg.UnstableIO
The UnstableIO type is missing the `closewrite` method which causes it to throw this error (which gets usually swallowed by precompile, but here I made it visible while checking for API mistakes)
```
    nested task error: MethodError: no method matching closewrite(::Pkg.UnstableIO)
    
    Closest candidates are:
      closewrite(!Matched::Base.BufferStream)
       @ Base stream.jl:1487
      closewrite(!Matched::Base.LibuvStream)
       @ Base stream.jl:437
      closewrite(!Matched::Base.GenericIOBuffer)
       @ Base iobuffer.jl:350
      ...
    
    Stacktrace:
     [1] (::Base.var"#848#849"{Pkg.UnstableIO, Bool, Base.PipeEndpoint, Pkg.UnstableIO, Base.PipeEndpoint})()
       @ Base ./process.jl:318
in expression starting at /home/vtjnash/julia/usr/share/julia/stdlib/v1.11/Pkg/src/precompile.jl:6
in expression starting at /home/vtjnash/julia/usr/share/julia/stdlib/v1.11/Pkg/src/Pkg.jl:3
in expression starting at stdin:3
```

mkitti:
While trying to send UnstableIO into Julia base, two suggestions arose:
1. Use `Base.inferencebarrier` to prevent specialization: 
   https://github.com/JuliaLang/julia/pull/52249#discussion_r1401199265
2. Use `IOContext{IO}` rather than introducing a new type:
   https://github.com/JuliaLang/julia/pull/52249#discussion_r1425332525

Fix #3722 (`IOContext` has `closewrite`)
Addresses part of #3702 by removing `UnstableIO` and its associated invalidations.


src/Pkg.jl
```diff
@@ -41,14 +41,18 @@ const RESPECT_SYSIMAGE_VERSIONS = Ref(true)
...
-struct UnstableIO <: IO
-    io::IO
+# See discussion in https://github.com/JuliaLang/julia/pull/52249
+const UnstableIO = IOContext{IO}
+function unstableio(@nospecialize(io::IO))
+    # Needed to prevent specialization https://github.com/JuliaLang/julia/pull/52249#discussion_r1401199265
+    _io = Base.inferencebarrier(io)
+    IOContext{IO}(
+        _io,
+        get(_io,:color,false) ? Base.ImmutableDict{Symbol,Any}(:color, true) : Base.ImmutableDict{Symbol,Any}()
```
KristofferC, line 51: Why is this needed?

mkitti: Do you have another suggestion as how one might construct a `IOContext{IO}`?

```julia
julia> IOContext{IO}(stdout)
ERROR: MethodError: no method matching IOContext{IO}(::Base.TTY)

Closest candidates are:
  IOContext{IO_t}(::IO_t, ::Base.ImmutableDict{Symbol, Any}) where IO_t<:IO
   @ Base show.jl:291
```

src/Pkg.jl
```diff
@@ -41,14 +41,18 @@ const RESPECT_SYSIMAGE_VERSIONS = Ref(true)
 # For globally overriding in e.g. tests
 const DEFAULT_IO = Ref{Union{IO,Nothing}}(nothing)
 
-struct UnstableIO <: IO
-    io::IO
+# See discussion in https://github.com/JuliaLang/julia/pull/52249
+const UnstableIO = IOContext{IO}
```
KristofferC, line 45: This can be removed I think? Or is it used somewhere else?

mkitti: There are a few other references to `UnstableIO` in Pkg.jl. I could get rid of the alias as its uses. Is that what you mean?

mkitti: I removed references to `UnstableIO` in 9d57162

src/Pkg.jl
mkitti: An alternative is that I use `Base.unwrapcontext` instead of copying the code. I'm not sure if that complicates the inference barrier or not.

mkitti: This does cause compilation of `get` for the `IO` though.

KristofferC:
Looks like this has to be rebased to update for another use of `UnstableIO` (looking at the CI):

```
┌ Pkg
│  WARNING: could not import Pkg.UnstableIO into Operations
└  
     Testing Running tests...
ERROR: LoadError: UndefVarError: `UnstableIO` not defined in `Pkg.Operations`
Stacktrace:
  [1] subprocess_handler(cmd::Cmd, io::IOContext{IO}, error_msg::String)
    @ Pkg.Operations ~/work/Pkg.jl/Pkg.jl/src/Operations.jl:2223
  [2] (::Pkg.Operations.var"#138#143"{Pkg.Types.Context})()
    @ Pkg.Operations ~/work/Pkg.jl/Pkg.jl/src/Operations.jl:2129
```

KristofferC:
👍

Issue #3737, opened 2023-12-19
1.11: Handling of --compiled-modules=existing is broken


maleadt:
After https://github.com/JuliaLang/julia/pull/50586, `Base.JLOptions().use_compiled_modules` can be 2 for `--compiled-modules=existing`:

```
❯ ../julia --compiled-modules=existing --project -e 'using Pkg; Pkg.test()'
     Testing Running tests...
ERROR: InexactError: Bool(2)
Stacktrace:
  [1] Bool
    @ ./float.jl:233 [inlined]
  [2] gen_subprocess_cmd(code::String, source_path::String; coverage::Bool, julia_args::Cmd)
    @ Pkg.Operations ~/Julia/src/julia/build/dev/usr/share/julia/stdlib/v1.11/Pkg/src/Operations.jl:1685
  [3] gen_subprocess_cmd
    @ ~/Julia/src/julia/build/dev/usr/share/julia/stdlib/v1.11/Pkg/src/Operations.jl:1677 [inlined]
```

https://github.com/JuliaLang/julia/pull/52573 adds something similar for `--pkgimages`, but that doesn't seem to get forwarded by Pkg. I guess it should?

PR #3738, opened 2023-12-19
rm old manual handling of `--compiled-modules`

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3737

`julia_cmd()` now handles this
https://github.com/JuliaLang/julia/blob/abb7e3074b75c85b409320ccc7d9f5e9e9adb4a3/base/util.jl#L208-L209

@maleadt


maleadt:
Thanks!

Issue #3739, opened 2023-12-20
Precompilation progress output ignores --color


maleadt:
Cased by:

https://github.com/JuliaLang/Pkg.jl/blob/03de920b383c783705f2f7910065c9ef13f2a456/src/API.jl#L1431

MWE:

```
❯ ./julia -q --color=no
julia> using Pkg
julia> string(Pkg.API.color_string("  ✓ ", :green), "Example")
"\e[32m  ✓ \e[39mExample"
```

This affects PkgEval, which runs under `--color=no`. Apart from the color, maybe the output shouldn't be as verbose for non-interactive sessions?

PR #3740, opened 2023-12-20
respect --color=no in Pkg.precompile

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3739

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1110,8 +1110,9 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     end
     fancyprint = can_fancyprint(io) && !timing
 
+    hascolor = get(io, :color, false)::Bool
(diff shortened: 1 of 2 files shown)


src/API.jl
```diff
@@ -1110,6 +1110,16 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
     end
     fancyprint = can_fancyprint(io) && !timing
 
+    function color_string(cstr::String, col::Union{Int64, Symbol})
+        if get(io, :color, false)::Bool
```
IanButterworth, line 1114:
```suggestion
    hascolor = get(io, :color, false)::Bool
    function color_string(cstr::String, col::Union{Int64, Symbol})
        if hascolor
```

maleadt:
Thanks!

PR #3745, opened 2024-01-04
check symlinks work on windows on artifact hash mismatch

IanButterworth:
For if https://github.com/JuliaLang/Pkg.jl/issues/3643 isn't otherwise fixable

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -268,7 +268,7 @@ end
 if Sys.iswindows()
     # check whether symlinks can be made
     function check_symlinks()
-        PlatformEngines.copy_symlinks() && return false # for testing
+        PlatformEngines.copy_symlinks() === true && return false # for testing
         tmpdir = mktempdir()
         fpath = joinpath(tmpdir, "f")
         touch(fpath)
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -266,7 +266,6 @@ function unbind_artifact!(artifacts_toml::String, name::String;
 end
 
 if Sys.iswindows()
-    symlinks_work = true
     function check_symlinks()
         tmpdir = mktempdir()
         fpath = joinpath(tmpdir, "f")
@@ -275,7 +274,6 @@ if Sys.iswindows()
             symlink(fpath, joinpath(tmpdir, "s"))
             return true
         catch
-            global symlinks_work = false
             return false
         end
     end
@@ -331,7 +329,7 @@ function download_artifact(
         msg *= "  Expected git-tree-sha1:   $(bytes2hex(tree_hash.bytes))\n"
         msg *= "  Calculated git-tree-sha1: $(bytes2hex(calc_hash.bytes))"
         @static if Sys.iswindows()
-            if !symlinks_work || !check_symlinks()
+            if !check_symlinks()
                 msg *= """"
                 Note: Julia cannot create symlinks, which may be the reason for the hash mismatch.
                 You may need to activate Developer Mode in Windows.
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -337,8 +337,8 @@ function download_artifact(
                 Note: Julia cannot create symlinks, which may be the reason for the hash mismatch.
                 One solution is to activate Developer Mode in Windows, for instructions on how to do that, see:
                     https://learn.microsoft.com/en-us/gaming/game-bar/guide/developer-mode
-                An alternative path is to set the environment variable  `JULIA_PKG_IGNORE_HASHES=1`
-                to ignore artifact hash mismatches, however this is generally not recommended.
+                An alternative path is to temporarily set `ENV["JULIA_PKG_IGNORE_HASHES"]=true`
+                to ignore artifact hash mismatches in this session, however this is generally not recommended.
                 """
             end
         end
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -265,6 +265,22 @@ function unbind_artifact!(artifacts_toml::String, name::String;
     return
 end
 
+if Sys.iswindows()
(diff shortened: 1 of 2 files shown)


src/Artifacts.jl
```diff
@@ -265,6 +265,20 @@ function unbind_artifact!(artifacts_toml::String, name::String;
...
+    function check_symlinks()
+        tmpdir = mktempdir()
+        fpath = joinpath(tmpdir, "f")
+        touch(fpath)
+        try
+            symlink(fpath, joinpath(tmpdir, "s"))
+            return true
+        catch
+            return false
+        end
```
Octogonapus, line 278:
```suggestion
        try
            symlink(fpath, joinpath(tmpdir, "s"))
            return true
        catch
            return false
        finally
            rm(fpath)
        end
```

IanButterworth: `rm`-ing a dir soon after it's been created often fails on windows because of antivirus. I think it's better to leave this up to the atextit cleanup

Octogonapus: :scream:

nhz2:
This works on my Windows computer. If I turn off "Developer Mode" and delete the existing artifacts I get:

```julia
julia> using Pkg

julia> Pkg.add("NativeFileDialog")
   Resolving package versions...
ERROR: Unable to automatically download/install artifact 'iso_codes' from sources listed in 'C:\Users\nzimm\.julia\packages\iso_codes_jll\wcUPz\Artifacts.toml'.
Sources attempted:
- https://pkg.julialang.org/artifact/71f68a3d55d73f2e15a3969c241fae2349b1feb5
    Error: Tree Hash Mismatch!
  Expected git-tree-sha1:   71f68a3d55d73f2e15a3969c241fae2349b1feb5
  Calculated git-tree-sha1: 409d6ac4c02dae43ff4fe576b5c5820d0386fb3f"
Note: Julia cannot create symlinks, which may be the reason for the hash mismatch.
You may need to activate Developer Mode in Windows.

- https://github.com/JuliaBinaryWrappers/iso_codes_jll.jl/releases/download/iso_codes-v4.11.0+0/iso_codes.v4.11.0.any.tar.gz
    Error: Tree Hash Mismatch!
  Expected git-tree-sha1:   71f68a3d55d73f2e15a3969c241fae2349b1feb5
  Calculated git-tree-sha1: 409d6ac4c02dae43ff4fe576b5c5820d0386fb3f"
Note: Julia cannot create symlinks, which may be the reason for the hash mismatch.
You may need to activate Developer Mode in Windows.

```

When I turn "Developer Mode" back on I can add `NativeFileDialog` without any issues.

src/Artifacts.jl
staticfloat: You might want to `readlink()` the `fpath` back, just to ensure that it's actually doing what you expect.

src/Artifacts.jl
```diff
@@ -314,6 +330,16 @@ function download_artifact(
         msg  = "Tree Hash Mismatch!\n"
         msg *= "  Expected git-tree-sha1:   $(bytes2hex(tree_hash.bytes))\n"
         msg *= "  Calculated git-tree-sha1: $(bytes2hex(calc_hash.bytes))"
+        @static if Sys.iswindows()
+            if !check_symlinks()
+                msg *= """
+
+                Note: Julia cannot create symlinks, which may be the reason for the hash mismatch.
+                You may need to activate Developer Mode in Windows.
+                If you cannot you can set the env var `JULIA_PKG_IGNORE_HASHES=1` to ignore artifact hash mismatches.
```
staticfloat, line 339:
```suggestion
                Note: Julia cannot create symlinks, which may be the reason for the hash mismatch.
                One solution is to activate Developer Mode in Windows, for instructions on how to do that, see:
                    https://learn.microsoft.com/en-us/gaming/game-bar/guide/developer-mode
                An alternative path is to set the environment variable  `JULIA_PKG_IGNORE_HASHES=1`
                to ignore artifact hash mismatches, however this is generally not recommended.
```

staticfloat (dismissed):

src/Artifacts.jl
IanButterworth: SGTM but this might be a more appropriate link? https://learn.microsoft.com/en-us/windows/apps/get-started/enable-your-device-for-development

staticfloat: I don't like that my link has "gaming" in it, but the link you posted is also really confusing because it doesn't actually tell you how to enable developer mode, the only instructions it gives are for responding to a popup inside of visual studio.  There's no instructions telling you to go to "updates and security" and flip the toggle, that's why I didn't use it.

Also, there are lots of warnings on that site that say things like `If you're not creating your own applications on your PC, you don't need to enable Developer Mode.`, or `If you're using your computer for ordinary day-to-day activities (such as gaming, web browsing, email, or Office apps), you don't need to activate Developer Mode, and in fact, you shouldn't activate it.`.

IanButterworth: yeah fair

mkitti: If we're going to tell people to set an environment variable on Windows, stating  `ENV["JULIA_PKG_IGNORE_HASHES"] = 1` explicitly, might be good. Otherwise, we will need to get into the details on how to do that in either the Cmd.exe shell or PowerShell.
* Cmd.exe: `set JULIA_PKG_IGNORE_HASHES=1`
* PowerShell: `$env:JULIA_PKG_IGNORE_HASHES=1`

IanButterworth: We could also add an easier way to do it given we're changing julia in adding this message

IanButterworth: I tweaked the wording to make it clearer to users and guide them to use ENV so they're more likely to do it temporarily

PR #3346, opened 2023-01-28
make `pin/free --all` apply to all deps

IanButterworth:
Alternative to https://github.com/JuliaLang/Pkg.jl/pull/3336

This behavior here seems to make more sense as a bug fix for `--all` rather than a new mode.

PR #3746, opened 2024-01-05
Failing test for windows artifact hash mismatch

nhz2:
An attempt to reproduce issue #3643 in GitHub actions.

I am using `withenv("BINARYPROVIDER_COPYDEREF" => "true")`
to cause symlinks to be copied when the artifact is extracted.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -34,19 +34,12 @@ jobs:
         pkg-server:
           - ""
           - "pkg.julialang.org"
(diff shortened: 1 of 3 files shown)


IanButterworth:
I guess it would be interesting to see the effect of reverting #3346

IanButterworth:
Thanks. I tweaked and added the test to here https://github.com/JuliaLang/Pkg.jl/pull/3745

PR #3748, opened 2024-01-07
Add skip_symlink keyword to Pkg.GitTools.tree_hash

mkitti:
This adds a keyword option, skip_symlink, allowing the git tree hash to be calculated without considering symlinks.

xref: https://github.com/JuliaLang/Pkg.jl/issues/3643

Bike shedding: Should it be skip_symlinks (plural)?

### DIFF
--- a/src/GitTools.jl
+++ b/src/GitTools.jl
@@ -283,7 +283,7 @@ end
 
 Calculate the git tree hash of a given path.
 """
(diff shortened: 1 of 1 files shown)


mkitti:
We actually do want to try to follow the symlinks and copy them to simulate the extraction process.

PR #3754, opened 2024-01-10
Send verbose test output to devnull by default

IanButterworth:
So that `make testall` is reasonable by default


Keno:
Honestly, I think this can be `devnull` by default (at least when run from Base). It's not really nice to leave log files in random locations that the user didn't ask for.

Keno:
LGTM, but somebody who more regularly works on Pkg should review also.

PR #3760, opened 2024-01-14
Reset overrides after artifact tests

IanButterworth:
@staticfloat could this be why we're seeing https://github.com/JuliaLang/julia/issues/52679
More discussion here https://github.com/JuliaLang/julia/pull/52899

### DIFF
--- a/test/artifacts.jl
+++ b/test/artifacts.jl
@@ -779,8 +779,11 @@ end
             Dict("0"^40 => ["not", "a", "string", "or", "dict"]),
             r"failed to parse entry",
         )
+
+        # reset DEPOT_PATH and force Pkg to reload what it knows about artifact overrides
         empty!(DEPOT_PATH)
         append!(DEPOT_PATH, old_depot_path)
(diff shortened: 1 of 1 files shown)


test/artifacts.jl
```diff
@@ -779,8 +779,11 @@ end
             Dict("0"^40 => ["not", "a", "string", "or", "dict"]),
             r"failed to parse entry",
         )
+
+        # reset DEPOT_PATH and force Pkg to reload what it knows about artifact overrides
         empty!(DEPOT_PATH)
         append!(DEPOT_PATH, old_depot_path)
+        Pkg.Artifacts.load_overrides(;force=true)
```
staticfloat, line 786: Yikes, this is a nasty gotcha.  Although overrides are kind of not that widely used (and should generally be eschewed in favor of JLL preferences) it is still annoying that you need to make sure to clear this cache when the depot path changes.  Ideally, we would have a `set_depot_path!()` function that automatically does this kind of clearing for you.

staticfloat (approved):

Issue #3758, opened 2024-01-12
Mention `preserve` strategy used during `add` 


jariji:
```jl
julia> Pkg.add("Cthulhu")
   Resolving package versions...
```

In https://github.com/JuliaLang/Pkg.jl/issues/3757 I forgot I had set `JULIA_PKG_PRESERVE_TIERED_INSTALLED` and I didn't notice I was using `tiered_installed`. It would be useful to mention in the `add` output which strategy is being used

```
Resolving package versions using strategy: tiered_installed (JULIA_PKG_PRESERVE_TIERED_INSTALLED is set)
```

PR #3761, opened 2024-01-14
Show preserve strategy in resolve

jariji:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3758


KristofferC:
Maybe only do this if it is not the default. I feel it would be a bit verbose/confusing to showing it every time.

PR #3764, opened 2024-01-17
ignore tree hashes on Windows w/o symlink capability

StefanKarpinski:
Step one of addressing https://github.com/JuliaLang/Pkg.jl/issues/3643. I implemented this by providing a default for the `ignore_hash` value in `download_artifact`, namely the default value is false unless the OS is Windows and the user cannot create symlinks in the artifact directory, in which case we default `ignore_hash` to true instead. This still emits a warning, which seems desirable based on the discussion in #3643.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -2,6 +2,7 @@ module Artifacts
 
 using Artifacts, Base.BinaryPlatforms, SHA
 using ..MiniProgressBars, ..PlatformEngines
+using Tar: can_symlink
 
 import ..set_readonly, ..GitTools, ..TOML, ..pkg_server, ..can_fancyprint,
        ..stderr_f, ..printpkgstyle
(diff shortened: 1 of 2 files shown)


staticfloat (dismissed):

src/Artifacts.jl
```diff
@@ -314,16 +315,27 @@ function download_artifact(
...
+            # default: false except Windows users who can't symlink
+            Sys.iswindows() && !mktempdir(can_symlink, dirname(src))
+        else
+            Base.get_bool_env("JULIA_PKG_IGNORE_HASHES", ignore_hash)
+            ignore_hash === nothing && @error(
+                "Invalid ENV[\"JULIA_PKG_IGNORE_HASHES\"] value",
+                ENV["JULIA_PKG_IGNORE_HASHES"],
+            )
+            ignore_hash = something(ignore_hash, false)
+        end
```
IanButterworth, line 334: I think this does what you want?
```suggestion
	default = Sys.iswindows() && !mktempdir(can_symlink, dirname(src))
	ignore_hash = Base.get_bool_env("JULIA_PKG_IGNORE_HASHES", default)
```

StefanKarpinski: That's what I had at first, but there are two issues:

1. I don't want to even try to make a symlink if the variable is set.
2. If the variable is set but has an invalid value, I don't want this code to inscrutably error.

IanButterworth: Sounds reasonable https://github.com/JuliaLang/julia/pull/52950

src/Artifacts.jl
```diff
@@ -311,19 +312,33 @@ function download_artifact(
...
+            ignore_hash = Base.get_bool_env("JULIA_PKG_IGNORE_HASHES", false)
+            ignore_hash === nothing && @error(
+                "Invalid ENV[\"JULIA_PKG_IGNORE_HASHES\"] value",
+                ENV["JULIA_PKG_IGNORE_HASHES"],
+            )
+            ignore_hash = something(ignore_hash, false)
+        end
         if ignore_hash
             msg *= "\n\$JULIA_PKG_IGNORE_HASHES is set to 1: ignoring error and moving artifact to the expected location"
             @error(msg)
```
IanButterworth, line 340:
```suggestion
        msg = """
        Tree Hash Mismatch!
          Expected git-tree-sha1:   $(bytes2hex(tree_hash.bytes))
          Calculated git-tree-sha1: $(bytes2hex(calc_hash.bytes))
        """
        # actual and expected artifiact paths
        src = artifact_path(calc_hash; honor_overrides=false)
        dst = artifact_path(tree_hash; honor_overrides=false)
        # Since tree hash calculation is rather fragile and file system dependent,
        # we allow setting JULIA_PKG_IGNORE_HASHES=1 to ignore the error and move
        # the artifact to the expected location and return true
        if get(ENV, "JULIA_PKG_IGNORE_HASHES", "") == ""
            # default: false except Windows users who can't symlink
            ignore_hash = Sys.iswindows() &&
                !mktempdir(can_symlink, dirname(src))
            if ignore_hash
                msg *= "\nWindows cannot make symlinks which can cause hash mismatches. Enable Developer Mode to allow symlinks. Ignoring hash mismatch."
            end
        else
            ignore_hash = Base.get_bool_env("JULIA_PKG_IGNORE_HASHES", false)
            ignore_hash === nothing && @error(
                "Invalid ENV[\"JULIA_PKG_IGNORE_HASHES\"] value",
                ENV["JULIA_PKG_IGNORE_HASHES"],
            )
            ignore_hash = something(ignore_hash, false)
            if ignore_hash 
                msg *= "\n\$JULIA_PKG_IGNORE_HASHES is set to $(repr(ENV["JULIA_PKG_IGNORE_HASHES"])): ignoring error and moving artifact to the expected location"
            end
        end
        if ignore_hash
            
            @error(msg)
```

IanButterworth: Otherwise the messaging makes it seem that the ENV var has been set when it hasn't been

StefanKarpinski:
If someone wants to help me debug this, I'm slammed and haven't and won't have the time (kid was sick last week, I was sick over the weekend, childcare is sick now, plus have to work), but this needs to go into 1.10.1. An intermediate version did work, now it's failing tests, probably because the tests expect some specific error message. If someone could help track down why this is failing now, that would really help.

IanButterworth:
This is the issue. 
```
julia> ENV["JULIA_PKG_IGNORE_HASHES"] = nothing

julia> ENV["JULIA_PKG_IGNORE_HASHES"]
"nothing"
```
The test setup is bad. It should avoid the `withenv` [here](https://github.com/IanButterworth/Pkg.jl/blob/736ac32732a3a2a0d3cb7ef5f1a398cd51e215cf/test/artifacts.jl#L293) rather than set it to nothing.

I'll push a fix

IanButterworth:
Ok. The logic was just inverted on `ignore_hash_env_set`. This passes artifacts tests locally now

IanButterworth (approved):

StefanKarpinski:
Thank you so much, @IanButterworth! You're the best ❤️

PR #3768, opened 2024-01-23
Avoid deleting existing artifacts when ignoring hashes.

nhz2:
Followup to https://github.com/JuliaLang/Pkg.jl/pull/3764
Fixes bug reported in https://github.com/JuliaLang/Pkg.jl/issues/3643#issuecomment-1905048298

This PR modifies `download_artifact` to avoid calling `mv(src, dst; force=true)` where `src` and `dst` are both possible valid artifacts.

The main steps of `download_artifact` are
1. Return `true` if the artifact already exists.
1. Determine the expected artifact path and parent directory.
2. Make a temporary directory.
3. `download_verify_unpack` into the temporary directory.
4. Calculate the tree hash of the temporary directory.
5. If the tree hash doesn't match, either throw an error or log an error, depending on the enviroment.
6. Atomically move the temporary directory to the expected path.
7. Try to clean up the temporary directory.

If steps 4 to 7 throw an error, that error gets returned, otherwise `true` is returned.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -49,6 +49,7 @@ function create_artifact(f::Function)
         # as something that was foolishly overridden.  This should be virtually impossible
         # unless the user has been very unwise, but let's be cautious.
         new_path = artifact_path(artifact_hash; honor_overrides=false)
+        _mv_temp_artifact_dir(temp_dir, new_path)
         if !isdir(new_path)
             # Move this generated directory to its final destination, set it to read-only
             mv(temp_dir, new_path)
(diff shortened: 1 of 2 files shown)


IanButterworth:
@staticfloat does this look ok to you?

IanButterworth:
Given you have a reproducer in https://github.com/JuliaLang/Pkg.jl/issues/3643#issuecomment-1905048298 could that be added as a test?

nhz2:
Yes, I'll try and add a test. I should be able to reproduce the issue on Linux using the right environment variables and artifacts.

staticfloat (approved):

PR #3775, opened 2024-02-01
Prompt the user on a manifest version mismatch

tecosaur:
Currently when the current project's Manifest's `julia_version` doesn't match the current Julia version, Pkg either:
- says nothing, or
- warns that there might be funny behaviour

This PR adds a check that results in a prompt when the versions differ, offering three possible courses of action:
- **refresh** the manifest, by removing the current one and regenerating the manifest
- **version** the manifest, by renaming the current one to `(Julia)Manifest-v{major}.{minor}.toml` and regenerating the manifest (corresponding to https://github.com/JuliaLang/julia/pull/43845)
- **ignore** the mismatch

image

The default behaviour can be set via the new environment variable `JULIA_PKG_MANIFEST_MISMATCH_ACTION`

This PR was prompted by this Zulip gripe thread: https://julialang.zulipchat.com/#narrow/stream/225540-gripes/topic/Manifests.20annoyances

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -488,6 +488,27 @@ To fix such errors, you have a number of options:
 - try fixing the problem yourself.
   This becomes easier once you understand `Project.toml` files and how they declare their compatibility requirements. We'll return to this example in [Fixing conflicts](@ref).
 
+## Manifest version mismatches
+
+Manifest files also record the Julia version used to resolve the project
+dependencies. As such, using an different Julia version can result in a
+different resolved manifest, and using an older Julia version than the one the
+manifest was generated with risks using packages incompatible with the current
+Julia version.
+
+When Pkg notices that there is a mismatch between the manifest Julia version and
+the current Julia version, it will prompt you and ask if you would like to:
+- **refresh** the manifest, by removing the current one and regenerating the manifest
+- **version** the manifest, by renaming the current one to `(Julia)Manifest-v{major}.{minor}.toml` and regenerating the manifest
+- **ignore** the mismatch
+
+The prompt only is shown in interactive sessions. In non-interactive sessions
+the default behaviour is to *ignore* the mismatch.
+
+The default behaviour of interactive an non-interactive sessions alike can be
+set through the `JULIA_PKG_MANIFEST_MISMATCH_ACTION` environment variable, which
+recognises the values: `ask`, `refresh`, `version`, and `ignore`.
+
 ## Garbage collecting old, unused packages
 
 As packages are updated and projects are deleted, installed package versions and artifacts that were
--- a/src/manifest.jl
+++ b/src/manifest.jl
@@ -219,7 +219,14 @@ function read_manifest(f_or_io::Union{String, IO})
     if Base.is_v1_format_manifest(raw)
         raw = convert_v1_format_manifest(raw)
     end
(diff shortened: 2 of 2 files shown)


tecosaur:
I have noticed one bug with this PR that I could do with help resolving, it seems that when using a mismatched manifest, and you just enter the `pkg>` repl and type `rm<space>` the prompt is shown, but you can't respond to it (and `^C` does nothing). I'm not sure what's happening here?

image

IanButterworth:
The bug you're seeing is because of the hint tab completion populating packages to tab complete by reading the manifest.

Note that adding just a warning at a similar code level stalled because it might be too annoying https://github.com/JuliaLang/Pkg.jl/pull/3024

Maybe this is more acceptable given it is actionable now we have `Manifest-v1.11.toml` handling, as your prompt suggests.

IanButterworth:
To handle the tab completion hint issue, maybe suppress the error if it's coming from the completion handler and prevent the manifest being read so that it's read in during the next actual prompt execution, at which point an interactive prompt makes more sense IMO.

MasonProtter:
> Note that adding just a warning at a similar code level stalled because it might be too annoying #3024

Didn't some version of that end up making it's way into julia? I get that warning from time to time, e.g.:
```julia
$ julia-1.10 --project=TSc1D
┌ Warning: The active manifest file has dependencies that were resolved with a different julia version (1.7.2). Unexpected behavior may occur.
└ @ ~/TSc1D/Manifest.toml:0
```

KristofferC:
I am personally not a big fan of:

- More environment variables to control small details of default behavior
- Surprising interactive input from commands that are not always interactive.

tecosaur:
Understandable. Have you got your own thoughts on how the situation with mis-matched Manifests could be improved? Or is it just those two points you have quibbles with?

> I am personally not a big fan of:
> More environment variables to control small details of default behavior

I thought this might be nice, but I don't mind getting rid of it at all.

> Surprising interactive input from commands that are not always interactive.

The interactive ask is gated behind `isinteractive()`, so this shouldn't happen.

KristofferC:
> The interactive ask is gated behind isinteractive(), so this shouldn't happen.

My point is that `add` typically does not have interactive behavior so I often run it, tab away, and then expect it to actually do the thing I told it to during that time.

MasonProtter:
I think `add` is probably just the wrong use-case for this, since as Kristoffer mentioned, the project will re-resolve. I think this is quite nice for `activate` though.

IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/3024 proposed adding the warning to `activate`.
Note that julia/Pkg does not `Pkg.activate` the `active_project` that julia starts in. 

So if your intention is to avoid nasty errors, they won't be avoided before a `Pkg.activate` unless you do something like add a hook to this check inside any possibly related error throw path.

KristofferC:
I'd say that having interactive input deep inside functions like this is non-workable because you don't know what contexts they are called in.

bvdmitri:
> I'd say that having interactive input deep inside functions like this is non-workable because you don't know what contexts they are called in.

Maybe not make it interactive but just do the `v` option by default all the time? What are the downsides? To me it seems like "move to the versioned manifest and refresh" is the best option to do by default and I don't see any immediate downside of this

PR #3776, opened 2024-02-05
make `pkgstr` independent of REPL

KristofferC:
Will make https://github.com/JuliaLang/Pkg.jl/pull/3724 easier.

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -377,21 +377,19 @@ end
 #############
 # Execution #
 #############
-function do_cmd(repl::REPL.AbstractREPL, input::String; do_rethrow=false)
-    if !isinteractive() && !TEST_MODE[] && !PRINTED_REPL_WARNING[]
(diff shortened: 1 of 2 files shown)


src/REPLMode/REPLMode.jl
```diff
@@ -400,9 +398,22 @@ function do_cmd(repl::REPL.AbstractREPL, input::String; do_rethrow=false)
     end
 end
 
-function do_cmd!(command::Command, repl)
+
+function do_cmds(commands::Vector{Command}, io)
+    if !isinteractive() && !TEST_MODE[] && !PRINTED_REPL_WARNING[]
+        @warn "The Pkg REPL mode is intended for interactive use only, and should not be used from scripts. It is recommended to use the functional API instead."
+        PRINTED_REPL_WARNING[] = true
+    end
```
IanButterworth, line 406: maxlog=1 instead?

KristofferC: Probably, but maybe that can be another PR to keep this part a pure code movement.

IanButterworth: Ah yeah ok

Issue #3774, opened 2024-02-01
Base.runtests(["Pkg"]) hangs


vtjnash:
```
julia> Base.runtests(["Pkg"])
Running parallel tests with:
  getpid() = 5539
  nworkers() = 1
  nthreads() = 1
  Sys.total_memory() = 7.274 GiB
  Sys.free_memory() = 1.777 GiB

Test  (Worker) | Time (s) | GC (s) | GC % | Alloc (MB) | RSS (MB)
Pkg        (1) |        started at 2023-12-17T08:10:38.155
┌ Info: Pkg Server metadata:
│ HTTP/2 301 
│ server: Varnish
│ retry-after: 0
│ location: https://us-east.pkg.julialang.org/
[...]
    Updating `/tmp/jl_CXNkw2/environments/v1.11/Project.toml`
  [7876af07] + Example v0.5.4 `https://github.com/JuliaLang/Example.jl#master`
    Updating `/tmp/jl_CXNkw2/environments/v1.11/Manifest.toml`
  [7876af07] + Example v0.5.4 `https://github.com/JuliaLang/Example.jl#master`
    Updating `/tmp/jl_CXNkw2/environments/v1.11/Project.toml`
  [7876af07] - Example v0.5.4 `https://github.com/JuliaLang/Example.jl#master`
    Updating `/tmp/jl_CXNkw2/environments/v1.11/Manifest.toml`
  [7876af07] - Example v0.5.4 `https://github.com/JuliaLang/Example.jl#master`
     Cloning git-repo `https://github.com/DoesNotExist/DoesNotExist.jl`
Username for 'https://github.com': 
```

`Base.runtests(["sorting", "Pkg"]; ncores=2)`, on the other hand, works fine. Presumably because in the latter case Pkg tests are run in a separate worker with a stdin other than stdin, so Pkg can detect that prompting for credentials is fruitless.

_Originally posted by @LilithHafner in https://github.com/JuliaLang/julia/issues/52562_

IanButterworth:
To fix this we can either:
- Go back to not testing `JULIA_PKG_USE_CLI_GIT=1` https://github.com/IanButterworth/Pkg.jl/blob/b13bd2ddf6049865ddd8f7984bc9aa2790c14d21/test/new.jl#L2730
- Figure out how to tell git on CI that it should never prompt for credentials

vtjnash:
[...]

PR #3778, opened 2024-02-05
avoid prompting for git creds on CI

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3774


test/new.jl
```diff
@@ -2727,7 +2727,7 @@ end
 #
 # Note: these tests should be run on clean depots
 for v in (nothing, "true")
-    withenv("JULIA_PKG_USE_CLI_GIT" => v, "GIT_TERMINAL_PROMPT" => 0) do
+    withenv("JULIA_PKG_USE_CLI_GIT" => v, "GIT_TERMINAL_PROMPT" => 0, "GIT_ASKPASS" => "true") do
```
IanButterworth, line 2730:
```suggestion
    # On CI when JULIA_PKG_USE_CLI_GIT=true we need to tell the cli git to not prompt for credentials
    # GIT_ASKPASS=true forces the credential provider to return "" https://stackoverflow.com/a/71057440
    # GIT_TERMINAL_PROMPT=0 is also supposed to avoid the prompt but doesn't reliably
    withenv("JULIA_PKG_USE_CLI_GIT" => v, "GIT_TERMINAL_PROMPT" => 0, "GIT_ASKPASS" => "true") do
```

test/new.jl
```diff
@@ -2727,7 +2727,10 @@ end
 #
 # Note: these tests should be run on clean depots
 for v in (nothing, "true")
-    withenv("JULIA_PKG_USE_CLI_GIT" => v, "GIT_TERMINAL_PROMPT" => 0) do
+    # On CI when JULIA_PKG_USE_CLI_GIT=true we need to tell the cli git to not prompt for credentials
+    # GIT_ASKPASS=true forces the credential provider to return "" https://stackoverflow.com/a/71057440
+    # GIT_TERMINAL_PROMPT=0 is also supposed to avoid the prompt but doesn't reliably
```
IanButterworth, line 2732:
```suggestion
    # GIT_TERMINAL_PROMPT=0 is also supposed to avoid the prompt but doesn't reliably https://github.com/JuliaLang/Pkg.jl/issues/3774
```

KristofferC (approved):

PR #3782, opened 2024-02-07
use the correct `TOML.Parser` coupled to `Base.TOMLCache`

KristofferC:
Fixes https://github.com/JuliaLang/julia/issues/53196

### DIFF
--- a/src/Registry/registry_instance.jl
+++ b/src/Registry/registry_instance.jl
@@ -15,7 +15,7 @@ function to_tar_path_format(file::AbstractString)
 end
 
 # See loading.jl
-const TOML_CACHE = Base.TOMLCache(TOML.Parser(), Dict{String, Dict{String, Any}}())
+const TOML_CACHE = Base.TOMLCache(Base.TOML.Parser(), Dict{String, Dict{String, Any}}())
 const TOML_LOCK = ReentrantLock()
 _parsefile(toml_file::AbstractString) = Base.parsed_toml(toml_file, TOML_CACHE, TOML_LOCK)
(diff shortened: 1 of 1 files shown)


src/Registry/registry_instance.jl
```diff
@@ -15,15 +15,15 @@ function to_tar_path_format(file::AbstractString)
...
+const TOML_CACHE = Base.TOMLCache(Base.TOML.Parser(), Dict{String, Dict{String, Any}}())
 const TOML_LOCK = ReentrantLock()
 _parsefile(toml_file::AbstractString) = Base.parsed_toml(toml_file, TOML_CACHE, TOML_LOCK)
 function parsefile(in_memory_registry::Union{Dict, Nothing}, folder::AbstractString, file::AbstractString)
     if in_memory_registry === nothing
         return _parsefile(joinpath(folder, file))
     else
         content = in_memory_registry[to_tar_path_format(file)]
-        return TOML.Internals.parse(TOML.Parser(content; filepath=file))
+        return TOML.Internals.parse(Base.TOML.Parser(content; filepath=file))
```
vtjnash, line 26: Seems odd to use `TOML.Internals` here but pass `Base.TOML` as an argument. Maybe instead change `using TOML` to `using Base.TOML` at the top of the file?

vtjnash: bump?

JeffBezanson:
Bump

vtjnash:
bump?

KristofferC:
I don't know why this was put on the milestone, it's been like this for years.

PR #3783, opened 2024-02-07
Support a `[sources]` section in Project.toml for specifying paths and repo locations for dependencies

KristofferC:
Extracted and tweaked based on https://github.com/JuliaLang/Pkg.jl/pull/3263.

Currently requires https://github.com/JuliaLang/julia/pull/53233

Some questions:

- What happens if you have a source for a URL and then you `add` the package with another URL? Error or update the source?
- Same as above but with a `path`.
- What should `free` do when there is a source specified.

TODO:
- [ ] Docs
- [ ] More thorough tests
- [x] Verify parsing of source section in project file
- [x] remove source in `free` (`update_package_free`)
- [x] update source on `add` (`update_package_add`)
- [x] remove source in `rm`
- [x] Test when `resolve` with non source in manifest and source in project.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,8 @@
+Pkg v1.12 Release Notes
+
+- It is now possible to specify "sources" for packages in a `[sources]` section in Project.toml.
+  This can be used to add non-registered normal or test dependencies. Packages will also automatically be added to `[sources]` when they are added by url or devved.
+
 Pkg v1.11 Release Notes
 =======================
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,7 +1,8 @@
 Pkg v1.12 Release Notes
+=======================
 
 - It is now possible to specify "sources" for packages in a `[sources]` section in Project.toml.
-  This can be used to add non-registered normal or test dependencies. Packages will also automatically be added to `[sources]` when they are added by url or devved.
+  This can be used to add non-registered normal or test dependencies.
 
 Pkg v1.11 Release Notes
 =======================
--- a/docs/src/toml-files.md
+++ b/docs/src/toml-files.md
@@ -91,6 +91,19 @@ Test = "8dfed614-e22c-5e08-85e1-65c5234f0b40"
 Typically it is not needed to manually add entries to the `[deps]` section; this is instead
 handled by Pkg operations such as `add`.
 
+### The `[sources]` section
+
+Specifiying a path or repo (+ branch) for a dependency is done in the `[sources]` section.
+These are especially useful for controlling unregistered dependencies without having to bundle a
+corresponding manifest file.
+
+```toml
+[sources]
+Example = {url = "https://github.com/JuliaLang/Example.jl", rev = "custom_branch"}
+SomeDependency = {path = "deps/SomeDependency.jl"}
+```
+
+Note that this information is only used when this environment is active, i.e. it is not used if this project is a package that is being used as a dependency.
 
 ### The `[compat]` section
 
@@ -135,7 +148,7 @@ Julia will then preferentially use the version-specific manifest file if availab
 For example, if both `Manifest-v1.11.toml` and `Manifest.toml` exist, Julia 1.11 will prioritize using `Manifest-v1.11.toml`.
 However, Julia versions 1.10, 1.12, and all others will default to using `Manifest.toml`.
 This feature allows for easier management of different instantiated versions of dependencies for various Julia versions.
-Note that there can only be one `Project.toml` file. While `Manifest-v{major}.{minor}.toml` files are not automatically 
+Note that there can only be one `Project.toml` file. While `Manifest-v{major}.{minor}.toml` files are not automatically
 created by Pkg, users can manually rename a `Manifest.toml` file to match
 the versioned format, and Pkg will subsequently maintain it through its operations.
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -181,6 +181,31 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why, :
     end
 end
 
+function update_source_if_set(project, pkg)
+    source = get(project.sources, pkg.name, nothing)
+    source === nothing && return
+    # This should probably not modify the dicts directly...
+    if pkg.repo.source !== nothing
+        source["url"] = pkg.repo.source
+    end
+    if pkg.repo.rev !== nothing
+        source["rev"] = pkg.repo.rev
+    end
+    if pkg.path !== nothing
+        source["path"] = pkg.path
+    end
+    path, repo = get_path_repo(project, pkg.name)
+    if path !== nothing
+        pkg.path = path
+    end
+    if repo.source !== nothing
+        pkg.repo.source = repo.source
+    end
+    if repo.rev !== nothing
+        pkg.repo.rev = repo.rev
+    end
+end
+
 function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
                  preserve::PreserveLevel=Operations.default_preserve(), platform::AbstractPlatform=HostPlatform(), kwargs...)
     require_not_empty(pkgs, :develop)
@@ -212,6 +237,7 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
 
     new_git = handle_repos_develop!(ctx, pkgs, shared)
 
+
     for pkg in pkgs
         if Types.collides_with_project(ctx.env, pkg)
             pkgerror("package $(err_rep(pkg)) has the same name or UUID as the active project")
@@ -219,6 +245,7 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
         if length(findall(x -> x.uuid == pkg.uuid, pkgs)) > 1
             pkgerror("it is invalid to specify multiple packages with the same UUID: $(err_rep(pkg))")
         end
+        update_source_if_set(ctx.env.project, pkg)
     end
 
     Operations.develop(ctx, pkgs, new_git; preserve=preserve, platform=platform)
@@ -272,6 +299,7 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}; preserve::PreserveLevel=Op
         if length(findall(x -> x.uuid == pkg.uuid, pkgs)) > 1
             pkgerror("it is invalid to specify multiple packages with the same UUID: $(err_rep(pkg))")
         end
+        update_source_if_set(ctx.env.project, pkg)
     end
 
     Operations.add(ctx, pkgs, new_git; preserve, platform, target)
@@ -311,12 +339,14 @@ end
 function append_all_pkgs!(pkgs, ctx, mode)
     if mode == PKGMODE_PROJECT || mode == PKGMODE_COMBINED
         for (name::String, uuid::UUID) in ctx.env.project.deps
-            push!(pkgs, PackageSpec(name=name, uuid=uuid))
+            path, repo = get_path_repo(ctx.env.project, name)
+            push!(pkgs, PackageSpec(name=name, uuid=uuid, path=path, repo=repo))
         end
     end
     if mode == PKGMODE_MANIFEST || mode == PKGMODE_COMBINED
         for (uuid, entry) in ctx.env.manifest
-            push!(pkgs, PackageSpec(name=entry.name, uuid=uuid))
+            path, repo = get_path_repo(ctx.env.project, entry.name)
+            push!(pkgs, PackageSpec(name=entry.name, uuid=uuid, path=path, repo=repo))
         end
     end
     return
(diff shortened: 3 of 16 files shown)


MilesCranmer:
Personal views on these:

> What happens if you have a source for a URL and then you add the package with another URL? Error or update the source?

Updating the source seems the most reasonable thing to do here. 

> Same as above but with a path.

Also update the source.

> What should free do when there is a source specified.

I think `free` should remove the source, only because currently the docstring says *If pkg is tracking a path, e.g. after Pkg.develop, go back to tracking registered versions.* Since a path is a type of source I think this behavior would be in line with the docstring.

KristofferC:
More questions:

- What happens if you have a source of a package that is not in `[deps]`? Is that an error, or what should happen when you end up adding a package with that name?
- Related to the above, should `rm` of a package remove its source entry?
- What happens if you have a manifest with one path and then modify the project source to take another path? What should take precedence? Presumably the project source?

MilesCranmer:
> * What happens if you have a source of a package that is not in `[deps]`? Is that an error, or what should happen when you end up adding a package with that name?

I think it should error, *unless* it is in `[weakdeps]` or `[extras]`. That would let you set sources for packages you don't directly depend on. i.e., the same behavior as for `[compat]`:

```julia
# With MyFakePackage appearing in [compat]
ERROR: Compat `MyFakePackage` not listed in `deps`, `weakdeps` or `extras` section at ...
```

> * Related to the above, should `rm` of a package remove its source entry?

I think it should (similar to `[compat]`)

> * What happens if you have a manifest with one path and then modify the project source to take another path? What should take precedence? Presumably the project source?

I think that the manifest would take precedence but there should be a warning message that the manifest file is out of date and should be updated. i.e., similar behavior to when you change the `[compat]` without updating the manifest:

```julia
┌ Warning: The project dependencies or compat requirements have changed since the manifest was last resolved.
│ It is recommended to `Pkg.resolve()` or consider `Pkg.update()` if necessary.
└ @ Pkg.API ~/.julia/juliaup/julia-1.10.0+0.aarch64.apple.darwin14/share/julia/stdlib/v1.10/Pkg/src/API.jl:1800
```

KristofferC:
> I think that the manifest would take precedence but there should be a warning message that the manifest file is out of date and should be updated. i.e., similar behavior to when you change the [compat] without updating the manifest:

I underspecified a bit perhaps. What I means was, you have a manifest with some `path` to a package. Then you set a `source` to another path. Now you do a `resolve`. What should be in the manifest after that? Having the non-source path entry from the old manifest would be bad IMO in a similar way of having a package that breaks the compat specification.

MilesCranmer:
Oh, I see. Then yes I would agree with you the `[sources]` should take precedence and overwrite the Manifest

giordano:
I haven't tried this PR yet, but a possible use-case I see for this feature is having test-only dependencies which aren't registered in General, sometimes they aren't really worth being registered on their own.  This came up for example in https://github.com/JuliaHEP/Geant4.jl/issues/8, where the current solution is to manually check out the repository.

Krastanov:
Will this be useful for `dev` packages. E.g. I `dev ./LocalPackage.jl` which provides a source entry `LocalPackage = "file:/.../LocalPackage.jl"`. The I delete my `Manifest.toml` and finally I do `instantiate`. Will `LocalPackage`:

- be added or deved?
- still refer to `file:/.../LocalPackage.jl`?

KristofferC:
> feature is having test-only dependencies which aren't registered in General, sometimes they aren't really worth being registered on their own. 

This should work now, see added testr.

mortenpi:
I am cautiously excited about this! I'm curious about two use cases though, and it's not immediately clear if this helps with these or not:

1. The Documenter-type use case. I have e.g. a `docs/Project.toml`:

   ```toml
   [deps]
   MyPackage = "5217a498-cd5d-4ec6-b8c2-9b85a09b6e3e"
   Documenter = "e30172f5-a6a5-5a46-863b-614d45cd2de4"

   [source]
   MyPackage = {path = ".."}
   ```

   Just doing `julia --project=docs/ -e 'using Pkg; Pkg.instantiate()'` should work, right? But will the Manifest now point to `MyPackage` as a dev-dependency (i.e. pointing to `..`) or will it copy the source to into the depot (i.e. `~/.julia/packages`)?

2. Will it handle the case where the local dependency has itself a `[sources]` section that points to another local "sub-dependency"? I.e. suppose I have a top-level environment I want to instantiate / work in:

   ```toml
   [deps]
   LocalPackage = "5217a498-cd5d-4ec6-b8c2-9b85a09b6e3e"

   [source]
   LocalPackage = {path = "packages/LocalPackage"}
   ```

   And `packages/LocalPackage/Project.toml` itself uses another local package:

   ```toml
   [deps]
   LocalSubPackage = "e30172f5-a6a5-5a46-863b-614d45cd2de4"

   [source]
   LocalSubPackage = {path = "../LocalSubPackage/"}
   ```

   Will it resolve it correctly? I.e. will `LocalSubPackage` get added to the manifest automatically?

   The use case here is the monorepo case, where you have one or more top-level dependencies, and then a bunch of vendored local packages that have a complex dependency graph among themselves, but you would pretty much like to treat them as normal package.

KristofferC:
> The Documenter-type use case. 

Yes, that should work fine and it will point to it as a "dev-dependency".

> Will it handle the case where the local dependency has itself a [sources] section that points to another local "sub-dependency"? 

At least for now, I didn't plan on making it "recursive". It's possible but it also has many failure modes where you encounter conflicting information and you may also have to keep downloading stuff as you "discover" more and more URL sources. As a first cut I don't think that will be enabled but it can be further discussed later of course.

mortenpi:
> At least for now, I didn't plan on making it "recursive".

Would it by any chance work if I specify the subpackages in the top-level project though, even if they are not direct dependencies of the top-level project themselves? So instead of the example above, you'd have something like this:

```toml
[deps]
LocalPackage = "5217a498-cd5d-4ec6-b8c2-9b85a09b6e3e"

[source]
LocalPackage = {path = "packages/LocalPackage"}
LocalSubPackage = {path = "packages/LocalSubPackage/"}
```

And `packages/LocalPackage/Project.toml` itself uses another local package:

```toml
[deps]
LocalSubPackage = "e30172f5-a6a5-5a46-863b-614d45cd2de4"
```

Adding some CI checks and simple tooling to keep a global `[source]` list updated across a few top-level projects would be easy.

KristofferC:
> Would it by any chance work if I specify the subpackages in the top-level project though, even if they are not direct dependencies of the top-level project themselves? 

In theory, I think it could be made to work if that package is also added to `[extras]`.

MasonProtter:
Just tested this out, and the use-case I had in mind seems to work great!  I'm very excited, thank you for the awesome work @KristofferC. This, together with the app stuff, and pkgimages are all coalescing into a very cool story for julia code management and sharing. 

___

Regarding the usage questions, I think I agree with @MilesCranmer's answers, they all seem reasonable. Generally, my mental model is that the Project.toml should describe what we desire, and the Manifest.toml should describe what specifically was done. 

So being able to describe paths, urls, and branches in the Project, and have those take priority over what is written in the Manifest seems correct and ergonomic to me. 

____


Regarding API, is there a way we can make `Pkg.add(path=some_local_path)` add the package at `some_local_path` to the `[sources]` section? Or maybe we should make a new function like `Pkg.add_source`?

KristofferC:
> Regarding API, is there a way we can make Pkg.add(path=some_local_path) add the package at some_local_path to the [sources] section?

I think that makes sense to do automatically because a `free` or `rm` will also remove it from the source section.

LilithHafner:
The design decision here all look good to me. Thanks for this lovely feature!

quinnj:
Thanks for picking this up and pushing it forward @KristofferC; sorry I let it drop/droop so long ago. I agree it will be a great feature!

tpgillam:
To echo, thank you for working on this!

I'm very interested in whether this might simplify the 'monorepo workflow' discussed here: https://github.com/JuliaLang/Pkg.jl/pull/3263#issuecomment-1890426685

Am I right in thinking that a top-level, single Manifest.toml will be required in this case? 

IIUC, I think 'yes' because `[sources]` won't be followed recursively. i.e. suppose we have a directory containing multiple packages that depend on each other non-trivially. Ideally one would be able to go into a leaf dependency and say `] instantiate`. But if A depends on B, and B depends on C, then I can't instantiate A because it won't find C?

Or is there a mechanism for all of A, B & C etc. to share a single `[sources]` section "included" into their Project.tomls from somewhere else?

goerz:
> But in test/Project.toml you don't need to put the parent package at all I thought?

The depends on your workflow. If you want to open up are REPL with `julia --project=test` to interactively run specific tests or to build your documentation, you'll generally end up with the main project listed in the `test/Project.toml`. This should make that easier.

nhz2:
This error also happens if I add a local unregistered test utility package to the `test/Project.toml` sources.

KristofferC:
Okay, I'll see what is going on with that.

ericphanson:
Could it make sense to be able to specify a registry in the sources block? Maybe with a separate registries block to have the url. That way a Project/Manifest does not need to depend on the registries already being installed.

Seelengrab:
Yeah, I like that too - `]instantiate` can then just `]registry add` the relevant registries too, if they're not already added.

KristofferC:
I'm going to back out the change to make adding things by url and devving to automatically add it to `[sources]` because i don't have time to look into CI for that (need to focus a bit on 1.11 now). I'll open a PR for the WIP on that though.

KristofferC:
> This error also happens if I add a local unregistered test utility package to the test/Project.toml sources.

I will fix this in an upcoming PR.

goerz:
Wow, so excited this is merged! Does this automatically show up in the "nightly" Julia release, or do I need to do something to play around with this (since Pkg is developed sort-of-independently)?

KristofferC:
When Pkg is bumped on nightly it will be in the nightly release. Probably easiest to wait for that.

KristofferC:
This is the PR to add thins automatically to `[sources]` if someone wants to look through the CI stuff. Might be easy. Also needs tests.

musoke:
> Wow, so excited this is merged! Does this automatically show up in the "nightly" Julia release, or do I need to do something to play around with this (since Pkg is developed sort-of-independently)?

Current nightly has these changes - this Project.toml acted as expected.
```
[deps]

[sources]
Example = {url = "https://github.com/JuliaLang/Example.jl", rev = "custom_branch"}
```

```
julia> versioninfo()
Julia Version 1.12.0-DEV.369
Commit 2f9096218bf (2024-04-19 12:25 UTC)
Build Info:
```

sylvaticus:
>  Generally, my mental model is that the Project.toml should describe what we desire, and the Manifest.toml should describe what specifically was done.
> So being able to describe paths, urls, and branches in the Project, and have those take priority over what is written in the Manifest seems correct and ergonomic to me.

I have the feeling that the second statement a bit contradicts the first one. Project.toml describes what [in general] we want, and Manifest.toml should describe what [in the specific installation of the package] was done. So Manifest should take precedence.
This leads to an issue with a couple of unregistered packages I am developing in parallel (julia-1.11.0-beta1). I want the main package project to specify that the dependent package source is on github. However I want then be able to specify that in my particular installation the source is in  `.julia/dev`:

Main package (GenFSM) Project.toml:
```
[deps]
GenFSM_resources = "731fd045-8cb1-4036-8c04-a2a477f3265a"

[sources]
GenFSM_resources = {url = "git@github.com:forestmod/GenFSM_resources.jl.git", rev = "main"}
```

But when I try to specify to use a local version of the dependant package, instead of modify Manifest.toml and use that path in my particular installation, it returns an error:
```
(GenFSM) pkg> dev GenFSM_resources
ERROR: `path` and `url` are conflicting specifications
```

KristofferC:
What do you want `dev GenFSM_resources` to do? Do you want it to rewrite the entry in your `[sources]` or just start ignoring that source entry completely? When is the source entry used?

sylvaticus:
Granted, I could be completely wrong, but I don't think that `dev GenFSM_resources`, ran in the context of the `GenFSM` environment, should change the `[source]` of `GenFSM_resources` (in `GenFSM` `Project.toml`), but rather add the local path of `GenFSM_resources` in  `GenFSM` `Manifest.toml`. 
Then the `[source]` of `GenFSM_resources`  would be used when there is no `Manifest`, for example to instantiate the environment of the `GenFSM` package...

KristofferC:
So say you add:

```
[sources]
Example = {url = "[https:](https://github.com/JuliaLang/Example.jl)"}
```

and then do a `Pkg.add("Example")`. Should the source be respected?

sylvaticus:
I think the difference between `add` and `dev` is that the first one is supposed to change the package itself (let's name it `Foo`). `Foo` to work needs `Example`, so you `add Example` for all the users of the package and you change the  `Foo` `Project.toml`.
If you instead do `dev Example`, it's for yourself, the guy in front of the machine, so `Example` `Project.toml` shouldn't be touched, only its manifest (and this should prevail when actual package loading happens to lookup `Example`).
To answer your question, provided that there is a way to do `Pkg.add("Example",url="[https:](https://github.com/JuliaLang/Example.jl)`, then doing `add` without this information for me should remove it from `[source]`.

PR #3793, opened 2024-02-13
CI: Add Apple Silicon (macOS aarch64) to the CI matrix

IanButterworth:


DilumAluthge:
For contrast, if I look at https://github.com/JuliaSparse/SparseArrays.jl/actions/runs/7891594032/job/21536261812?pr=505 (whcih is from https://github.com/JuliaSparse/SparseArrays.jl/pull/505), I see:

```
Platform Info:
  OS: macOS (arm64-apple-darwin22.4.0)
  CPU: 3 × Apple M1 (Virtual)
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, apple-m1)
```

IanButterworth:
Ok yeah, that's more like it. 20 min on Apple silicon. 35 min on Intel

DilumAluthge (approved):
LGTM.

It would be nice to have a way of programmatically checking that we're not in the "running x86_64 Julia under Rosetta on Apple Silicon" situation, instead of having to manually inspect the logs. But I think we can do that in a follow-up PR.

PR #3800, opened 2024-02-15
Prevent repl crash on invalid command

oscardssmith:
Previously, `prepare_cmd` was outside the lock.
fixes https://github.com/JuliaLang/Pkg.jl/issues/3787


oscardssmith:
This needs tests, but I'm not entirely sure where the test should go.

KristofferC:
Testing this is a bit awkward since it would require firing up and simulating a REPL and I don't think we should block that on getting this fixed.

Issue #1062, opened 2019-02-14
Introduce `down` as opposite of `up`


oxinabox:
I wish to test that I have set the lower bounds in my project.toml correctly.
I thus want to check that with the oldest package versions that can be compatibly installed,
my tests still pass.

So I want to do the oppoite of `]up`

fredrikekre:
This should probably just be flag to `pkg> test` rather than a `pkg> down` command.

oxinabox:
No, because when debugging I want to instantiate the set of package versions I am testing on.
And then I can incrementally raise my lower bound, and run `pkg> resolve` and then test again.

StefanKarpinski:
`pkg> down` seems reasonable to me although it might just be a shortcut for `pkg> up --lower` or something like that.

cossio:
Related post from rust: https://users.rust-lang.org/t/psa-please-specify-precise-dependency-versions-in-cargo-toml/71277.

marius311:
Are there any hacks to do this right now? I agree it'd be really useful for exactly the reason stated by OP. The place I'd really want to use it is for my CI; if available, I'd test with both the highest and lowest possible compatible dependency versions.

KristofferC:
@carlobaldassi, what would be required in the resolver to support this? I am guessing just redefining some stuff in https://github.com/JuliaLang/Pkg.jl/blob/master/src/Resolve/versionweights.jl but are there more things? If you could have some example code I could integrate it easily.

carlobaldassi:
@KristofferC I think that the main change would be to reverse the sorting order of versions of the required packages (while leaving the rest unchanged) here: https://github.com/JuliaLang/Pkg.jl/blob/76070d295fc4a1f27f852e05400bbc956962e084/src/Resolve/graphtype.jl#L128

Then I think it might be sufficient to negate the second argument `vweight[p0][v0]` here: https://github.com/JuliaLang/Pkg.jl/blob/76070d295fc4a1f27f852e05400bbc956962e084/src/Resolve/maxsum.jl#L61

I suspect this may be all (plus of course passing around some flags), but it's hard to tell without testing. Unfortunately, due to personal matters, I don't have the time for doing that myself at the moment.

On the other hand, I wonder whether it would be best to simply call the solver with the versions of the explicitly required dependencies fixed to their declared lower bounds. In my mind, that would be simpler, safer, and more precise.
[...]

PR #3802, opened 2024-02-19
quick proof of concept of `down`

KristofferC:
Very quick implementation of https://github.com/JuliaLang/Pkg.jl/issues/1062#issuecomment-1952014608, (with some debug logging in it to show what happens), specifically:

> On the other hand, I wonder whether it would be best to simply call the solver with the versions of the explicitly required dependencies fixed to their declared lower bounds. In my mind, that would be simpler, safer, and more precise.
There would be no need to further increase the complexity of the resolver code, and it would actually test that the declared lower bounds work. After all, if the resolver finds that it cannot set a package version to its declared lower bound due to an incompatibility with some other required package, then the lower bound should be raised, no? Or am I missing something?

An example of its usage:

```
(PGFPlotsX) pkg> st
Project PGFPlotsX v1.6.1
...

julia> Pkg.API.down()
[ Info: Fixing OrderedCollections to version 1.4.0
[ Info: Fixing Tables to version 1.0.0
[ Info: Fixing DefaultApplication to version 1.0.0
[ Info: Fixing Parameters to version 0.12.0
[ Info: Fixing ArgCheck to version 1.0.0
[ Info: Fixing DocStringExtensions to version 0.8.0
[ Info: Fixing Requires to version 0.5.0
[ Info: Fixing MacroTools to version 0.5.0

    Updating `~/JuliaPkgs/PGFPlotsX.jl/Project.toml`
⌃ [dce04be8] ↓ ArgCheck v2.3.0 ⇒ v1.0.0
⌃ [3f0dd361] ↓ DefaultApplication v1.1.0 ⇒ v1.0.0
⌃ [ffbed154] ↓ DocStringExtensions v0.9.3 ⇒ v0.8.0
⌃ [1914dd2f] ↓ MacroTools v0.5.13 ⇒ v0.5.0
⌃ [bac558e1] ↓ OrderedCollections v1.6.3 ⇒ v1.4.0
⌃ [d96e819e] ↓ Parameters v0.12.3 ⇒ v0.12.0
⌃ [ae029012] ↓ Requires v1.3.0 ⇒ v0.5.0
⌃ [bd369af6] ↓ Tables v1.11.1 ⇒ v1.0.0
    Updating `~/JuliaPkgs/PGFPlotsX.jl/Manifest.toml`
⌃ [dce04be8] ↓ ArgCheck v2.3.0 ⇒ v1.0.0
⌅ [00ebfdb7] + CSTParser v0.6.2
⌅ [34da2185] + Compat v2.2.1
⌅ [864edb3b] + DataStructures v0.17.20
⌃ [3f0dd361] ↓ DefaultApplication v1.1.0 ⇒ v1.0.0
  [8bb1440f] + DelimitedFiles v1.9.1
⌃ [ffbed154] ↓ DocStringExtensions v0.9.3 ⇒ v0.8.0
⌃ [1914dd2f] ↓ MacroTools v0.5.13 ⇒ v0.5.0
⌃ [bac558e1] ↓ OrderedCollections v1.6.3 ⇒ v1.4.0
⌃ [d96e819e] ↓ Parameters v0.12.3 ⇒ v0.12.0
⌃ [ae029012] ↓ Requires v1.3.0 ⇒ v0.5.0
  [10745b16] + Statistics v1.11.1
⌃ [bd369af6] ↓ Tables v1.11.1 ⇒ v1.0.0
  [0796e94c] + Tokenize v0.5.28
  [3a884ed6] - UnPack v1.0.2
```

Trying this on Plots.jl we get the as @oxinabox predicted resolver errors:

```
pkg> activate ../Plots.jl/
  Activating project at `~/JuliaPkgs/Plots.jl`

julia> Pkg.API.down()
[ Info: Fixing Showoff to version 0.3.1
[ Info: Fixing GR to version 0.69.5
[ Info: Fixing Statistics to version 1.11.0
[ Info: Fixing FixedPointNumbers to version 0.6.0
[ Info: Fixing JLFzf to version 0.1.0
[ Info: Fixing PrecompileTools to version 1.0.0
[ Info: Fixing Unzip to version 0.1.0
[ Info: Fixing UnitfulLatexify to version 1.0.0
[ Info: Fixing RecipesPipeline to version 0.6.10
[ Info: Fixing LaTeXStrings to version 1.0.1
[ Info: Fixing PlotUtils to version 1.0.0
[ Info: Fixing JSON to version 0.21.0
[ Info: Fixing StatsBase to version 0.33.0
[ Info: Fixing RelocatableFolders to version 0.3.0
[ Info: Fixing Scratch to version 1.0.0
[ Info: Fixing Latexify to version 0.14.0
[ Info: Fixing FFMPEG to version 0.2.0
[ Info: Fixing Measures to version 0.3.0
[ Info: Fixing RecipesBase to version 1.3.1
[ Info: Fixing UnicodeFun to version 0.4.0
[ Info: Fixing PlotThemes to version 2.0.0
[ Info: Fixing Contour to version 0.5.0
[ Info: Fixing Reexport to version 0.2.0
[ Info: Fixing Requires to version 1.0.0
[ Info: Fixing NaNMath to version 0.3.2
ERROR: Unsatisfiable requirements detected for package Scratch [6c6a2e73]:
 Scratch [6c6a2e73] log:
 ├─possible versions are: 1.0.0 - 1.2.1 or uninstalled
 ├─restricted to versions 1 by Plots [91a5bcdd], leaving only versions: 1.0.0 - 1.2.1
 │ └─Plots [91a5bcdd] log:
 │   ├─possible versions are: 1.40.1 or uninstalled
 │   └─Plots [91a5bcdd] is fixed to version 1.40.1
 ├─restricted to versions 1.0.0 by an explicit requirement, leaving only versions: 1.0.0
 └─restricted by compatibility requirements with RelocatableFolders [05181044] to versions: 1.1.0 - 1.2.1 — no versions left
   └─RelocatableFolders [05181044] log:
     ├─possible versions are: 0.1.0 - 1.0.1 or uninstalled
     ├─restricted to versions [0.3, 1] by Plots [91a5bcdd], leaving only versions: 0.3.0 - 1.0.1
     │ └─Plots [91a5bcdd] log: see above
     └─restricted to versions 0.3.0 by an explicit requirement, leaving only versions: 0.3.0
```

cc @carlobaldassi, @oxinabox

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -140,7 +140,7 @@ function require_not_empty(pkgs, f::Symbol)
 end
 
 # Provide some convenience calls
-for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why, :precompile)
+for f in (:develop, :add, :rm, :up, :down, :pin, :free, :test, :build, :status, :why, :precompile)
     @eval begin
         $f(pkg::Union{AbstractString, PackageSpec}; kwargs...) = $f([pkg]; kwargs...)
         $f(pkgs::Vector{<:AbstractString}; kwargs...)          = $f([PackageSpec(pkg) for pkg in pkgs]; kwargs...)
@@ -156,8 +156,8 @@ for f in (:develop, :add, :rm, :up, :pin, :free, :test, :build, :status, :why, :
             pkgs = deepcopy(pkgs) # don't mutate input
             foreach(handle_package_input!, pkgs)
             ret = $f(ctx, pkgs; kwargs...)
-            $(f in (:up, :pin, :free, :build)) && Pkg._auto_precompile(ctx)
-            $(f in (:up, :pin, :free, :rm)) && Pkg._auto_gc(ctx)
+            $(f in (:up, :pin, :free, :build, :down)) && Pkg._auto_precompile(ctx)
+            $(f in (:up, :pin, :free, :rm, :down)) && Pkg._auto_gc(ctx)
             return ret
         end
         $f(ctx::Context; kwargs...) = $f(ctx, PackageSpec[]; kwargs...)
@@ -333,7 +333,7 @@ function up(ctx::Context, pkgs::Vector{PackageSpec};
         printpkgstyle(ctx.io, :Update, "All dependencies are pinned - nothing to update.", color = Base.info_color())
         return
     end
-    if update_registry
+    if false # update_registry
         Registry.download_default_registries(ctx.io)
         Operations.update_registries(ctx; force=true)
     end
@@ -349,6 +349,14 @@ function up(ctx::Context, pkgs::Vector{PackageSpec};
     end
     Operations.up(ctx, pkgs, level; skip_writing_project, preserve)
     return
+end
+
+function down(ctx::Context, pkgs::Vector{PackageSpec};
+            kwargs...)
+    if !isempty(pkgs)
+        pkgerror("`down` can only be called without specifying packages")
+    end
+    return up(ctx, pkgs; level=UPLEVEL_DOWN)
 end
 
 resolve(; io::IO=stderr_f(), kwargs...) = resolve(Context(;io); kwargs...)
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1480,10 +1480,38 @@ function up_load_versions!(ctx::Context, pkg::PackageSpec, entry::PackageEntry, 
         r = level == UPLEVEL_PATCH ? VersionRange(ver.major, ver.minor) :
             level == UPLEVEL_MINOR ? VersionRange(ver.major) :
             level == UPLEVEL_MAJOR ? VersionRange() :
-                error("unexpected upgrade level: $level")
+            level == UPLEVEL_DOWN ? begin
+                if !is_tracking_registry(pkg) || is_stdlib(pkg.uuid)
+                    VersionRange()
+                else
+                    compat = get(ctx.env.project.compat, pkg.name, nothing)
+                    vr = compat === nothing ? VersionRange() : compat.val
+                    lowest_version(pkg, ctx.registries, vr)
+                end
+            end :
+            error("unexpected upgrade level: $level")
         pkg.version = VersionSpec(r)
     end
     return false
+end
+
+function lowest_version(pkg, registries, vr::VersionSpec)
+    minv = nothing
+    for reg in registries
+        rpkg = get(reg, pkg.uuid, nothing)
+        rpkg === nothing && continue
+        pkginfo = Registry.registry_info(rpkg)
+        matching_versions = filter(in(vr), keys(pkginfo.version_info))
+        if !isempty(matching_versions)
+            min_match = minimum(matching_versions)
+            minv = minv === nothing ? min_match : minv < min_match ? minv : min_match
+        end
+    end
+    if minv === nothing
+        pkgerror("Did not find any version matching $vr in registries")
+    end
+    @info "Fixing $(pkg.name) to version $minv"
+    return minv
 end
 
 up_load_manifest_info!(pkg::PackageSpec, ::Nothing) = nothing
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -13,7 +13,7 @@ using Dates
 export @pkg_str
 export PackageSpec
 export PackageMode, PKGMODE_MANIFEST, PKGMODE_PROJECT
-export UpgradeLevel, UPLEVEL_MAJOR, UPLEVEL_MINOR, UPLEVEL_PATCH
+export UpgradeLevel, UPLEVEL_MAJOR, UPLEVEL_MINOR, UPLEVEL_PATCH, UPLEVEL_DOWN
 export PreserveLevel, PRESERVE_TIERED_INSTALLED, PRESERVE_TIERED, PRESERVE_ALL_INSTALLED, PRESERVE_ALL, PRESERVE_DIRECT, PRESERVE_SEMVER, PRESERVE_NONE
 export Registry, RegistrySpec
 
@@ -68,7 +68,7 @@ include("API.jl")
 include("REPLMode/REPLMode.jl")
 
 import .REPLMode: @pkg_str
-import .Types: UPLEVEL_MAJOR, UPLEVEL_MINOR, UPLEVEL_PATCH, UPLEVEL_FIXED
+import .Types: UPLEVEL_MAJOR, UPLEVEL_MINOR, UPLEVEL_PATCH, UPLEVEL_FIXED, UPLEVEL_DOWN
 import .Types: PKGMODE_MANIFEST, PKGMODE_PROJECT
 import .Types: PRESERVE_TIERED_INSTALLED, PRESERVE_TIERED, PRESERVE_ALL_INSTALLED, PRESERVE_ALL, PRESERVE_DIRECT, PRESERVE_SEMVER, PRESERVE_NONE
 
@@ -99,6 +99,7 @@ An enum with the instances
   * `UPLEVEL_PATCH`
   * `UPLEVEL_MINOR`
   * `UPLEVEL_MAJOR`
+  * `UPLEVEL_DOWN`
 
 Determines how much a package is allowed to be updated.
 Used as an argument to  [`PackageSpec`](@ref) or as an argument to [`Pkg.update`](@ref).
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -18,7 +18,7 @@ import FileWatching
 import Base: SHA1
 using SHA
 
-export UUID, SHA1, VersionRange, VersionSpec,
+export UUID, SHA1, VersionRange, VersionSpec, Compat,
     PackageSpec, PackageEntry, EnvCache, Context, GitRepo, Context!, Manifest, Project, err_rep,
     PkgError, pkgerror, PkgPrecompileError,
     has_name, has_uuid, is_stdlib, is_or_was_stdlib, stdlib_version, is_unregistered_stdlib, stdlibs, write_env, write_env_usage, parse_toml,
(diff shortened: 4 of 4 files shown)


KristofferC:
Also cc @ChrisRackauckas who has thought about these stuff.

MasonProtter:
> Trying this on Plots.jl we get resolver errors (as @oxinabox predicted) due to Scratch.jl being lower bounded to 1.0.0 while one of our dependencies (RelocatableFolders has a lower bound of 1.1):

Hm, shouldn't there be a way to work around that though? Like if we flip the situation and look at `up`, if Plots.jl upper bounded Scratch.jl to `1.0.0`, and `RelocatableFolders` upper bounded it to `1.1`, then the resolver would just install it at version `1.0.0` instead of erroring right?

KristofferC:
> I might be missing something, but I think it should be possible for down to be fully symmetric with up.

Yes, but as I wrote in https://github.com/JuliaLang/Pkg.jl/issues/1062#issuecomment-1952096504, the whole reason to do this is to exactly test that your code works with the given lower bounds. So if you aren't hitting them exactly you are just testing some semi-random version configuration that have low versions "on average". But you cannot really say anything about the test result in that case.

MasonProtter:
Ah okay, sorry I missed that comment, so this is a (nicer implementation of) the idea in https://github.com/julia-actions/julia-downgrade-compat 

I'll say that I didn't really find downgrade-compat very useful because 

1) the complains from the resolver when it fails are indecipheralbe and very hard to turn into something actionable if you're not an expert

2) If the lower bounds of your package isn't actually installbe due to being un-resolvable, then I don't see why that's actually problematic. It's not a state a user would end up in and so it doesn't really need to be tested. What I *do* want to be able to test is the oldest set of *compatible* packages.

KristofferC:
>  What I do want to be able to test is the oldest set of compatible packages.

Okay, I think this can be extended to that. I will try...

MasonProtter:
> Edit: Thinking about this some more, I think it requires basically a full resolver since there is for example not a unique answer to this question of the "oldest set of compatible packages".

Yeah, I was imagining  the exact same behaviour as `up`, but instead of seeking newer versions it seeks older versions. It won't catch everything since the set of versions isn't unique, but that seems "good enough" for trying to keep track of appropriate lower bounds, and much less annoying than having the resolver simply error out because of a incompatible set of lower bounds.

MasonProtter:
Afterall, we also think it's okay that when I `]test` a package with a given set of upper bounds (either explicit or implicit), that we don't end up testing the unique newest set of packages, and it'd be really aggravating if the resolver threw an error just because it wasn't able to install the maximally newest compatible version of every dependency.

KristofferC:
> Yeah, I was imagining the exact same behaviour as up, but instead of seeking newer versions it seeks older versions.

Again, I am not sure that is useful in answering the question of if our lower bounds are correct. If you run the tests with this version of `down` you might have the following situation:

```toml
[compat]
A = "0.5.1"
B = "0.7.1"
```

and you get when testing `A=v0.5.1` and `B=v0.7.3`. However, you could in another environment end up with another pareto optimal, say `A=v0.5.3` and `B=v0.7.1` (the resolver is free to give any pareto optimal resolution). So the fact that we have multiple pareto optimal means that our testing will be insufficient to say that our lower bounds are correct.

KristofferC:
> Afterall, we also think it's okay that when I ]test a package with a given set of upper bounds (either explicit or implicit), that we don't end up testing the unique newest set of packages, 

I think this is different because when we are normally testing our package we are not explicitly testing that our upper bounds are correct because these are talking about an unknown future which can only be discussed from a "contract" p.o.v like semver.

MasonProtter:
That's fair. 

I guess though I feel that having a `]down` that acts like `]up` at least *seems* like an easier path forward to getting people (imperfectly) testing their lower bounds than having something very strict and fussy. 

If we do go with something as strict and fussy as pinning all the lower bounds, then I think nobody is going to use it unless we get much more informative and actionable error messages than we currently have, but making those error messages informative and actionable seems like a very difficult task.

KristofferC:
I'm not sure if the error messages are so bad. They will basically always be of the form that one of the direct dependencies limits another of the direct dependencies to something higher than the lower bounds of the package. For example in the Plots example the relevant stuff are:

```
ERROR: Unsatisfiable requirements detected for package Scratch [6c6a2e73]:
 Scratch [6c6a2e73] log:
 ├─possible versions are: 1.0.0 - 1.2.1 or uninstalled
 ├─restricted to versions 1.0.0 by an explicit requirement, leaving only versions: 1.0.0
 └─restricted by compatibility requirements with RelocatableFolders [05181044] to versions: 1.1.0 - 1.2.1 — no versions left
```

MasonProtter:
That said, maybe with this implementation it's not so critical that the error messages are good since the feedback loop is so much faster than with the `downgrade-ci`. Looking back, the main reason I gave up after trying to use it was that I'd need to go and semi-blindly adjust my bounds, run the ci process, get a new error, adjust them again, get a new error, etc.

The process here would be fundamentally the same, but in practice not so painful because I just need the `Project.toml` file open and a repl, and I can just tweak the bounds and continue hitting `]down` until I get something resolveable...



> I'm not sure if the error messages will be so bad. They will basically always be of the form that one of the direct dependencies limits another of the direct dependencies to something higher than the lower bounds that the package itself sets on it. For example in the Plots example the relevant stuff are:
> 
> ```
> ERROR: Unsatisfiable requirements detected for package Scratch [6c6a2e73]:
>  Scratch [6c6a2e73] log:
>  ├─possible versions are: 1.0.0 - 1.2.1 or uninstalled
>  ├─restricted to versions 1.0.0 by an explicit requirement, leaving only versions: 1.0.0
>  └─restricted by compatibility requirements with RelocatableFolders [05181044] to versions: 1.1.0 - 1.2.1 — no versions left
> ```

In my experience, I couldn't figure out what was going on. Sure, it's not so bad when direct dependancies start getting angry at eachother, but once you have nested indirect dependancies fighting, I get thoroughly confused. But as said above, maybe it's not so bad with this implementation.

MasonProtter:
You can set which mode it uses with a command line arg, but yeah I can't comment on the specific choices made there.

KristofferC:
Ok you expect this but why would it be *useful*?

carlobaldassi:
I guess that the name `down` will (rightfully) generate the expectation of an opposite operation to `up`. Maybe something like `pin_down`?

MasonProtter:
> That can't be true in general (unless I'm misunderstanding precisely what you mean). E.g.,
> 
> ```
> [compat]
> PkgA = 1.2.3
> ```
> 
> Means I'm compatible with `v1.2.4`, `v1.2.5`, ..., `v1.3.1`, ..., but I'm incompatible with `v1.2.2` or anything lower. So, it's not symmetric, right?

Where is the asymmetry you're pointing to? There's only a finite number of versions of any package in the general registry, so you can't end up with `v1.2.∞`. That means there is some `N` where the versions are bounded between `v1.2.3` and `v1.2.N`. 

Kristoffer is right in that there's an asymmetry induced by the fact that new features can only appear in one direction, so it's good to have a function that demands the absolute oldest version set, whereas it makes less sense to demand the absolute newest version set, but even then I have some doubts.

You could say that it's important that you get the absolute newest version set if you want to test for potentially breaking changes in your dependencies. But we're okay with testing on just a reasonably new set of compatible packages, so I'd also be okay with testing on a reasonably old set.

KristofferC:
> Having a less fussy ]down that gives a reasonable set of old versions I can test on without having to comb through and meticulously purge my Project.toml of incompatible lower bounds would give *most of the testing utility*..

I disagree, you would have failed to verify your lower bounds so even if your tests pass your lower bounds can still be buggy and give runtime errors for users (which is what you wanted to avoid in the first place).

oxinabox:
Yes, but perfect can be the enemy of good.

There are also other positions in the compat interval not tested.
Like if you declare compat for `1,2,3` but actually v2 removed a feature that v1 had and that v3 brought back.
Testing both top and bottom isn't going to be enough.

Testing (potential one of multiple) lowest possible version that can be achieved in practice -- that a user might actually see,
seems like a good point in the trade-off space of time spent chasing dependencies (which becomes zero as it isn't required to transcript stuff from your dependencies lower bounded lower bounds wherever higher than your own) vs  coverage of potential mistakes and problems.

adrhill:
> Again, I am not sure that is useful in answering the question of if our lower bounds are correct.

I understand the argument that ideally, we should test every point of the "Pareto front" of resolvable lower bounds. But realistically speaking, this front can be very large and very expensive to test.

As far as I can tell, `julia-downgrade-compat` **only** works if the exact lower bounds specified in the compat entries are resolvable. This is equivalent to the "Pareto front" being a singular point.

From the perspective of a package developer with multiple dependencies (and possibly hundreds of indirect dependencies), this means that `julia-downgrade-compat` doesn't work out of the box. Developers must first manually set a resolvable set of lower compat entries. This manual resolving must be done by developers every time a lower bound of a compat entry is updated.

I therefore have to agree with @oxinabox on "perfect can be the enemy of good". 
Personally, I would be happy with either of these options:
1. Have `down` resolve an arbitrary lower bound environment in the same way that `up` resolves a somewhat arbitrary upper bound environment. Being able to test a single point on the "Pareto front" is already better than not testing any point at all.
2. Have `down` resolve an arbitrary lower bound environment and update the lower-bound compat entries such that the "Pareto front" is reduced to a single point and can be fully tested in CI. This is most likely overly restrictive, but it is what package developers currently have to do manually.

adrhill:
> It works, you just have to bump the lower bounds so they agree with your dependencies so you can actually run your tests with the lower bounds you claim to support.

In theory you are right, but in practice, this manual bisection can take hours – [I've tried](https://github.com/gdalle/DifferentiationInterface.jl/pull/380) and [given up](https://github.com/gdalle/DifferentiationInterface.jl/issues/378#issuecomment-2262864492) before. 
And this soul-crushing task must be repeated every time one of the compat entries has to be changed, e.g. due to a new breaking release in one of the dependencies.

KristofferC:
It could maybe be automated somehow?

Issue #3066, opened 2022-04-22
"No Changes to `~/PkgA/Manifest.toml`" is misleading since there can be changes


ericphanson:
This is a "known" issue but I couldn't find it filed anywhere. If `resolve` fixes the dependencies of a dev'd package in a way that doesn't add new dependencies to the overall Manifest, then "no changes" are reported, even if the Manifest is indeed changed.

```julia
❯ julia -q
(@v1.7) pkg> generate PkgA
  Generating  project PkgA:
    PkgA/Project.toml
    PkgA/src/PkgA.jl

(@v1.7) pkg> generate PkgB
  Generating  project PkgB:
    PkgB/Project.toml
    PkgB/src/PkgB.jl

(@v1.7) pkg> activate PkgB
  Activating project at `~/PkgB`

(PkgB) pkg> add Dates
    Updating registry at `~/.julia/registries/General.toml`
   Resolving package versions...
    Updating `~/PkgB/Project.toml`
  [ade2ca70] + Dates
    Updating `~/PkgB/Manifest.toml`
  [ade2ca70] + Dates
  [de0858da] + Printf
  [4ec0a83e] + Unicode
Precompiling project...
  1 dependency successfully precompiled in 1 seconds

(PkgB) pkg> activate PkgA
  Activating project at `~/PkgA`

(PkgA) pkg> dev ./PkgB
   Resolving package versions...
    Updating `~/PkgA/Project.toml`
  [1821aa1f] + PkgB v0.1.0 `../PkgB`
    Updating `~/PkgA/Manifest.toml`
  [1821aa1f] + PkgB v0.1.0 `../PkgB`
  [ade2ca70] + Dates
  [de0858da] + Printf
  [4ec0a83e] + Unicode
```

Ok, now we have PkgA which dev's PkgB, which depends on Dates. All is well. Now edit PkgB to do `using Unicode`. Then
```julia
> julia -q --project=PkgA
julia> using PkgB
[ Info: Precompiling PkgB [1821aa1f-13ce-41b2-bd12-ba9ec74d35d8]
┌ Warning: Package PkgB does not have Unicode in its dependencies:
│ - If you have PkgB checked out for development and have
│   added Unicode as a dependency but haven't updated your primary
│   environment's manifest file, try `Pkg.resolve()`.
│ - Otherwise you may need to report an issue with PkgB
└ Loading Unicode into PkgB from project dependency, future warnings for PkgB are suppressed.
```
[...]

PR #3806, opened 2024-02-23
Tweak wording to make "no changes" message more accurate

ericphanson:
fixes https://github.com/JuliaLang/Pkg.jl/issues/3066

(cc @haberdashPI)

edit: I'll need to update the tests

### DIFF
--- a/docs/src/environments.md
+++ b/docs/src/environments.md
@@ -70,8 +70,8 @@ Cloning into 'Example.jl'...
 Activating project at `~/Example.jl`
 
 (Example) pkg> instantiate
-  No added or removed packages to `~/Example.jl/Project.toml`
-  No added or removed packages to `~/Example.jl/Manifest.toml`
+  No packages added to or removed from `~/Example.jl/Project.toml`
+  No packages added to or removed from `~/Example.jl/Manifest.toml`
 ```
 
 If the project contains a manifest, this will install the packages in the same state that is given by that manifest.
--- a/docs/src/environments.md
+++ b/docs/src/environments.md
@@ -66,7 +66,7 @@ shell> git clone https://github.com/JuliaLang/Example.jl.git
 Cloning into 'Example.jl'...
 ...
 
-(@v1.9) pkg> activate Example.jl
+(@v1.12) pkg> activate Example.jl
 Activating project at `~/Example.jl`
 
 (Example) pkg> instantiate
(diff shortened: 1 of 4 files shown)


IanButterworth:
> No packages added to or removed from ..

Changes could be a lot more things than just added or removed i.e. version changes, path changes, pin status etc.

I had proposed

> No changes to the dependencies in /path/to/Manifest.toml but metadata was updated

If we don't want to mention the metadata part because it's hard to track and know whether it's changed could just say 

> No changes to the dependencies in /path/to/Manifest.toml

haberdashPI:
> No changes to the dependencies in /path/to/Manifest.toml

This is quite misleading as a user with no deep knowledge of Pkg internals (and the metadata part only adds a bit to my confusion). I think it has the same problems as the extant message. 

I think the main problem with the current wording is that as a user "change" can be understood to mean any change to the file on disk regarding a dependency. But e.g. `resolve` might change whether a dev'ed package can be compiled successfully, which clearly means there was a change to one of the dependencies in Manifest.toml (right??), even though this message will show up in that case.

What about:

> No dependencies needed to be installed or uninstalled due to changes in Project.toml/Manifest.toml

This makes it clear that there were indeed changes to the toml files.

ericphanson:
> Changes could be a lot more things than just added or removed i.e. version changes, path changes, pin status etc.

So this message prints when `all(p-> p[2] == p[3], xs)` holds, which is some statement about the difference in env caches. I think you are saying roughly: "if there is no difference in env caches, we know more than no packages were added or removed! we also know you didn't pin something, or change a version, or change a path!" And therefore the message should reflect this stronger notion of lack of change.

My counter is: the current message is basically wrong (there are, in fact, changes to the manifest/project in many cases). The new message I propose here is actually 100% correct, it is just not as strong as it possibly could be; we could be making even bigger claims about not only packages not changing but also pin status not changing etc.

So the question is, how much does making bigger claims here help the user? Well, Pkg helpfully prints most changes/updates/etc, so this message is mostly in lieu of printing *nothing*, so the user knows the command actually ran. So what we need to tell the user is: "we ran your command and there's nothing worth reporting about specifically". I don't think that needs the strongest possible wording. I think that just shouldn't *overstate* the case.

Therefore, in my opinion, the current simple wording I propose is an improvement over the status quo (saying false things -> saying true things), and that a more nuanced printing about exactly how little has changed isn't super useful.

haberdashPI:
I'm good with your wording @ericphanson, but I could also imagine:

> No dependencies were installed or uninstalled

ericphanson:
Take another look @IanButterworth ? Seems like this bothers others too, #4031

IanButterworth:
Seems well reasoned and worth giving a go

PR #3807, opened 2024-02-23
Append bundled depots to places where we empty it during tests

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3969

I am thinking we might be precompiling stdlibs over and over since we won't find the precompile files for those if we do not have the bundled depots in the depot path.

If this is useful this should have a bit smaller CI runtime.

### DIFF
--- a/test/artifacts.jl
+++ b/test/artifacts.jl
@@ -652,6 +652,7 @@ end
         old_depot_path = copy(DEPOT_PATH)
         empty!(DEPOT_PATH)
         append!(DEPOT_PATH, [depot1, depot2, depot3])
+        Base.append_bundled_depot_path!(DEPOT_PATH)
 
         # First sanity check; does our depot path searching code actually work properly?
         @test startswith(artifact_path(foo_hash), depot3)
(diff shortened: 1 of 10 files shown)


KristofferC:
I don't think this matters since we turn off auto precompile for the tests.

IanButterworth:
Trying a fix. I think that test was installing the registry into the shared depot. First time this was seen is after that test
```
┌ Info: /Users/runner/hostedtoolcache/julia/nightly/aarch64/share/julia
│   readdir(d) =
│    8-element Vector{String}:
│     "base"
│     "base.cache"
│     "cert.pem"
│     "compiled"
│     "julia-config.jl"
│     "registries"
│     "stdlib"
└     "test"
```

IanButterworth:
Depends on https://github.com/JuliaLang/Pkg.jl/pull/3972

IanButterworth:
Not sure why this one is still happening. We appear to be adding the correct Pkg there
```
Precompiling ArtifactOverrideLoading...
  67218.6 ms  ✓ Pkg
  1 dependency successfully precompiled in 67 seconds. 24 already precompiled.
  1 dependency precompiled but a different version is currently loaded. Restart julia to access the new version
```

IanButterworth:
27m down to 15m on MacOS CI. Nice

KristofferC:
Nice indeed!

Issue #3808, opened 2024-02-24
Make spinners spin more slowly


eschnett:
I was updating Julia packages over an unstable internet connection and I didn't like how fast the spinners were running. I think they spin about twice per second.

I looked at the Pkg.jl code, and I found that setting the environment variable `CI=true` would disable all progress output. That's good, but I was looking for something fancier – just slowing down the progress output, not eliminating it.

I think it would be possible to make the spinner speed configurable with a rather small code change. I think this line https://github.com/JuliaLang/Pkg.jl/blob/e7e8ce38359330441b1340046add367761035f69/src/API.jl#L1389 defines how fast the spinners are.

One could add a new global variable
```Julia
const PRECOMPILE_SPINNER_INTERVAL = Ref(0.1)
```
to the code, and then create the Timer via `Timer(0; interval=PRECOMPILE_SPINNER_INTERVAL[])`. This would allow changing the spinners by setting a global variable. A similar mechanism is already used in `MiniProgressBars.jl` where two variables are defined https://github.com/JuliaLang/Pkg.jl/blob/e7e8ce38359330441b1340046add367761035f69/src/MiniProgressBars.jl#L21-L22

KristofferC:
How does the precompile spinner relate to your internet connection?

eschnett:
I was logged in via a slow connection and issued an `]update` command from the Julia REPL, which lead to precompiling a few package that take a minute to do so (CairoMakie?) while showing the spinners. The spinners didn't spin smoothly.

PR #3809, opened 2024-02-24
New variable PRECOMPILE_SPINNER_INTERVAL

eschnett:
Closes https://github.com/JuliaLang/Pkg.jl/issues/3808


IanButterworth:
If still desired, this needs to be reopened against the julia repo as this code has moved.

PR #3811, opened 2024-02-24
avoid registry warning during precompile

IanButterworth:
Avoids 
```
    JULIA stdlib/REPLExt.release.image
    JULIA stdlib/LazyArtifacts.release.image
    JULIA stdlib/LazyArtifacts.release.image
 │ Attempted to find missing packages in package registries but no registries are installed.
 └ Use package mode to install a registry. `pkg> registry add` will install the default registries.

    JULIA stdlib/REPLExt.release.image
 │ Attempted to find missing packages in package registries but no registries are installed.
 └ Use package mode to install a registry. `pkg> registry add` will install the default registries.

```

Pkg.jl precompile.jl does this, which is why it didn't show before, I believe.

I guess it means the registry stuff isn't baked in properly.. but that would need a better fix i.e. https://github.com/JuliaLang/Pkg.jl/pull/2833

### DIFF
--- a/ext/REPLExt/precompile.jl
+++ b/ext/REPLExt/precompile.jl
@@ -10,14 +10,24 @@ let
     REPL.raw!(::FakeTerminal, raw::Bool) = raw
 
     function pkgreplmode_precompile()
+        original_depot_path = copy(DEPOT_PATH)
+        original_load_path = copy(LOAD_PATH)
         __init__()
         Pkg.UPDATED_REGISTRY_THIS_SESSION[] = true
-        try_prompt_pkg_add(Symbol[:notapackage])
-        promptf()
-        term = FakeTerminal()
-        repl = REPL.LineEditREPL(term, true)
-        REPL.run_repl(repl)
-        repl_init(repl)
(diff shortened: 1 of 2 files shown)


KristofferC:
> I guess it means the registry stuff isn't baked in properly.

What does this mean?

IanButterworth:
Actually this isn't enough.. I checked with a julia build and still see the messages.

> > I guess it means the registry stuff isn't baked in properly.
> What does this mean?

I mean that if the Pkg precompile process is being told not to update the registry, then the registry update code paths aren't being hit during precompilation

IanButterworth:
This looks clean in a julia build now
```
    JULIA stdlib/Pkg.release.image
    JULIA stdlib/Pkg.release.image
    JULIA stdlib/REPLExt.release.image
    JULIA stdlib/LazyArtifacts.release.image
    JULIA stdlib/LazyArtifacts.release.image
    JULIA stdlib/REPLExt.release.image
ian@Ians-MBP julia %
```

IanButterworth:
@KristofferC this look ok to you?

KristofferC:
👍

PR #3792, opened 2024-02-13
do not start a new process for precompiling the test env

KristofferC:
Requires https://github.com/JuliaLang/julia/pull/53316

PR #3813, opened 2024-02-26
Revert "do not start a new process for precompiling the test env"

KristofferC:
Reverts JuliaLang/Pkg.jl#3792

See if this restores CI on master to green...


KristofferC:
Actually, I think this is https://github.com/JuliaLang/julia/pull/53439#issuecomment-1962174524

PR #3820, opened 2024-02-29
move to using Base parallel precompile

KristofferC:
Can be merged after https://github.com/JuliaLang/julia/pull/53403.

cc @IanButterworth


src/API.jl
```diff
@@ -1099,703 +1098,16 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
-                    end
-                end
-            end
-        catch err
-            err isa InterruptException || rethrow()
-        end
+    activate(dirname(ctx.env.project_file)) do
+        pkgs_name = String[pkg.name for pkg in pkgs]
+        return Base.PrecompilePkgs.precompilepkgs(pkgs_name; internal_call, strict, warn_loaded, timing, _from_loading, flags_cacheflags, io=ctx.io)
     end
```
IanButterworth, line 1109: Why is the activate needed? Pkg.precompile calls `instantiate` before this so if this is needed it needs to happen earlier?

KristofferC: I think this was an issue before because the Base loading code needs to have the packages getting precompiled in the load path to look stuff up but if you create a Context with some random environment path, that won't be the case (unless we activate it here)

IanButterworth (approved):

PR #3821, opened 2024-03-01
Activate special environments similar to Julia's CLI arg `--project`

ven-k:
Similar to Julia's command line argument `--project`,
- `activate @temp` activates temp environment
(See https://github.com/JuliaLang/julia/pull/51149)
- `activate @stdlib` activates `Sys.STDLIB`
- `activate @.` activates current project
- `activate @` activates active project

This also changes how Pkg.activate behaves wrt `temp` (disallows) and `stdlib` (activates `@stdlib/Project.toml`)

Related to these, the PR adds tests and updates docs.

### DIFF
--- a/docs/src/environments.md
+++ b/docs/src/environments.md
@@ -91,6 +91,8 @@ In short, `instantiate` is your friend to make sure an environment is ready to u
     ```bash
     $ julia --project=. myscript.jl
     ```
+    For Julia 1.7 or later, `--project=@stdlib` activates stdlib environment.
+    For Julia 1.11 or later, `--project=@temp` activates temporary environment.
 
 
 ## Temporary environments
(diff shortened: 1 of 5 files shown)


ven-k:
While here, I wonder if [this](https://github.com/JuliaLang/julia/blob/0918cf12d549a53da02265ec1005aec7fce87e5f/base/initdefs.jl#L269-L303) part of `load_path_expand` should be separated out as an internal function which Pkg can then use in `parse_activate `and `activate `(for `shared = <>` dispatch)

IanButterworth:
I don't understand the need for this as we have `activate --temp`, `activate .` etc.

Please feel free to clarify/disagree if I'm being shortsighted.

PR #3826, opened 2024-03-05
WIP: Add entries to `[sources]` automatically if package is added by URL or devved

KristofferC:
Some CI errors here and I probably won't get to that for a while.


KristofferC:
Subsumed by https://github.com/JuliaLang/Pkg.jl/pull/4225

PR #3829, opened 2024-03-06
load the set of stdlibs during precompilation

KristofferC:
Otherwise Pkg becomes non relocatable. I remember doing this before but not sure why it didn't happen.


IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/2916

PR #3837, opened 2024-03-09
remove line about changing UUID to dev in readme

IanButterworth:


DilumAluthge:
Why is this no longer necessary? Because Pkg is not in the default sysimage?

PR #3841, opened 2024-03-12
Add workspace feature

KristofferC:
This PR requires https://github.com/JuliaLang/julia/pull/53653.

This allows one to define a subproject which is an incremental addition to the dependency graph of the "base project". When the sub project is active it can load any dependency from the base project and the sub project itself. When Pkg resolves, it resolves among the base project and all sub projects and the full dependency graph is stored in a single manifest next to the base project's manifest.

Subprojects are defined by specifying a list of folders in the project file as:

```
subprojects = ["test", "benchmarks"]
```

Right now, the base project has to be one level above any sub project.

TODO: 

- [x] More thorough tests especially regarding merging of compats and sources among projects.
- [x] Docs
- [x] Hook up Pkg.status output
- [x] Test nested subprojects
- [x] Show Pkg prompt as `Pkg/test>` if `test` is a subproject to Pkg.

As a follow-up to this I want to make it so that the test subproject is special and that running `Pkg.test` if this is active should be equivalent to:

```
Pkg.activate("test")
Pkg.resolve()
include("test/runtests.jl")
```

Other follow up is to perhaps enhance `Pkg.status()` to show more clearly what packages comes from the base project and what comes from the sub project.


test/subprojects.jl
```diff
@@ -0,0 +1,60 @@
+module SubProjectsTest
+
+import ..Pkg # ensure we are using the correct Pkg
```
KristofferC, line 3: Should use a fresh depot in these tests.

KristofferC:
It would be nice if `dev Foo` in a subproject would just work in case it is already devved in the base project's manifest.

jakobnissen:
Some feedback:
* In a project where the main project has Automa added with compat "1", and the sub-project has Automa added with compat "=1.0.2" (not the newest version), when showing `(mypkg) pkg> status`, it falsely displays that Automa can be upgraded:
```
⌃ [67c07d97] Automa v1.0.2
```
This also occurs the other way around: If the parent project has a restrictive dependency but the subproject does not, the subproject's version is held back, but the green `^` still appears in the subproject's `status`, suggesting the package can be updated.

* When the subproject depends on `BenchmarkTools` but the parent does not, `BenchmarkTools` shows up in `pkg> status`. However, it's not part of the project, nor can it be loaded. This might be a bug

KristofferC:
@jakobnissen, both your issues should be solved now. There is also an update on the julia side for the corresponding PR.

KristofferC:
Should check that `manifest = ` works when specified in the root base project.

oxinabox:
To confirm I am understanding correctly.
the core feature that the resolver would *always* be restricted to using packages from parent/siblings/nephew/etc packages in the mono-repo (so you couldn't accidentally get an independent version of `Foo.jl` within `Foo/Bar.jl`).
Which is advantage over putting stuff like relative paths in the Manifest, or fiddling with LOAD_PATH.

KristofferC:
> Which is advantage over putting stuff like relative paths in the Manifest, or fiddling with LOAD_PATH.

I'll say some things that would work with this:

### No need for TestEnv.jl

Instead, make `test/Project.toml` a subproject of the package, and add the test-specific deps to the test-project. `Pkg.test` is now just `julia --project=test runtests.jl`, no need for a temporary manifest etc.
Also, the `[targets]` is now not special for `test`, you can do the same thing for `docs` or `benchmarks`. 

### Relative paths in Manifest.toml.

I'm not sure what this means. Sure, you can dev things with your "root" project active but as soon as you activate one of the packages in the monorepo and do a resolve you are going to have to dev everything unregistered there again.
These might have different versions etc then what you have in your root project.


### LOAD_PATH pushing

Packages are not resolved together in the different environments in the load path so you might load incompatible versions here (if your root project has a required dependency version that is different from what that same dependency resolved to in the other environment). You also need to mutate this global state at the top of the script file if you want to mutate `LOAD_PATH` which is a bit inelegant.

oxinabox:
To be clear my comment was in agreement that those were good things, for the reasons you listed.
My *question* was: am i correct in understanding that the way subprojects work is that it makes sure  parent/siblings/nephew/etc packages in the mono-repo are always what is resolved to

KristofferC:
> am i correct in understanding that the way subprojects work is that it makes sure parent/siblings/nephew/etc packages in the mono-repo are always what is resolved to

Every project in the "workspace" will resolve together and generate a single manifest, yes.

KristofferC:
- [ ] Should write packages in workspace to manifest
- [ ] Also write extensions for package to manifest

KristofferC:
The diff output for the manifest is weird when removing the base package:

```
(test) pkg> rm Pkg
  No Changes to `~/JuliaPkgs/Pkg.jl/test/Project.toml`
    Updating `~/JuliaPkgs/Pkg.jl/Manifest.toml`
  [0dad84c5] - ArgTools v1.1.2
  [56f22d72] - Artifacts v1.11.0
  [f43a241f] - Downloads v1.6.0
  [7b1f6079] - FileWatching v1.11.0
  [b27032c2] - LibCURL v0.6.4
  [76f85450] - LibGit2 v1.11.0
  [8f399da3] - Libdl v1.11.0
  [ca575930] - NetworkOptions v1.2.0
  [44cfe95a] - Pkg v1.12.0 `.`
  [a4e569a6] - Tar v1.10.0
  [cf7118a7] - UUIDs v1.11.0
  [deac9b47] - LibCURL_jll v8.6.0+0
  [e37daf67] - LibGit2_jll v1.7.2+0
  [29816b5a] - LibSSH2_jll v1.11.0+1
  [c8ffd9c3] - MbedTLS_jll v2.28.6+0
  [14a3606d] - MozillaCACerts_jll v2023.12.12
  [83775a58] - Zlib_jll v1.2.13+1
  [8e850ede] - nghttp2_jll v1.59.0+0
  [3f19e933] - p7zip_jll v17.4.0+2

(test) pkg> st -m --workspace
Status `~/JuliaPkgs/Pkg.jl/Manifest.toml`
...
  [8f399da3] Libdl v1.11.0
  [56ddb016] Logging v1.11.0
  [d6f4376e] Markdown v1.11.0
  [a63ad114] Mmap v1.11.0
  [ca575930] NetworkOptions v1.2.0
  [44cfe95a] Pkg v1.12.0 `.`
  [de0858da] Printf v1.11.0
  [3fa0cd96] REPL v1.11.0
  [9a3f8284] Random v1.11.0
```

Note that the packages shown as removed are still in the manifest.

KristofferC:
```
(test) pkg> dev .
   Resolving package versions...
  No Changes to `~/JuliaPkgs/Pkg.jl/test/Project.toml`
    Updating `~/JuliaPkgs/Pkg.jl/Manifest.toml`
  [0dad84c5] + ArgTools v1.1.2
  [56f22d72] + Artifacts v1.11.0
  [f43a241f] + Downloads v1.6.0
  [7b1f6079] + FileWatching v1.11.0
  [b27032c2] + LibCURL v0.6.4
  [76f85450] + LibGit2 v1.11.0
```

Missing showing that Pkg got added to the project file for some reason.

KristofferC:
```
(MonoRepo/PrivatePackage/test) pkg> st
Status `~/MonoRepo/PrivatePackage/test/Project.toml`
  [a23bb17a] PrivatePackage v0.1.0 `PrivatePackage` # this is now relative to manifest
  [8dfed614] Test v1.11.0
```

should print paths relative to project when doing a `st` and it is showing direct deps

KristofferC:
I'm seeing some (I think) spurious warnings like:

```
Warning The project dependencies or compat requirements have changed since the manifest was last resolved. It is recommended to `Pkg.resolve()` or consider `Pkg.update()` if necessary.
```

lassepe:
I'm a bit confused about the workflow here, in particular in the context of `test/` being a subproject:

From the discussion above, I understand that one has to specify `test/` as a subproject in the parent `Project.toml`. Would that mean that, when instantiating the parent, it gets resolved together with all the test dependencies (even when not testing). If that is true, couldn't it lead to confusing downgrades of packages due to test dependencies?

KristofferC:
> If that is true, couldn't it lead to confusing downgrades of packages due to test dependencies?

I don't know if they are confusing or not. But being able to run the tests with the same versions as you actually run your code seems like a feature to me. Of course, you are not forced to put `test` into the workspace.

lassepe:
I fully agree that having `julia --project=test/ runtests.jl` "just work" would be amazing. I am just a little bit surprised that the "parent project" would have to know about the subprojects at all. I would have thought that the `test/Project.toml` may include a pointer to the parent (maybe via the new `[source]` feature) to make `julia --project=test/ runtests.jl` work without affecting `julia --project .`

That said, the fact that the proposed design has worked well for Rust is probably a good indicator that the idea is sound and I don't want to add noise to this discussion. So please feel free to ignore my feedback if you don't share these concerns.

KristofferC:
> I fully agree that having julia --project=test/ runtests.jl "just work" would be amazing. I am just a little bit surprised that the "parent project" would have to know about the subprojects at all. 

There is another possible (somewhat orthogonal) feature that could be explored (which could probably use the name "subproject") which is a project whose dependencies are the union of those listed in its own project and that of its "parent" (which is specified in the project file of the subproject).

I am not fully clear exactly what the rules should be for resolving etc for that.

docs/src/toml-files.md
```diff
@@ -128,6 +128,24 @@ constraints in detail. It is also possible to list constraints on `julia` itself
...
+A workspace is defined in the base project by giving a list of the projects in it:
+
+```toml
+[workspace]
+projects = ["test", "docs", "benchmarks", "PrivatePackage"]
+```
+
+This structure is particularly beneficial for developers using a monorepo approach, where a large number of unregistered packages may be involved. It's also useful for adding documentation or benchmarks to a package by including additional dependencies beyond those of the package itself.
+
+Workspace can be nested, that ism a project that itself defines a workspace can also be part of another workspace.
```
Datseris, line 147: Thanks for this wonderful feature, just pointing out the typo `ism` above!

PR #3847, opened 2024-03-17
Add example of manually checking specifications

MilesCranmer:
While the Compatibility guide is very helpful, I find the syntax itself easy to forget, and find myself repeatedly checking this guide whenever I'm creating a complex version specification.

I recently found the `Pkg.Versions.semver_spec` as a way to manually parse the version string. I thought it would be helpful to give an example of how to use this to explicitly check versions, and make sure you aren't too constrained or too flexible in a given version spec.

### DIFF
--- a/docs/src/compatibility.md
+++ b/docs/src/compatibility.md
@@ -163,6 +163,28 @@ PkgA = "0.2 - 0.5"     # 0.2.0 - 0.5.* = [0.2.0, 0.6.0)
 PkgA = "0.2 - 0"       # 0.2.0 - 0.*.* = [0.2.0, 1.0.0)
 ```
 
+### Checking specification
+
+Pkg.jl parses a given version specification using `Pkg.Versions.semver_spec`.
+You can check if a particular version of a package is contained in a particular
+range by using this command. For example:
+
+```julia
+julia> v"0.1.0" in Pkg.Versions.semver_spec("=0.1")
+true
+
+julia> v"0.1.0" in Pkg.Versions.semver_spec("=0.1.1")
+false
+
+julia> v"0.1.0" in Pkg.Versions.semver_spec("0.1 - 0.2")
+true
+
+julia> v"0.3.0" in Pkg.Versions.semver_spec("~0.3.2")
+false
+
+julia> v"0.3.3" in Pkg.Versions.semver_spec("0.1 - 0.2, ~0.3.2")
+true
+```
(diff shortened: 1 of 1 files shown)


KristofferC:
I think this is a good idea but we should imo:

- Give it a docstring. 
- Make it public.
- Give it a better name.

MilesCranmer:
That all sounds good to me. 

For naming, maybe just something like `parse_compat`?

DilumAluthge:
Does this mean that we have to make `Pkg.Versions.VersionSpec` part of the public API? Because `semver_spec` returns a `VersionSpec`?

MilesCranmer:
Ping on this, I ran into this again and would have found it useful to have this in the docs

MilesCranmer:
Is `parse_compat` a good name? I can add it if that sounds good.

IanButterworth:
`parse_compat` sounds good to me 👍

DilumAluthge:
Just bumping my comment here: https://github.com/JuliaLang/Pkg.jl/pull/3847#issuecomment-2002662230

> Does this mean that we have to make `Pkg.Versions.VersionSpec` part of the public API? Because `semver_spec` returns a `VersionSpec`?

---

It seems like all of the use cases here are for calling `x in semver_spec("foo")`, right? Not for using the return value of `semver_spec("foo")` directly? So how about we keep `semver_spec("foo")` private, and we have a new public function `satisfies_compat_entry(x, "foo") = x in semver_spec("foo")`? This allows us to avoid needing to make `VersionSpec` part of the public API.

DilumAluthge:
We can pick whatever name we want, I just gave `satisfies_compat_entry()` as an example.

MilesCranmer:
@DilumAluthge who was your question addressed to? Just want to make sure I didn't miss it

DilumAluthge:
Just the group as a whole.

DilumAluthge:
> It seems like all of the use cases here are for calling `x in semver_spec("foo")`, right? Not for using the return value of `semver_spec("foo")` directly? So how about we keep `semver_spec("foo")` private, and we have a new public function `satisfies_compat_entry(x, "foo") = x in semver_spec("foo")`? This allows us to avoid needing to make `VersionSpec` part of the public API.

Implemented in #4292.

IanButterworth:
#4292 is on the way to merge so we can close this.

PR #3848, opened 2024-03-18
RFC: only auto precompile the added package (+ deps) after an `add` operation

KristofferC:
Many times in the global env I add some small package to try it out and then Pkg starts precompiling everything (even things I did not intend to load). This means I often have to interrupt it (which is a bit brittle) and it feels like it is doing something I didn't ask it to do.

With this, it will only compile the deps graph of the added packages. If a user want to load some other package later, the load precompile step will pick that up.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1424,7 +1424,7 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}, new_git=Set{UUID}();
         write_env(ctx.env) # write env before building
         show_update(ctx.env, ctx.registries; io=ctx.io)
         build_versions(ctx, union(new_apply, new_git))
-        Pkg._auto_precompile(ctx)
+        Pkg._auto_precompile(ctx, pkgs)
     else
         record_project_hash(ctx.env)
         write_env(ctx.env)


IanButterworth:
The current approach aims at having the only thing that should cause load-time precompilation being changes to dev-ed packages. This moves more precompilation back to load time, when deps of other packages are changed by an add, which IMO would be a UX regression. But perhaps this is popular?

Personally I prefer the current arrangement and using the tiered_installed resolve strategy. Less changes, precompilation stays at install time.

KristofferC:
> The current approach aims at having the only thing that should cause load-time precompilation being changes to dev-ed packages.

Yeah, of course, if you precompile the world once you don't have to precompile it later.

> which IMO would be a UX regression

Well, it is not a straight-up regression, it is a trade off. It is a big win in the cases where you do `add Foo; using Foo; Foo.do_what_I_want()` in that that takes 10s, and not 10 minute because I happened to have OrdinaryDiffEq also in the env (which I wasn't going to use).

IanButterworth:
I think we might have opposite opinions on this. Might be worth fielding opinions on slack?

IanButterworth:
A fancy solution would be to be specific like this PR, but spawn off the rest of the Project work as a background thing afterwards. 

The UI of that needs figuring out, but it might not be too tricky..
- notify that `the rest of the project is being precompiled in the background. Use pkg> precompile --stop to interrupt`
- always be silent, don't report success or errors

Pidfile locking should protect against races and we already report if the current process is holding the lock in an async task.

KristofferC:
Right now, there is no way to turn off full precompilation of env on `add` while keeping precompilation of test env. :(

KristofferC:
I must say this is sooo annoying. 

```
(@v1.12) pkg> add Cthulhu
   Resolving package versions...
   Installed AbstractTrees ── v0.4.5
   Installed TypedSyntax ──── v1.3.1
   Installed WidthLimitedIO ─ v1.0.1
   Installed FoldingTrees ─── v1.2.1
   Installed JuliaSyntax ──── v0.4.9
   Installed Cthulhu ──────── v2.12.7
    Updating `~/.julia/environments/v1.12/Project.toml`
  [f68482b8] + Cthulhu v2.12.7
    Updating `~/.julia/environments/v1.12/Manifest.toml`
  [1520ce14] + AbstractTrees v0.4.5
  [f68482b8] + Cthulhu v2.12.7
  [1eca21be] + FoldingTrees v1.2.1
  [70703baa] + JuliaSyntax v0.4.9
  [d265eb64] + TypedSyntax v1.3.1
  [b8c1c048] + WidthLimitedIO v1.0.1
Precompiling all packages...
  Progress [==================================>      ]  237/287
```

Argh!! All the time.

IanButterworth:
IMO it's an issue for people that work on master. We could turn it off for non tagged builds?

KristofferC:
If `add Foo` precompiles everything why doesn't `dev ./Foo` precompile anything? Why would I want to be more likely to want to precompile `IrrelevantPackage.jl` if I do an add on some other package vs a dev?

IanButterworth:
Why does everything else need re-precompiling? Because you're changing julia version a lot, no?

IanButterworth:
Otherwise it's handled at install time for those packages.

IanButterworth:
And IMO we should https://github.com/JuliaLang/Pkg.jl/issues/3905 to help everyone avoid re-precompilation they don't really need

IanButterworth:
Perhaps this could be merged behind a switch?

One reason I don't like it is that it dislocates cause and effect for precompilation activity. Like, add a package, updates a dep that is shared by another top level package. With this change the user will hit precompilation at a later stage and not know why, whereas currently they can see that a dep changed. For normal non package dev users, it reverts to the olden days of Pkg not really fully finishing the installation.

KristofferC:
How about this:

- Keep the same behavior for package projects
- Do this behavior for non-package projects (like the default environment).

IanButterworth:
I see your suggestion and raise you:

- Keep the same behavior for package projects
- Do this behavior for non-package projects (like the default environment), but if the manifest version has changed prompt to ask if the full env should be precompiled

KristofferC:
Just as another data point, I've still never found the current behavior desirable, and in combination with Ctrl+C not working well, it causes many fists to hit the table.

IanButterworth (approved):

IanButterworth:
Might need docs change?

KristofferC:
Which docs need changing specifically?

IanButterworth:
I was asking. If you didn't find any mention of the previous behavior then we're good.

PR #3849, opened 2024-03-20
small fixes in GitTools module

lklein-btig:
Making sure the progress bar for cloning actually states that its cloning, rather than fetching.
Removing unused credentials in fetch function.


src/GitTools.jl
```diff
@@ -159,9 +159,6 @@ function fetch(io::IO, repo::LibGit2.GitRepo, remoteurl=nothing; header=nothing,
         LibGit2.Callbacks()
     end
     fancyprint && start_progress(io, bar)
-    if credentials === nothing
```
KristofferC, line 162: Shouldn't this be used instead of being removed?

lklein-btig: I was unsure which direction we want to go with this.
I went with removing it as that seemed more conservative to me, but I am open to using it.
I will add a commit to use it.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/4286

PR #3850, opened 2024-03-20
fix how entry point to package is computed with `path` provided in project file

KristofferC:


Keno:
Probably could use a test, but as the bug-reporter, LGTM ;)

KristofferC:
Requires https://github.com/JuliaLang/julia/pull/53939 to pass.

Issue #3814, opened 2024-02-26
Weak dependency compat not respected


yha:
With this `Project.toml` file
```
[deps]
WebIO = "0f1e0344-ec1d-5b48-a673-e5cf874b6c29"

[weakdeps]
WebSockets = "104b5d7c-a370-577a-8038-80a2059c5097"

[compat]
WebSockets = "~1.5"
```
WebSockets 1.6 ends up being installed. 
This seems like the simplest possible use case of weak dependencies as described here:
https://pkgdocs.julialang.org/v1/creating-packages/#Weak-dependencies 
> A weak dependency is a dependency that will not automatically install when the package is installed but you can still control what versions of that package are allowed to be installed by setting compatibility on it

so I find it quite surprising that it doesn't work.

Example:
```
(@v1.10) pkg> activate --temp
  Activating new project at `C:\Users\sternlab\AppData\Local\Temp\jl_iiJvns`

(jl_iiJvns) pkg> add WebIO
   Resolving package versions...
    Updating `C:\Users\sternlab\AppData\Local\Temp\jl_iiJvns\Project.toml`
  [0f1e0344] + WebIO v0.8.21
    Updating `C:\Users\sternlab\AppData\Local\Temp\jl_iiJvns\Manifest.toml`
[...]
  [104b5d7c] + WebSockets v1.6.0
[...]

(jl_iiJvns) pkg> # edit Project.toml to add `weakdeps` and `compat` section

(jl_iiJvns) pkg> status
Status `C:\Users\sternlab\AppData\Local\Temp\jl_iiJvns\Project.toml`
  [0f1e0344] WebIO v0.8.21
Warning The project dependencies or compat requirements have changed since the manifest was last resolved. It is recommended to `Pkg.resolve()` or consider `Pkg.update()` if necessary.

(jl_iiJvns) pkg> up
    Updating registry at `C:\Users\sternlab\.julia\registries\General`
    Updating git-repo `https://github.com/JuliaRegistries/General.git`
  No Changes to `C:\Users\sternlab\AppData\Local\Temp\jl_iiJvns\Project.toml`
  No Changes to `C:\Users\sternlab\AppData\Local\Temp\jl_iiJvns\Manifest.toml`

(jl_iiJvns) pkg> resolve
ERROR: empty intersection between WebSockets@1.6.0 and project compatibility 1.5

(jl_iiJvns) pkg> status -m WebSockets
[...]

PR #3852, opened 2024-03-21
collect e.g. weak deps from project even if it is not a package

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3814

Need to add a test (and fix CI apparently).

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -249,8 +249,6 @@ function collect_project(pkg::Union{PackageSpec, Nothing}, path::String)
     deps = PackageSpec[]
     weakdeps = Set{UUID}()
     project_file = projectfile_path(path; strict=true)
-    weakdeps = Set{UUID}()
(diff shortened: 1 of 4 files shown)


src/Operations.jl
```diff
@@ -309,13 +308,13 @@ end
...
     weak_map = Dict{UUID,Set{UUID}}()
-    if env.pkg !== nothing
-        pkg = env.pkg
-        deps, weakdeps = collect_project(pkg, dirname(env.project_file))
-        deps_map[pkg.uuid] = deps
-        weak_map[pkg.uuid] = weakdeps
-        names[pkg.uuid] = pkg.name
-    end
+
+    uuid = env.pkg === nothing ? Base.dummy_uuid(env.project_file) : env.pkg.uuid
```
KristofferC, line 312:
```suggestion
    uuid = project_uuid(env)
```

Issue #3863, opened 2024-04-04
Precompile after `Pkg.add` tries to precompile dependencies that were removed


Keno:
In the following, the precompile after the second add should not have attempted to precompile `Enzyme`, because it got removed from the manifest. Indeed, and explicit `precompile` after the `add` succeeds. It looks like the precompile after the `add` may be issued with the wrong (outdated) list of packages.
```
(test) pkg> add SciMLSensitivity
   Resolving package versions...
    Updating `~/test/Project.toml`
  [1ed8b502] + SciMLSensitivity v7.56.2
    Updating `~/test/Manifest.toml`
  [47edcb42] + ADTypes v0.2.7
  [621f4979] + AbstractFFTs v1.5.0
  [7d9f7c33] + Accessors v0.1.36
  [79e6a3ab] + Adapt v4.0.4
⌅ [ec485272] + ArnoldiMethod v0.2.0
  [4fba245c] + ArrayInterface v7.9.0
  [4c555306] + ArrayLayouts v1.8.0
  [a9b6321e] + Atomix v0.1.0
  [62783981] + BitTwiddlingConvenienceFunctions v0.1.5
  [fa961155] + CEnum v0.5.0
  [2a0fbf3d] + CPUSummary v0.2.4
  [49dc2e85] + Calculus v0.5.1
  [7057c7e9] + Cassette v0.3.13
  [082447d4] + ChainRules v1.63.0
  [d360d2e6] + ChainRulesCore v1.23.0
  [fb6a15b2] + CloseOpenIntervals v0.1.12
  [38540f10] + CommonSolve v0.2.4
  [bbf7d656] + CommonSubexpressions v0.3.0
  [34da2185] + Compat v4.14.0
  [a33af91c] + CompositionsBase v0.1.2
  [2569d6c7] + ConcreteStructs v0.2.3
  [187b0558] + ConstructionBase v1.5.5
  [adafc99b] + CpuId v0.3.1
  [9a962f9c] + DataAPI v1.16.0
  [864edb3b] + DataStructures v0.18.18
  [e2d170a0] + DataValueInterfaces v1.0.0
  [2b5f629d] + DiffEqBase v6.149.0
  [459566f4] + DiffEqCallbacks v3.5.0
  [77a26b50] + DiffEqNoiseProcess v5.21.0
  [163ba53b] + DiffResults v1.1.0
  [b552c78f] + DiffRules v1.15.1
  [b4f34e82] + Distances v0.10.11
[...]

PR #3864, opened 2024-04-05
prune manifest after the set of some deps have been "demoted" to weakdeps

KristofferC:
Fixes the manifest pruning issue in https://github.com/JuliaLang/Pkg.jl/issues/3863.

Before we run this fixup code we do not know which deps are weak and not (all are considered normal deps). Or well, we kind of know but we don't have the information easily retrievable. Manifest pruning before this stage will not prune orphan entries which are weak deps of some package which is why in #3863 Enzyme was not pruned away (it is a weak dep of DiffEqBase).

Solve this by just pruning after we have categorized things into weak/normal deps.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -242,6 +242,7 @@ function fixup_ext!(env, pkgs)
             end
         end
     end
+    prune_manifest(env)
 end
 
 ####################


KristofferC:
This is not 100% satisfactory though because we download the packages before this step and IIUC it means we would download Enzyme in this case and only then prune the manifest to remove it. So perhaps we need to do this before we have downloaded anything.

KristofferC:
At least this is better than status quo.

Issue #3797, opened 2024-02-14
`pkg> add` for a `weakdep` package does not permanently install package


topolarity:
If you `pkg> add` a package that's already in your Project.toml as a weakdep, Pkg gets mildly indecisive and installs all of the package + deps only to remove them on the next package operation.

**MWE:**
```console
$ cat Foo/Project.toml
name = "Foo"
uuid = "721b2572-63d0-4bb6-a91e-fe9952250d90"
authors = ["Cody Tapscott <topolarity@tapscott.me>"]
version = "0.1.0"

[weakdeps]
CUDA = "052768ef-5323-5732-b1bb-66c8b64840ba"

[compat]
CUDA = "4"
$ julia --project=Foo -q
(Foo) pkg> add CUDA
<adds a ton of CUDA + dependent packages>
(Foo) pkg> add TOML
<immediately removes all of CUDA + its dependent packages>
```

It is a bit concerning that `> add Foo` can have an inconsistent concept of the state after an add versus the rest of Pkg.

KristofferC:
The reason for this is that when Pkg reads a project and sees that it is both a weakdep and a dep it ignores the dep entry (so that extensions can be used in a backwards compatible manner). 

The solution here might be to just delete the weak dep entry when an explicit `add` is made.

KristofferC:
But on the other hand, having extensions while having strong deps on the triggers for those extensions is very likely to run into a cycle analogous to https://github.com/JuliaLang/julia/issues/52511#issuecomment-1983388150.

PR #3865, opened 2024-04-09
make `add` and `dev` on a package remove it from the set of weak dependencies

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3797


topolarity:
I've been running into this on 1.10 - Would it be possible to get a backport?

KristofferC:
👍

PR #3867, opened 2024-04-10
add "t" option for autoinstall into a temp env

aplavin:


IanButterworth:
See original PR for why this wasn't added https://github.com/JuliaLang/Pkg.jl/pull/2669

IanButterworth:
Yeah, agreed, adding that to docs would be great

PR #3872, opened 2024-04-14
Pkg.BinaryPlatforms: Explicitly itemize Base.BinaryPlatforms usage

mkitti:
Currently, Pkg.BinaryPlatforms does `using Base.BinaryPlatforms`. This pull request itemizes the symbols used explicitly.


mkitti:
This is still useful. Imports in the standard library should be explicit.

PR #3876, opened 2024-04-18
do not precompile full env on `using` callback from REPL

KristofferC:
@Keno was annoyed that calling `using` on some small package and having it auto install compiled the whole env. This disables that. I still think https://github.com/JuliaLang/Pkg.jl/pull/3848 is better but since that has some push back, this solves the immediate issue.


IanButterworth (approved):
Yeah this makes sense to me. The user is in the julia REPL and in a loading and using immediately mentality.

PR #3877, opened 2024-04-19
guard against isdir EACESS in completions

IanButterworth:
Fixes https://github.com/JuliaLang/julia/issues/54154

### DIFF
--- a/ext/REPLExt/completions.jl
+++ b/ext/REPLExt/completions.jl
@@ -42,6 +42,7 @@ function complete_expanded_local_dir(s, i1, i2, expanded_user, oldi2)
             isdir(s[1:prevind(s, first(cmp2)-i1+1)]*x)
         catch e
             (e isa Base.IOError && e.code == Base.UV_EACCES) || rethrow()
+            return false
         end
     end
     if expanded_user
--- a/ext/REPLExt/completions.jl
+++ b/ext/REPLExt/completions.jl
@@ -41,8 +41,11 @@ function complete_expanded_local_dir(s, i1, i2, expanded_user, oldi2)
         try
             isdir(s[1:prevind(s, first(cmp2)-i1+1)]*x)
         catch e
-            (e isa Base.IOError && e.code == Base.UV_EACCES) || rethrow()
-            return false
+            if e isa Base.IOError && e.code == Base.UV_EACCES
+                return false
+            else
+                rethrow()
+            end
         end
     end
     if expanded_user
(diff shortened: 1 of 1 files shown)


ext/REPLExt/completions.jl
```diff
@@ -37,7 +37,13 @@ function complete_expanded_local_dir(s, i1, i2, expanded_user, oldi2)
     cmp = REPL.REPLCompletions.complete_path(s, i2, shell_escape=true)
     cmp2 = cmp[2]
     completions = [REPL.REPLCompletions.completion_text(p) for p in cmp[1]]
-    completions = filter!(x -> isdir(s[1:prevind(s, first(cmp2)-i1+1)]*x), completions)
+    completions = filter!(completions) do x
+        try
+            isdir(s[1:prevind(s, first(cmp2)-i1+1)]*x)
```
KristofferC, line 42: Shouldn't this function return `true` or `false`? Does it always do that now or is the fall through returning `nothing` which would error in `filter!`?

IanButterworth: It throws if it doesn't have perms to stat the path

KristofferC: Yes, and it should return `false` in that case, but right now it isn't I think?

IanButterworth: @vtjnash perhaps you have an opinion on what to do here?

vtjnash (commented):
Yes, the catch needs to end with a bool so that the filter doesn't error

IanButterworth:
Oh right, I thought @KristofferC was talking about changing the behavior of `isdir` not the filter anonymous function.

ext/REPLExt/completions.jl
```diff
@@ -37,7 +37,13 @@ function complete_expanded_local_dir(s, i1, i2, expanded_user, oldi2)
...
+        catch e
+            (e isa Base.IOError && e.code == Base.UV_EACCES) || rethrow()
```
IanButterworth, line 44:
```suggestion
            (e isa Base.IOError && e.code == Base.UV_EACCES) || rethrow()

            return false
```

vtjnash (commented):
Though really I think we should later add a kwarg to these functions because this pattern is getting really annoying

Eg `isdir(path, default=false)`
Or `default=nothing` or `default=throw` (the current behavior) or `default=Returns`, or such

ext/REPLExt/completions.jl
```diff
@@ -37,7 +37,14 @@ function complete_expanded_local_dir(s, i1, i2, expanded_user, oldi2)
...
+            return false
```
IanButterworth, line 45:
```suggestion
            if e isa Base.IOError && e.code == Base.UV_EACCES
                return false
            else
                rethrow()
            end
```

PR #3911, opened 2024-06-03
Use historical stdlib dependency information

staticfloat:
This makes use of the new functionality in HistoricalStdlibVersions [0]
to properly pull out the dependency information in old stdlibs.  This
fixes https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl/issues/9.

[0] https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl/pull/23

### DIFF
--- a/src/HistoricalStdlibs.jl
+++ b/src/HistoricalStdlibs.jl
@@ -1,6 +1,17 @@
 using Base: UUID
 
-const DictStdLibs = Dict{UUID,Tuple{String,Union{VersionNumber,Nothing}}}
+struct StdlibInfo
+    name::String
+    uuid::UUID
+
+    # This can be `nothing` if it's an unregistered stdlib
+    version::Union{Nothing,VersionNumber}
+
+    deps::Vector{UUID}
+    weakdeps::Vector{UUID}
+end
+
+const DictStdLibs = Dict{UUID,StdlibInfo}
 
 # Julia standard libraries with duplicate entries removed so as to store only the
 # first release in a set of releases that all contain the same set of stdlibs.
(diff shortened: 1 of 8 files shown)


staticfloat:
One point of contention here; the `Pkg.Types.stdlibs()` function seems to be [used in the ecosystem](https://juliahub.com/ui/Search?q=Pkg.Types.stdlibs()&type=code), but I like the `StdlibInfo` struct a lot more than just a random tuple.  Should I maintain backwards compatibility, or do we just require package authors to do something like:

```
if isdefined(Pkg.Types, :StdlibInfo)
    stdlib_names = [info.name for (uuid, info) in Pkg.Types.stdlibs()]
else
    stdlib_names = [name for (uuid, (name, version)) in Pkg.Types.stdlibs()]
end
```

staticfloat:
@KristofferC If you are happy with this, I will make a breaking `v2.0.0` release of HistoricalStdlibVersions, register that, then drop the commit here that manually installs that branch.  (Kudos on the `[sources]` section, super convenient!)

DilumAluthge:
Just to double-check: the `Pkg.Types.stdlibs()` function is not currently public, right? So in theory we are allowed to make whatever breaking changes we want?

> the `Pkg.Types.stdlibs()` function seems to be [used in the ecosystem](https://juliahub.com/ui/Search?q=Pkg.Types.stdlibs()&type=code),

Skimming that list, it looks like I'm responsible for 4 out of 7 of the usages. I'd be fine with doing a breaking change and then patching my packages to do the requisite `if isdefined ...`.

DilumAluthge:
Just a quick note: a few additional usages pop up if you change the `Pkg.Types.stdlibs()` (in your search) to `Pkg.Types.stdlibs`: https://juliahub.com/ui/Search?q=Pkg.Types.stdlibs&type=code

staticfloat:
> Just to double-check: the Pkg.Types.stdlibs() function is not currently public, right? So in theory we are allowed to make whatever breaking changes we want?

Yes, and perhaps the _right_ thing to do here is to provide a `stdlibs_names()::Vector{String}` method for people, as it seems that's most of what people want to know.  There do seem to be some people who want to check `is_stdlib(::String)`, which we already have inside of Pkg, so it shouldn't be too hard to convert them to a new API if need be.

KristofferC:
LGTM but I would prefer leaving the `stdlibs` etc functions as they were and make new functions for the new info. Just because I know people use Pkg internals a bit all over the place and it seems like there is not a huge gain to modify the existing functions over adding a new one.

DilumAluthge:
Converting to draft so we don't accidentally merge before we remove the "drop me" commits.

staticfloat:
Alright, I'm happy with this, the new HSG version has been published and as long as @KristofferC's happy with it, I think we're good to merge.

src/HistoricalStdlibs.jl
```diff
@@ -13,32 +24,4 @@ const STDLIBS_BY_VERSION = Pair{VersionNumber, DictStdLibs}[]
...
-    UUID("1a1011a3-84de-559e-8e89-a11a2f7dc383") => ("SharedArrays", nothing),
-    UUID("6462fe0b-24de-5631-8697-dd941f90decc") => ("Sockets", nothing),
-    UUID("2f01184e-e22b-5df5-ae63-d93ebab69eaf") => ("SparseArrays", nothing),
-    UUID("10745b16-79ce-11e8-11f9-7d13ad32a3b2") => ("Statistics", nothing),
-    UUID("4607b0f0-06f3-5cda-b6b1-a6196a1729e9") => ("SuiteSparse", nothing),
-    UUID("8dfed614-e22c-5e08-85e1-65c5234f0b40") => ("Test", nothing),
-    UUID("cf7118a7-6976-5b1a-9a39-7adc72f591a4") => ("UUIDs", nothing),
-    UUID("4ec0a83e-493e-50e2-b9ac-8f72acf5a8f5") => ("Unicode", nothing),
-)
+const UNREGISTERED_STDLIBS = DictStdLibs()
```
IanButterworth, line 27: Is this populated by HSV now? Might be worth a comment?

staticfloat: This was actually always populated by HSV, it was never used as-is.

IanButterworth:
@KristofferC merge?

KristofferC:
Go for it!

KristofferC:
(I went for it!)

PR #3913, opened 2024-06-03
abort querying more packages for hint auto complete

### LINKED ISSUE #3909: Early abandonment of remote package name completion
As this is related to the tremendous efforts for TTFX, the TTFCompletion of remote package names is still stressful in some cases. (**Edited**)

The main source of latency (with **uncompressed** registry) is to check all relevant TOML files for compatibility of candidate packages.
You can see what is happening while typing a package name with the following monkey patch:
```julia
using Pkg
function Pkg.Registry._parsefile(toml_file::AbstractString)
    println(toml_file) # print file path to stdout
    Base.parsed_toml(toml_file, Pkg.Registry.TOML_CACHE, Pkg.Registry.TOML_LOCK)
end
```

However, in terms of user input completion, you should be able to give up when the final candidate is no longer determinable.
Therefore, I believe that the following monkey patch would be useful.

```julia
using Pkg
using Pkg: Registry

REPLExt = Base.get_extension(Pkg, :REPLExt)
JULIA_UUID = REPLExt.JULIA_UUID
function REPLExt.complete_remote_package(partial)
    isempty(partial) && return String[]
    cmp = Set{String}()
    nextchar::Union{Nothing, Char} = nothing
    nextcharidx = length(partial) + 1
    for reg in Registry.reachable_registries()
        for (uuid, regpkg) in reg
            name = regpkg.name
            name in cmp && continue
            if startswith(name, partial) # case-sensitive?
                if length(name) >= nextcharidx # check the next char (In terms of consistency, this might be just before `push!`.)
                   if nextchar === nothing
                      nextchar = name[nextcharidx]
                   elseif nextchar != name[nextcharidx]
                      return String[] # give up
                   end
                end
                pkg = Registry.registry_info(regpkg)
                compat_info = Registry.compat_info(pkg)
                # Filter versions
                for (v, uncompressed_compat) in compat_info
                    Registry.isyanked(pkg, v) && continue
                    # TODO: Filter based on offline mode
                    is_julia_compat = nothing
                    for (pkg_uuid, vspec) in uncompressed_compat
                        if pkg_uuid == JULIA_UUID
                            found_julia_compat = true
                            is_julia_compat = VERSION in vspec
                            is_julia_compat && continue
                        end
                    end
                    # Found a compatible version or compat on julia at all => compatible
                    if is_julia_compat === nothing || is_julia_compat
                        push!(cmp, name)
                        break
                    end
                end
            end
        end
    end
    return sort!(collect(cmp))
end
```

```julia
julia> versioninfo()
Julia Version 1.11.0-beta2
Commit edb3c92d6a (2024-05-29 09:37 UTC)
Build Info:
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, tigerlake)
```

KristofferC:
the result of hint auto complete is only used if a unique entry is returned

fixes #3909

### DIFF
--- a/ext/REPLExt/REPLExt.jl
+++ b/ext/REPLExt/REPLExt.jl
@@ -26,7 +26,7 @@ struct PkgCompletionProvider <: LineEdit.CompletionProvider end
 function LineEdit.complete_line(c::PkgCompletionProvider, s; hint::Bool=false)
     partial = REPL.beforecursor(s.input_buffer)
     full = LineEdit.input_string(s)
-    ret, range, should_complete = completions(full, lastindex(partial))
+    ret, range, should_complete = completions(full, lastindex(partial); hint)
     return ret, partial[range], should_complete
 end
 
--- a/ext/REPLExt/completions.jl
+++ b/ext/REPLExt/completions.jl
@@ -11,7 +11,7 @@ function _shared_envs()
     return possible
 end
 
-function complete_activate(options, partial, i1, i2)
+function complete_activate(options, partial, i1, i2; hint::Bool)
     shared = get(options, :shared, false)
     if shared
         return _shared_envs()
@@ -62,7 +62,8 @@ end
 
 
 const JULIA_UUID = UUID("1222c4b2-2114-5bfd-aeef-88e4692bbb3e")
-function complete_remote_package(partial)
+function complete_remote_package(partial; hint::Bool)
+    found_match = false
     isempty(partial) && return String[]
     cmp = Set{String}()
     for reg in Registry.reachable_registries()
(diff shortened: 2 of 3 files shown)


IanButterworth (approved):

kimikage:
If an abort occurs while entering the name of a remote package, will it display the only local directory (or something) with the same starting string as a hint?
I would find that acceptable, but I want to know the behavior just to be sure.

KristofferC:
> If an abort occurs while entering the name of a remote package, will it display the only local directory (or something) with the same starting string as a hint?

I don't understand what this means. There should be no observable difference between the behavior here and the one on the current 1.11 julia.

KristofferC:
Local directories are handled by `complete_local_dir` which is not touched here.

PR #3917, opened 2024-06-04
avoid returning empty completion when aborting to not have it inline directories

### LINKED ISSUE #3916: Inconsistent preference of local directories over package names in auto hints
PR #3913 introduces inconsistency in automatic hinting under certain conditions.
>  For example, When there is a directory named "Documents" in the active directory, and I type "Doc", will it not display a hint "Documents/" and then complete it when I press `<TAB>`?
I.e., "Documenter" or "DocumenterTools", etc. will not be ignored?
>
> **Edit:** I am a Windows user, but I believe macOS also has multiple directories in the home directory which start with uppercase.

_Originally posted by @kimikage in https://github.com/JuliaLang/Pkg.jl/pull/3913#issuecomment-2145313089

The above concerns are actually realized in #master.
336280522-4d26d03a-bc4e-4661-ae9d-01d0f2986c7f

Personally, I don't think this is a high priority issue. However, I file this as an issue because the conditions for reproducing the problem depend on the specific situation.

KristofferC:
Should fix #3916 (haven't tested)

### DIFF
--- a/ext/REPLExt/completions.jl
+++ b/ext/REPLExt/completions.jl
@@ -88,9 +88,9 @@ function complete_remote_package(partial; hint::Bool)
                     if is_julia_compat === nothing || is_julia_compat
                         push!(cmp, name)
                         # In hint mode the result is only used if there is a single matching entry
-                        # so we can return no matches in case of more than one match
+                        # so we abort the search
                         if hint && found_match
-                            return String[]
+                            return sort!(collect(cmp))
                         end
                         found_match = true
                         break


kimikage:
I believe I still lack the composure to review this PR.

As a matter of fact, my REPL did not hint "Documents/" when "Doc" was typed, and hinted "DocumenterTools" when "DocumenterT" was typed.
pkg_3917

kimikage:
This is definitely an improvement, but I realized that this is not a perfect solution after looking at https://github.com/JuliaLang/julia/issues/54906.

If you type "add Cha", "ChainRules" and "ChainRulesCore" will be candidates, and "ChainRules" will be displayed as a hint. Given the existence of "Chairmarks", the hint is somewhat misleading.

KristofferC:
I didn't know that the auto complete did partial completions like that. I even thought I tried it...

Issue #3929, opened 2024-06-18
New HistoricalStdlibVersions version is breaking the Pkg tests


staticfloat:
I'm confused; I tagged a v2.0 breaking version of HistoricalStdlibVersions, but it appears that new version is [breaking existing Pkg tests](https://buildkite.com/julialang/julia-master/builds/37467#01902945-bb30-47fa-ab51-7abc5e6a94c0/835-1765).  Is [this compat entry](https://github.com/JuliaLang/Pkg.jl/blob/master/test/Project.toml#L18) insufficient?

KristofferC:
It seems Pkg adds HistoricalStdlibVersions.jl during testing in a temp env:

https://github.com/JuliaLang/Pkg.jl/blob/db8ce8919c3f4a51d3e9307fa56a83932bf9d2c1/test/runtests.jl#L43

Which will not have that compat entry in it.

PR #3930, opened 2024-06-18
Fix HistoricalStdlibVersions install to version

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3929

Given this branch isn't touched on this CI
```
% julia  
               _
  (_)     | (_) (_)    |
  | | | | | | |/ _` |  |
  | | |_| | | | (_| |  |  Version 1.10.4 (2024-06-04)
|__/                   |

(@v1.10) pkg> activate --temp
  Activating new project at `/var/folders/1z/jf841bdj73bdj3vk7kc7f_3w0000gn/T/jl_0iHV55`

julia> import Pkg

julia> Pkg.add(name="HistoricalStdlibVersions", version="1.2")
   Resolving package versions...
    Updating `/private/var/folders/1z/jf841bdj73bdj3vk7kc7f_3w0000gn/T/jl_0iHV55/Project.toml`
⌃ [6df8b67a] + HistoricalStdlibVersions v1.2.4
    Updating `/private/var/folders/1z/jf841bdj73bdj3vk7kc7f_3w0000gn/T/jl_0iHV55/Manifest.toml`
⌃ [6df8b67a] + HistoricalStdlibVersions v1.2.4

julia> 
```
```
julia> Pkg.add(name="HistoricalStdlibVersions", version="1.2", io=iob)

julia>
```

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -42,7 +42,7 @@ Logging.with_logger((islogging || Pkg.DEFAULT_IO[] == devnull) ? Logging.Console
         iob = IOBuffer()
         Pkg.activate(; temp = true)
         try
-            Pkg.add("HistoricalStdlibVersions", io=iob) # Needed for custom julia version resolve tests
+            Pkg.add(name="HistoricalStdlibVersions", version="1.2", io=iob) # Needed for custom julia version resolve tests
         catch
             println(String(take!(iob)))
             rethrow()
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -42,7 +42,7 @@ Logging.with_logger((islogging || Pkg.DEFAULT_IO[] == devnull) ? Logging.Console
         iob = IOBuffer()
         Pkg.activate(; temp = true)
         try
-            Pkg.add(name="HistoricalStdlibVersions", version="1.2", io=iob) # Needed for custom julia version resolve tests
+            Pkg.add(name="HistoricalStdlibVersions", version="1.2", uuid="6df8b67a-e8a0-4029-b4b7-ac196fe72102", io=iob) # Needed for custom julia version resolve tests
         catch
             println(String(take!(iob)))
             rethrow()


test/runtests.jl
```diff
@@ -42,7 +42,7 @@ Logging.with_logger((islogging || Pkg.DEFAULT_IO[] == devnull) ? Logging.Console
         iob = IOBuffer()
         Pkg.activate(; temp = true)
         try
-            Pkg.add("HistoricalStdlibVersions", io=iob) # Needed for custom julia version resolve tests
+            Pkg.add(name="HistoricalStdlibVersions", version="1.2", io=iob) # Needed for custom julia version resolve tests
```
DilumAluthge, line 45: Can we also take this opportunity to specify the UUID?

DilumAluthge (approved):

test/runtests.jl
IanButterworth: Go ahead

DilumAluthge, line 45:
```suggestion
            Pkg.add(name="HistoricalStdlibVersions", version="1.2", uuid="6df8b67a-e8a0-4029-b4b7-ac196fe72102", io=iob) # Needed for custom julia version resolve tests
```

DilumAluthge (approved):

DilumAluthge:
@IanButterworth The backport label for 1.10 has no hyphen, but the backport label for 1.11 does have a hyphen. Is that something we need to fix? Does this break the Backporter script?

IanButterworth:
Good catch https://github.com/KristofferC/Backporter/blob/f366a87a94c9d38fd107a1c53ccc4550f3687625/backporter.jl#L23

Fixed

test/runtests.jl
```diff
@@ -42,7 +42,7 @@ Logging.with_logger((islogging || Pkg.DEFAULT_IO[] == devnull) ? Logging.Console
         iob = IOBuffer()
         Pkg.activate(; temp = true)
         try
-            Pkg.add("HistoricalStdlibVersions", io=iob) # Needed for custom julia version resolve tests
+            Pkg.add(name="HistoricalStdlibVersions", version="1.2", uuid="6df8b67a-e8a0-4029-b4b7-ac196fe72102", io=iob) # Needed for custom julia version resolve tests
```
staticfloat, line 45: Maybe we should parse the compat bound from the Project.toml?

IanButterworth: I thought about that, but what if we start doing `"1.2,2"`. Part of the problem is this code path isn't hit on Pkg's own CI.

We might just want to teach `Base.runtests` to run stdlib tests via `Pkg.test` to handle test deps properly

DilumAluthge: Is `Pkg.test` guaranteed to pick up the correct version of the stdlib? (Because we're not doing the "modify the stdlib UUID trick.)

If we can get `Pkg.test("StdlibName")` to work correctly in `Base.runtests`, I agree that's probably the best solution.

staticfloat: > I thought about that, but what if we start doing `"1.2,2"`.

That's exactly why I want this.  `Pkg.add(;..., version=VersionSpec(...))` works, so you can do something like the following:

```
hsv_compat_val = TOML.parsefile(project_path)["compat"]["HistoricalStdlibVersions"]
Pkg.add(;name="HistoricalStdlibVersions", version=Pkg.Types.semver_spec(hsv_compat_val))
```

DilumAluthge: IIRC, we do some shenanigans with the `JULIA_PROJECT` and `JULIA_LOAD_PATH`. So we need to make sure that `Pkg.test` still works correctly in that case.

Also, how will this work with testgroups? For stdlibs, `Base.runtests` is supposed to look at the `test/testgroups` file, and run each testgroup in parallel. Would we have to teach `Pkg.test` to understand stdlib testgroups? I'm not sure if we want to add that kind of functionality as public API in Pkg, but maybe we can add it as a private/internal Pkg API that we only use for stdlib tests in `Base.runtests`.

DilumAluthge: Something like `Pkg.test("StdlibName"; _stdlib_testgroup = "desired_test_group")` might be sufficient. And we keep `_stdlib_testgroup` as private/internal API only.

PR #3938, opened 2024-07-03
Use `Base.TOML.Parser{Dates}` for TOML parsing w/ Dates support

topolarity:
Adapting to https://github.com/JuliaLang/julia/pull/54755/ and https://github.com/JuliaLang/julia/pull/55017

### DIFF
--- a/src/Registry/registry_instance.jl
+++ b/src/Registry/registry_instance.jl
@@ -1,5 +1,6 @@
 using Base: UUID, SHA1
 using TOML
+using Dates
 using Tar
 using ..Versions: VersionSpec, VersionRange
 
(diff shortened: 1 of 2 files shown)


src/Registry/registry_instance.jl
```diff
@@ -15,9 +15,8 @@ function to_tar_path_format(file::AbstractString)
 end
 
 # See loading.jl
-const TOML_CACHE = let parser = Base.TOML.Parser()
-    parser.Dates = Dates
-    Base.TOMLCache(parser, Dict{String, Dict{String, Any}}())
+const TOML_CACHE = let parser = TOML.DTParser()
```
vtjnash, line 18: You are not allowed to parameterize the Base.TOML module with the unrelated code from TOML, as they only share identity by occasional accident as an implementation detail

topolarity: Where do you see `Base.TOML` here?

If you mean `Base.TOMLCache` then we can re-export that via `TOML.Internals.TOMLCache` if need be

vtjnash: Base.TOMLCache has a typeassert that you use Base.TOML and not the TOML package

topolarity:
Probably need to wait for a new nightly first 😉

IanButterworth:
Nightlies are updated as soon as master commits are green, not nightly. The powerpc builds are lagging but otherwise all other nightlies should be up to date.

IanButterworth:
The 64 bit windows nightly is a day old even though there are multiple green CI runs on julia master since.. odd.

I'll give it time to resolve itself

PR #3941, opened 2024-07-04
Use more internal Pkg.add api to bypass auto-registry-install

### LINKED ISSUE #3940: Pkg tests puts an outdated copy of `General` in `usr/share/registries`
Before:
```
keno@demeter2:~$ ls -la julia/usr/share/julia/registries
total 0
drwxrwxr-x 1 keno keno   0 Jul  4 18:41 .
drwxrwxr-x 1 keno keno 180 Jul  4 18:29 ..
```
Then
```
keno@demeter2:~/julia$ make test-Pkg
Whitespace check found no issues.
This build contains [GPL-2.0+] libs: libcholmod librbio libspqr libumfpack
    JULIA test/Pkg
Running parallel tests with:
  getpid() = 3098280
  nworkers() = 1
  nthreads() = 1
  Sys.CPU_THREADS = 64
  Sys.total_memory() = 755.458 GiB
  Sys.free_memory() = 727.929 GiB

Test  (Worker) | Time (s) | GC (s) | GC % | Alloc (MB) | RSS (MB)
Pkg        (1) |        started at 2024-07-04T18:41:40.414
```
and shortly thereafter:
```
total 7608
drwxrwxr-x 1 keno keno      52 Jul  4 18:41 .
drwxrwxr-x 1 keno keno     180 Jul  4 18:29 ..
-rw-rw-r-- 1 keno keno 7786491 Jul  4 18:41 General.tar.gz
-rw-rw-r-- 1 keno keno     129 Jul  4 18:41 General.toml
```
This is problematic because:
1. This copy of General doesn't get updated
2. Having this copy of General present prevents julia from installing the `~/.julia/registries` copy (if not already present)
3. Having only this copy is confusing (#3939)

Keno:
Fixes #3940. HistoryicalStdlibVersions has no dependencies, so the registry should not be required to install it.

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -48,7 +48,7 @@ Logging.with_logger((islogging || Pkg.DEFAULT_IO[] == devnull) ? Logging.Console
             spec = Pkg.PackageSpec(
                 name="HistoricalStdlibVersions", 
                 url="https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl",
-                rev="39d6ab6e71b614242976ab014635ac7ef039cb58", #= version="1.2", =#
+                rev="5879c5f690795208481c60b904f4af4e8c1eeef8", #= version="2.0.0", =#
                 uuid="6df8b67a-e8a0-4029-b4b7-ac196fe72102")
             Pkg.API.handle_package_input!(spec)
             Pkg.add(Pkg.API.Context(), [spec], io=iob)
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -42,7 +42,15 @@ Logging.with_logger((islogging || Pkg.DEFAULT_IO[] == devnull) ? Logging.Console
         iob = IOBuffer()
         Pkg.activate(; temp = true)
         try
-            Pkg.add(name="HistoricalStdlibVersions", version="1.2", uuid="6df8b67a-e8a0-4029-b4b7-ac196fe72102", io=iob) # Needed for custom julia version resolve tests
+            # Needed for custom julia version resolve tests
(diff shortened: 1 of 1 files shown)


IanButterworth:
Might a cleaner way to do this be to git clone HistoryicalStdlibVersions at the right commit then `Pkg.dev` it?

IanButterworth:
Even `develop` installs the default registries if not present https://github.com/IanButterworth/Pkg.jl/blob/aa6e43ff5dbd6e88202b966e9d55221b15825d4e/src/API.jl#L149

So this seems to be the way. Just a little brittle as this code isn't hit in the Pkg repo CI, just julia CI

test/runtests.jl
```diff
@@ -43,7 +43,15 @@ Logging.with_logger((islogging || Pkg.DEFAULT_IO[] == devnull) ? Logging.Console
         iob = IOBuffer()
         Pkg.activate(; temp = true)
         try
-            Pkg.add(name="HistoricalStdlibVersions", version="2", uuid="6df8b67a-e8a0-4029-b4b7-ac196fe72102", io=iob) # Needed for custom julia version resolve tests
+            # Needed for custom julia version resolve tests
+            # Don't use the toplevel PKg.add() command to avoid accidentally installing another copy of the registry
+            spec = Pkg.PackageSpec(
+                name="HistoricalStdlibVersions", 
+                url="https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl",
+                rev="39d6ab6e71b614242976ab014635ac7ef039cb58", #= version="1.2", =#
```
IanButterworth, line 51:
```suggestion
                rev="5879c5f690795208481c60b904f4af4e8c1eeef8", #= version="2.0.0", =#
```

PR #3946, opened 2024-07-11
[sources] is available in 1.11

j-fu:
Move the remark in the changelog to 1.11

According to Kristoffer Carlsson at State of Julia 2024, this is in 1.11.
Confirmed by testing with 1.11-rc1


KristofferC:
Thanks!

PR #3951, opened 2024-07-13
fix artifact printing quiet rule

IanButterworth:
Without this artifact downloads have not been printed...


KristofferC:
Hah, funny.

PR #3955, opened 2024-07-17
Fix codeblock language and prompt in Pkg.status() docstring

### LINKED ISSUE #3954: Weird syntax highlighting in docstrings
Happens on `julia +nightly`. Issue is likely that the stdout output is included in a codeblock (cc @fredrikekre).

image

fatteneder:
Fixes #3954

image

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -518,13 +518,14 @@ Setting `outdated=true` will only show packages that are not on the latest versi
 their maximum version and why they are not on the latest version (either due to other
 packages holding them back due to compatibility constraints, or due to compatibility in the project file).
 As an example, a status output like:
-```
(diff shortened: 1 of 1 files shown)


src/Pkg.jl
```diff
@@ -520,11 +520,14 @@ packages holding them back due to compatibility constraints, or due to compatibi
 As an example, a status output like:
 ```
 pkg> Pkg.status(; outdated=true)
```
fredrikekre, line 522: This should be a julia prompt then?

fatteneder: Julia prompt is `julia-repl`?
Adding this does not highlight anything then.

fatteneder: Now I see what you mean. I fixed it.

Issue #3960, opened 2024-07-22
Manifest usage log error should be non-fatal


fonsp:
I got this error today in a production server, which caused the process to crash.

```
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]: ERROR: TOML Parser error:
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]: /home/fons/.julia/logs/manifest_usage.toml:17338:25 error: string literal ended unexpectedly
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:   [["/tmp/jl_UekhXS/Manife
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:                          ^
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]: Stacktrace:
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:   [1] parse
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:     @ ./toml_parser.jl:445 [inlined]
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:   [2] parsefile(f::String)
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:     @ TOML ~/julia-1.10.4/share/julia/stdlib/v1.10/TOML/src/TOML.jl:43
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:   [3] (::Pkg.Types.var"#51#54"{String, String, Dates.DateTime, String})()
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:     @ Pkg.Types ~/julia-1.10.4/share/julia/stdlib/v1.10/Pkg/src/Types.jl:541
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:   [4] mkpidlock(f::Pkg.Types.var"#51#54"{String, String, Dates.DateTime, String}, at::String, pid::Int32; kwopts::@Kwargs{stale_age::Int6>
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:     @ FileWatching.Pidfile ~/julia-1.10.4/share/julia/stdlib/v1.10/FileWatching/src/pidfile.jl:93
Jul 22 14:59:49 ct-bind-4 pluto-slider-server.sh[941]:   [5] mkpidlock
[...]

PR #3961, opened 2024-07-22
Buffer writes to usage and orphanage files to avoid corruption

fonsp:
The error in https://github.com/JuliaLang/Pkg.jl/issues/3960 is caused by a partially written TOML file. I suspect that this can be mostly avoided by buffering the file write (writing TOML into a memory buffer, then writing the buffer to file).


IanButterworth:
Note that I added https://github.com/JuliaLang/Pkg.jl/pull/3928 on master, which I hope had improved this somewhat

IanButterworth:
i.e. that may be enough already?

fonsp:
Nice!

PR #3962, opened 2024-07-22
Make manifest usage log errors non-fatal

fonsp:
Implements #3960

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -622,7 +622,12 @@ function write_env_usage(source_file::AbstractString, usage_filepath::AbstractSt
     ## Atomically write usage file using process id locking
     FileWatching.mkpidlock(usage_file * ".pid", stale_age = 3) do
         usage = if isfile(usage_file)
-            TOML.parsefile(usage_file)
+            try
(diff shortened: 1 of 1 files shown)


IanButterworth:
I think it's probably still good to have this, even though we now have https://github.com/JuliaLang/Pkg.jl/pull/3928 because those files are changed by all julia versions that might not have that fix

src/Types.jl
```diff
@@ -622,7 +622,12 @@ function write_env_usage(source_file::AbstractString, usage_filepath::AbstractSt
     ## Atomically write usage file using process id locking
     FileWatching.mkpidlock(usage_file * ".pid", stale_age = 3) do
         usage = if isfile(usage_file)
-            TOML.parsefile(usage_file)
+            try
+                TOML.parsefile(usage_file)
+            catch err
+                @warn "Failed to parse usage file `$usage_file`, ignoring." err
```
IanButterworth, line 628:
```suggestion
                @warn "Failed to parse updated usage file `$usage_file`, so not saving." err
```

Issue #3885, opened 2024-05-01
`MethodError` upon using `Pkg.add()` with a `PackageSpec` with unspecified `version`


topolarity:
```julia
$ cat Project.toml
name = "FooPackage"
uuid = "7da9e4b2-06fe-11ef-02e0-95e6aca98d3b"
$ julia +nightly --project=. -e "using Pkg; Pkg.add(Pkg.PackageSpec(name=\"SuiteSparse\", uuid=\"4607b0f0-06f3-5cda-b6b1-a6196a1729e9\"))"
    Updating registry at `~/.julia/registries/JuliaComputingRegistry.toml`
    Updating registry at `~/.julia/registries/JuliaHubRegistry.toml`
    Updating registry at `~/.julia/registries/JuliaSimRegistry.toml`
    Updating registry at `~/.julia/registries/PumasRegistry.toml`
    Updating registry at `~/.julia/registries/TestRegistry.toml`
    Updating registry at `~/.julia/registries/githubjuliaregistryjdb.toml`
    Updating registry at `~/.julia/registries/gitlabjuliaregistryjdb.toml`
   Resolving package versions...
ERROR: MethodError: no method matching thispatch(::Nothing)
The function `thispatch` exists, but no method is defined for this combination of argument types.

Closest candidates are:
  thispatch(::VersionNumber)
   @ Base version.jl:233

Stacktrace:
 [1] add(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}, new_git::Set{Base.UUID}; preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform, target::Symbol)
   @ Pkg.Operations ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.12/Pkg/src/Operations.jl:1523
 [2] add
   @ ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.12/Pkg/src/Operations.jl:1481 [inlined]
 [3] add(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}; preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform, target::Symbol, kwargs::@Kwargs{io::Pkg.UnstableIO})
   @ Pkg.API ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.12/Pkg/src/API.jl:306
[...]

PR #3970, opened 2024-07-25
Fix handling of `PackageSpec` with un-specified version

topolarity:
Resolves https://github.com/JuliaLang/Pkg.jl/issues/3885.

I could use a bit of help to add an appropriate test - Looks like this was regressed in https://github.com/JuliaLang/Pkg.jl/pull/3732

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1524,8 +1524,10 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}, new_git=Set{UUID}();
             compat_names = String[]
             for pkg in pkgs
                 haskey(ctx.env.project.compat, pkg.name) && continue
-                pkgversion = Base.thispatch(ctx.env.manifest[pkg.uuid].version)
-                set_compat(ctx.env.project, pkg.name, string(pkgversion))
+                v = ctx.env.manifest[pkg.uuid].version
+                v === nothing && continue
+                pkgversion = string(Base.thispatch(v))
+                set_compat(ctx.env.project, pkg.name, pkgversion)
                 push!(compat_names, pkg.name)
             end
             printpkgstyle(ctx.io, :Compat, """entries added for $(join(compat_names, ", "))""")


src/Operations.jl
```diff
@@ -1524,8 +1524,10 @@ function add(ctx::Context, pkgs::Vector{PackageSpec}, new_git=Set{UUID}();
             compat_names = String[]
             for pkg in pkgs
                 haskey(ctx.env.project.compat, pkg.name) && continue
-                pkgversion = Base.thispatch(ctx.env.manifest[pkg.uuid].version)
-                set_compat(ctx.env.project, pkg.name, string(pkgversion))
+                v = ctx.env.manifest[pkg.uuid].version
+                v === nothing && continue
```
IanButterworth, line 1528: I think we'd expect the package version to be populated here by then by e.g. `update_manifest!`.
The scenario you have isn't that the manifest ends up without a version, right?

topolarity: Yeah, that makes sense - I was worried that might be the case.

The MWE is just `Pkg.add(Pkg.PackageSpec(name="SuiteSparse", uuid="4607b0f0-06f3-5cda-b6b1-a6196a1729e9"))`

If I apply this workaround to get past the error message it looks like we do end up with a missing version in the Manifest:
```julia
[[deps.SuiteSparse]]
deps = ["Libdl", "LinearAlgebra", "Serialization", "SparseArrays"]
uuid = "4607b0f0-06f3-5cda-b6b1-a6196a1729e9"

[[deps.SuiteSparse_jll]]
deps = ["Artifacts", "Libdl", "libblastrampoline_jll"]
uuid = "bea87d4a-7f5b-5778-9afe-8cc45184846c"
version = "7.7.0+0"
```

topolarity: Looks like the same thing happens on 1.10 too

IanButterworth: @KristofferC I think we made it that stdlibs get versions recorded in the manifest, so this seems like a bug?

KristofferC: I feel we have a real issue in Pkg as described in https://lexi-lambda.github.io/blog/2019/11/05/parse-don-t-validate/ denoted as "shotgun parsing".

> Shotgun parsing is a programming antipattern whereby parsing and input-validating code is mixed with and spread across processing code—throwing a cloud of checks at the input, and hoping, without any systematic justification, that one or another would catch all the “bad” cases.

We have this `PackageSpec` object that:

- Users add fields to from the constructor.
- During the progress of a Pkg operation we discover various things and update those fields.

But it is hard to know what state the PackageSpec is in at some arbitrary point in the code. Is it valid or do more things have to be "discovered" on the way. In https://github.com/JuliaLang/Pkg.jl/pull/3783 this was particularly annoying because this was yet another way where new information was discovered and that surprised the existing code etc.

topolarity: This at least gets us back to the (admittedly a bit incorrect) behavior that we have on 1.10

Can we merge this and open up an issue for the deeper problem here?

PR #3976, opened 2024-08-02
mark command help blocks as julia-repl

IanButterworth:
Fixes some of but not all of
image

# This PR

Screenshot 2024-08-01 at 10 41 14 PM

### DIFF
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -153,7 +153,7 @@ precompiled before, or the precompile cache has been deleted by the LRU cache st
 (see `JULIA_MAX_NUM_PRECOMPILE_FILES`).
 
 **Examples**
-```
+```julia-repl
 pkg> add Example
 pkg> add --preserve=all Example
 pkg> add --weak Example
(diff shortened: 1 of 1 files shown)


src/REPLMode/command_declarations.jl
```diff
@@ -153,7 +153,7 @@ precompiled before, or the precompile cache has been deleted by the LRU cache st
 (see `JULIA_MAX_NUM_PRECOMPILE_FILES`).
 
 **Examples**
-```
+```julia-repl
```
giordano, line 156: Shouldn't this be text or some other neutral "language"? This isn't  really julia-repl.

IanButterworth: Ah I expected it to detect the other repl modes and color appropriately.

tecosaur: Hmm, we could extend

https://github.com/JuliaLang/julia/blob/19165bec4a3e2ed0de4f2f43bccc3fcac2d76662/stdlib/Markdown/src/render/terminal/render.jl#L123-L124

or it might be nice to make it possible for other "languages" to register their own highlighting functions (with a Markdown + REPL pkg ext for instance) then you could have `pkg-repl` for instance.

KristofferC:
I think the fix should be elsewhere. There isn't anything wrong with this markdown.

tecosaur:
This is an extension of https://github.com/JuliaLang/julia/issues/55253

I think what's really needed is a way to treat docstring markdown a little differently from other `Markdown.MD` objects. I'll look into this if nobody else does, but I can't make any promises on the timeline.

PR #3982, opened 2024-08-07
Backport #3276 to 1.10

gdalle:
Following a discussion on Slack with @KristofferC and @MasonProtter, it would be good to have better test env merging in the future LTS.

This PR takes the changes of #3276 and adds them to the `release-1.10` branch. Disclaimer: I have zero clue what this code does, this is pure copy-pasting.


gdalle:
@MasonProtter I think Julia 1.10.5 (just released) contains this

PR #3983, opened 2024-08-08
Add `julia_version_strict` option to `instantiate`

Octogonapus:
This PR adds a new kwarg to `instantiate`, `manifest_version_check_mode`, which can be used to turn the existing manifest version warnings into errors. We keep strict control over our Julia versions and dependencies, and never want to have a mismatch between the running Julia version and the manifest's Julia version. This kwarg would greatly simplify this behavior for that use case.

I'm not tied to this exact implementation; I could also implement the kwarg as a bool, or add a `:none` mode, etc.

### DIFF
--- a/test/manifests.jl
+++ b/test/manifests.jl
@@ -92,12 +92,12 @@ end
             m.julia_version = v"1.5.0"
             msg = r"The active manifest file has dependencies that were resolved with a different julia version"
             @test_logs (:warn, msg) Pkg.Types.check_manifest_julia_version_compat(m, env_manifest)
-            @test_throws PkgError Pkg.Types.check_manifest_julia_version_compat(m, env_manifest, julia_version_strict=true)
+            @test_throws Pkg.Types.PkgError Pkg.Types.check_manifest_julia_version_compat(m, env_manifest, julia_version_strict=true)
 
             m.julia_version = nothing
             msg = r"The active manifest file is missing a julia version entry"
             @test_logs (:warn, msg) Pkg.Types.check_manifest_julia_version_compat(m, env_manifest)
-            @test_throws PkgError Pkg.Types.check_manifest_julia_version_compat(m, env_manifest, julia_version_strict=true)
+            @test_throws Pkg.Types.PkgError Pkg.Types.check_manifest_julia_version_compat(m, env_manifest, julia_version_strict=true)
         end
     end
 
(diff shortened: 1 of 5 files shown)


IanButterworth:
I think this would make sense as `julia_strict::Bool=false`

Octogonapus:
`julia_strict` sounds too broad to me, unless I misunderstand and that's already an existing flag? I'm not opposed to making it a `bool`, though.

IanButterworth:
`julia_version_strict`?

test/manifests.jl
```diff
@@ -91,11 +91,11 @@ end
             m = Pkg.Types.read_manifest(env_manifest)
             m.julia_version = v"1.5.0"
             msg = r"The active manifest file has dependencies that were resolved with a different julia version"
-            @test_logs (:warn, msg) Pkg.Types.check_warn_manifest_julia_version_compat(m, env_manifest)
+            @test_logs (:warn, msg) Pkg.Types.check_manifest_julia_version_compat(m, env_manifest)
```
IanButterworth, line 94:
```suggestion
            @test_logs (:warn, msg) Pkg.Types.check_manifest_julia_version_compat(m, env_manifest)
            @test_throws PkgError Pkg.Types.check_manifest_julia_version_compat(m, env_manifest, julia_version_strict=true)
```

test/manifests.jl
```diff
@@ -91,11 +91,11 @@ end
...
 
             m.julia_version = nothing
             msg = r"The active manifest file is missing a julia version entry"
-            @test_logs (:warn, msg) Pkg.Types.check_warn_manifest_julia_version_compat(m, env_manifest)
+            @test_logs (:warn, msg) Pkg.Types.check_manifest_julia_version_compat(m, env_manifest)
```
IanButterworth, line 98:
```suggestion
            @test_logs (:warn, msg) Pkg.Types.check_manifest_julia_version_compat(m, env_manifest)
            @test_throws PkgError Pkg.Types.check_manifest_julia_version_compat(m, env_manifest, julia_version_strict=true)
```

IanButterworth (dismissed):

IanButterworth:
Just occurred to me that we should add a compat note to the docstring for the new kwarg.

IanButterworth (approved):
It'd be good to get another review on this, but LGTM

PR #3990, opened 2024-08-13
Revert "Backport #3276 to 1.10"

KristofferC:
Reverts JuliaLang/Pkg.jl#3982

This caused a scary failure mode where a devved package seems to not actually be the one getting used when testing stuff:

```
(1.10dev) pkg> st
Status `~/PkgEval/1.10dev/Project.toml`
  [31c24e10] Distributions v0.25.110
  [b5a8929a] RiskMeasures v0.4.0 `~/.julia/dev/RiskMeasures`

(1.10dev) pkg> test RiskMeasures
     Testing RiskMeasures
      Status `/tmp/jl_wsLGsE/Project.toml`
⌃ [31c24e10] Distributions v0.25.104
⌃ [b5a8929a] RiskMeasures v0.3.0
```

Note how RiskMeasures is devved but we use RiskMeasures 0.3.0 anyway in the testing.

A new 1.10 needs to go out soon so I will just revert this but we should look more into what is going on here (and add a test for it).

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1746,6 +1746,7 @@ function sandbox(fn::Function, ctx::Context, target::PackageSpec, target_path::S
                  force_latest_compatible_version::Bool=false,
                  allow_earlier_backwards_compatible_versions::Bool=true,
                  allow_reresolve::Bool=true)
+    active_manifest = manifestfile_path(dirname(ctx.env.manifest_file))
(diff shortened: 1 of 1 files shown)


KristofferC:
This bug also happens on 1.11. So I guess https://github.com/JuliaLang/Pkg.jl/pull/3276 is buggy!

Issue #3992, opened 2024-08-13
Download progress bar for packages flicker


KristofferC:
[Screencast from 08-13-2024 06:00:24 PM.webm](https://github.com/user-attachments/assets/ee18622d-d420-421c-b279-3f504332b338)

KristofferC:
On Julia 1.6 it is rock solid so that at least gives us some clue.

KristofferC:
Seems to be due to a change in Julia. Running on `release-1.6` with julia 1.6 it is solid, running it with julia 1.7 it flickers. So probably have to bisect julia itself for this.

KristofferC:
Very annoying to try to bisect this because I am hitting julia compile errors everywhere between 1.6 and 1.7.

KristofferC:
Could be something libuv related? I don't have a good guess.

KristofferC:
Actually bisected to https://github.com/JuliaLang/julia/pull/40684 so it is a Pkg change.

https://github.com/JuliaLang/Pkg.jl/commit/abe90860 looks like a possible candidate.

Issue #3581, opened 2023-08-11
Package installation overwrites too many terminal lines


IanButterworth:
I thought there was an issue about this but can't find it.

Something in this process is overwriting too many lines and eating up terminal history

```
$ julia
               _
   Installed AbstractFFTs ──────────── v1.5.0
   Installed TimeZones ─────────────── v1.11.0
   Installed LoopVectorization ─────── v0.12.165
   Installed OrderedCollections ────── v1.6.2
   Installed KernelAbstractions ────── v0.9.8
```

I can't reliably reproduce this, but I do see it from time to time

PR #3993, opened 2024-08-14
fix flickering of progress bar on package installation

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3992.
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3581

Sigh.


IanButterworth:
I think this fixes https://github.com/JuliaLang/Pkg.jl/issues/3581

Because the print bottom function always prints.

PR #3994, opened 2024-08-14
fix stdlib recompile test

IanButterworth:
Missed this one before somehow

Screenshot 2024-08-14 at 12 36 26 PM


IanButterworth:
This is a build system bug https://github.com/JuliaLang/julia/issues/55491

PR #3952, opened 2024-07-13
Parallelize artifact downloads. Update progress bar styling.

IanButterworth:
## This PR
```
% JULIA_DEPOT_PATH=$(mktemp -d): JULIA_PKG_PRECOMPILE_AUTO=0 ./julia --project=../Pkg.jl -t1 -e 'import Pkg; Pkg.activate(temp=true); @time Pkg.add("FFMPEG")'
...
 12.999424 seconds (9.81 M allocations: 915.741 MiB, 2.30% gc time, 73 lock conflicts, 11.69% compilation time: 16% of which was recompilation)
```

## Master
```
% JULIA_DEPOT_PATH=$(mktemp -d): JULIA_PKG_PRECOMPILE_AUTO=0 ./julia-t1 -e 'import Pkg; Pkg.activate(temp=true); @time Pkg.add("FFMPEG")'
...
 23.540455 seconds (9.37 M allocations: 887.289 MiB, 0.82% gc time, 123 lock conflicts, 5.85% compilation time: 18% of which was recompilation)

```

https://github.com/user-attachments/assets/8f1024d6-f255-420c-a70e-becdbfa1dd86

Todo
- [x] parallel progress printing 
- [ ] effective interrupt handling (proven to be hard to do because of Downloads.jl being so async)
  - [ ] needs testing
- [x] error handling & reporting
  - [x] needs testing
- [x] code tidy and remove duplication (will implement a new MiniMultiProgressBar type in a follow-on)
- [x] simple printing for CI etc.
- [x] show progress in bytes 
  - [ ] rate/s ?
  - [ ] total across all artifacts?

PR #3995, opened 2024-08-14
show file size in progress bar when downloading artifacts

KristofferC:
Maybe nice?

[Screencast from 08-14-2024 03:58:27 PM.webm](https://github.com/user-attachments/assets/214af246-92d0-41bd-bf40-9a5aaebdca07)


IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/3952 has this so I guess it's good to do this for all artifact downloads like lazy ones outside of Pkg install time.

IanButterworth:
But note that I changed the `percentage` field to a `mode` symbol, as these progress bars are internal to Pkg (we even copied them over to Base for precompilepkgs)

PR #3997, opened 2024-08-16
Allow comma separated packages for ]dev

mhauru:
In line with how e.g. `]add` works.

### DIFF
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -122,6 +122,8 @@ temp_pkg_dir(;rm=false) do project_path; cd(project_path) do;
     mktempdir() do tmp_dev_dir
     withenv("JULIA_PKG_DEVDIR" => tmp_dev_dir) do
         pkg"develop Example"
+        pkg"develop Example,PackageCompiler"
+        pkg"develop Example PackageCompiler"
 
         # Copy the manifest + project and see that we can resolve it in a new environment
         # and get all the packages installed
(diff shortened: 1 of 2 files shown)


test/repl.jl
```diff
@@ -122,6 +122,8 @@ temp_pkg_dir(;rm=false) do project_path; cd(project_path) do;
     mktempdir() do tmp_dev_dir
     withenv("JULIA_PKG_DEVDIR" => tmp_dev_dir) do
         pkg"develop Example"
+        pkg"develop Example,PackageCompiler"
+        pkg"develop Example PackageCompiler"
```
IanButterworth, line 126:
```suggestion
        pkg"develop Example1, Example2"
        pkg"develop Example1 Example2"
```

mhauru: I thought the use of `Example` relied on this package: https://github.com/JuliaLang/Example.jl i.e. that these had to actually be real packages. Very happy to change this if that's not the case, the use of PackageCompiler here is very clunky.

IanButterworth: It's not a big deal. `Package1 Package2` also works

IanButterworth: I missed that this was in tests. Thought it was the docstring.. whoops. 

LGTM then

IanButterworth:
I'm marking this for backport as the omission of this seems like a bug

PR #4000, opened 2024-08-19
update package extension naming docs

aplavin:
As discussed in https://github.com/JuliaLang/Pkg.jl/pull/3600, Pkg output always displays the parent package name next to the extension name. The naming from the current pkg docs example (`PlottingContourExt`) makes Pkg output super verbose, like `Plotting -> PlottingContourExt`. Especially for longer package names, it can easily exceed typical line length.

Here, I update the docs to show shorter nonredundant naming, `ContourExt` (leading to `Plotting -> ContourExt` output). Many packages already follow this short naming, and presumably more will do if it becomes the official recommended way.

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -416,7 +416,7 @@ Contour = "d38c429a-6771-53c6-b99e-75d170b6e991"
 # name of extension to the left
 # extension dependencies required to load the extension to the right
 # use a list for multiple extension dependencies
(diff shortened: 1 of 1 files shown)


IanButterworth (approved):
LGTM but will hold off merging for @KristofferC's review

IanButterworth:
I'll merge tomorrow if there are no objections.

KristofferC:
Fine by me

PR #4001, opened 2024-08-20
Retry artifact rename if it fails

### LINKED ISSUE #3822: Artifact folder could not be made on Windows.
When running `Pkg.add("CairoMakie")` in an empty project I'm getting this, `Error: "C:\\Users\\meyer\\.julia\\artifacts\\a8244d6d23cbb895fcd39dd3eddb859a0c05d1c6" could not be made`, which I seems to occur as the atomic rename fails after https://github.com/JuliaLang/Pkg.jl/pull/3768. 

However, before getting there, I hit https://github.com/JuliaLang/julia/issues/34700, but that could be due to a failure related to above, but the `rm` is in a final block...

To "fix" the problem, I first hacked `Base.rm`
```julia
function Base.rm(path::String; force::Bool=false, recursive::Bool=false)
   if !(force && recursive)
       return invoke(Base.rm, Tuple{AbstractString}, path; force, recursive)
   else
       max_attempts = 3
       attempts = 0
       while attempts < max_attempts
           attempts += 1
           try
               invoke(Base.rm, Tuple{AbstractString}, path; recursive=true, force=true)
           catch err
               if isa(err, Base.IOError) && attempts < max_attempts
                   println("Trying again for \"", path, "\"")
                   attempts == (max_attempts - 1) && sleep(1.0)
                   continue
               else
                   println("Failed for \"", path, "\"")
                   rethrow(err)
               end
           end
       end
   end
end
```
(but this could just as well be done in `create_artifact`) 

And modified `_mv_temp_artifact_dir` (which I couldn't hack since the dispatch is concrete...)
```julia
function _mv_temp_artifact_dir(temp_dir::String, new_path::String)::Nothing
    if !isdir(new_path)
        # This next step is like
        # `mv(temp_dir, new_path)`.
        # However, `mv` defaults to `cp` if `rename` returns an error.
        # `cp` is not atomic, so avoid the potential of calling it.
        err = ccall(:jl_fs_rename, Int32, (Cstring, Cstring), temp_dir, new_path)
        # Ignore rename error, but ensure `new_path` exists.
        if !isdir(new_path)
            println("Just do cp 💀")
            mv(temp_dir, new_path)
        end
        if !isdir(new_path)
            error("$(repr(new_path)) could not be made")
        end
        chmod(new_path, filemode(dirname(new_path)))
        set_readonly(new_path)
    end
    nothing
end
```

`Just do cp 💀` is printed twice, along with a couple of `Trying again for ...`. 
In fact, it seems like the `x264_jll` and `Pixman_jll` are the ones that needs retries, not sure if by chance or anything special about these. 

Strangely, trying many times, it seemed to work, but fails again if I empty the artifacts folder...

<details> <summary> Versioninfo </summary>

```julia
Julia Version 1.10.2
Commit bd47eca2c8 (2024-03-01 10:14 UTC)
Build Info:
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 8 × 11th Gen Intel(R) Core(TM) i7-1185G7 @ 3.00GHz
  WORD_SIZE: 64
  LIBM: libopenlibm
  LLVM: libLLVM-15.0.7 (ORCJIT, tigerlake)
Environment:
  JULIA_EDITOR = code
  JULIA_NUM_THREADS =
```
</details>

nhz2:
Hopefully fixes #3822 

This uses a strategy from https://github.com/isaacs/node-graceful-fs/blob/234379906b7d2f4c9cfeb412d2516f42b0fb4953/polyfills.js#L87-L95 to retry the rename for about a minute to work around anti-virus software temporarily locking a directory with newly created files. This strategy is also used in vscode, Ref: https://github.com/microsoft/vscode/pull/188899

I am not restricting this fix to only Windows, because similar issues can happen on Linux, Ref: https://github.com/isaacs/node-graceful-fs/issues/133

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -65,7 +65,18 @@ Either rename the directory at `temp_dir` to `new_path` and set it to read-only
 or if `new_path` artifact already exists try to do nothing.
 """
 function _mv_temp_artifact_dir(temp_dir::String, new_path::String)::Nothing
-    if !isdir(new_path)
+    # Sometimes a rename can fail because the temp_dir is locked by
+    # anti-virus software scanning the new files.
+    # In this case we want to sleep and try again.
+    # I am using the list of error codes to retry from:
+    # https://github.com/isaacs/node-graceful-fs/blob/234379906b7d2f4c9cfeb412d2516f42b0fb4953/polyfills.js#L87
+    # Retry for up to about 60 seconds by retrying 20 times with exponential backoff.
+    retry = 0
+    max_num_retries = 20 # maybe this should be configurable?
+    sleep_amount = 0.01 # seconds
+    max_sleep_amount = 5.0 # seconds
+    while true
+        isdir(new_path) && return
         # This next step is like
         # `mv(temp_dir, new_path)`.
         # However, `mv` defaults to `cp` if `rename` returns an error.
@@ -75,14 +86,19 @@ function _mv_temp_artifact_dir(temp_dir::String, new_path::String)::Nothing
             # rename worked
             chmod(new_path, filemode(dirname(new_path)))
             set_readonly(new_path)
+            return
         else
-            # Ignore rename error, if `new_path` exists.
-            if !isdir(new_path)
+            # Ignore rename error if `new_path` exists.
+            isdir(new_path) && return
+            if retry < max_num_retries && err ∈ (Base.UV_EACCES, Base.UV_EPERM, Base.UV_EBUSY)
+                sleep(sleep_amount)
+                sleep_amount = min(sleep_amount*2.0, max_sleep_amount)
+                retry += 1
+            else
                 Base.uv_error("rename of $(repr(temp_dir)) to $(repr(new_path))", err)
             end
         end
     end
-    nothing
 end
 
 """


nhz2:
@xlxs4 if you can still reproduce the problem you were having in #3822, could you give this PR a try to see if it helps at all?

src/Artifacts.jl
```diff
@@ -75,14 +86,19 @@ function _mv_temp_artifact_dir(temp_dir::String, new_path::String)::Nothing
...
+            return
         else
-            # Ignore rename error, if `new_path` exists.
-            if !isdir(new_path)
+            # Ignore rename error if `new_path` exists.
+            isdir(new_path) && return
+            if retry < max_num_retries && err ∈ (Base.UV_EACCES, Base.UV_EPERM, Base.UV_EBUSY)
+                sleep(sleep_amount)
+                sleep_amount = min(sleep_amount*2.0, max_sleep_amount)
+                retry += 1
```
staticfloat, line 96: If this can take up to 60s, I think we should print something out to the user.  Something like:

```
# Warn once, and only if we've slept a non-trivial amount of time already
if retry == 8
    @warn("Artifact installation failed, retrying for up to 60s.  Check to see if an anti-virus may be interfering.")
end
```

IanButterworth: If this is hit during the new parallel artifact download process, then outputting a log will mess with the multiprogress printing (sorry). We'll need some way to keep that tidy

nhz2: image
How about adding other status messages to the right of the progress bar where it currently has "XX MiB/XX MiB" like "unpacking", "verifying tree hash", and "installing"?

IanButterworth: I think the location works but in the default happy path I wouldn't show any further detail

xlxs4:
> @xlxs4 if you can still reproduce the problem you were having in #3822, could you give this PR a try to see if it helps at all?

I haven't been able to repro the issue... will let you know if I do

src/Artifacts.jl
IanButterworth: Sounds good but I'd do it such that if status is non empty just show the status string after the name and not the progress bar or ending ratio. Just so that we're more likely to fit in the terminal width.

IanButterworth (approved):

src/Artifacts.jl
IanButterworth: I think we want to backport this to 1.10 & 1.11, but when we do we'll want to add a log message about it taking a long time, as suggested above, given the parallel download thing isn't going to be backported.

IanButterworth:
As this is currently, the status that will print will explain the step that's taking a long time, but won't mention that it could be the antivirus causing it. Should that be changed? Might be annoying to see the same long explanation multiple times as I assume the issue will affect multiple parallel install tasks?

src/Artifacts.jl
nhz2: Compared to the time it already takes to download and unpack artifacts on Windows with an aggressive anti-virus I don't think one minute extra is something to create a warning about, as long as the artifact eventually gets installed.

nhz2:
I was able to test this PR using `LD_PRELOAD` to simulate an unreliable `rename`. If I make `rename` only sometimes fail, installing works. If I make `rename` always error with `EBUSY`, I get the `moving to artifact store` message for a minute, then Pkg tries to download from the next URL, and I get `moving to artifact store` again for a minute. Finally, I get an error:

```julia-repl
julia> Pkg.pkg"add x265_jll"
   Resolving package versions...
  Installing artifacts  ╸━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 0/1
ERROR: Failed to install some artifacts:

Unable to automatically download/install artifact 'x265' from sources listed in '/home/nathan/.julia/packages/x265_jll/70izA/Artifacts.toml'.
Sources attempted:
- https://pkg.julialang.org/artifact/00023f2f62a04af7531278ce1a689ff93800ffa1
    Error: IOError: rename of "/home/nathan/.julia/artifacts/jl_Aly3ZD" to "/home/nathan/.julia/artifacts/00023f2f62a04af7531278ce1a689ff93800ffa1": resource busy or locked (EBUSY)
- https://github.com/JuliaBinaryWrappers/x265_jll.jl/releases/download/x265-v3.6.0+0/x265.v3.6.0.x86_64-linux-gnu.tar.gz
    Error: IOError: rename of "/home/nathan/.julia/artifacts/jl_p180ux" to "/home/nathan/.julia/artifacts/00023f2f62a04af7531278ce1a689ff93800ffa1": resource busy or locked (EBUSY)
Stacktrace:
  [1] pkgerror(msg::String)
    @ Pkg.Types ~/github/Pkg.jl/src/Types.jl:68
  [2] download_artifacts(ctx::Pkg.Types.Context; platform::Base.BinaryPlatforms.Platform, julia_version::VersionNumber, verbose::Bool)
    @ Pkg.Operations ~/github/Pkg.jl/src/Operations.jl:978
  [3] download_artifacts
    @ ~/github/Pkg.jl/src/Operations.jl:827 [inlined]
  [4] add(ctx::Pkg.Types.Context, pkgs::Vector{…}, new_git::Set{…}; allow_autoprecomp::Bool, preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform, target::Symbol)
    @ Pkg.Operations ~/github/Pkg.jl/src/Operations.jl:1656
  [5] add
    @ ~/github/Pkg.jl/src/Operations.jl:1621 [inlined]
  [6] add(ctx::Pkg.Types.Context, pkgs::Vector{…}; preserve::Pkg.Types.PreserveLevel, platform::Base.BinaryPlatforms.Platform, target::Symbol, allow_autoprecomp::Bool, kwargs::@Kwargs{…})
    @ Pkg.API ~/github/Pkg.jl/src/API.jl:306
  [7] add(pkgs::Vector{Pkg.Types.PackageSpec}; io::IOContext{IO}, kwargs::@Kwargs{})
    @ Pkg.API ~/github/Pkg.jl/src/API.jl:159
  [8] add(pkgs::Vector{Pkg.Types.PackageSpec})
    @ Pkg.API ~/github/Pkg.jl/src/API.jl:148
  [9] do_cmd(command::Pkg.REPLMode.Command, io::IOContext{IO})
    @ Pkg.REPLMode ~/github/Pkg.jl/src/REPLMode/REPLMode.jl:407
 [10] do_cmds(commands::Vector{Pkg.REPLMode.Command}, io::IOContext{IO})
    @ Pkg.REPLMode ~/github/Pkg.jl/src/REPLMode/REPLMode.jl:393
 [11] do_cmds(input::String, io::IOContext{IO})
    @ Pkg.REPLMode ~/github/Pkg.jl/src/REPLMode/REPLMode.jl:383
 [12] pkgstr(str::String)
    @ Pkg.REPLMode ~/github/Pkg.jl/src/REPLMode/REPLMode.jl:450
 [13] top-level scope
    @ REPL[3]:1
Some type information was truncated. Use `show(err)` to see complete types.
```

IanButterworth:
Great. Seems like a big improvement. 

A thought for a follow on PR would be adding something to the docs about why that could happen.

PR #4002, opened 2024-08-22
advertise `undo` and `redo` when there are changes to an env

IanButterworth:
Seems a lot of people haven't discovered this functionality, so advertise it.
Screenshot 2024-08-22 at 9 32 26 PM


fredrikekre:
Will probably feel old/spammy after the first time you see it.

IanButterworth:
Maybe we could:
1. Only advertise once per session
2. Remember if the user has ever used `undo/redo` across julia versions and not advertise if they have. Pkg repl history search?

PR #4015, opened 2024-09-04
REPLMode.jl: set compiler_options optimize = 1

PallHaraldsson:
Fixes 3x PythonCall startup regression because of: https://github.com/JuliaPy/CondaPkg.jl/issues/145 (at least least if backported to 1.11).

```
$ julia +1.11 --compile=min

julia> @time Pkg.REPLMode.gen_help()
  0.180643 seconds (91.88 k allocations: 2.218 MiB, 17.97% compilation time)

vs.
julia> @time Pkg.REPLMode.gen_help()
  2.129547 seconds (871.64 k allocations: 44.875 MiB, 5.32% gc time, 99.94% compilation time)
```

That one function slows down CondaPkg, and therefore also PythoCall, by that amount on 1.11

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -2,11 +2,7 @@
 
 module REPLMode
 
-if isdefined(Base, :Experimental) &&
-   isdefined(Base.Experimental, Symbol("@compiler_options"))
-    # Note: compile=min makes --code-coverage not work
-    @eval Base.Experimental.@compiler_options optimize = 1 infer = false  # compile=min
-end
+@eval Base.Experimental.@compiler_options optimize = 1
 
 using Markdown, UUIDs, Dates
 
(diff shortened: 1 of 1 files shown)


PallHaraldsson:
I think I can only selectively change opt for a module, not one function. That's the one I care about, do you think any of the others are speed-critical?

KristofferC:
What version are you on? This is for me:

```julia-rpel
julia> @time Pkg.REPLMode.gen_help()^C

julia> VERSION
v"1.11.0-rc3"

julia> using Pkg

julia> @time Pkg.REPLMode.gen_help()
  0.000287 seconds (2.50 k allocations: 134.836 KiB)
  Welcome to the Pkg REPL-mode. To return to the julia>
...
```

KristofferC:
The optimize=1 is most likely worth it. I am a bit worried about the `infer=false` because that can give a lot of invalidations.

Issue #3549, opened 2023-07-14
`add --weak` option


oscardssmith:
It would be great if there were an easy way to add weakdeps from the repl rather than editing the Project.toml file.

liuyxpp:
So if I am not missed anything, currently I still have to edit Project.toml directly to add weak deps?

oscardssmith:
correct

IanButterworth:
Technically you can already add weak deps via `--weak`. What cannot be done is automatically create an extension that depends on them, or add that weak dep to an existing extension.

IanButterworth:
Proposal here https://github.com/JuliaLang/Pkg.jl/pull/4016

liuyxpp:
> Technically you can already add weak deps via `--weak`. That's [new in 1.11](https://github.com/JuliaLang/Pkg.jl/pull/3708). What cannot be done is automatically create an extension that depends on them, or add that weak dep to an existing extension.

That's great! I just don't want to deal those UUID manually.

PR #4016, opened 2024-09-08
add option to create ext when adding weakdeps

IanButterworth:
Closes https://github.com/JuliaLang/Pkg.jl/issues/3549

Screenshot 2024-09-08 at 8 24 31 AM

Creates
```
...
[weakdeps]
Chairmarks = "0ca39b1e-fe0b-4e98-acfc-b1656634c4de"
ImageIO = "82e4d734-157c-48bb-816b-45c225c6df19"

[extensions]
ChairmarksImageIOExt = ["Chairmarks", "ImageIO"]
ExampleExt = ["Example"]
```


KristofferC:
I'm bit sure this should be part of  `add`. It feels more like a `generate` thing.

IanButterworth:
I think there's elements of both. Not sure generate should add a weak dep, conversely. 

It could be two steps:
```
pkg> add --weak Foo
pkg> generate --extension Foo
```

So those each are logical to add. 

But for convenience I think offering to do the generate as part of `add` seems common enough to make sense.

Perhaps the default for noninteractive should be `"n"` though, so that would require two api calls.

tecosaur:
Could `--extension` be made a flag of `add` instead? E.g.

```
pkg> add --weak Foo --ext FooExt
```

PR #4022, opened 2024-09-18
warn if General is installed via the old slow methods

IanButterworth:
Seeing https://discourse.julialang.org/t/adding-package-to-new-environment-takes-long-time/119518/5 makes me wonder how many people out there might be stuck with General installed via the old methods, so this adds a note if so.

Screenshot 2024-09-17 at 11 23 31 PM

~~I can't imagine anyone intentionally uses either on newer versions for General..~~
If anyone is intentionally using an old format this check can be disabled via
`JULIA_PKG_GEN_REG_FMT_CHECK=0`

### DIFF
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -447,6 +447,15 @@ function update(regs::Vector{RegistrySpec}; io::IO=stderr_f(), force::Bool=true,
                                     registry_update_log[string(reg.uuid)] = now()
                                     @label done_tarball_read
                                 else
+                                    if reg.name == "General"
+                                        @info """
+                                            The General registry is installed via unpacked tarball.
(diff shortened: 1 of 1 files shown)


src/Registry/Registry.jl
```diff
@@ -447,6 +447,12 @@ function update(regs::Vector{RegistrySpec}; io::IO=stderr_f(), force::Bool=true,
                                     registry_update_log[string(reg.uuid)] = now()
                                     @label done_tarball_read
                                 else
+                                    @info """
```
IanButterworth, line 450:
```suggestion
                                    reg.name == "General"&& @info """
```

KristofferC:
Seems like a good idea.

IanButterworth:
I'll add an env var disable just in case there are people that wish to stick with the old format.

PR #4024, opened 2024-09-18
Use same PrecompileTools.jl approach as REPL to precompile

IanButterworth:
Without this: (exposing `pkg_precompile` which it usually isn't)
```
julia> @time Pkg.pkg_precompile()
  1.796371 seconds (2.23 M allocations: 125.344 MiB, 4.53% gc time, 66.34% compilation time: 9% of which was recompilation)
```

With
```
julia> @time Pkg.pkg_precompile()
  0.929547 seconds (555.18 k allocations: 35.390 MiB, 2.02% gc time, 29.93% compilation time: 32% of which was recompilation)
```

The code is quite small. Do we want to make a utility in Base or just copy it here?

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -165,7 +165,31 @@ let
         return nothing
     end
 
-    if Base.generating_output()
-        pkg_precompile()
+    # Copied from REPL (originally PrecompileTools.jl)
+    function check_edges(node)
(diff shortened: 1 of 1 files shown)


KristofferC:
Nice nice. For now I think we can just keep it here?

PR #4025, opened 2024-09-24
Fix julia#55850 by using safe_realpath instead of abspath in projname

christiangnrd:
Replaces `REPLExt.projname` with `REPL.projname` since they are identical.

Will solve https://github.com/JuliaLang/julia/issues/55850 for situations where Pkg is loaded once https://github.com/JuliaLang/julia/pull/55851 is merged.

Test will be covered by https://github.com/JuliaLang/julia/pull/55851.

### DIFF
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -733,4 +733,14 @@ end
     end
 end
 
+@testset "JuliaLang/julia #55850" begin
+    tmp_55850 = mktempdir()
+    tmp_sym_link = joinpath(tmp_55850, "sym")
+    symlink(tmp_55850, tmp_sym_link; dir_target=true)
+    withenv("JULIA_DEPOT_PATH" => tmp_sym_link, "JULIA_LOAD_PATH" => nothing) do
+        prompt = readchomp(`$(Base.julia_cmd()[1]) --startup-file=no -e "using Pkg: Pkg, Types; import REPL; const REPLExt = Base.get_extension(Pkg, :REPLExt); print(REPLExt.projname(Types.find_project_file()))"`)
+        @test prompt == "@v$(VERSION.major).$(VERSION.minor)"
+    end
+end
+
 end # module
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -737,7 +737,7 @@ end
     tmp_55850 = mktempdir()
     tmp_sym_link = joinpath(tmp_55850, "sym")
     symlink(tmp_55850, tmp_sym_link; dir_target=true)
-    withenv("JULIA_DEPOT_PATH" => tmp_sym_link * ":", "JULIA_LOAD_PATH" => nothing) do
+    withenv("JULIA_DEPOT_PATH" => tmp_sym_link * (Sys.iswindows() ? ";" : ":"), "JULIA_LOAD_PATH" => nothing) do
         prompt = readchomp(`$(Base.julia_cmd()[1]) --project=$(dirname(@__DIR__)) --startup-file=no -e "using Pkg, REPL; Pkg.activate(io=devnull); REPLExt = Base.get_extension(Pkg, :REPLExt); print(REPLExt.promptf())"`)
         @test prompt == "(@v$(VERSION.major).$(VERSION.minor)) pkg> "
     end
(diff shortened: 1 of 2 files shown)


christiangnrd:
@IanButterworth This is also ready for review.

IanButterworth:
Just to highlight because the diff isn't clear in either PR, the difference between the two projnames is that Pkg uses
`Types.read_project` to get the project name, where REPL's directly uses `Base.TOML.Parser`.

I don't see an issue switching over because there's no side effects/caching of the project in either.

Pkg's
```
function projname(project_file::String)
    project = try
        Types.read_project(project_file)
    catch
        nothing
    end
    if project === nothing || project.name === nothing
        name = basename(dirname(project_file))
    else
        name = project.name::String
    end
    for depot in Base.DEPOT_PATH
        envdir = joinpath(depot, "environments")
        if startswith(abspath(project_file), abspath(envdir))
            return "@" * name
        end
    end
    return name
end
```
REPL's
```
function projname(project_file::String)
    if isfile(project_file)
        name = try
            # The `nothing` here means that this TOML parser does not return proper Dates.jl
            # objects - but that's OK since we're just checking the name here.
            p = Base.TOML.Parser{nothing}()
            Base.TOML.reinit!(p, read(project_file, String); filepath=project_file)
            proj = Base.TOML.parse(p)
            get(proj, "name", nothing)
        catch
            nothing
        end
    else
        name = nothing
    end
    if name === nothing
        name = basename(dirname(project_file))
    end
    for depot in Base.DEPOT_PATH
        envdir = joinpath(depot, "environments")
        if startswith(abspath(project_file), abspath(envdir))
            return "@" * name
        end
    end
    return name
end
```

IanButterworth (dismissed):

KristofferC:
In my opinion I rather have some code duplication than start reaching into the internals of things. But that's just my opinion.

IanButterworth:
Ok, let's just implement the https://github.com/JuliaLang/julia/pull/55851 fix in both then.

Sorry @christiangnrd for the flip flop

test/repl.jl
```diff
@@ -733,4 +733,14 @@ end
     end
 end
 
+@testset "JuliaLang/julia #55850" begin
+    tmp_55850 = mktempdir()
+    tmp_sym_link = joinpath(tmp_55850, "sym")
+    symlink(tmp_55850, tmp_sym_link; dir_target=true)
+    withenv("JULIA_DEPOT_PATH" => tmp_sym_link, "JULIA_LOAD_PATH" => nothing) do
```
IanButterworth, line 740: Without a trailing ":" the shipped stdlib caches aren't reachable so this test will needlessly precompile Pkg etc.
```suggestion
    withenv("JULIA_DEPOT_PATH" => tmp_sym_link * ":", "JULIA_LOAD_PATH" => nothing) do
```

christiangnrd: Should I also fix it in the julia PR (and the test where I borrowed this code from)?

IanButterworth: Yeah needed there too I believe

christiangnrd:
I'm stuck trying to figure out how to use the in-development version of Pkg while loading the default environment for the test.

Or is it fine as-is since this won't be a problem once merged?

test/repl.jl
```diff
@@ -733,4 +733,14 @@ end
     end
 end
 
+@testset "JuliaLang/julia #55850" begin
+    tmp_55850 = mktempdir()
+    tmp_sym_link = joinpath(tmp_55850, "sym")
+    symlink(tmp_55850, tmp_sym_link; dir_target=true)
+    withenv("JULIA_DEPOT_PATH" => tmp_sym_link * ":", "JULIA_LOAD_PATH" => nothing) do
+        prompt = readchomp(`$(Base.julia_cmd()[1]) --startup-file=no -e "using Pkg: Pkg, Types; import REPL; const REPLExt = Base.get_extension(Pkg, :REPLExt); print(REPLExt.projname(Types.find_project_file()))"`)
+        @test prompt == "@v$(VERSION.major).$(VERSION.minor)"
```
IanButterworth, line 742:
```suggestion
        prompt = readchomp(`$(Base.julia_cmd()[1]) --project=$(dirname(@__DIR__)) --startup-file=no -e "using Pkg, REPL; Pkg.activate(io=devnull); REPLExt = Base.get_extension(Pkg, :REPLExt); print(REPLExt.promptf())"`)
        @test prompt == "(@v$(VERSION.major).$(VERSION.minor)) pkg> "
```

christiangnrd:
This seems good to go on all platforms now.

IanButterworth (approved):

PR #4027, opened 2024-09-25
[release-1.10] Manually backport #4025

christiangnrd:
I'm investigating. Seems like appending `:` to `JULIA_DEPOT_PATH` hides the bug.

### DIFF
--- a/src/REPLMode/REPLMode.jl
+++ b/src/REPLMode/REPLMode.jl
@@ -10,7 +10,7 @@ import REPL: TerminalMenus
 
 import ..casesensitive_isdir, ..OFFLINE_MODE, ..linewrap, ..pathrepr
 using ..Types, ..Operations, ..API, ..Registry, ..Resolve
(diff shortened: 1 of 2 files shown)


IanButterworth:
Pkg is in the sysimage on 1.10. Using the development version is more complicated.. I might suggest backporting this without tests 🙈

christiangnrd:
> Pkg is in the sysimage on 1.10. Using the development version is more complicated.. I might suggest backporting this without tests 🙈

Hahaha after the amount of fixes it took me to get the main PR test right, that's a terrifying proposition.

The test should be valid now.

christiangnrd:
Luckily the test fails even with the `:` on master so I didn't just add a minute to the 1.12 Pkg test suite 😅.

I think I should remove the test here. Now that I think about it, recompiling the whole library just to test a bug so obscure it's been in the language for 4.5 years might not be worth it.

Lmk what you think.

IanButterworth:
Thanks!

PR #4029, opened 2024-09-26
test fixes because Example v0.5.4 and v0.5.5 were registered

IanButterworth:
See https://buildkite.com/julialang/julia-master/builds/40302#01922f49-90f7-4410-9c2b-53fda2abb7a6/847-875

because https://github.com/JuliaLang/Example.jl/releases/tag/v0.5.5 and https://github.com/JuliaLang/Example.jl/releases/tag/v0.5.4

Our tests assumed that there was a version on the tip of Example master that was unregistered...

### DIFF
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -319,7 +319,8 @@ temp_pkg_dir() do project_path
                 @test isfile(joinpath(devdir, TEST_PKG.name, "deps", "deps.jl"))
                 Pkg.test(TEST_PKG.name)
                 Pkg.free(TEST_PKG.name)
-                @test Pkg.dependencies()[TEST_PKG.uuid].version == old_v
+                @test Pkg.dependencies()[TEST_PKG.uuid].version < v"100.0.0"
+                @test Pkg.dependencies()[TEST_PKG.uuid].version >= old_v
             end
         end
     end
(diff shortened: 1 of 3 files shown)


test/pkg.jl
```diff
@@ -277,7 +277,10 @@ temp_pkg_dir() do project_path
                 @test_throws PkgError Pkg.develop(Pkg.PackageSpec(url="bleh", rev="blurg"))
                 Pkg.develop(TEST_PKG.name)
                 @test isinstalled(TEST_PKG)
-                @test Pkg.dependencies()[TEST_PKG.uuid].version > old_v
+                # This test is disabled because it relied on Example.jl having a version
+                # number in the Project.toml file on master that wasn't released, which
+                # is no longer true.
+                # @test Pkg.dependencies()[TEST_PKG.uuid].version > old_v
```
DilumAluthge, line 283: I suppose we could do `@test Pkg.dependencies()[TEST_PKG.uuid].version >= old_v`, but I don't know if that's a useful test to have.

DilumAluthge (dismissed):

test/pkg.jl
IanButterworth: We could add an old version above then restore this check

DilumAluthge: Seems reasonable to me.

DilumAluthge (approved):

DilumAluthge:
@IanButterworth Thanks for fixing this so quick.

DilumAluthge (dismissed):

DilumAluthge (approved):

PR #4038, opened 2024-10-07
smoothen progress bar

KristofferC:
After suggestion by @tecosaur, also use the heavy right box drawing character.  Can be tested with:

```julia
function test_progress(n, dt)
   io = stdout
   bar = Pkg.MiniProgressBars.MiniProgressBar(; indent=1, header = "Downloading packages", color = Base.info_color(),
                                    mode=:int, always_reprint=true)
   bar.max = n
   for i in 1:n
      bar.current = i
      Pkg.MiniProgressBars.show_progress(io, bar)
      sleep(dt)
   end
end
```

Doing something like `test_progress(1000, 0.001)` shows some difference vs the old behavior.

### DIFF
--- a/src/MiniProgressBars.jl
+++ b/src/MiniProgressBars.jl
@@ -72,8 +72,9 @@ function show_progress(io::IO, p::MiniProgressBar; termwidth=nothing, carriagere
     end
     termwidth = @something termwidth displaysize(io)[2]
     max_progress_width = max(0, min(termwidth - textwidth(p.header) - textwidth(progress_text) - 10 , p.width))
-    n_filled = ceil(Int, max_progress_width * perc / 100)
-    n_left = max_progress_width - n_filled
+    n_filled = floor(Int, max_progress_width * perc / 100)
+    partial_filled = (max_progress_width * perc / 100) - n_filled
+    n_left = max_progress_width - n_filled - 1
     headers = split(p.header)
     to_print = sprint(; context=io) do io
         print(io, " "^p.indent)
@@ -88,8 +89,15 @@ function show_progress(io::IO, p::MiniProgressBar; termwidth=nothing, carriagere
             print(io, p.status)
         else
             printstyled(io, "━"^n_filled; color=p.color)
-            printstyled(io, perc >= 95 ? "━" : "╸"; color=p.color)
-            printstyled(io, "━"^n_left, " "; color=:light_black)
+            if n_left > 0
+                if partial_filled > 0.5
+                    printstyled(io, "╸"; color=p.color) # More filled, use ╸
+                else
+                    printstyled(io, "╺"; color=:light_black) # Less filled, use ╺
+                end
+                printstyled(io, "━"^n_left; color=:light_black)
+            end
+            printstyled(io, " "; color=:light_black)
             print(io, progress_text)
         end
         carriagereturn && print(io, "\r")


tecosaur:
Nice! To engage in some blatant scope-creep, I've been thinking it would be nice for the bar to also show the packages currently being precompiled, with something like this:

image

Similarly, I think it could be fun to show the proportion that error/warn with a red/yellow bar, like so:

image

To actually help instead of keeping on repeating "this would be nice", I've put together what I think's actually a fairly nice function for displaying multi-segment progress bars.

```julia
"""
    multibar(io::IO, specs::Vector{<:Pair{Symbol, <:Real}}, width::Int=30)

Print a bar of certain `width` to `io`, split into segments according to `specs`.

Each element of `specs` is a pair `(color, weight)`, where `color` is
a named color recognised by `printstyled` and `weight` is a number.

Each segment is draw with a width proportional to its weight.

All segments with a non-zero weight are drawn.
"""
"""
    multibar(io::IO, specs::Vector{<:Pair{Symbol, <:Real}}, width::Int=30)

Print a bar of certain `width` to `io`, split into segments according to `specs`.

Each element of `specs` is a pair `(color, weight)`, where `color` is
a named color recognised by `printstyled` and `weight` is a number.

Each segment is draw with a width proportional to its weight.

All segments with a non-zero weight are drawn.
"""
function multibar(io::IO, specs::Vector{<:Pair{Symbol, <:Real}}, width::Int=min(30, last(displaysize(io))))
    bars = (whole='━', lpart='╺', rpart='╸')
    specs = filter(s -> !iszero(last(s)), specs)
    isempty(specs) && return
    totalhalves = 2 * width - length(specs) + 1
    duowidth = totalhalves / sum(last, specs, init=0)
    halfwidths = [max(1, round(Int, weight * duowidth)) for (_, weight) in specs]
    if sum(halfwidths) > totalhalves
        rawproportions = map(last, specs) / sum(last, specs)
        stolenwidths = zeros(Int, length(halfwidths))
        while sum(halfwidths) > totalhalves
            stolenproportions = rawproportions ./ (stolenwidths .+ 1)
            i = argmax(stolenproportions)
            halfwidths[i] -= 1
            stolenwidths[i] += 1
        end
    end
    partial = false
    for ((color, _), halves) in zip(specs, halfwidths)
        partial && printstyled(io, bars.lpart; color)
        nbars = (halves - partial) ÷ 2
        partial = halves - partial - 2 * nbars == 0
        printstyled(io, bars.whole^nbars; color)
        !partial && printstyled(io, bars.rpart; color)
    end
end
```

What do you think?

IanButterworth (approved):

KristofferC:
I personally am not sure I like it so much because it becomes confusing when the progress bar double dips as progress and as a counter. To me it looks like the red part is anither progress overplayed the green one and is supposed to progress all the way to the right. Just my initial thoughts.

tecosaur:
I find it difficult to see as double dipping. Ultimately a progress bar is just a visual representation of category sizes. Often there are two categories: "done" and "not done", and Pkg could keep with that but I think it could also expand without becoming confusing. Red `x`s and yellow warnings are already shown in the list of packages, so to me this proposal seems like a natural extension.

I actually see the current use of green checks put next to successfully precompiled packages as forming a natural association with the green section of the progress bar, and in that respect it's unintuitive that error/warn results also form part of the green progress bar.

christiangnrd:
While playing around with this I noticed that the double space before the progress bar is still present even after #4008. Seems like #4008 might have accidentally reintroduced it. I'd submit a PR but then we risk the same thing happening again so I think removing the `print(io, " ")` on line 87 as part of this PR may be less risky.

KristofferC:
> While segmented state bars may not be particularly common, I don't think they're rare either:

But those aren't progress bars so then my double dip argument doesn't apply.

tecosaur:
The line of thought I'm trying to articulate is that progress bars are just a special case of segmented state bars.

fredrikekre:
> I personally am not sure I like it so much because it becomes confusing when the progress bar double dips as progress and as a counter. To me it looks like the red part is anither progress overplayed the green one and is supposed to progress all the way to the right.

I agree. The extra grey bar for in progress ones would be fine though, since that will be eaten up by the green. On the other hand, it might not be obvious what it means so perhaps just the current thing is best.

tecosaur:
Perhaps I'm overly optimistic, but I suspect that in practice it should be fairly intuitive to interpret since you don't just look at the progress bar by itself and have to guess what it means, you actually get the context of the precompile-log displayed underneath it.

Perhaps it would be worth whipping up a MWE and then asking some people who haven't seen this thread how much sense it makes/do they find it intuitive?

KristofferC:
> The line of thought I'm trying to articulate is that progress bars are just a special case of segmented state bars.

I disagree, the examples you showed are different in that they try to show how a whole (the code base, the hard drive, etc.) is divided into different categories. There is no progress toward completion and there isn't a continuous filling from left to right as time passes.

> The extra grey bar for in progress ones would be fine though, since that will be eaten up by the green.

Okay, but do I really care to see in bar form how many packages are currently being precompiled out of a total? Like if I have two packages (out of two) that are currently precompiling the whole bar would be gray. It is giving the information  "x% of packages out of the total number of packages I will precompile is currently precompiling" in a visual form. Even if it is intuitive it's a bit like "so what?"

tecosaur:
> I disagree, the examples you showed are different in that they try to show how a whole (the code base, the hard drive, etc.) is divided into different categories

...like the current progress bars, which split a number of packages up by state into "tried to precompile" and "haven't tried yet" :wink: 

> Okay, but do I really care to see in bar form how many packages are currently being precompiled out of a total?

It's just visual pizazz :shrug: I thought it could be nice, but it seems like I might be the  odd one out.

KristofferC:
> ...like the current progress bars, which split a number of packages up by state into "tried to precompile" and "haven't tried yet" 

This is why I specifically emphasized the time and progress aspect, that we are watching something go from incomplete to complete. Anyway, I don't have much more to say about this.

PR #4039, opened 2024-10-07
Allow use of a url and subdir in [sources]

### LINKED ISSUE #4026: sources section of Project.toml does not support subdir
As discussed in Slack, attempting to specify a non-top-level package in a GitHub repo in the `sources` section of Project.toml errors.

MWE:

```toml
[deps]
DuckDB = "d2f5444f-75bc-4fdf-ac35-56f514c445e1"

[sources]
DuckDB = {url = "https://github.com/duckdb/duckdb", subdir = "tools/juliapkg", rev = "main"}
```

Calling `Pkg.resolve()` on the above project fails with

```
ERROR: Invalid key `subdir` in `source` section
```

For reference, DuckDB is a monorepo where the julia package is in the `tools/juliapkg` folder.

Version information (release candidate installed via juliaup):

```julia-repl
julia> versioninfo()
Julia Version 1.11.0-rc3
Commit 616e45539db (2024-08-26 15:46 UTC)
Build Info:
Platform Info:
  OS: Linux (x86_64-linux-gnu)
  CPU: 8 × Intel(R) Core(TM) i7-7700HQ CPU @ 2.80GHz
  WORD_SIZE: 64
  LLVM: libLLVM-16.0.6 (ORCJIT, skylake)
```

tecosaur:
Monorepos exist, so we shouldn't complain when somebody tries to combine the url and subdir parameters.

I think this is all that's needed to close #4026. These additions are just from looking through  #3783 for suspicious omissions.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -195,6 +195,9 @@ function update_source_if_set(project, pkg)
     if pkg.path !== nothing
         source["path"] = pkg.path
     end
+    if pkg.subdir !== nothing
+        source["subdir"] = pkg.subdir
+    end
     path, repo = get_path_repo(project, pkg.name)
     if path !== nothing
         pkg.path = path
@@ -204,6 +207,9 @@ function update_source_if_set(project, pkg)
     end
     if repo.rev !== nothing
         pkg.repo.rev = repo.rev
+    end
+    if repo.subdir !== nothing
+        pkg.repo.subdir = repo.subdir
     end
 end
 
--- a/src/project.jl
+++ b/src/project.jl
@@ -92,7 +92,7 @@ read_project_compat(raw, project::Project) =
 
 read_project_sources(::Nothing, project::Project) = Dict{String,Any}()
 function read_project_sources(raw::Dict{String,Any}, project::Project)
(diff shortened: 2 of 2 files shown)


KristofferC:
This might be all that is needed but would be good with a test to confirm.

tecosaur:
I notice in the tests in `test/sources.jl` https://github.com/JuliaLang/Example.jl is used, and that actually exists. Is `https://github.com/JuliaLang/ExampleMonorepo.jl` needed or similar?

fatteneder:
duplicate of #3826?

tecosaur:
Huh, looks a bit like it. I assumed no fix was around from the state of #4026.

tecosaur:
It would be good to get a fix into 1.11.2 if possible. If sorting out the test is the main blocker, can't we grab this for now and add a test later: it's not like it will make the current situation (not working at all) any worse.

#3826 also includes this fix, but it's marked as WIP.

tecosaur:
That's an idea, I'll try using something from the `test_packages` folder and push something tonight.

KristofferC:
Excellent, thank you!

PR #4042, opened 2024-10-11
Allow configuring can_fancyprint(io::IO) using IOContext

fonsp:
Hi!

This PR adds the ability to override `can_fancyprint(io::IO)` for a self-provided IO by setting an IOContext flag:  `:Pkg_force_fancyprint` can be set to true or false.

# Why
My use case is Pluto, where we call Pkg functions like `Pkg.precompile()` with a **buffer as `io`**. Our display (in the browser) supports fancy-printing, but because the buffer is not a `<: Base.TTY`, Pkg will not fancy-print.

The current workaround in Pluto is to add type piracy to the `Pkg.can_fancyprint` function, defining a new, more specific method for `can_fancyprint(io::IOContext{BufferStream})` where we set it to `true`, but I want to avoid this solution.

What do you think?

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -57,7 +57,7 @@ stdout_f() = something(DEFAULT_IO[], unstableio(stdout))
 const PREV_ENV_PATH = Ref{String}("")
 
 usable_io(io) = (io isa Base.TTY) || (io isa IOContext{IO} && io.io isa Base.TTY)
-can_fancyprint(io::IO) = (usable_io(io)) && (get(ENV, "CI", nothing) != "true")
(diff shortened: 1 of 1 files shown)


fonsp:
Hmm I see that this function was moved/copied to Base in https://github.com/JuliaLang/julia/pull/53403, so I would make a matching PR there. But I'll wait for feedback here before I continue :)

IanButterworth:
Makes sense to me, and yeah needs opening against the julia repo now.

fonsp:
@IanButterworth Moved to https://github.com/JuliaLang/julia/pull/58887

PR #4045, opened 2024-10-12
Fix end of progress bar

christiangnrd:
The progress bar would "bounce" by not printing any `:light_black` characters when `n_left == 0` but `n_filled != max_progress_width`.

```
#Before
julia> bar.current=980;show_progress(stdout,bar)
 Downloading packages ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━╺ 980/1000
#After
julia> bar.current=980;show_progress(stdout,bar)
 Downloading packages ━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━━ 980/1000
```

### DIFF
--- a/src/MiniProgressBars.jl
+++ b/src/MiniProgressBars.jl
@@ -74,7 +74,7 @@ function show_progress(io::IO, p::MiniProgressBar; termwidth=nothing, carriagere
     max_progress_width = max(0, min(termwidth - textwidth(p.header) - textwidth(progress_text) - 10 , p.width))
     n_filled = floor(Int, max_progress_width * perc / 100)
     partial_filled = (max_progress_width * perc / 100) - n_filled
-    n_left = max_progress_width - n_filled - 1
+    n_left = max_progress_width - n_filled
     headers = split(p.header)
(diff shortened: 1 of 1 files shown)


src/MiniProgressBars.jl
```diff
@@ -74,7 +74,7 @@ function show_progress(io::IO, p::MiniProgressBar; termwidth=nothing, carriagere
     max_progress_width = max(0, min(termwidth - textwidth(p.header) - textwidth(progress_text) - 10 , p.width))
     n_filled = floor(Int, max_progress_width * perc / 100)
     partial_filled = (max_progress_width * perc / 100) - n_filled
-    n_left = max_progress_width - n_filled - 1
+    n_left = max_progress_width - n_filled
```
KristofferC, line 77: Is line 91 unconditionally true now?

christiangnrd: No it needs to stay for when the bar is at 100%. Without it, an extra unwanted `'╺'` gets printed and then line 97 errors for trying to repeat a string -1 times.

PR #4057, opened 2024-10-25
fix Base.unreference_module call

vtjnash:
No need to commit privacy violations here, when we have a API method for doing this for this purpose


IanButterworth:
I invoked https://github.com/JuliaLang/BumpStdlibs.jl/actions/workflows/BumpStdlibs.yml for this

PR #4064, opened 2024-10-29
[release-1.10] Allow extensions to trigger from packages in [deps]

topolarity:
This is a partial backport of julialang/julia#54009, with just the pre-compilation piece that lives in Pkg on 1.10


topolarity:
I should mention this is a blind backport - I don't actually know why this change was included in https://github.com/JuliaLang/julia/pull/54009

KristofferC:
I think on Julia versions without https://github.com/JuliaLang/julia/pull/56234 this should be kept.

topolarity:
Got it - I thought that might be what this was doing

Thanks @KristofferC !

Issue #4048, opened 2024-10-14
type BasicREPL has no field interface


fonsp:
A user posted https://github.com/fonsp/Pluto.jl/issues/3059 where the REPL printed the following error after running `using Pluto`:

```
ERROR: InitError: type BasicREPL has no field interface
Stacktrace:
  [1] getproperty
    @ ./Base.jl:49 [inlined]
  [2] repl_init(repl::REPL.BasicREPL)
    @ REPLExt ~/.julia/juliaup/julia-1.11.0+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/ext/REPLExt/REPLExt.jl:164
  [3] __init__()
    @ REPLExt ~/.julia/juliaup/julia-1.11.0+0.x64.linux.gnu/share/julia/stdlib/v1.11/Pkg/ext/REPLExt/REPLExt.jl:305
  [4] run_module_init(mod::Module, i::Int64)
    @ Base ./loading.jl:1336
  [5] register_restored_modules(sv::Core.SimpleVector, pkg::Base.PkgId, path::String)
    @ Base ./loading.jl:1324
  [6] _include_from_serialized(pkg::Base.PkgId, path::String, ocachepath::String, depmods::Vector{…}, ignore_native::Nothing; register::Bool)
    @ Base ./loading.jl:1213
  [7] _include_from_serialized (repeats 2 times)
    @ ./loading.jl:1169 [inlined]
  [8] _require_search_from_serialized(pkg::Base.PkgId, sourcepath::String, build_id::UInt128, stalecheck::Bool; reasons::Dict{…}, DEPOT_PATH::Vector{…})
    @ Base ./loading.jl:1959
  [9] _require(pkg::Base.PkgId, env::String)
    @ Base ./loading.jl:2435
 [10] __require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:2300
 [11] #invoke_in_world#3
    @ ./essentials.jl:1088 [inlined]
 [12] invoke_in_world
    @ ./essentials.jl:1085 [inlined]
 [13] _require_prelocked(uuidkey::Base.PkgId, env::String)
    @ Base ./loading.jl:2287
 [14] macro expansion
    @ ./loading.jl:2226 [inlined]
 [15] macro expansion
    @ ./lock.jl:273 [inlined]
 [16] __require(into::Module, mod::Symbol)
    @ Base ./loading.jl:2183
[...]

PR #4067, opened 2024-10-30
REPLExt: check for compliant repl mode during repl init

IanButterworth:
I don't think this fixes https://github.com/JuliaLang/julia/issues/56216 but it seems safer given other unsupported repl modes exist.

### DIFF
--- a/ext/REPLExt/REPLExt.jl
+++ b/ext/REPLExt/REPLExt.jl
@@ -168,7 +168,7 @@ function create_mode(repl::REPL.AbstractREPL, main::LineEdit.Prompt)
     return pkg_mode
 end
 
-function repl_init(repl::REPL.AbstractREPL)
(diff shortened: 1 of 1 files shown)


ext/REPLExt/REPLExt.jl
```diff
@@ -311,7 +311,13 @@ end
 
 function __init__()
     if isdefined(Base, :active_repl)
-        repl_init(Base.active_repl)
+        if Base.active_repl isa REPL.LineEditREPL
+            repl_init(Base.active_repl)
+        else
+            # not sure what to do here..
```
IanButterworth, line 317:
```suggestion
            # TODO: not sure what to do here..
```

fonsp:
Does this also fix https://github.com/JuliaLang/Pkg.jl/issues/4048?

PR #4070, opened 2024-11-04
rename FORMER_STDLIBS -> UPGRADABLE_STDLIBS

IanButterworth:
Calling these former stdlibs is confusing, I think.
My understanding:
- stdlib - package that ships with julia (no git-tree-sha1 in the manifest)
- upgradable stdlib - package that ships with julia, which can optionally be upgraded (gets a git-tree-sha1 if upgraded)
- former stdlibs - a package that used to be shipped with julia but isn't any longer (no packages yet, AFAIK)

If a julia 2.0 is cut, then upgradable stdlibs are primed to become former stdlibs, i.e. not shipped with julia, but before that it's confusing I think.

Also I don't believe we document that some stdlibs can be upgraded.

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -460,8 +460,8 @@ is_project_uuid(env::EnvCache, uuid::UUID) = project_uuid(env) == uuid
 # Context #
 ###########
 
-const FORMER_STDLIBS = ["DelimitedFiles", "Statistics"]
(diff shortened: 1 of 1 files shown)


ViralBShah:
This seems good to merge.

KristofferC:
I think this name was made when these stdlibs were in fact removed from the Julia repo. But that was reverted so the name here makes more sense.

PR #4075, opened 2024-11-05
Fix artifact directories not having traversal permissions

staticfloat:
It turns out that Windows requires the executable bit set if the `BYPASS_TRAVERSE_CHECKING` privilege is not attached to a user's account.  It's simply more correct to ensure that our directories have this bit set, and unfortunately our `filemode()` function call is not complete on Windows and does not include this bit, so we manually add it in on Windows.


giordano:
Is this related to https://github.com/JuliaLang/julia/issues/52272?

StefanKarpinski:
Should we also be setting this bit on directories in Tar.extract?

staticfloat:
In my tests, Tar.extract didn’t set any permissions on directories on windows.

ViralBShah:
Shall we merge this and have it backported to the upcoming 1.10.7 (if there is still time?) and 1.11 as well?

davidanthoff:
Why are we still modifying file permissions on Windows at all? I'm still completely convinced that we should do absolutely nothing to modify file permissions on Windows systems, see https://github.com/JuliaLang/julia/issues/52272#issuecomment-2089115803 for my explanation of that.

topolarity:
Libuv seems to agree with you: https://github.com/libuv/libuv/blob/31ea3411ccab6db63a3f573523be61f80f984d50/src/win/fs.c#L1840-L1849

I'm a bit surprised this even does anything.. I think libuv calls `_wchmod`, which is [documented as ignoring anything that's not _S_IWRITE or _S_IREAD](https://learn.microsoft.com/en-us/cpp/c-runtime-library/reference/chmod-wchmod?view=msvc-170) (in particular `_S_IEXEC` is being set here)

src/Artifacts.jl
```diff
@@ -84,7 +84,14 @@ function _mv_temp_artifact_dir(temp_dir::String, new_path::String)::Nothing
         err = ccall(:jl_fs_rename, Int32, (Cstring, Cstring), temp_dir, new_path)
         if err ≥ 0
             # rename worked
-            chmod(new_path, filemode(dirname(new_path)))
+            new_path_mode = filemode(dirname(new_path))
+            if Sys.iswindows()
+                # If this is Windows, ensure the directory mode is executable,
+                # as `filemode()` is incomplete.  Some day, that may not be the
+                # case, there exists a test that will fail if this is changes.
+                new_path_mode |= 0o001
```
staticfloat, line 92: Looking at this again, shouldn't this actually be `0o111`?

staticfloat:
I continue to believe that it's better to have `chmod()` do something than nothing, and that bugs in implementation should just be fixed so the behavior is as close to reasonable as possible.  However, I also understand the argument that this is an unusual position for opensource-on-windows-software to take, and I could see the argument for reverting this behavior in future releases of Julia.

However, at least for the present releases, I think this change is purely an improvement.  (`filemode()` not returning `+x` on Windows is pretty clearly a bug, working around that seems like a fine thing to do.  Alternatively, fixing `filemode()` could also be a good thing to do, but that's more invasive.)

topolarity:
> I continue to believe that it's better to have `chmod()` do something than nothing...
> However, at least for the present releases, I think this change is purely an improvement.  

I'm confused.. maybe I chased through the source code wrong, but based on the libuv implementation it really looks like the `chmod` call (the one implemented by Microsoft) that this is passed to ought to do nothing with the new flag.

staticfloat:
> I'm confused.. maybe I chased through the source code wrong, but based on the libuv implementation it really looks like the `chmod` call (the one implemented by Microsoft) that this is passed to ought to do nothing with the new flag.

You want to take a look at [this function](https://github.com/JuliaLang/libuv/blob/af4172ec713ee986ba1a989b9e33993a07c60c9e/src/win/fs.c#L2301), our fork has diverged from upstream libuv on this.

davidanthoff:
> I continue to believe that it's better to have chmod() do something than nothing,

We should just do exactly what MS is doing, this is perfectly speced out by the behavior of `chmod` in WSL when operating on NTFS file systems: https://learn.microsoft.com/en-us/windows/wsl/file-permissions#changing-file-permissions-on-an-existing-windows-file-using-chmod. Apparently that also has been the long-standing behavior of CIFS.

staticfloat:
If I'm reading that link right, CIFS and the "default" (non-metadata) behavior outlined in that link both completely ignore everything except removing all `read` permissions on a file; so `chmod("foo", 0o755)` and `chmod("foo", 0o400)` do the exact same thing, and there's no way to check things like `isexecutable()`, etc...  IMO, that's not a good solution.

The metadata behavior looks appealing to me, I suppose that on our end we could just implement that directly?  There's not some option we need to turn on on the NTFS volume?  (I believe the option that they're talking about turning on is for how WSL interacts with NTFS, not in NTFS itself?)

StefanKarpinski:
Can we have the "What to do about Windows permissions" conversation somewhere a little harder to forget than this? Maybe a discussion on the @JuliaLang/julia repo? (unless Viral has turned that off again)

PR #4080, opened 2024-11-08
Actually switch to "Resolving Deltas"

Keno:
Libgit2's "Resolving Deltas" code is extremely slow (https://github.com/libgit2/libgit2/issues/4674) on larger repositories, so it is important to have an accurate progress bar to avoid users thinking the download is stuck. We had this implemented. However, we were never actually switching to it, because the progress meter thought the progress was jumping backwards and wouldn't actually update because of it. Fix that by resetting it on the first switch to resolving deltas.


KristofferC:
I have been wondering why I never saw this... I thought it was just too fast I think 😆

IanButterworth:
Merge?

PR #4091, opened 2024-11-18
Automatically upgrade empty manifest files to v2 format

IanButterworth:
So one can do
```
$ touch Manifest-v1.11.toml
$ julia --project
pkg> add Foo
```
and not end up with a V1 manifest and associated warnings.

See https://github.com/JuliaLang/julia/pull/43845#issuecomment-1943995168

### DIFF
--- a/test/manifests.jl
+++ b/test/manifests.jl
@@ -54,14 +54,14 @@ end
             output = String(take!(io))
             @test occursin(r"Activating.*project at.*", output)
             env_manifest = Pkg.Types.Context().env.manifest_file
-            @test env_manifest == manifest
+            @test samefile(env_manifest, manifest)
             # an empty manifest is still technically considered to be v1 manifest
             @test Base.is_v1_format_manifest(Base.parsed_toml(env_manifest))
             @test Pkg.Types.Context().env.manifest.manifest_format == v"2.0.0"
 
             Pkg.add("Profile"; io=io)
             env_manifest = Pkg.Types.Context().env.manifest_file
-            @test env_manifest == manifest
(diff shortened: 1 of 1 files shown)


src/manifest.jl
```diff
@@ -219,7 +219,11 @@ function read_manifest(f_or_io::Union{String, IO})
         rethrow()
     end
     if Base.is_v1_format_manifest(raw)
-        raw = convert_v1_format_manifest(raw)
+        if isempty(raw) # treat an empty Manifest file as v2 format for convenience
```
DilumAluthge, line 222: If someone has a newline or something in there for some reason (maybe they did `echo "" > Manifest-v1.11.toml`), seems like we can still treat that as v2 format?

```suggestion
        if isempty(strip(raw)) # treat an empty Manifest file as v2 format for convenience
```

DilumAluthge: Oh wait, I completely misread this code.

DilumAluthge: `raw` is already a dict, so this is irrelevant.

DilumAluthge (approved):

src/manifest.jl
```diff
@@ -219,7 +219,11 @@ function read_manifest(f_or_io::Union{String, IO})
...
+            raw["manifest_format"] => "2.0.0"
```
IanButterworth, line 223:
```suggestion
            raw["manifest_format"] = "2.0.0"
```

DilumAluthge (dismissed):

vchuravy (dismissed):

PR #4093, opened 2024-11-23
[backports-release-1.10] Backport julia/#55910 and julia/#56621

christiangnrd:
Don't squash?

I backported them while looking into https://github.com/JuliaConcurrent/Atomix.jl/issues/44 and figured I might as well submit a PR.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1144,15 +1144,15 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
         end
     end
 
-    # An extension effectively depends on another extension if it has all the the
-    # dependencies of that other extension
-    function expand_dependencies(depsmap)
+    # A package/extension effectively depends on another extension if it (transitively)
+    # has all the dependencies of that other extension
+    function expand_indirect_dependencies(direct_deps)
         function visit!(visited, node, all_deps)
             if node in visited
                 return
             end
             push!(visited, node)
-            for dep in get(Set{Base.PkgId}, depsmap, node)
+            for dep in get(Set{Base.PkgId}, direct_deps, node)
                 if !(dep in all_deps)
                     push!(all_deps, dep)
                     visit!(visited, dep, all_deps)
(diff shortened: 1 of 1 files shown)


IanButterworth:
This is great. Thanks! 
Minor thing that helps track backports. Can you make the commit titles include the name of the original PR.

src/API.jl
```diff
@@ -1189,14 +1232,48 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
         end
     end
 
+    function expand_indirect_dependencies(direct_deps)
```
topolarity, line 1235: This is a duplicate of the existing `expand_dependencies` function above

christiangnrd:
> Can you make the commit titles include the name of the original PR.

Done!

IanButterworth:
Thanks. CI isn't running because this is merging into a backports branch. Should be fixed by https://github.com/JuliaLang/Pkg.jl/pull/4094

christiangnrd:
I accidentally merged the UUID modification I'll push a fix right away

IanButterworth:
If you rebase this now on the base branch it should run CI

src/API.jl
```diff
@@ -1189,14 +1232,48 @@ function precompile(ctx::Context, pkgs::Vector{PackageSpec}; internal_call::Bool
...
-            if !in(_pkg, keys(exts)) && pkg in deps # if not an extension and depends on pkg
-                append!(deps, pkg_exts) # add the package extensions to deps
-                filter!(!isequal(pkg), deps) # remove the pkg from deps
+    for ext in keys(exts)
+        ext_loadable_in_pkg = Dict{Base.PkgId,Bool}()
+        for pkg in keys(depsmap)
+            is_trigger = in(pkg, depsmap[ext])
+            is_extension = in(pkg, keys(exts))
+            has_triggers = issubset(depsmap[ext], indirect_deps[pkg])
+            ext_loadable_in_pkg[pkg] = !is_extension && has_triggers && !is_trigger
```
topolarity, line 1270: For 1.10, I think we want to delete the `!is_extension` condition
```suggestion
            # In contrast to 1.11+, on 1.10 both "pkg → ext" and "ext → ext" dependency edges
            # are implied based on transitive dependencies.
            #
            # This condition is inconsistent for "ext → ext" edges, leading to dependency
            # cycles on 1.10, but this behavior is intentionally preserved for now to avoid
            # breaking packages that depend on this (bad) implicit behavior.
            #
            # See https://github.com/JuliaLang/julia/issues/56204#issuecomment-2442652997
            # for the improved behavior this was replaced with in 1.11
            ext_loadable_in_pkg[pkg] = has_triggers && !is_trigger
```

Technically the check you have here is OK, but I prefer deleting the `!is_extension` condition and getting rid of the for loop just above this that was computing these "ext → ext" dependency edges separately

topolarity:
Thanks for the backport!

For clarity, this will fix the "false cycles" we were adding on accident before https://github.com/JuliaLang/julia/pull/55910, but it leaves in any true cycles that come from the bad "ext → ext" edge definition we have on 1.10

I think that's the right thing to do - just wanted to make it clear what this does / doesn't fix

src/API.jl
topolarity: Took the liberty of pushing this

topolarity (approved):

topolarity:
I think this might fix a significant majority of our cycle problems on 1.10

Circularity should now only occur when two extensions have exactly the same* set of triggers, and the cycles should be a lot shorter / more direct (e.g. "ExtAB ↔ extBA")

\* technically the "same" triggers after transitive reduction, meaning not counting any triggers in the dependencies of another

PR #4096, opened 2024-11-25
Clarify guidelines

rayegun:
Clarify that according to the General README these are guidelines *not* rules, and as such may be overridden with sufficient justification.

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -562,10 +562,10 @@ duplicated into `[extras]`. This is an unfortunate duplication, but without
 doing this the project verifier under older Julia versions will throw an error
 if it finds packages under `[compat]` that is not listed in `[extras]`.
 
-## Package naming rules
+## Package naming guidelines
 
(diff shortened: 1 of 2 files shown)


rayegun:
I believe this was changed at some point from guidelines to rules, given that there are broken links on General/README.md.

I am strongly in favor of maintaining the word guidelines, not the word rules. Names are by nature a deeply personal part of the package development process. With sufficient justification and/or rebuttals to community name feedback we should feel comfortable merging package author preferred names.

In addition it must be clear that renaming suggestions ***are suggestions***. Users who point to Pkg.jl docs as "rules" rather than guidelines are empowered to outright deny names, rather than offer constructive suggestions. I receive frequent complaints about weaponization of the rules/guidelines from new and existing community members. 

Explicitly stating that these are guidelines provides more support for package authors who feel they are being demanded to make changes rather than being suggested.

ViralBShah (dismissed):

PR #3827, opened 2024-03-05
Better error messages if artifact rename fails

nhz2:
This PR should improve error messages if an anti-virus prevents the temporary directory the artifact is unpacked into from being renamed or deleted. Ref #3822 

The issue is that an error that occurs in a `finally` after a `try` and `catch` hides any error that occurred in the `try`. 

The following is the message @xlxs4 [got when testing this PR](https://discourse.julialang.org/t/artifacts-folder-not-empty-error-when-installing-plots-jl/117289/7?u=nhz2)

```julia
julia> Pkg.pkg"add Plots"
   Resolving package versions...
   Installed Statistics ─ v1.11.1
┌ Error: your anti-virus may be interfering with artifact installation
│ See https://github.com/JuliaLang/Pkg.jl/issues/3822
└ @ Main.Pkg.Artifacts C:\Users\ores.ousoultzoglou\GitHub\Pkg.jl\src\Artifacts.jl:81
┌ Warning: Failed to clean up temporary directory "C:\\Users\\ores.ousoultzoglou\\.julia\\artifacts\\jl_fukAt4"
│   exception = IOError: rm("C:\\Users\\ores.ousoultzoglou\\.julia\\artifacts\\jl_fukAt4\\bin"): directory not empty (ENOTEMPTY)
└ @ Main.Pkg.Artifacts C:\Users\ores.ousoultzoglou\GitHub\Pkg.jl\src\Artifacts.jl:386
┌ Error: your anti-virus may be interfering with artifact installation
│ See https://github.com/JuliaLang/Pkg.jl/issues/3822
└ @ Main.Pkg.Artifacts C:\Users\ores.ousoultzoglou\GitHub\Pkg.jl\src\Artifacts.jl:81
┌ Warning: Failed to clean up temporary directory "C:\\Users\\ores.ousoultzoglou\\.julia\\artifacts\\jl_nEUv5J"
│   exception = IOError: rm("C:\\Users\\ores.ousoultzoglou\\.julia\\artifacts\\jl_nEUv5J\\bin"): directory not empty (ENOTEMPTY)
└ @ Main.Pkg.Artifacts C:\Users\ores.ousoultzoglou\GitHub\Pkg.jl\src\Artifacts.jl:386
ERROR: Unable to automatically download/install artifact 'fzf' from sources listed in 'C:\Users\ores.ousoultzoglou\.julia\packages\fzf_jll\okrgy\Artifacts.toml'.
Sources attempted:
[...]

PR #4099, opened 2024-11-28
Backport artifact rename to 1.11

visr:
#4001 was marked with a backport-1.11 label but wasn't part of #4037. It did get in 1.10 with #4036.

This also includes #3827 since it is a related improvement and by adding both it avoids merge conflicts.

As suggested by @IanButterworth in https://discourse.julialang.org/t/error-ioerror-rm-c-users-35879-julia-artifacts-jl-wek1vy-bin-directory-not-empty-enotempty/123142/6?u=visr.


IanButterworth:
Thanks. Not sure what happened with my backports PR. Thanks for catching and fixing.

PR #4100, opened 2024-11-29
Update Manifest to avoid MbedTLS_jll

eschnett:
I'm working on replacing MbedTLS by OpenSSL in Julia. See https://github.com/JuliaLang/julia/pull/56708. This PR avoids MbedTLS_jll.


eschnett:
These tests will fail because they are run with the old version of Julia. The Julia tests (see https://github.com/JuliaLang/julia/pull/56708) actually run the Pkg tests as well, and they succeed there. Thus the test failures here should be ignored.

fxcoudert:
kindly pinging @KristofferC on this? This is required for the julia PR at https://github.com/JuliaLang/julia/pull/56708 given that MbedTLS is reaching end-of-support.

IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/4123 is an alternative that should work both here and on the updated version, so we can check CI on both sides.

PR #4101, opened 2024-11-29
Automatically add compat bounds when adding package in non-shared env

### LINKED ISSUE #4069: Add automatically compat entry when adding a dependency to non-package projects
#3732 enabled automatic addition of compat entries when adding a new dependency to a package project, but I believe we should do that for all environments, or at least the non shared ones.  Manifests are certainly useful when you have a fixed pipeline which is always running with a fixed version of Julia, but in other cases they're more annoying to carry around, as they are julia version-specific. Now we have versioned manifests which help in certain situations, but do absolutely nothing for instantiating a compatible project in one or two year time with a version of julia which didn't exist when one first created the environment.

Certainly the risk of this proposal is for compat bounds to go stale over a long time, but I think that ensuring that instantiating an environment with (presumably) compatible packages at a later time outweighs the benefit of always having the super latest version of all packages which comes with the risk of breaking the workflow anyway.  Automatically adding compat bounds on all non-shared environments would help promote this practice.

giordano:
Doing some tests locally, this seems to work as advertised.

Do I understand correctly that to test the Pkg REPL mode I need to start julia with
```
julia --project=${HOME}/.julia/dev/Pkg -e 'using Pkg' -i
```
instead of just
```
julia --project=${HOME}/.julia/dev/Pkg
```
as [suggested by the README](https://github.com/JuliaLang/Pkg.jl/blob/83e13631e712384340ca5dff8c390f4dc6ad2479/README.md#using-the-development-version-of-pkgjl)?

Fix #4069.


wilf-phasecraft:
I noticed that this moved to Backlog and just want to check whether this is still planned for a future release?

PR #4109, opened 2024-12-06
status: highlight when deps have different loaded versions

IanButterworth:
See https://github.com/JuliaLang/julia/pull/56769

Screenshot 2024-12-09 at 9 39 09 AM

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -2747,9 +2747,11 @@ function print_status(env::EnvCache, old_env::Union{Nothing,EnvCache}, registrie
                 loaded_version = pkgversion(m)
                 env_version = pkg_spec.version
                 if loaded_version !== env_version
-                    printstyled(io, " [different version loaded: v$loaded_version]"; color=:light_yellow)
+                    printstyled(io, " [loaded: v$loaded_version]"; color=:light_yellow)
                 else
-                    printstyled(io, " [different path loaded, but same version: `$loaded_path` expected `$env_path`]"; color=:light_yellow)
+                    loaded_version_str = loaded_version === nothing ? "" : " (v$loaded_version)"
+                    env_version_str = env_version === nothing ? "" : " (v$env_version)"
+                    printstyled(io, " [loaded: `$loaded_path`$loaded_version_str expected `$env_path`$env_version_str]"; color=:light_yellow)
                 end
             end
         end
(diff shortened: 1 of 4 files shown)


src/Operations.jl
```diff
@@ -2736,6 +2736,16 @@ function print_status(env::EnvCache, old_env::Union{Nothing,EnvCache}, registrie
                 printstyled(io, pkg_str; color=Base.warn_color())
             end
         end
+        # show if loaded and version doesn't match
+        pkg_spec = something(pkg.new, pkg.old)
+        pkgid = Base.PkgId(pkg.uuid, pkg_spec.name)
+        m = get(Base.loaded_modules, pkgid, nothing)
+        if m isa Module && pkg_spec.version !== nothing
+            v = pkgversion(m)
+            if v !== pkg_spec.version
```
KristofferC, line 2745: Would using the path to the package be more robust? That would include the version but also for example if you have dvved something that happens to have the same version as the one in your main env.

IanButterworth: Makes sense. Updated

tecosaur:
Nit/preference: I feel like the "different version" loaded bit of the "different version loaded" is a bit redundant, just having `[loaded: v1.6.3]` in yellow would be enough IMO.

src/Operations.jl
```diff
@@ -2736,6 +2736,25 @@ function print_status(env::EnvCache, old_env::Union{Nothing,EnvCache}, registrie
                 printstyled(io, pkg_str; color=Base.warn_color())
             end
         end
+        # show if loaded version and version in the manifest doesn't match
+        pkg_spec = something(pkg.new, pkg.old)
+        pkgid = Base.PkgId(pkg.uuid, pkg_spec.name)
+        m = get(Base.loaded_modules, pkgid, nothing)
+        if m isa Module && pkg_spec.version !== nothing
+            loaded_path = pathof(m)
+            env_path = Base.locate_package(pkgid) # nothing if not installed
```
KristofferC, line 2745: There are so many ways to get the path to a package, could also for example use `pkgorigins` but this should be fine as well.

KristofferC (approved):

src/Operations.jl
IanButterworth: `pkgorigins` only works for loaded packages? This line's the src as found based on the env, i.e. what loading would find. I thought `locate_package` would be the closest to the actual loading mechanics? If there's a more proper way then let's use it

PR #4115, opened 2024-12-16
Improve latency for `Pkg.add`

KristofferC:
Before:

```
  1.379429 seconds (3.39 M allocations: 211.476 MiB, 2.36% gc time, 60.64% compilation time: 38% of which was recompilation)
```

 After:

```
   0.963989 seconds (2.83 M allocations: 182.510 MiB, 1.67% gc time, 42.22% compilation time: 75% of which was recompilation)
```

### DIFF
--- a/src/precompile.jl
+++ b/src/precompile.jl
@@ -118,7 +118,6 @@ let
                     Pkg.add(Pkg.PackageSpec(path = "TestPkg.jl/"))
                     Pkg.update(; update_registry = false)
                     Pkg.status()
-                    Pkg.Registry.update()
                     pkgs_path = pkgdir(Pkg, "test", "test_packages")
                     # Precompile a diverse set of test packages
                     # Check all test packages occasionally if anything has been missed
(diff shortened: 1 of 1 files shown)


src/precompile.jl
```diff
@@ -118,6 +118,7 @@ let
                     Pkg.add(Pkg.PackageSpec(path = "TestPkg.jl/"))
                     Pkg.update(; update_registry = false)
                     Pkg.status()
+                    Pkg.Registry.update()
```
IanButterworth, line 121: Aren't we intentionally avoiding any internet activity? Or at least trying to i.e. `Pkg.UPDATED_REGISTRY_THIS_SESSION[] = true` above.

I think there's an issue open but can't find it.

KristofferC: I thought that since we already install a local registry it wouldn't go out to the internet to update anything.

IanButterworth: huh. Ok. I guess we can delete `Pkg.UPDATED_REGISTRY_THIS_SESSION[] = true` and the `update_registry = false` kwarg on `Pkg.update`. Worth checking offline precompilation doesn't error.

KristofferC: I did get

```
│   exception = RequestError: Could not resolve host: pkg.julialang.org while requesting https://pkg.julialang.org/registries
└ @ Pkg.Registry ~/JuliaPkgs/Pkg.jl/src/Registry/Registry.jl:77
```

without internet. With the explicit `precompile` for updating registries it seems it wasn't super important so I took it away.

PR #4123, opened 2024-12-29
Support before & after MbedTLS > OpenSSL switch

IanButterworth:
Alternative to https://github.com/JuliaLang/Pkg.jl/pull/4100 that should work before & after the switch.


IanButterworth:
This is passing at https://github.com/JuliaLang/julia/pull/56708 so I'll go ahead

PR #4128, opened 2025-01-05
Don't set up LibGit2 callbacks if using cli git

IanButterworth:
Just an observation. I'm not aware of any issues this caused


KristofferC:
Makes sense.

KristofferC (approved):

PR #4134, opened 2025-01-12
Use `_` for unused argument [NFC]

LilithHafner:


LilithHafner:
I read this source code and was briefly confused; replacing with `_` would have alleviated my confusion. NBD, feel free to close if maintainers here would rather not.

Issue #4160, opened 2025-02-15
Error if `julia_version` is specified through `Context`


IanButterworth:
As far as I understand, these should be equivalent.
```
dependencies = [PackageSpec(; name="CMake_jll", version = "3.24.3")]
platform = Base.BinaryPlatforms.Platform("x86_64", "linux"; libc="musl")

@testset "good" begin
    Pkg.activate(temp=true)
    Pkg.add(deepcopy(dependencies); platform, julia_version=nothing)
end

@testset "bad" begin
    Pkg.activate(temp=true)
    ctx = Pkg.Types.Context(;julia_version=nothing)
    Pkg.add(ctx, deepcopy(dependencies); platform)
end
```
But the latter errors with
```
  Activating new project at `/var/folders/1z/jf841bdj73bdj3vk7kc7f_3w0000gn/T/jl_C6cT0a`
   Resolving package versions...
bad: Error During Test at REPL[20]:1
  Got exception outside of a @test
  MethodError: no method matching length(::Pkg.Versions.VersionSpec)
  The function `length` exists, but no method is defined for this combination of argument types.

  Closest candidates are:
    length(::Base.MethodSpecializations)
     @ Base runtime_internals.jl:1643
    length(::Core.SimpleVector)
     @ Base essentials.jl:934
    length(::Markdown.MD)
     @ Markdown ~/Documents/GitHub/julia/usr/share/julia/stdlib/v1.13/Markdown/src/parse/parse.jl:35
    ...

  Stacktrace:
    [1] union!(s::Set{Any}, itr::Pkg.Versions.VersionSpec)
      @ Base ./abstractset.jl:104
    [2] intersect!(s::Set{Any}, itr::Pkg.Versions.VersionSpec)
      @ Base ./abstractset.jl:197
    [3] _shrink(shrinker!::typeof(intersect!), itr::String, itrs::Tuple{Pkg.Versions.VersionSpec})
      @ Base ./array.jl:3113
    [4] intersect(itr::String, itrs::Pkg.Versions.VersionSpec)
      @ Base ./array.jl:3117
    [5] resolve_versions!(env::Pkg.Types.EnvCache, registries::Vector{Pkg.Registry.RegistryInstance}, pkgs::Vector{Pkg.Types.PackageSpec}, julia_version::Nothing, installed_only::Bool)
      @ Pkg.Operations ~/Documents/GitHub/Pkg.jl/src/Operations.jl:502
[...]

PR #4151, opened 2025-02-09
Expand HistoricalStdlibVersions tests

IanButterworth:
Building ontop of https://github.com/JuliaLang/Pkg.jl/pull/4150

- [x] Expands HistoricalStdlibVersions tests based on needs outlined in https://github.com/JuliaPackaging/JLLPrefixes.jl/issues/6 thanks to the tests constructed by @mkitti [here](https://github.com/JuliaPackaging/JLLPrefixes.jl/issues/6#issuecomment-1978200959)
- [x] Stops installing HistoricalStdlibVersions manually if not installed, so any such tests would not be run via `Base.runtests`
- [ ] Consolidate HistoricalStdlibVersions tests into a single file to easier maintenance

PR #4161, opened 2025-02-16
Rename context API methods as internal functions

IanButterworth:
The `Pkg.add(ctx::Context.....` methods are internal methods that require preconditioning the input args using internal functions 
https://github.com/JuliaLang/Pkg.jl/blob/c54ee594d1c16270d4f67e09788839d9618faf18/src/API.jl#L162-L163

This renames those methods to underscore versions to make it clearer that they're internal.

BinaryBuilderBase is known to use the `Pkg.add(ctx` method and during development saw the confusing `length` errors #4160  that you get if  you don't precondition the input, which added to the impression that Pkg was broken. @giordano and I have talked about moving that to the public way of controlling the julia version via the `julia_version` kwarg, which has better test coverage now and should be in better shape now since #4151 and #4156


DilumAluthge:
Makes sense to me - we should make it easy to notice when a generic function is private or internal.

PR #4166, opened 2025-02-22
Add note about character minimums

wylie39:
I didn't know this was a requirement until the automated registration bot flagged it, feels like it should be documented.

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -621,9 +621,12 @@ may fit your package better.
        there's no copyright or trademark infringement etc.)
 
 9. Packages should follow the [Stylistic Conventions](https://docs.julialang.org/en/v1/manual/variables/#Stylistic-Conventions).
-     * The package name begin with a capital letter and word separation is shown with upper camel case
+     * The package name begins with a capital letter and word separation is shown with upper camel case
      * Packages that provide the functionality of a project from another language should use the Julia convention
-     * Packages that [provide pre-built libraries and executables](https://docs.binarybuilder.org/stable/jll/) can keep orignal name, but should get `_jll`as a suffix. For example `pandoc_jll` wraps pandoc. However, note that the generation and release of most JLL packages is handled by the [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil) system. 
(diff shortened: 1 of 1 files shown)


ViralBShah:
Merge?

digital-carver:
This only covers the one specific rule - that still leaves other rules to be added piecemeal later, for eg. that the name should have only ASCII characters. 

It seems better to just have the final point be "For the complete list of rules for automatic merging into the General registry, see [these guidelines](https://juliaregistries.github.io/RegistryCI.jl/stable/guidelines/)."

docs/src/creating-packages.md
```diff
@@ -621,9 +622,12 @@ may fit your package better.
        there's no copyright or trademark infringement etc.)
 
 9. Packages should follow the [Stylistic Conventions](https://docs.julialang.org/en/v1/manual/variables/#Stylistic-Conventions).
-     * The package name begin with a capital letter and word separation is shown with upper camel case
+     * The package name begins with a capital letter and word separation is shown with upper camel case
```
digital-carver, line 625:
```suggestion
     * The package name should begin with a capital letter and word separation is shown with upper camel case
```

To be consistent with the rest of the text (which all use the "should" form)

docs/src/creating-packages.md
```diff
@@ -621,9 +622,12 @@ may fit your package better.
...
      * Packages that provide the functionality of a project from another language should use the Julia convention
-     * Packages that [provide pre-built libraries and executables](https://docs.binarybuilder.org/stable/jll/) can keep orignal name, but should get `_jll`as a suffix. For example `pandoc_jll` wraps pandoc. However, note that the generation and release of most JLL packages is handled by the [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil) system. 
+     * Packages that [provide pre-built libraries and executables](https://docs.binarybuilder.org/stable/jll/) can keep orignal name, but should get `_jll`as a suffix. For example `pandoc_jll` wraps pandoc. However, note that the generation and release of most JLL packages is handled by the [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil) system.
```
digital-carver, line 627:
```suggestion
     * Packages that [provide pre-built libraries and executables](https://docs.binarybuilder.org/stable/jll/) can keep their original name, but should get `_jll`as a suffix. For example `pandoc_jll` wraps pandoc. However, note that the generation and release of most JLL packages is handled by the [Yggdrasil](https://github.com/JuliaPackaging/Yggdrasil) system.
```

digital-carver:
This seems good to merge now.

Issue #2677, opened 2021-07-26
Use WSL style NTFS extended attributes to store permissions of extracted artifacts on Windows


davidanthoff:
The idea would be to follow https://docs.microsoft.com/en-us/windows/wsl/file-permissions#wsl-metadata-on-windows-files to store unix file permissions instead of changing ACLs of extracted files on Windows. The functions to interact with these extended attributes are https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-zwqueryeafile and https://docs.microsoft.com/en-us/windows-hardware/drivers/ddi/ntifs/nf-ntifs-zwseteafile.

@staticfloat, @StefanKarpinski, giordano had a brief chat about this on Slack, and I'm just opening this issue to keep a record of the links we dug out.

staticfloat:
Okay, had a talk with @vtjnash about this, and we're on board with trying this out.

The concrete steps forward are:

- [ ] Write the EA utility functions in libuv to read/write WSL2 permissions strings, use those in `uv_fs__chmod()` and `uv_fs__stat()` instead of the ACL code.
- [ ] Create Julia branch that uses that libuv code, so `chmod()` and `stat()` use EAs instead of ACLs
- [ ] See if our `access(X_OK)` code can detect whether `~/.julia` inherits an executable permission, and if it does not, surface a warning to the user that they may need to do something like `icacls .julia /grant ${USER}:(OI)(CI)F /T`.

PR #4167, opened 2025-02-23
Enable windows tree_hash checking within install_archive

IanButterworth:
Given the TODO and referenced issue https://github.com/JuliaLang/julia/issues/33212 is fixed/closed. Has been since 2020.

cc. @davidanthoff

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -750,15 +750,12 @@ function install_archive(
             unpacked = joinpath(dir, dirs[1])
         end
         # Assert that the tarball unpacked to the tree sha we wanted
-        # TODO: Enable on Windows when tree_hash handles
-        # executable bits correctly, see JuliaLang/julia #33212.
-        if !Sys.iswindows()
-            if SHA1(GitTools.tree_hash(unpacked)) != hash
-                @warn "tarball content does not match git-tree-sha1"
-                url_success = false
-            end
-            url_success || continue
+        if SHA1(GitTools.tree_hash(unpacked)) != hash
+            @warn "tarball content does not match git-tree-sha1"
+            url_success = false
         end
+        url_success || continue
+
         # Move content to version path
         !isdir(version_path) && mkpath(version_path)
         mv(unpacked, version_path; force=true)


davidanthoff:
I think we should _not_ do this. We should never have started to use Windows ACLs to store Unix file permissions in the first place, but at least so far nothing depended on that (super fragile) state of affairs. With this PR merged, if an admin sees the mess we are creating with individual file permissions and tries to fix the ACL situation by modifying permissions back to inherited permissions (as Windows is designed), then that will now completely break with this PR (I think). Or just think about the "add another users to a shared folder" scenario I outline in the first link below, that would now break everything with this PR merged.

I think we should wait with enabling this check until we have moved away from ACL modifications towards the WSL attribute modification approach.

Background issues:
- Explanation why the current approach is so bad (https://github.com/JuliaLang/julia/issues/52272#issuecomment-2089115803)
- Outline what we should do (https://github.com/JuliaLang/Pkg.jl/issues/2677)

davidanthoff:
Another reason to revert this PR is that it would make a transition to the WSL permission style much, much more difficult. If we have a Julia version that shipped that relies on the ACL modifications, then we can't remove that in a future Julia version (as we should). Say 1.12 checks the ACLs, and then 1.13 moves to the WSL permission style and no longer modifies ACLs, then any artifact/package installed by Julia 1.13 will appear broken to Julia 1.12...

IanButterworth:
To help my understanding, specifically when are we setting the problematic file permissions?

I know there's `set_readonly` but that's after this function has returned.

staticfloat:
> To help my understanding, specifically when are we setting the problematic file permissions?

When we `chmod()` as part of `Tar.extract()`

> Say 1.12 checks the ACLs, and then 1.13 moves to the WSL permission style and no longer modifies ACLs, then any artifact/package installed by Julia 1.13 will appear broken to Julia 1.12...

Hmmm, but this check is only occurring upon install by the version of Julia that is currently running; we never verify something after it has already been installed.  If we did do that, I would agree with you, that this needs to be reverted.

Issue #2219, opened 2020-11-19
Adding packages is thread-unsafe?


giordano:
In Yggdrasil, where we happen to install the same packages in parallel, we have often issues with files that disappear.  [Last example](https://dev.azure.com/JuliaPackaging/Yggdrasil/_build/results?buildId=7399&view=logs&j=572c5e49-83d5-5271-390a-e6dc77f89c6b&t=25aa3cdb-e20d-599e-657c-8776cc8a525f&s=d654deb9-056d-50a2-1717-90c08683d50a):
```
ERROR: LoadError: Error when installing package SuiteSparse32_jll:
IOError: unlink: no such file or directory (ENOENT)
Stacktrace:
 [1] uv_error
   @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:802 [inlined]
 [9] (::Pkg.Operations.var"#62#65"{Bool, Pkg.Types.Context, Dict{Base.UUID, Vector{String}}, Channel{Any}, Channel{Tuple{Pkg.Types.PackageSpec, String}}})()
   @ Pkg.Operations ./task.jl:395
Stacktrace:
  [1] pkgerror(::String, ::Vararg{String, N} where N)
    @ Pkg.Types /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Types.jl:52
  [2] macro expansion
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:819 [inlined]
  [3] macro expansion
    @ ./task.jl:371 [inlined]
  [4] download_source(ctx::Pkg.Types.Context, pkgs::Vector{Pkg.Types.PackageSpec}, urls::Dict{Base.UUID, Vector{String}}; readonly::Bool)
    @ Pkg.Operations /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:775
  [5] #download_source#58
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:750 [inlined]
  [6] download_source
    @ /buildworker/worker/package_linux64/build/usr/share/julia/stdlib/v1.6/Pkg/src/Operations.jl:748 [inlined]
[...]

PR #4168, opened 2025-02-23
Pidlock source and artifact installation

### LINKED ISSUE #3495: Feature request: Support instantiation of multiple julia environments in parallel
With Julia 1.9 it gets more important to do proper precompilation ahead of time. I am using Pluto and like to instantiate all notebook environments with high speed.

Currently, when just spawning multiple julia processes and instantiating several environments, I get a GitError as soon as two environments have the same dependency (there is probably some randomness involved, but it failes constantly for me - there are several overlaps)

```julia
GitError(Code:ERROR, Class:OS, could not open '/home/myhome/.julia/packages/MyPackage/1opWe/Project.toml' for writing: Permission denied)
```

It would be really great to have a performant way of instantiating multiple environments at once.

~~EDIT: Just realized, that I only tested this Julia 1.8.5, I am going to test on Julia 1.9 immediately~~ Applies to 1.9 as well

### LINKED ISSUE #4174: Artifact download & installation isn't pidlocked
Currently multiple concurrent workers will race to download and install artifacts, which is inefficient at best.

Highlighted in #4168

IanButterworth:
Fixes #3495
Fixes #4174 
Also fixes registry installation which wasn't checking after acquiring the pidlock which registries still needed installing.
Makes progress on https://github.com/JuliaLang/Pkg.jl/issues/2219

But only if all concurrent julia process installations have this fix.. I'm not sure there's a way to protect against older versions..

Source:
- Pidlocks the source path during check/download/installation
- Protects against an interrupted `mv` from temp dir to final path by checking `tree_hash` before and after and deleting the destination if bad.

Artifacts:
- Pidlocks artifact downloads and updates the progress meter to indicate another process is doing the downloading

Registries:
- Checks after acquiring the lock whether it still needs installing

https://github.com/user-attachments/assets/0aecef62-4bce-4548-bc5b-adb5220dc20d

Todo:
- [x] Disable test debug prints
- [x] Decide how much testing to keep - One test with 3 concurrent processes


IanButterworth:
Artifact installation isn't pidfile orchestrated. It seems to be ok, but each worker is doing the same work, then one wins by doing the final mv.

Should be fixed.

i.e. these shouldn't report as `1 byte`
```
   Installed artifact LLVMOpenMP   1 byte
   Installed artifact Libiconv     1 byte
   Installed artifact Pixman       1 byte
   Installed artifact LAME         1 byte
   Installed artifact Expat        1 byte
   Installed artifact Opus         309.2 KiB
   Installed artifact Graphite2    1 byte
   Installed artifact XSLT         1 byte
   Installed artifact libfdk_aac   857.9 KiB
   Installed artifact Libgpg_error 1 byte
```

src/Operations.jl
```diff
@@ -758,7 +759,17 @@ function install_archive(
 
         # Move content to version path
         !isdir(version_path) && mkpath(version_path)
-        mv(unpacked, version_path; force=true)
+        unpacked_tree_hash = GitTools.tree_hash(unpacked)
+        try # protect against interrupted moves
+            mv(unpacked, version_path; force=true)
```
staticfloat, line 764: I think the actually correct thing to do here is to instead change line 733 to:

```
dir = mktempdir(dirname(version_path))
```

Then change this `mv()` to a `rename()`.  I believe this should be guaranteed atomic, because we can safely assume that users have not created separate filesystems to hold specific versions of packages.  The only caveat to be aware of here I believe is that `rename()` can fail if `version_path` exists, which could happen if, for instance, an older Julia without this pidlocking is installing the same thing at the same time.  So I guess we just have to handle that exception, check that the directory exists, then consider that a success.  If you want to be paranoid, you could check the treehash in such an event.

staticfloat: Looking through the Artifacts codepath, I think you basically want to call `_mv_temp_artifact_dir(unpacked, version_path)` here.

src/Operations.jl
```diff
@@ -1082,40 +1094,46 @@ function download_source(ctx::Context, pkgs; readonly=true)
...
-                                        url = "$server/package/$(pkg.uuid)/$(pkg.tree_hash)"
-                                        push!(archive_urls, url => true)
-                                        break
+                    ispath(path) && continue
+                    mkpath(dirname(path)) # the `packages/Package` dir needs to exist for the pidfile to be created
+                    FileWatching.mkpidlock(path * ".pid", stale_age = 20) do
+                        ispath(path) && @goto done
+                        if ctx.use_git_for_all_downloads
+                            put!(results, (pkg, false, (urls, path)))
+                            @goto done
```
staticfloat, line 1103: I think these `@goto`'s can be replaced with `return`, right?

staticfloat (dismissed):
This is a very good change overall.  I have mostly one comment to improve the bulletproof-ness of package installation.

src/Operations.jl
IanButterworth: Sounds good. `_mv_temp_artifact_dir` internally set permissions, which we do after `install_archive` has returned, so I just disabled that here.

IanButterworth:
These tests add between 1m40s and 4m (linux git registry).

I'm thinking of merging this and trying out in a Pkg bump PR to see the impact on Julia CI. Adjusting as necessary.

IanButterworth:
I saw the same hang locally on macOS when testing two concurrent processes, so I don't think the test failures are just CI peculiarities.

https://github.com/user-attachments/assets/8764844b-cfd2-4299-9b0e-cbb0056055db

IanButterworth:
@staticfloat this should be good to go now

PatrickHaecker:
Thanks for the fix, @IanButterworth. This just hit us on Julia 1.12. Shouldn't it be backported?

IanButterworth:
Yeah. Can you see if it backports cleanly, or figure out the fix on that branch. Thanks!

PR #4170, opened 2025-02-24
Add error that explains PackageSpec.repo is a private field

IanButterworth:
See https://github.com/JuliaPackaging/BinaryBuilderBase.jl/pull/416#discussion_r1967008740

Alternatively we could just use the `repo` entry if it's already populated? But it's not in the docs currently https://pkgdocs.julialang.org/v1/api/#Pkg.PackageSpec


test/new.jl
```diff
@@ -1831,6 +1831,8 @@ end
         Pkg.dependencies(exuuid) do pkg
             @test pkg.version > v"0.3.0"
         end
+
+        @test_throws PkgError("`repo` is a private field of PackageSpec and should not be set directly") Pkg.add([PackageSpec(;repo=LibGit2.GitRepo(url="someurl"))])
```
IanButterworth, line 1835:
```suggestion
        @test_throws PkgError("`repo` is a private field of PackageSpec and should not be set directly") Pkg.add([PackageSpec(;repo=Pkg.Types.GitRepo(url="someurl"))])
```

IanButterworth:
@KristofferC what do you think? This or should we use it if set? Silently ignoring it is confusing.

test/new.jl
```diff
@@ -2164,6 +2164,8 @@ end
         Pkg.dependencies(exuuid) do pkg
             @test pkg.version > v"0.3.0"
         end
+
+        @test_throws PkgError("`repo` is a private field of PackageSpec and should not be set directly") Pkg.add([PackageSpec(;repo=Pkg.Types.GitRepo(source="someurl"))])
```
IanButterworth:
```suggestion
        @test_throws PkgError("`repo` is a private field of PackageSpec and should not be set directly") Pkg.add([Pkg.PackageSpec(;repo=Pkg.Types.GitRepo(source="someurl"))])
```

PR #4171, opened 2025-02-24
Artifacts: Add download size to `Artifacts.toml`

staticfloat:
This shows one possible way of adding an optional download size value to an artifact binding.  While this value is not used anywhere yet, it could be used to create warnings when downloading large artifacts, or to feed progress meters when the server does not provide content size headers.

### DIFF
--- a/src/Artifacts.jl
+++ b/src/Artifacts.jl
@@ -18,7 +18,7 @@ import ..Types: write_env_usage, parse_toml
 export create_artifact, artifact_exists, artifact_path, remove_artifact, verify_artifact,
        artifact_meta, artifact_hash, bind_artifact!, unbind_artifact!, download_artifact,
        find_artifacts_toml, ensure_artifact_installed, @artifact_str, archive_artifact,
-       select_downloadable_artifacts
+       select_downloadable_artifacts, ArtifactDownloadInfo
 
 """
     create_artifact(f::Function)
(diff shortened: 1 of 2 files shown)


src/Artifacts.jl
```diff
@@ -186,6 +186,56 @@ function archive_artifact(hash::SHA1, tarball_path::String; honor_overrides::Boo
...
+"""
+struct ArtifactDownloadInfo
+    # URL the artifact is available at as a gzip-compressed tarball
+    url::String
+
+    # SHA256 hash of the tarball
+    hash::Vector{UInt8}
+
+    # Size in bytes of the tarball.  A size of `0` means unknown.
+    size::UInt64
```
nhz2, line 203: To prevent confusion with an empty file, `size` could be an `Int64` with `-1`  used to mean unknown. `size::Union{UInt64, Nothing}` would also work.

staticfloat: Empty tarball files are not allowed.

nhz2: That makes sense.

nhz2:
I made https://github.com/JuliaPackaging/ArtifactUtils.jl/pull/25 to try this out.

KristofferC:
> feed progress meters when the server does not provide content size headers.

How common is this? Is there any concrete case of this happening right now?

nhz2:
Very important bikeshedding question: should the "size" field be in hex or base 10?

For example: `size = 0xd38b3c` vs `size = 13863740`.

staticfloat:
We don't put anything else in hex, I think we should just stick with the most human-readable format.

src/Artifacts.jl
```diff
@@ -186,6 +186,56 @@ function archive_artifact(hash::SHA1, tarball_path::String; honor_overrides::Boo
...
+ArtifactDownloadInfo(adi::ArtifactDownloadInfo) = adi
+
+# Make the dict that will be embedded in the TOML
+function make_dict(adi::ArtifactDownloadInfo)
+    ret = Dict{String,Any}(
+        "url" => adi.url,
+        "sha256" => bytes2hex(adi.hash),
+    )
+    if adi.size > 0
+        ret["size"] = adi.size
```
nhz2, line 234:
```suggestion
        ret["size"] = Int64(adi.size)
```

Converting `UInt64` to `Int64` here prevents TOML.jl from writing the number in hex.

staticfloat: Darn, I was hoping it would just call `string(x)`, but I guess not.  I'll just make everything an `Int64` then.

nhz2 (approved):

PR #4176, opened 2025-02-28
Disallow precompiling Pkg in Pkg tests

IanButterworth:
Used to happen a lot. Easy to regress, so test for it.


ViralBShah:
Merge?

IanButterworth:
REPLExt is being re-precompiled during tests. I added the same check for that and will investigate
https://github.com/JuliaLang/Pkg.jl/actions/runs/15168228801/job/42651517879?pr=4176#step:5:216

PR #4180, opened 2025-03-01
Use temp dir on same filesystem for source install

IanButterworth:
Fixup #4168 
I missed this detail in https://github.com/JuliaLang/Pkg.jl/pull/4168#discussion_r1970925988

See failures in https://github.com/JuliaLang/julia/pull/57598

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -717,10 +717,13 @@ function install_archive(
     version_path::String;
     io::IO=stderr_f()
 )::Bool
+    depot_temp = mkpath(joinpath(dirname(dirname(dirname(version_path))), "tmp"))
     tmp_objects = String[]
     url_success = false
     for (url, top) in urls
-        path = tempname() * randstring(6)
+        # the temp dir should be in the same depot because the `rename` operation in `mv_temp_dir_retries`
+        # is possible only if the source and destination are on the same filesystem
+        path = tempname(depot_temp) * randstring(6)
         push!(tmp_objects, path) # for cleanup
         url_success = true
         try
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -65,7 +65,8 @@ set_readonly(::Nothing) = nothing
     mv_temp_dir_retries(temp_dir::String, new_path::String; set_permissions::Bool=true)::Nothing
 
 Either rename the directory at `temp_dir` to `new_path` and set it to read-only
-or if `new_path` already exists try to do nothing.
+or if `new_path` already exists try to do nothing. Both `temp_dir` and `new_path` must
+be on the same filesystem.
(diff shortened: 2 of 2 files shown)


maleadt:
Looks like PkgEval is stumbling on this change, or any from the preceding PRs. Lots of new skips in https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_date/2025-03/04/report.html, e.g., https://s3.amazonaws.com/julialang-reports/nanosoldier/pkgeval/by_date/2025-03/04/PyCall.primary.log:

```
ERROR: LoadError: Error when installing package Conda:
IOError: rename of "/home/pkgeval/.julia/tmp/jl_wNc5SF4dC1wSxoe4" to "/home/pkgeval/.julia/packages/Conda/zReqD": cross-device link not permitted (EXDEV)
Stacktrace:
 [1] uv_error
   @ ./libuv.jl:106 [inlined]
 [2] mv_temp_dir_retries(temp_dir::String, new_path::String; set_permissions::Bool)
   @ Pkg /opt/julia/share/julia/stdlib/v1.13/Pkg/src/utils.jl:111
 [3] mv_temp_dir_retries
   @ /opt/julia/share/julia/stdlib/v1.13/Pkg/src/utils.jl:71 [inlined]
 [4] install_archive(urls::Vector{Pair{String, Bool}}, hash::Base.SHA1, version_path::String; io::IOContext{IO})
   @ Pkg.Operations /opt/julia/share/julia/stdlib/v1.13/Pkg/src/Operations.jl:765
```

Any thoughts? PkgEval does have a somewhat peculiar filesystem set-up (with some directories mounted from an overlay filesystem), but Pkg shouldn't stumble on that.

 cc @KristofferC

IanButterworth:
So `.depot/tmp` could be on a different fs to `.depot/packages`? That seems quite unusual.

The temp write could be moved to within `packages`?

IanButterworth:
If it's not clear, the reason it's stumbling is because we now use `rename` not `mv`.

maleadt:
`compiled`, `artifacts`, and `packages` are mounted from an overlay filesystem indeed: https://github.com/JuliaCI/PkgEval.jl/blob/ff3294147c090023b6b491fcd025500abdec8b82/src/evaluate.jl#L289-L300

IanButterworth:
I'm happy to move the temp write to `packages/temp`. I think a single temp dir in there with any potential orphans is clear-enough.

maleadt:
> I'm happy to move the temp write to `packages/temp`.

That seems safer in general; users doing something special with `depot/packages` (like PkgEval does) doesn't seem that outrageous, so it'd be great to support that.

PR #4186, opened 2025-03-07
Update registries if necessary during `dev`

### LINKED ISSUE #4185: Resolver error if a manifest contains a version that isn't in registries or local packages
If a Manifest contains a newer version that the current session doesn't know about, say because the manifest is checked in and the local registry hasn't been updated recently, you can end up with resolver errors during a `dev` on the local package.

It'd probably be good to try updating registries beforehand, either every time `dev` is called, or just try again before throwing a resolver error.

IanButterworth:
Fixes #4185 

Otherwise resolve can fail if the manifest that the package is being dev-ed into is in a newer state than the local machine's registry.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -249,6 +249,7 @@ function develop(ctx::Context, pkgs::Vector{PackageSpec}; shared::Bool=true,
 
     new_git = handle_repos_develop!(ctx, pkgs, shared)
 
+    Operations.update_registries(ctx; force=false, update_cooldown=Day(1))
 
     for pkg in pkgs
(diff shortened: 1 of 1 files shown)


src/API.jl
```diff
@@ -281,6 +281,7 @@ function develop(
 
     new_git = handle_repos_develop!(ctx, pkgs, shared)
 
+    Operations.update_registries(ctx; force=false, update_cooldown=Day(1))
```
IanButterworth:
```suggestion
    Operations.update_registries(ctx; force = false, update_cooldown = Day(1))
```

PR #4188, opened 2025-03-14
Store temp packages in packages/temp

IanButterworth:
Fixes issue reported here https://github.com/JuliaLang/Pkg.jl/pull/4180#issuecomment-2715408525

cc. @maleadt


maleadt (approved):
Thanks!

Issue #4197, opened 2025-03-28
[sources] fails to pull remote deps when nested


mortenpi:
So if you have "nested" local deps like this (full MWE here: https://github.com/mortenpi/pkg-4197-mwe):

```toml
name = "PackageA"
uuid = "bbcb243f-f034-2038-cf56-a1fff5ae3881"
authors = ["JuliaHub Inc"]
version = "0.0.0"

[deps]
PackageB = "35109ab6-39ae-8b6d-e3b4-cc633fdcfb67"

[sources]
PackageB = { path = "../PackageB" }
```

```toml
name = "PackageB"
uuid = "35109ab6-39ae-8b6d-e3b4-cc633fdcfb67"
authors = ["JuliaHub Inc"]
version = "0.0.0"

[deps]
PackageC = "b160dae6-bb54-ebdb-3ad7-d2b4dbb778f3"
ReturnScope = "3caac13a-2753-4d5a-becf-3271e52727bc"

[sources]
PackageC = { path = "../PackageC" }
ReturnScope = { url = "https://github.com/mortenpi/ReturnScope.jl" }
```

Pkg correctly figures out where to find `PackageC`, but not how to pull `ReturnScope` (i.e. any remote dep provided via `url`). If you comment out `ReturnScope` it works the following works as expected:

```
$ julia +nightly --project=PackageA/ -e 'using Pkg; Pkg.instantiate()'
    Updating registry at `~/.julia/registries/General.toml`
ERROR: Unsatisfiable requirements detected for package ReturnScope [3caac13a]:
 ReturnScope [3caac13a] log:
 ├─ReturnScope [3caac13a] has no known versions!
 └─restricted to versions * by PackageB [35109ab6] — no versions left
   └─PackageB [35109ab6] log:
     ├─possible versions are: 0.0.0 or uninstalled
     ├─restricted to versions * by PackageA [bbcb243f], leaving only versions: 0.0.0
     │ └─PackageA [bbcb243f] log:
     │   ├─possible versions are: 0.0.0 or uninstalled
     │   └─PackageA [bbcb243f] is fixed to version 0.0.0
     └─PackageB [35109ab6] is fixed to version 0.0.0
Stacktrace:
  [1] check_constraints(graph::Pkg.Resolve.Graph)
    @ Pkg.Resolve ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/src/Resolve/graphtype.jl:1046
[...]

PR #4198, opened 2025-03-28
also collect packages tracking a repo when "traversing" developed packages

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4197.

Should add a test.


mortenpi:
Works! Both on the MWE and the real-world use case that prompted the issue.

PR #4202, opened 2025-03-30
Lock io within printpkgstyle to avoid print tearing

IanButterworth:
This was sometimes breaking up printing so the test failed. i.e.
```
Installed 2025-03-29T21:02:12.769
artifact FFMPEG                10.6 MiB
```

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -1,7 +1,8 @@
+# "Precompiling" is the longest operation
+const pkgstyle_indent = textwidth(string(:Precompiling))
 
 function printpkgstyle(io::IO, cmd::Symbol, text::String, ignore_indent::Bool=false; color=:green)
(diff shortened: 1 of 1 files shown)


IanButterworth:
I guess `printpkgstyle` could have been more robust against this, by not yielding the io between these two prints, but I'm not sure how to achieve that

https://github.com/JuliaLang/Pkg.jl/blob/07235d0e9a4bc75ddcb36eff6165806cec78b049/src/utils.jl#L5-L6

DilumAluthge:
Is there a way for us to "lock" the `io`?

IanButterworth:
Yeah, TIL https://github.com/IanButterworth/julia/blob/a046da5f259f74011d5dab6716ac9c9fbe4424d7/base/strings/io.jl#L32-L40

Issue #3460, opened 2023-05-08
TestEnv.jl functionality/support


oxinabox:
The code in TestEnv.jl is made from hacking up the internal code from Pkg.
This means it often breaks, due to changes in Pkg that are not mirrored in TestEnv yet.
Though for the last few releases the Pkg internals seem to be back in a period of stability.

I think at this point, TestEnv.jl has proven the utility of being able to activate the test enviroment. 
And that it would be reasonable to add `] activate --test` or similar.
If we did that and stdlib-ized the the TestEnv functionality then it wouldn't need 
Could keep TestEnv.jl around for compat reasons, but it could just delegate to Pkg if version is beyond 1.10

An alternative would be to expose just enough functionality to do this.
This probably means a public API of:
 - extract UUID from `[extras]` 
 -  easily get lists of packages from `target.test` 
 - and easily subset `[compat]`
 - Merge  Projects/Manifests

fredrikekre:
Duplicate of https://github.com/JuliaLang/Pkg.jl/issues/1233 ?

oxinabox:
no, it is related but distinct.
In that the TestEnv thing *also* needs to keep support for the way tests are specified Julia 1.0 style. with `[extras]`. Which is not going to be removed any time soon. As well as the 1.2 approximation to sub-proects with `test/Project.toml`.

IanButterworth:
FWIW I took a look at implementing `pkg> activate --test` (and `pkg> activate --build` given its a supported target) but it's relatively involved.

It would be nice to get those in though, I agree that TestEnv has shown how helpful it would be.

ufechner7:
Any update on this? Is there a consensus that this would be useful?

IanButterworth:
WIP here https://github.com/JuliaLang/Pkg.jl/pull/4218

KristofferC:
At least to me, the workspace feature is kind of supposed to subsume the implicit test env creation and those projects you can just activate as a normal julia project (`--project`).

oxinabox:
I do think, even if the goal is people use the new workspaces feature.
(I remain unconvinced personally, seems too complex for a little package)
it would still be nice to have these features needed to build TestEnv.jl.
Those features would let people do lots of interesting things with their dependencies.

The key one that is nontrivial to implement is the _Merge Projects/Manifests_, in the same way Pkg does right now for tests.
There is a lot of complexity and corner cases.
[...]

PR #4218, opened 2025-04-11
Add ability to activate test/build target envs

IanButterworth:
Closes https://github.com/JuliaLang/Pkg.jl/issues/3460

```
(Pkg) pkg> activate /Users/ian/Documents/GitHub/ObjectDetector.jl
  Activating project at `~/Documents/GitHub/ObjectDetector.jl`

(ObjectDetector) pkg> activate --test
    Creating temporary project based on the target `test` in `~/Documents/GitHub/ObjectDetector.jl/Project.toml`
  Activating project at `/var/folders/1z/jf841bdj73bdj3vk7kc7f_3w0000gn/T/jl_wHY4q9`

(jl_wHY4q9) pkg> st
Status `/private/var/folders/1z/jf841bdj73bdj3vk7kc7f_3w0000gn/T/jl_wHY4q9/Project.toml`
  [79e6a3ab] Adapt
  [5c00bae2] AllocArrays
  [6e4b80f9] BenchmarkTools
  [052768ef] CUDA
  [e2912957] Darknet
  [5789e2e9] FileIO
  [587475ba] Flux
  [a09fc81d] ImageCore
  [4381153b] ImageDraw
  [6a3955dd] ImageFiltering
  [82e4d734] ImageIO
  [02fcd773] ImageTransformations
  [aea7be01] PrecompileTools
  [08abe8d2] PrettyTables
  [324d217c] ReferenceTests
  [a759f4b9] TimerOutputs
  [c4a57d5a] UnsafeArrays
  [02a925ec] cuDNN
  [4af54fe1] LazyArtifacts
  [8dfed614] Test
```


davidanthoff:
This would be amazing to have! Lots of use for it in the VS Code extension. Probably too late for 1.12?

IanButterworth:
@KristofferC mentioned in https://github.com/JuliaLang/Pkg.jl/issues/3460 that the new workspaces feature was a better way to handle this in 1.12+

We could add this functionality, or we could not and focus on prodding people towards using workspaces in 1.12+. 

It came up in the #testing slack channel that this section of the docs should be updated to highlight workspaces being the best way to do test envs. https://pkgdocs.julialang.org/v1/creating-packages/#Alternative-approach:-test/Project.toml-file-test-specific-dependencies

IanButterworth:
Also so I don't forget if we go with this PR.. there's a bug. It's currently not including the parent package in the test env.

davidanthoff:
It would still be great to have just one function that handles all the complexities of deciding how to set up the test env. I.e. if the workspace is there, is that, if the old style separate Project.toml is there, us that, or the target thing. We will essentially have three different ways to specify test env in 1.12 that all need to be supported, and it would be nice if Pkg.jl handled all cases.

Moelf:
with workspaces, this should be expanded to be able to activate any workspace or...? `--docs` would be useful, and `--benchmarks`

PR #4225, opened 2025-04-30
Re-land: Add entries to [sources] automatically if package is added by URL or devved 

lassepe:
This is a rebased version of #3826. Locally, all tests pass for me.


KristofferC:
Let's try this :)

IanButterworth:
Partial backport needed. See https://github.com/JuliaLang/Pkg.jl/pull/4370#issuecomment-3210907907

lassepe:
I'm not familiar with the backporting process and generally quite swamped right now with other things. So if someone else has the bandwidth for the backport, please go ahead :)

Issue #4211, opened 2025-04-06
`PackageSpec` implements `Base.:(==)` but not `Base.hash`


nhz2:
This causes the following strange behavior:

```julia-repl
julia> using Pkg

julia> allunique(unique([PackageSpec(path="foo"), PackageSpec(path="foo")]))
false
```

PR #4228, opened 2025-05-05
 add some missing hash definitions

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4211


nhz2 (approved):
Thank you for fixing this.

PR #4233, opened 2025-05-09
Support parsing version 0.0.0 internally for package versions

serenity4:
As was agreed in https://github.com/JuliaLang/julia/pull/57520, we decided to introduce a special `v"0.0.0"` version for Compiler as a fallback to `Base.Compiler`.

The issue is that Pkg internals [explicitly forbid](https://github.com/JuliaLang/Pkg.jl/blob/80d2e7505970b9f1a33b2a4e1c4c5bcea7a709b0/test/pkg.jl#L147) to use this version number, which shows up as

```julia
(Pkg) pkg> add Compiler@0.0.0
   Resolving package versions...
ERROR: invalid version: "0.0.0"
Stacktrace:
  [1] error(s::String)
    @ Base ./error.jl:44
  [2] semver_interval(m::RegexMatch{String})
    @ Pkg.Versions ~/.julia/dev/Pkg/src/Versions.jl:335
  [3] semver_spec(s::String; throw::Bool)
    @ Pkg.Versions ~/.julia/dev/Pkg/src/Versions.jl:310
  [4] semver_spec
    @ ~/.julia/dev/Pkg/src/Versions.jl:303 [inlined]
  [5] set_compat
    @ ~/.julia/dev/Pkg/src/Operations.jl:303 [inlined]
  [6] add(ctx::Pkg.Types.Context, pkgs::Vector{PackageSpec}, new_git::Set{Base.UUID}; allow_autoprecomp::Bool, preserve::PreserveLevel, platform::Base.BinaryPlatforms.Platform, target::Symbol)
    @ Pkg.Operations ~/.julia/dev/Pkg/src/Operations.jl:1725
  [...]
```

However, it seems that semantic versioning actually allows such a version number (see [this post](https://stackoverflow.com/a/66610835)). Furthermore, compat ranges generated by Pkg may in fact include `v"0.0.0"`, and are documented as such in https://pkgdocs.julialang.org/v1/compatibility/#Caret-specifiers. I also suspect that most things should "just work" with it.

For these reasons, it seems reasonable to me to add support for it. I would not be surprised however that it requires some discussion before a consensus is reached as to whether this is desirable or not.

Regarding the changes proposed in this PR, I would appreciate any pointers toward areas that require further modification or that are suspected to break.

(cc @aviatesk)

### DIFF
--- a/src/Versions.jl
+++ b/src/Versions.jl
@@ -331,9 +331,6 @@ function semver_interval(m::RegexMatch)
     major =                           parse(Int, _major)
     minor = (n_significant < 2) ? 0 : parse(Int, _minor)
     patch = (n_significant < 3) ? 0 : parse(Int, _patch)
-    if n_significant == 3 && major == 0 && minor == 0 && patch == 0
-        error("invalid version: \"0.0.0\"")
-    end
     # Default type is :caret
     vertyp = (typ == "" || typ == "^") ? :caret : :tilde
     v0 = VersionBound((major, minor, patch))
@@ -368,12 +365,10 @@ function inequality_interval(m::RegexMatch)
     major =                           parse(Int, _major)
     minor = (n_significant < 2) ? 0 : parse(Int, _minor)
     patch = (n_significant < 3) ? 0 : parse(Int, _patch)
-    if n_significant == 3 && major == 0 && minor == 0 && patch == 0
-        error("invalid version: 0.0.0")
-    end
     v = VersionBound(major, minor, patch)
     if occursin(r"^<\s*$", typ)
         nil = VersionBound(0, 0, 0)
+        v == v"0.0.0" && return VersionRange(nil, nil)
         if v[3] == 0
             if v[2] == 0
                 v1 = VersionBound(v[1]-1)
--- a/test/pkg.jl
+++ b/test/pkg.jl
@@ -90,6 +90,7 @@ import Pkg.Types: semver_spec, VersionSpec
     @test !(v"1.2.2" in semver_spec(">=1.2.3"))
 
     @test_throws ErrorException semver_spec("0.1.0-0.2.2")
+    @test semver_spec("0.0.0") == VersionSpec("0.0.0-0.0.0")
     @test semver_spec("0.1.0 - 0.2.2") == VersionSpec("0.1.0 - 0.2.2")
     @test semver_spec("1.2.3 - 4.5.6") == semver_spec("1.2.3  - 4.5.6") == semver_spec("1.2.3 -  4.5.6") == semver_spec("1.2.3  -  4.5.6")
     @test semver_spec("0.0.1 - 0.0.2") == VersionSpec("0.0.1 - 0.0.2")
@@ -144,7 +145,6 @@ import Pkg.Types: semver_spec, VersionSpec
 
     @test_throws ErrorException semver_spec("^^0.2.3")
     @test_throws ErrorException semver_spec("^^0.2.3.4")
-    @test_throws ErrorException semver_spec("0.0.0")
     @test_throws ErrorException semver_spec("0.7 1.0")
 
     @test Pkg.Types.isjoinable(Pkg.Types.VersionBound((1,5)), Pkg.Types.VersionBound((1,6)))


DilumAluthge:
I personally am not a fan of this approach. It's one thing when it was just the General registry, but now we are having to add hacks in Pkg? Personally I think it would be better to just have Compiler.jl start at 0.0.1.

But if @StefanKarpinski / @KristofferC / @fredrikekre are fine with this, I'll defer to them.

serenity4:
My expectation would be that we don't need any hacks, and that the implementation already mostly supports this "for free".

aviatesk:
As a compromise, how about we allow v0.0.0 on the Pkg.jl side, but only allow registration in the General registry in very specific cases?
In Pkg.jl, which follows Semver for versions, I don't think we need to reject something that Semver allows (v0.0.0).
Normally, v0.0.0 shouldn't be recommended, and I think it should definitely be avoided in the General registry. However, for example, people using private registries might want to use v0.0.0. And I don't think there's a need to reject such use cases.

KristofferC:
There seems to be a lot of confusion here. The version number 0.0.0 should be supported, it is only for the compat range specifier (`[compat]`) where it is not because it isn't ear what versions it should include. Should it only be 0.0.0?

serenity4:
A compat range of `[0.0.0, 0.0.1)` would seem consistent with `[0.0.1, - 0.0.2)`.

KristofferC:
When does this even come up? Why put a 0.0.0 compat specifier? Why not call it version 0.0.1?

serenity4:
I believe it is to indicate that this version is fundamentally different than any other, as in, `0.0.0` being a proxy for `Base.Compiler` (of the currently running Julia version, which can be anything), while versions `> 0.0.0` will most likely host a copy that matches the Compiler implementation of a given Julia version. In a sense, `0.0.0` stands for "it's not a real version" in this specific context for BaseCompiler.jl.

I would not be opposed myself to labeling it `0.0.1` though, and have downstream package authors know that it is special in that sense. I'm not sure these arguments in favor of `0.0.0` would bear enough justification to oppose a strong resistance. Perhaps @aviatesk has other arguments to bring.

IanButterworth:
How is this different to the way upgradable stdlibs work? Why does Compiler need something different?

KristofferC:
AFAIU, because it is so heavily tied to internals.

serenity4:
Yes, its tight integration with the Julia runtime implies that it regularly breaks after significant changes to the latter, which is every other week or so in the current state of development.

Release versions are not affected by this, but the development of libraries dependent on Compiler often requires changes to internals, which then requires `master`. For example, just last week we had https://github.com/JuliaLang/julia/pull/58324 and https://github.com/JuliaLang/julia/pull/58343 to fix a crash and a cache corruption for the development of DAECompiler.jl.

This special `0.0.0` version would provide better stability guarantees until a proper versioning and release scheme can be established, when the relevant internals become less of a moving target. At the same time, having it available through the package manager grants us the ability to `dev` it like a normal package, which is very important for debugging, and having a registered version allows CI to work with it reliably.

DilumAluthge:
I guess what I'm asking is why the special version can't just be called 0.0.1 instead of 0.0.0?

serenity4:
See the first paragraph of https://github.com/JuliaLang/Pkg.jl/pull/4233#issuecomment-2870218287.

I'm not opposed to calling it `0.0.1`, but I think it makes a bit more sense to call it `0.0.0`.

aviatesk:
Thank you @serenity4 for replying on my behalf.
My opinion is exactly the same as his. v0.0.0 is allowed by semver. Also, in cases like [BaseCompiler.jl](https://github.com/JuliaLang/BaseCompiler.jl), using `v0.0.0` feels more natural than using `v0.0.1`. For these reasons, I hope Pkg.jl will support v0.0.0.

StefanKarpinski:
To be clear, I'm fine with making a version of 0.0.0 work here like any other version. Why not? I'm even ok with people using it to have some special meaning. Where I draw the line is at changing the git tree hash that any version is associated with. Why is that relevant? Bc when the implementation of the special version inevitably changes, then you can't use the special 0.0.0 version number anymore because you already have it a meaning. So you could use 0.0.0+0 and then increment the build number. I'd be fine with that. But at that point you could also use a more normal version number. Given that Pkg doesn't handle build releases very well (resolver can't handle them, registry format can't bound on them), using only the build number to distinguish versions that you might need to put compat bounds on seems like a recipe for pain. If there's zero chance that you need to put compat bounds on different implementations of the special version, then using 0.0.0+0 format seems ok to me.

For the specific issue of supporting 0.0.0 versions, I'm totally fine with it.

KristofferC:
It seems this didn't end up being needed

serenity4:
The specific issue that prompted this PR used a different solution, that is correct. Nonetheless, this PR was also considered a minor improvement, so that's something we could merge.

I won't push more for it at this point though.

PR #4242, opened 2025-05-22
Don't rely on the default threads for the Pkg test

gbaraldi:


KristofferC:
I think this can be closed now that https://github.com/JuliaLang/Pkg.jl/pull/4243 is merged? Reopen if not.

PR #4245, opened 2025-05-22
WIP: `Pkg.Registry.update()`: throw if an error is encountered

### LINKED ISSUE #4244: Registry update failures should throw or return status
When using `Pkg.Registry.update()` if a registry fails to be updated an error is logged but there is no exception or return value to indicate this. This can make it difficult to work with this function call in automated scripts such as `Dockerfile`s where you may want to abort immediately if registries fail to update:

```
...
#14 1.239     Updating registry at `/usr/local/share/julia-depot/registries/Private`
#14 1.275     Updating git-repo `https://github.com/.../PrivateRegistry.git`
#14 1.621 Username for 'https://github.com' [git]:     Updating registry at `/usr/local/share/julia-depot/registries/General.toml`
#14 1.816 ┌ Error: Some registries failed to update:
#14 1.816 │     — /usr/local/share/julia-depot/registries/Private — failed to fetch from repo: failed to fetch from https://github.com/../PrivateRegistry.git, error: GitError(Code:EUSER, Class:Callback, Aborting, user cancelled credential request.)
#14 1.816 └ @ Pkg.Registry /usr/local/julia/share/julia/stdlib/v1.10/Pkg/src/Registry/Registry.jl:528
#14 2.026 ┌ Warning: The active manifest file has dependencies that were resolved with a different julia version (1.11.5). Unexpected behavior may occur.
#14 2.026 └ @ /project/Manifest.toml:0
...
```

DilumAluthge:
Fixes #4244 

With this PR, by default, if `Pkg.Registry.update()` encounters an exception, we will throw that exception.

However, the user can revert to the previous behavior (all exceptions are caught, none are rethrown) by specifying `throw_on_error = false`.

The name `throw_on_error` can be bikeshedded.

### DIFF
--- a/src/Registry/Registry.jl
+++ b/src/Registry/Registry.jl
@@ -392,7 +392,7 @@ function update(; name=nothing, uuid=nothing, url=nothing, path=nothing, linked=
         update([RegistrySpec(; name, uuid, url, path, linked)]; kwargs...)
     end
 end
-function update(regs::Vector{RegistrySpec}; io::IO=stderr_f(), force::Bool=true, depots = [depots1()], update_cooldown = Second(1))
+function update(regs::Vector{RegistrySpec}; io::IO=stderr_f(), force::Bool=true, depots = [depots1()], update_cooldown = Second(1), throw_on_error = true)
     registry_update_log = get_registry_update_log()
     for depot in depots
         depot_regs = isempty(regs) ? reachable_registries(; depots=depot) : regs
@@ -430,6 +430,7 @@ function update(regs::Vector{RegistrySpec}; io::IO=stderr_f(), force::Bool=true,
                                     try
                                         download_verify(url, nothing, tmp)
                                     catch err
+                                        throw_on_error && rethrow()
                                         push!(errors, (reg.path, "failed to download from $(url). Exception: $(sprint(showerror, err))"))
                                         @goto done_tarball_read
                                     end


DilumAluthge:
Hmmm. I think I should wait until the end, instead of throwing immediately.

KristofferC:
I don't really like this because `Registry.update` is automatically called in various places and you don't want the pkg operation to completely fail if one of the registries fail to update. For example, if you do `Pkg.add("Plots")` and some test registry fails to update you still want to install plots.

Keno:
In particular, the network might be down, but you still want to do the Project.toml modification.

DilumAluthge:
How about we set `throw_on_error = false` by default? So then, by default, people will get the existing behavior, but people can opt-in by specifying `throw_on_error = true`?

@omus I think that would still solve your use case, right?

DilumAluthge:
> How about we set `throw_on_error = false` by default? So then, by default, people will get the existing behavior, but people can opt-in by specifying `throw_on_error = true`?

I have implemented this.

DilumAluthge:
> Maybe could be:
> 
> We already have a `Pkg.precompile(; strict::Bool=false)` which reminds me of this

I think that `strict` would be fine, especially if we already use `strict` in other places.

So the options that we are considering are:

1. `throw_on_error`
2. `strict`

Are there any other options that we can think of?

IanButterworth:
I think just `throw` is enough. I think that's used elsewhere.

Issue #4086, opened 2024-11-15
Resolving a Project.toml with `[sources]` set for an existing dependency


fonsp:
Hi! I found an interesting case while experimenting for #4085:

Setup: I have a local clone of https://github.com/JuliaPluto/HypertextLiteral.jl checked out at some old commit, at `v0.9.4`. The current latest version is `v0.9.5`. 

I wrote this Project.toml:

```toml
[deps]
HypertextLiteral = "ac1192a8-f4b3-4bfe-ba22-af5b92cd3ab2"

[sources.HypertextLiteral]
path = "/Users/fons/Documents/HypertextLiteral.jl"
```

I make two empty directories that I activate.

## One
In the first environment, I write the Project.toml, and then `resolve`. The result is as expected:

```
(yay3) pkg> resolve
    Updating `~/Downloads/yay3/Project.toml`
  [ac1192a8] + HypertextLiteral v0.9.4 `~/Documents/HypertextLiteral.jl`
    Updating `~/Downloads/yay3/Manifest.toml`
  [ac1192a8] + HypertextLiteral v0.9.4 `~/Documents/HypertextLiteral.jl`
  [410a4b4d] + Tricks v0.1.9

shell> cat Project.toml
[deps]
HypertextLiteral = "ac1192a8-f4b3-4bfe-ba22-af5b92cd3ab2"

[sources.HypertextLiteral]
path = "/Users/fons/Documents/HypertextLiteral.jl"


shell> cat Manifest.toml
# This file is machine-generated - editing it directly is not advised

julia_version = "1.11.1"
manifest_format = "2.0"
project_hash = "312e4b9e605df01eeae246a2087d05a62416059a"

[[deps.HypertextLiteral]]
deps = ["Tricks"]
path = "/Users/fons/Documents/HypertextLiteral.jl"
uuid = "ac1192a8-f4b3-4bfe-ba22-af5b92cd3ab2"
version = "0.9.4"

[[deps.Tricks]]
git-tree-sha1 = "7822b97e99a1672bfb1b49b668a6d46d58d8cbcb"
uuid = "410a4b4d-49e4-4fbc-ab6d-cb71b17b3775"
version = "0.1.9"

(yay3) pkg> precompile

(yay3) pkg> st
Status `~/Downloads/yay3/Project.toml`
  [ac1192a8] HypertextLiteral v0.9.4 `~/Documents/HypertextLiteral.jl`
```

## Two
[...]

PR #4260, opened 2025-06-13
don't use tree hash from manifest if the path is set from sources

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4086


fonsp:
Awesome, thank you!!

PR #4266, opened 2025-06-24
Add a tip when trying to upgrade a specific package that's possible to upgrade, just not optimal

IanButterworth:
Screenshot 2025-06-24 at 4 54 02 PM


IanButterworth:
I think this can be done cleaner by using the old env in the env cache

IanButterworth:
I tried to come up with a test for this but it'd have to be quite complex, I think.

PR #4270, opened 2025-06-26
feat(errors): Improve error message for incorrect package UUID

### LINKED ISSUE #4267: "Expected package to be registered" could suggest that a UUID mismatch may be the cause
This might be a bit of a "you're holding it wrong" problem, but the various AI assistants will often hallucinate UUIDs, resulting in the `expected package to be registered` error. But this error is somewhat misleading - the package *is* registered, it just had the wrong UUID. This is made worse by the fact that the displayed part of the UUID is often the part that it hallucinates correctly. I think it would be nice to update the error message to suggest that there may be a UUID mismatch. Bonus points for checking for a registered package of the name in all available registries and printing out the full correct UUID.

KristofferC:
When a user attempts to add a package with a correct name but an incorrect UUID, the error message was previously a generic "package not registered". This could leave the user unsure of the root cause.

This commit enhances the error handling to detect this specific scenario. If the provided package name exists in the registry but with a different UUID, the error message will now include the correct UUID(s) found for that package name, guiding the user toward the correct command.

A new test case is added to ensure the new error message is triggered correctly and displays the expected information.

Fixes #4267

This commit was written by an AI assistant.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1594,7 +1594,15 @@ function check_registered(registries::Vector{Registry.RegistryInstance}, pkgs::V
     end
     pkg = is_all_registered(registries, pkgs)
     if pkg isa PackageSpec
-        pkgerror("expected package $(err_rep(pkg)) to be registered")
+        msg = "expected package $(err_rep(pkg)) to be registered"
+        # check if the name exists in the registry with a different uuid
+        if pkg.name !== nothing
+            uuids = Types.registered_uuids(registries, pkg.name)
+            if !isempty(uuids)
+                msg *= "\n You may have provided the wrong UUID for package $(pkg.name).\n Found the following UUIDs for that name: $(join(uuids, ", "))"
+            end
+        end
+        pkgerror(msg)
     end
     return nothing
 end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1597,9 +1597,18 @@ function check_registered(registries::Vector{Registry.RegistryInstance}, pkgs::V
         msg = "expected package $(err_rep(pkg)) to be registered"
         # check if the name exists in the registry with a different uuid
         if pkg.name !== nothing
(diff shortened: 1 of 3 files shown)


src/Operations.jl
```diff
@@ -1594,7 +1594,15 @@ function check_registered(registries::Vector{Registry.RegistryInstance}, pkgs::V
     end
     pkg = is_all_registered(registries, pkgs)
     if pkg isa PackageSpec
-        pkgerror("expected package $(err_rep(pkg)) to be registered")
+        msg = "expected package $(err_rep(pkg)) to be registered"
+        # check if the name exists in the registry with a different uuid
+        if pkg.name !== nothing
+            uuids = Types.registered_uuids(registries, pkg.name)
+            if !isempty(uuids)
+                msg *= "\n You may have provided the wrong UUID for package $(pkg.name).\n Found the following UUIDs for that name: $(join(uuids, ", "))"
```
Keno: Should this tell you what registry it came from?

KristofferC: Probably, but I got rate limited on Google Gemini CLI so I guess this is as good as it gets 🤣

Keno: Meh, I'll take it

KristofferC: You might be in luck, I might have been able to get some $300 free tier thingy.

KristofferC:
```
julia> Pkg.add(PackageSpec(name="Example", uuid=UUID(UInt128(2))))
   Resolving package versions...
ERROR: expected package `Example [00000000]` to be registered
 You may have provided the wrong UUID for package Example.
 Found the following UUIDs for that name:
  - 7876af07-990d-54b4-ab0e-23690620f79a from registry: General
  ```

Keno: Perfect

Keno (approved):

PR #4278, opened 2025-06-30
Various app improvements

KristofferC:
- Avoids directly interpolating external data into files that will be executed.
- Warn if an app is not found by `Sys.which` after being installed
- Remove an unused `AppInfo` field


src/Apps/Apps.jl
```diff
@@ -38,6 +71,18 @@ function overwrite_file_if_different(file, content)
     end
 end
 
+function check_apps_in_path(apps)
+    for app_name in keys(apps)
+        if Sys.which(app_name) === nothing
```
fredrikekre: Perhaps also warn if it points to another executable too?

KristofferC: Yeah, makes sense

KristofferC: Added in last commit

PR #4284, opened 2025-07-01
Add readonly field support for Project.toml environments

### LINKED ISSUE #4195: Read-only flag
It would be nice to be able to set an environment to "read only" mode, such that users would get a nice informative error message when they tried to modify it (compared to an access error when setting the toml files to readonly via chmod). This would be useful for cases where a pre-made environment is supplied and the admin wants to avoid users diverging from the intended set of preinstalled packages (because they're in some sysimage, or because they have precompilation files shipped, or any other reason).

KristofferC:
Implements readonly environment functionality to prevent modifications when readonly = true is set in Project.toml. Adds parsing, validation, and error handling for readonly environments with informative messages.

Fixes #4195

🤖 Generated with [Claude Code](https://claude.ai/code)


KristofferC:
I don't think so, the manifest is not user writable so what reason is there to have it there?

IanButterworth:
Just noticed this isn't in the changelog

PR #4285, opened 2025-07-01
update CHANGELOG with some new things

KristofferC:


CHANGELOG.md
```diff
@@ -1,10 +1,20 @@
+Pkg v1.13 Release Notes
+=======================
+
+- Project.toml environments now support a `readonly` field to mark environments as read-only, preventing modifications. ([#4284])
+- Pkg now automatically adds entries to `[sources]` when packages are added by URL or devved, improving workflow consistency. ([#4225])
```
GunnarFarneback, line 5: This is suspiciously similar to line 18.

KristofferC: Ah, thanks, I don't think that should be in 1.12 because it is not backported (https://github.com/JuliaLang/Pkg.jl/pull/4225)

PR #4288, opened 2025-07-02
Format code base with runic

KristofferC:
I think it is time for this. The code base uses a salad of different conventions and styles as it has been developed over time.  This also adds a check that runic formatting is adhered to to CI and a pre commit hook that can be installed.

We should add a `.git-blame-ignore-revs` after this.


KristofferC:
This will pick up merge conflicts infinitely fast so I suggest merging this when it is green next time.

IanButterworth:
Or can we delay it until we've got a few of these easy PRs in? It'll be quite disruptive to old PRs

KristofferC:
True, I didn't know there was a bunch of other PRs waiting to be merged.

IanButterworth:
We didn't, you just prompted me to do a pass at lining easy ones up.

IanButterworth:
I was just thinking that we should also format the active release branches, but only after all backports have been done from before master was formatted

KristofferC:
👍

PR #4289, opened 2025-07-02
fix unused imports reported by ExplicitImports.jl

KristofferC:
There are more `imports` we can move to `using :` but this cleans up quite a bit for now. Also adds some CI to catch future things.

### DIFF
--- a/.github/workflows/check.yml
+++ b/.github/workflows/check.yml
@@ -28,3 +28,30 @@ jobs:
       - uses: fredrikekre/runic-action@v1
         with:
           version: "1.4" # Keep version in sync with .pre-commit-config.yaml
+
+  explicit-imports:
+    runs-on: ubuntu-latest
+    name: "ExplicitImports.jl"
+    steps:
+      - uses: actions/checkout@v4
+      - uses: julia-actions/setup-julia@v2
+        with:
+          version: 'nightly'
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-buildpkg@v1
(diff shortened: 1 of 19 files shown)


KristofferC:
```
ERROR: LoadError: ImplicitImportsException
Module `Pkg.Registry` is relying on the following implicit imports:
* `print` which is exported by Base
```

not sure what that means, I also don't get it locally...

KristofferC:
@ericphanson any idea w.r.t the comment above?

.github/workflows/check.yml
```diff
@@ -28,3 +28,30 @@ jobs:
...
+          version: 'nightly'
+      - uses: julia-actions/cache@v2
+      - uses: julia-actions/julia-buildpkg@v1
+      - name: Install dependencies
+        shell: julia {0}
+        run: |
+          # Add ExplicitImports.jl
+          using Pkg
+          Pkg.add([
+            PackageSpec(name = "ExplicitImports", version = "1.6"),
```
ericphanson: maybe it's a bug that has been fixed? not sure but 1.6 is pretty old
```suggestion
            PackageSpec(name = "ExplicitImports", version = "1.12"),
```

ericphanson:
it is supposed to skip flagging usages of names from Base/Core by default, looks like that isn't happening here for some reason. If it doesn't repro locally that adds credence to it being a bug that was fixed since the old version used in the CI script

IanButterworth:
Looks good now

Issue #1263, opened 2019-07-25
Method to print valid range of a version specifier


CameronBieganek:
It would be handy if there was a method exported in **Pkg** that would take a version specifier and print the range of valid versions. In other words, it should work as follows:

```julia
julia> versionrange("^1.2")
[1.2.0, 2.0.0)

julia> versionrange("~1.2")
[1.2.0, 1.3.0)
```

PR #4292, opened 2025-07-02
Add new public function `satisfies_compat::VersionNumber, ::String)`, and add (to docs) some examples of checking `[compat]` entries

DilumAluthge:
Replacement for #3847 (closes #3847).

This is an alternative for #3847. AFAICT, it satisfies the use case of #3847, and has the advantages of not exposing internal types, and not renaming any internal functions.

I'm not at all married to the name `satisfies_compat` - I'm happy to change it.


DilumAluthge:
cc: @MilesCranmer

IanButterworth:
SGTM +1

DilumAluthge:
CHANGELOG entry added.

DilumAluthge:
I've added a few tests.

We also have a few doctests in the docstring.

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@ Pkg v1.13 Release Notes
 
 - Project.toml environments now support a `readonly` field to mark environments as read-only, preventing modifications. ([#4284])
 - Pkg now automatically adds entries to `[sources]` when packages are added by URL or devved, improving workflow consistency. ([#4225])
+- New public function `Pkg.satisfies_compat`. ([#4292])
```
IanButterworth:
```suggestion
- New public function `Pkg.satisfies_compat` for checking whether a VersionNumber satisfies a compat string definition. ([#4292])
```

IanButterworth (approved):
SGTM but worth getting @KristofferC review too

IanButterworth (approved):

docs/src/compatibility.md
```diff
@@ -163,6 +163,40 @@ PkgA = "0.2 - 0.5"     # 0.2.0 - 0.5.* = [0.2.0, 0.6.0)
...
+### Checking specifications
+
+You can check if a particular version of a package is contained in a particular
+range by using the `Pkg.satisfies_compat` function. For example:
+
+```julia
+julia> Pkg.satisfies_compat(v"0.1.0", "=0.1")
+true
+
+julia> Pkg.satisfies_compat(v"0.1.0", "=0.1.1")
```
KristofferC, line 175: I think we can reduce the number of examples a bit, I think people will understand how to use it after one or two examples.

KristofferC (commented):
LGTM, with the note that I think we can reduce the number of examples given a bit.

KristofferC (commented):
LGTM, with the note that I think we can reduce the number of examples given a bit.

KristofferC (commented):
LGTM, with the note that I think we can reduce the number of examples given a bit.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/issues/1263 wants to get the range returned. Perhaps we can also do that (but only document that you can call `in` on that range). Then you would also get the pretty printing.

IanButterworth:
Could that be a follow-on? I can tidy up the examples to get this merged if so.

KristofferC:
But if we go with that there is no need for the `satisfies_compat` method at all?

IanButterworth:
Can all semver specs be represented by a single range object?
```
julia> Pkg.Types.semver_spec("1, 2, =3.0, 3.2")
VersionSpec("[1 - 2, 3.0.0, 3.2.0 - 3]")
```

KristofferC:
A VersionSpec can contain a collection of ranges.

IanButterworth:
So how does the user test inclusion of a version in that collection of ranges?

You're proposing returning a range not a VersionSpec?

KristofferC:
A VersionSpec is a single object (that can internally have many ranges in it) , that you can test `in` with a VersionNumber with.

IanButterworth:
Ok, I was understanding that you wanted to return a single range, not the VersionSpec (which would need to become public, as discussed in https://github.com/JuliaLang/Pkg.jl/pull/3847#issuecomment-2002662230 which lead to this alternative PR)

> https://github.com/JuliaLang/Pkg.jl/issues/1263 wants to get the range returned. Perhaps we can also do that (but only document that you can call in on that range).

IanButterworth:
Making VersionSpec public seems fine to me.

KristofferC:
Ah sorry, I should have been more accurate and said `VersionSpec`.

Issue #4253, opened 2025-06-04
Why does Artifacts.jl contact JULIA_PKG_SERVER first (defaults to pkg.julialang.org)?


quachpas:
Hello! I have noticed that Artifacts.jl will always contact JULIA_PKG_SERVER (by default pkg.julialang.org) first, despite the specified source in Artifacts.toml.

I understand this is a change that dates from 6 years ago (258d74ecd4104aadb34ff65f384782eb9a8eacd5), but is there a valid reason to keep that behaviour? I don't think Artifacts.jl should contact unexpected registries on a opt-out basis, i.e., we have to set JULIA_PKG_SERVER _only for Artifacts_. Additionally, this probably adds some unnecessary load to the storage servers.

https://github.com/JuliaLang/Pkg.jl/blob/5577f68d612139693282c037d070f515bf160d1b/src/Artifacts.jl#L485-L502

KristofferC:
You mean that Pkg should only contact the package server if the package is in a registry provided by the server? If that is the case, I agree and we do that for packages already.

quachpas:
Thanks for the quick response. 
Registry-based downloads will reduce the number of missed hits for sure! 
My question was more on why does Artifacts need to contact the default package server first, instead of trying the download.url and failing? For Julia packages, it makes sense to use a default server, but for generic data/Artifacts, I don't see a good reason.

KristofferC:
Better and more consistent user experience since the julia community have control over the package servers and can ensure they have good regional coverage and caching and speed compared to some random URL. Same as for packages pretty much.

quachpas:
Thanks for the response! I understand why the default is here.

Would there be any point in adding a new configuration, e.g., JULIA_ARTIFACTS_SERVER, that will be point to https://pkg.julialang.org/artifacts by default? I guess not, because if someone distrusts JULIA_PKG_SERVER (override), then they might also not want the Artifacts.

johnnychen94:
> Would there be any point in adding a new configuration, e.g., JULIA_ARTIFACTS_SERVER, that will be point to https://pkg.julialang.org/artifacts by default?

IMO, this would be better handled on the infrastructure side, e.g., setting up an Nginx proxy that routes regular requests to pkg.julialang.org and any other necessary destinations to your own. Maximum customizability😆

PR #4297, opened 2025-07-03
only try download artifacts from pkg server if the corresponding package is in a registry from it

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4253

Does someone have an unregistered package with artifacts I can try this on?

(written with some help from Claude Code).


KristofferC:
I've confirmed this works locally.

Issue #3523, opened 2023-06-22
Error message could recommend a resolution


jariji:
```
pkgerror("expected package $(err_rep(pkg)) to exist at path `$path`")
```

I got this error from `add` but it didn't tell me how to fix the problem, which would be helpful.

Issue #2191, opened 2020-11-02
give better diagnostic for missing package


rfourquet:
I renamed a local package (was `InlineTest`), and then got stuck with the following Pkg error (e.g. after trying to `up` or `dev` a package):
```
ERROR: expected package `InlineTest [e0db7c4e]` to exist at path `/local/src/julia-packages/InlineTest`
```
It took me a while to find what the problem was: I had been using `InlineTest` in a package `P`, present in my global environment; then deleted manually `InlineTest` from `P`'s Project.toml file, and renamed `InlineTest` (would have been the same if it got deleted). The problem was that `InlineTest` was still referenced in `P`'s manifest file. It would be good if `Pkg` could give a hint as to where the error comes from, i.e. tell why it thinks `InlineTest` should exist.

As an aside, if possible, it could be great if this error didn't block other `Pkg` operations (like `dev`ing another package), as it forced me to look into this problem "right now" .

KristofferC:
We can give a better error message for sure but I think this has to block operations because the resolver result is based on the compat of all packages.

PR #4303, opened 2025-07-03
Add more info when package is missing

IanButterworth:
Closes https://github.com/JuliaLang/Pkg.jl/issues/2191
Closes https://github.com/JuliaLang/Pkg.jl/issues/3523

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -417,7 +417,25 @@ function collect_fixed!(env::EnvCache, pkgs::Vector{PackageSpec}, names::Dict{UU
             path = project_rel_path(env, source_path(env.manifest_file, pkg))
         end
         if !isdir(path)
-            pkgerror("expected package $(err_rep(pkg)) to exist at path `$path`")
+            # Find which packages depend on this missing package for better error reporting
+            dependents = String[]
+            for (dep_uuid, dep_entry) in env.manifest.deps
+                if pkg.uuid in values(dep_entry.deps) || pkg.uuid in values(dep_entry.weakdeps)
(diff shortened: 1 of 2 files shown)


src/Operations.jl
```diff
@@ -417,7 +417,25 @@ function collect_fixed!(env::EnvCache, pkgs::Vector{PackageSpec}, names::Dict{UU
...
+            end
+
+            error_msg = "expected package $(err_rep(pkg)) to exist at path `$path`"
+            error_msg *= "\n\nThis package is referenced in the manifest file: $(env.manifest_file)"
+
+            if !isempty(dependents)
+                if length(dependents) == 1
+                    error_msg *= "\nIt is required by: $(dependents[1])"
+                else
+                    error_msg *= "\nIt is required by:\n$(join(["  - $dep" for dep in dependents], "\n"))"
```
KristofferC, line 435: Should be fine but I'm always scared these O(n^2) string builders will pop-up and be slow in some odd case.

IanButterworth: I think this will only be direct dependents, so should be ok

test/pkg.jl
```diff
@@ -730,6 +730,39 @@ end
...
+        end
+
+        # Now remove the B directory to simulate the missing package scenario
+        rm("B", recursive=true)
+
+        # Try to perform an operation that would trigger the missing package error
+        cd("A") do
+            try
+                Pkg.resolve()
+                @test false "Expected a PkgError to be thrown"
```
IanButterworth:
```suggestion
                @test false # a PkgError should be thrown"
```

Issue #3817, opened 2024-02-28
Initialy display of possible versions  doesn't take into account yanked versions


oxinabox:
Consider:
```
(jl_ykwcvm) pkg> add KernelAbstractions@0.9.16
   Resolving package versions...
ERROR: Unsatisfiable requirements detected for package KernelAbstractions [63c18a36]:
 KernelAbstractions [63c18a36] log:
 ├─possible versions are: 0.1.0 - 0.9.17 or uninstalled
 └─restricted to versions 0.9.16 by an explicit requirement — no versions left
```

KernelAbstractions v0.9.16 was yanked.
The correct display is:
`possible versions are: 0.1.0 - 0.9.15,  0.9.17 or uninstalled`

I think the problem is here https://github.com/JuliaLang/Pkg.jl/blob/1f16df404a2fbe8642ea3eecc9f4d7064c400a73/src/Resolve/graphtype.jl#L625C21-L625C59

It does: `range_compressed_versionspec(versions)`
but I think it should do `range_compressed_versionspec(versions, [ver.v for ver in ... if !ver.isyanked])`
but idk what goes in the `...`

I wrote this display code (a long time ago), It is smart enough to know about skipped over versions if you tell it about them.
I think that should work for yanked versions too

KristofferC:
> but I think it should do range_compressed_versionspec(versions, [ver.v for ver in ... if !ver.isyanked])


To me, filtering out yanked versions should already have happened at this stage in https://github.com/JuliaLang/Pkg.jl/blob/1f16df404a2fbe8642ea3eecc9f4d7064c400a73/src/Operations.jl#L532 so that the resolver only has non-yanked versions as input.

KristofferC:
What goes into the resolver is:

```
[...]

PR #4310, opened 2025-07-06
Warn about yanked versions. Add docs about yanked versions.

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/3817
Or at least a proposal.

The implementation may be optimizable.

Screenshot 2025-07-06 at 1 22 02 PM

cc. @oxinabox


src/Types.jl
```diff
@@ -109,11 +110,12 @@ function PackageSpec(; name::Union{Nothing,AbstractString} = nothing,
                        repo::GitRepo = GitRepo(),
                        path::Union{Nothing,AbstractString} = nothing,
                        pinned::Bool = false,
+                       yanked::Bool = false,
```
KristofferC: I am not a big fan of adding this to the PackageSpec. 

I think it's better to look it up in the registry when printing in status.

PR #4326, opened 2025-07-16
add ExplicitImports to CI check

KristofferC:
Had this in https://github.com/JuliaLang/Pkg.jl/pull/4288


ericphanson:
we do have https://github.com/JuliaTesting/ExplicitImports.jl/blob/main/.pre-commit-hooks.yaml in case you want to use that, fyi. https://github.com/JuliaTesting/ExplicitImports.jl/tree/main?tab=readme-ov-file#pre-commit-hooks

IanButterworth:
@ericphanson I'm struggling to get the recommended script to work because it expects ExplicitImports to be a dep of the main Pkg Project.toml and I don't think exposes a way to specify a different Project.toml and retain the pwd that files should be evaluated from?

ericphanson:
Cc @abelsiqueira

abelsiqueira:
In the linting workflow, we explicitly add ExplicitImports and JuliaFormatter (https://github.com/JuliaBesties/BestieTemplate.jl/blob/main/.github/workflows/Lint.yml#L32).
For the local development, we install these packages globally.

The `language: julia` for pre-commit did not exist when we created the pre-commit hook for explicit imports, so it doesn't support `additional_dependencies` (https://pre-commit.com/#julia).
@ericphanson, we could update the hook to use the Julia language, like https://github.com/fredrikekre/runic-pre-commit/blob/master/.pre-commit-hooks.yaml. What do you think?

ericphanson:
I got it working with https://github.com/JuliaLang/Pkg.jl/pull/4329 but it needs a pre-commit change to support in-repo hooks for julia. So if we want it soon we'd have to move the hooks out I think. Or maybe the best course is to revert to https://github.com/JuliaLang/Pkg.jl/pull/4326/commits/2dcbe44479b53b32ab0683e4177a37f19c70f20c for now since it seems like that was working.

KristofferC:
I changed this to just run on CI. @ericphanson, are these bugs in ExplicitImports:

```julia
  using UUIDs: UUIDs # used at /Users/kc/JuliaPkgs/Pkg.jl/src/REPLMode/REPLMode.jl:8:7

  using Dates: @dateformat_str # used at /Users/kc/JuliaPkgs/Pkg.jl/src/Registry/Registry.jl:668:57
```

The first one seems wrong, at that code location the source code is

```julia
using UUIDs: UUID
```

The second one has an explicit import for that symbol:

```julia
using Dates: @dateformat_str, Dates, DateTime, Millisecond, Second, now
```

ericphanson:
> I changed this to just run on CI. @ericphanson, are these bugs in ExplicitImports:
> 
> ```julia
>   using UUIDs: UUIDs # used at /Users/kc/JuliaPkgs/Pkg.jl/src/REPLMode/REPLMode.jl:8:7
> 
>   using Dates: @dateformat_str # used at /Users/kc/JuliaPkgs/Pkg.jl/src/Registry/Registry.jl:668:57
> ```
> 
> The first one seems wrong, at that code location the source code is
> 
> ```julia
> using UUIDs: UUID
> ```
> 
> The second one has an explicit import for that symbol:
> 
> ```julia
> using Dates: @dateformat_str, Dates, DateTime, Millisecond, Second, now
> ```

probably... IIRC there is an issue where if I removed these it would fail to recognize that `using UUIDs: UUIDs` was needed at all and would call it stale. So I think they are buggy but kind of a workaround.

KristofferC:
I don't really understand this:

> artifacts.jl: Error During Test at /Users/runner/work/Pkg.jl/Pkg.jl/test/runtests.jl:83
  Got exception outside of a `@test`
  LoadError: ArgumentError: Package Artifacts not found in current path.
 > - Run `import Pkg; Pkg.add("Artifacts")` to install the Artifacts package.

Artifacts is a dependency... Maybe it collides with Pkg.Artifacts somehow.

IanButterworth:
Or PkgArtifacts? https://github.com/JuliaLang/Pkg.jl/pull/4230

KristofferC:
```
aqua.jl: Error During Test at /home/runner/work/Pkg.jl/Pkg.jl/test/runtests.jl:83
  LoadError: ArgumentError: Package Aqua [4c88cf16-eb10-579e-8560-4a9242c79595] is required but does not seem to be installed:
   - Run `Pkg.instantiate()` to install all reco

 Error During Test at /home/runner/work/Pkg.jl/Pkg.jl/test/runtests.jl:83
  LoadError: ArgumentError: Package HistoricalStdlibVersions [6df8b67a-e8a0-4029-b4b7-ac196fe72102] is required but does not seem to be installed:
   - Run `Pkg.instantiate()` to install all recorded dependencies.
  ```

what is going on here... why cant it load any packages

File: src/API.jl [hunk 1]
```diff
@@ -1435,10 +1434,10 @@ function activate(path::AbstractString; shared::Bool = false, temp::Bool = false
             end
         end
     else
-        # initialize `fullpath` in case of empty `Pkg.depots()`
+        # initialize `fullpath` in case of empty `Pkg.Base.DEPOT_PATH`
         fullpath = ""
         # loop over all depots to check if the shared environment already exists
-        for depot in Pkg.depots()
+        for depot in Pkg.Base.DEPOT_PATH
```
IanButterworth, line 1440: This seems odd?

KristofferC: Yeah, I removed `depots()` and replaced it with Base.DEPOT_PATH. This is weird (should just be `Base.DEPOT_PATH`) but I don't think it should change anything..

PR #4331, opened 2025-07-24
Add a pkg message if julia_version is being specified

IanButterworth:
Probably useful for BinaryBuilder logs generally, but also debugging

### DIFF
--- a/src/Types.jl
+++ b/src/Types.jl
@@ -597,7 +597,7 @@ function Context!(ctx::Context; kwargs...)
     if haskey(kwargs, :julia_version) && ctx.julia_version !== nothing && ctx.julia_version != VERSION
         Pkg.printpkgstyle(
             ctx.io, :Context,
-            "Pkg is being asked to operate using julia_version `$(ctx.julia_version)`",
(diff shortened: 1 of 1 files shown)


src/Types.jl
```diff
@@ -592,6 +592,16 @@ function Context!(ctx::Context; kwargs...)
     for (k, v) in kwargs
         setfield!(ctx, k, v)
     end
+
+    # Highlight for logging purposes if julia_version is set to a different version than current VERSION
+    if haskey(kwargs, :julia_version) && ctx.julia_version !== nothing && ctx.julia_version != VERSION
+        Pkg.printpkgstyle(
+            ctx.io, :Context,
+            "Pkg is being asked to operate using julia_version `$(ctx.julia_version)`",
```
IanButterworth:
```suggestion
            "Pkg is operating with julia_version set to `$(ctx.julia_version)`",
```

KristofferC:
Seems a bit odd to do this at context creation (which is just creating an object). Feels like it should be done during resolving or something?

IanButterworth:
I put it here so it gets printed early. Given that this stuff has been buggy in the past it means it's shown even if we don't get to `Pkg.status`.

PR #4336, opened 2025-07-26
integrate the SAT-based resolver into Pkg

KristofferC:
This makes the SAT resolver (https://github.com/StefanKarpinski/Resolver.jl) a new default (so that people actually try it). We still need to make libpicosat available from julia itself (https://github.com/JuliaLang/julia/pull/59119).

This adds an extra CI run that runs with the maxsum resolver by default so that we properly test that.


DilumAluthge:
Should we make this "pluggable"? E.g., instead of special-casing `StefanKarpinski/Resolver.jl`, could we have a way for the user to provide a `Resolver` struct, and then Pkg uses that? Obviously we'd need to write out the interface spec for `Resolver` objects. But that would allow anyone to experiment with their own Resolver and then drop it in?

KristofferC:
I think that is kind of a premature generalisation.

PR #4341, opened 2025-07-30
Registry: Add lock to lazy fields in RegistryInstance and PkgEntry

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4340

### DIFF
--- a/src/Registry/registry_instance.jl
+++ b/src/Registry/registry_instance.jl
@@ -175,83 +175,89 @@ mutable struct PkgEntry
     const uuid::UUID
 
     const in_memory_registry::Union{Dict{String, String}, Nothing}
+    # Lock for thread-safe lazy loading
+    const info_lock::ReentrantLock
     # Version.toml / (Compat.toml / Deps.toml):
     info::PkgInfo # lazily initialized
 
-    PkgEntry(path, registry_path, name, uuid, in_memory_registry) = new(path, registry_path, name, uuid, in_memory_registry #= undef =#)
+    PkgEntry(path, registry_path, name, uuid, in_memory_registry) = new(path, registry_path, name, uuid, in_memory_registry, ReentrantLock() #= undef =#)
 end
 
 registry_info(pkg::PkgEntry) = init_package_info!(pkg)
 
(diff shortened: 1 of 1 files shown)


KristofferC:
I'm worried this will be too slow.

KristofferC:
I think there is still possibility of a race since even when `.pkgs` is defined we still have work left to do in `_ensure_registry_loaded_slow!`, or?

IanButterworth:
There was one `isdefined(r, :pkgs)` outside the lock.

Also I think we should do the same for PkgEntry, so pushed that.

KristofferC:
Can we time how long `deps_graph` takes before and after this change? Just to be sure we don't tank performance.

IanButterworth:
For
```
% julia +nightly --project -ie 'using Pkg, REPL; Pkg.activate(temp=true); Pkg.add("DifferentialEquations")'
```
Timing
`(graph, compat_map) = deps_graph(env, registries, names, reqs, fixed, julia_version, installed_only)`

Master (3 separate runs)
```
0.430757 seconds (4.38 M allocations: 418.646 MiB, 37.38% gc time, 0.79% compilation time)
0.617611 seconds (8.08 M allocations: 593.018 MiB, 20.04% gc time, 5.59% compilation time)
0.657180 seconds (8.08 M allocations: 593.016 MiB, 18.94% gc time, 5.33% compilation time)
```

PR
```
0.680969 seconds (8.08 M allocations: 592.888 MiB, 20.19% gc time, 4.98% compilation time)
0.632186 seconds (8.08 M allocations: 592.887 MiB, 19.35% gc time, 5.14% compilation time)
0.644893 seconds (8.08 M allocations: 593.055 MiB, 20.05% gc time, 5.49% compilation time)
```

Seems about the same.

KristofferC:
Okay nice. Should be ok then

KristofferC:
I think this is only on nightly (original bug report was faulty with the version).

KristofferC:
1.12-rc1 accidentally used the nightly version of Pkg, with rc2 this issue does not exist so removing backport label.

Issue #4345, opened 2025-07-31
First depot in `Base.DEPOT_PATH` is not always respected in 1.12


JamesWrigley:
On 1.12 RC1 Pkg seems to use a lower-priority depot for some operations if a package is already installed there, without first checking to see if the depot is writable. Here's an MWE:
```julia-repl                                                                                                                                                                                                                         
# Make an empty depot
julia> mkdir("/tmp/bar")                                                                                                                                                                                                                       
"/tmp/bar"                                                                                                                                                                                                                                     

julia> empty!(Base.DEPOT_PATH)
String[]

julia> push!(Base.DEPOT_PATH, "/tmp/bar")
1-element Vector{String}:
 "/tmp/bar"

# Activate an empty environment and add Revise
(@v1.12) pkg> activate --temp
  Activating new project at `/tmp/jl_VZslN2`

(jl_VZslN2) pkg> add Revise

# Make the depot read-only
julia> run(`chmod -R -w /tmp/bar`)
Process(`chmod -R -w /tmp/bar`, ProcessExited(0))

# Make a new writable depot and make sure it's the first one
julia> mkdir("/tmp/foo")
"/tmp/foo"

julia> pushfirst!(Base.DEPOT_PATH, "/tmp/foo/")
2-element Vector{String}:
 "/tmp/foo/"
 "/tmp/bar"

# Adding a package fails
(jl_VZslN2) pkg> add Revise
   Resolving package versions...
ERROR: IOError: open("/tmp/bar/packages/Requires/1eCOK.pid", 194, 292): permission denied (EACCES)
Stacktrace:
  [1] _uv_error(prefix::String, c::Int32)
    @ Base ./libuv.jl:114
  [2] uv_error
    @ ./libuv.jl:113 [inlined]
  [3] open(path::String, flags::UInt16, mode::UInt16)
    @ Base.Filesystem ./filesystem.jl:190
  [4] tryopen_exclusive(path::String, mode::UInt16)
    @ FileWatching.Pidfile ~/.julia/juliaup/julia-1.12.0-rc1+0.x64.linux.gnu/share/julia/stdlib/v1.12/FileWatching/src/pidfile.jl:215
  [5] open_exclusive(path::String; mode::UInt16, poll_interval::Int64, wait::Bool, stale_age::Int64, refresh::Float64)
[...]

PR #4346, opened 2025-08-01
Make sure source pidfile is in a writable location

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4345

@JamesWrigley would you mind testing this i.e. from this branch `julia +nightly --project -e 'using Pkg, REPL'`

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1210,8 +1210,19 @@ function download_source(ctx::Context, pkgs; readonly = true)
         tracking_registered_version(pkg, ctx.julia_version) || continue
         path = source_path(ctx.env.manifest_file, pkg, ctx.julia_version)
         path === nothing && continue
-        mkpath(dirname(path)) # the `packages/Package` dir needs to exist for the pidfile to be created
-        FileWatching.mkpidlock(() -> ispath(path), path * ".pid", stale_age = pidfile_stale_age) && continue
+        if ispath(path) && iswritable(path)
+            pidfile = path * ".pid"
+        else
+            # If the path is not writable, we cannot create a pidfile there so use one in the first depot.
+            # (pidlocking probably isn't needed as in this case the package source logically is alredy installed
+            # in the readonly depot, but keep the pidfile logic for consistency)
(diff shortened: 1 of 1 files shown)


JamesWrigley:
Success, the MWE runs now :tada:

src/Operations.jl
```diff
@@ -1210,8 +1210,19 @@ function download_source(ctx::Context, pkgs; readonly = true)
         tracking_registered_version(pkg, ctx.julia_version) || continue
         path = source_path(ctx.env.manifest_file, pkg, ctx.julia_version)
         path === nothing && continue
-        mkpath(dirname(path)) # the `packages/Package` dir needs to exist for the pidfile to be created
-        FileWatching.mkpidlock(() -> ispath(path), path * ".pid", stale_age = pidfile_stale_age) && continue
+        if ispath(path) && iswritable(path)
```
KristofferC, line 1179: Why don't we just do a `ispath(path) && continue`?

IanButterworth: I think because it's a read only dir you're right, but the dir existing doesn't usually mean the contents exist. I erred on the side of caution, hence the comment.

KristofferC: What other check do we do for that the content exists?

KristofferC: It used to be like that:

https://github.com/IanButterworth/Pkg.jl/blob/2779d42f4991d09733bf52c9ca47478d1bef3a8d/src/Operations.jl#L1049.

So, why take the pid lock if we won't do anything?

IanButterworth: The path existing doesn't mean its contents are fully installed. So we pidlock checking the existence of that dir and installation into it.

IanButterworth: If it's not clear, this is about making installation work when multiple processes are doing the same thing.

KristofferC: Okay, that makes sense to me now.

KristofferC:
The mkpidlock code does not exist on the Pkg 1.12 branch (and 1.12-rc1 used Pkg nightly) so this does not need to be backported.

PR #4347, opened 2025-08-01
Base.parse CacheFlags from Julia process instead of reading UInt8

xal-0:
This is intended to support a widening of a CacheFlags from 8 -> 64 bits, while allowing us to change the representation of CacheFlags to be variable-length in the future.


Keno:
In general not a huge fan of `eval'ing` random things, but probably safe enough here. That said, if there was a stringification of this with a safe print/parse, that might make me even happier.

xal-0:
I went with the eval route so Pkg would work with the current version of Julia. How would I go about adding a safe parser for CacheFlags? Make a separate PR to Julia adding it first, then modify this PR to use it?

Keno (approved):

src/Operations.jl
```diff
@@ -2597,7 +2598,7 @@ function test(
             flags = gen_subprocess_flags(source_path; coverage, julia_args)
 
             if should_autoprecompile()
-                cacheflags = Base.CacheFlags(parse(UInt8, read(`$(Base.julia_cmd()) $(flags) --eval 'show(ccall(:jl_cache_flags, UInt8, ()))'`, String)))
+                cacheflags = parse(CacheFlags, read(`$(Base.julia_cmd()) $(flags) --eval 'show(Base.CacheFlags())'`, String))
```
vchuravy, line 2601: Should this use `repr`?

PR #4348, opened 2025-08-04
Nicer no changes formatting

IanButterworth:
@ericphanson what do you think? The length of the current one is a bit against the Pkg (cargo) prefix styling

Shorter prefix and blue, as green = action, blue = info ?

Currently
Screenshot 2025-08-04 at 4 11 32 PM

This PR
Screenshot 2025-08-04 at 4 11 28 PM

Re. 
https://github.com/JuliaLang/Pkg.jl/issues/3066
https://github.com/JuliaLang/Pkg.jl/pull/3806

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -2935,7 +2935,11 @@ function print_status(
     end
     no_changes = all(p -> p[2] == p[3], xs)
     if no_changes
-        printpkgstyle(io, Symbol("No packages added to or removed from"), "$(pathrepr(manifest ? env.manifest_file : env.project_file))", ignore_indent)
+        if manifest
+            printpkgstyle(io, :Manifest, "No packages added to or removed from $(pathrepr(env.manifest_file))", ignore_indent; color = Base.info_color())
+        else
+            printpkgstyle(io, :Project, "No packages added to or removed from $(pathrepr(env.project_file))", ignore_indent; color = Base.info_color())
+        end
     else
         xs = !filter ? xs : eltype(xs)[(id, old, new) for (id, old, new) in xs if (id in uuids || something(new, old).name in names)]
         if isempty(xs)


ericphanson:
hm, I agree it's a bit too long, though this looks a bit weird to my eye. ChatGPT suggested "No package additions or removals in Project.toml" but that's only 1 character shorter than "No packages added to or removed from Project.toml". It also suggested "Package set unchanged" but that seems a bit abstract to me, though maybe OK.

I forget what exactly _can_ change. I know the original issue was dev'd dependencies's metadata updates and can fix issues in the manifest. Could we say "Package versions unchanged"? that's a bit shorter

src/Operations.jl
```diff
@@ -2958,7 +2958,11 @@ function print_status(
     end
     no_changes = all(p -> p[2] == p[3], xs)
     if no_changes
-        printpkgstyle(io, Symbol("No packages added to or removed from"), "$(pathrepr(manifest ? env.manifest_file : env.project_file))", ignore_indent)
+        if manifest
+            printpkgstyle(io, :Manifest, "No packages added to or removed from $(pathrepr(env.manifest_file))", ignore_indent; color = Base.info_color())
+        else
+            printpkgstyle(io, :Project, "No packages added to or removed from $(pathrepr(env.project_file))", ignore_indent; color = Base.info_color())
```
IanButterworth, line 2941:
```suggestion
            printpkgstyle(io, :Manifest, "Package versions unchanged $(pathrepr(env.manifest_file))", ignore_indent; color = Base.info_color())
        else
            printpkgstyle(io, :Project, "Package versions unchanged $(pathrepr(env.project_file))", ignore_indent; color = Base.info_color())
```

IanButterworth: @ericphanson  I think we want the prefix to be one word (or two at most) and align with the longest prefix, which is currently "Precompiling"

ericphanson: this seems good as long as it's true aha

IanButterworth: It's actually not true.. I wasn't thinking straight. `Package set unchanged` is correct but does sound a bit technical.

I think the current state of the PR might be ok. It's visually aligned, clear and only slightly longer than before. My main gripe was the alignment of the colored prefix, which this solves.

ericphanson:
> hm, I agree it's a bit too long, though this looks a bit weird to my eye

Sorry, I mixed up which was PR and which was current. Yours looks much better of course :). Somehow I thought you were introducing the colored prefix instead of removing it.

PR #4335, opened 2025-07-25
Error if the stdlib version requested mismatches current version

IanButterworth:
Fixes #4334 

```
(jl_TIjQ4O) pkg> add Pkg@1.14
ERROR: Cannot add stdlib `Pkg` with version specification `1.14`.
The current Julia version 1.13.0-DEV.897 uses `Pkg` v1.13.0.

(jl_TIjQ4O) pkg> add Pkg@1.12
ERROR: Cannot add stdlib `Pkg` with version specification `1.12`.
The current Julia version 1.13.0-DEV.897 uses `Pkg` v1.13.0.
```

I think this is also the only time we actually test that upgradable stdlibs are upgradable (downgradable)?

Issue #4352, opened 2025-08-06
Cannot add stdlib message completely useless


vtjnash:
```
$ ./julia --project=~/.julia/dev/ClaudeBox -q
(ClaudeBox) pkg> up
    Updating registry at `~/.julia/registries/General.toml`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl`
    Updating git-repo `https://github.com/staticfloat/Glibc_jll.jl`
    Updating git-repo `https://github.com/staticfloat/GCC_crt_objects_jll.jl`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl.git`
    Updating git-repo `https://github.com/staticfloat/LLVMLibcxx_jll.jl`
    Updating git-repo `https://github.com/JuliaContainerization/Sandbox.jl`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl`
    Updating git-repo `https://github.com/staticfloat/macOSSDK_jll.jl`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl`
    Updating git-repo `https://github.com/staticfloat/libtapi_jll.jl`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl`
    Updating git-repo `https://github.com/staticfloat/Binutils_jll.jl`
    Updating git-repo `https://github.com/staticfloat/LinuxKernelHeaders_jll.jl`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl`
    Updating git-repo `https://github.com/staticfloat/CCTools_jll.jl`
    Updating git-repo `https://github.com/JuliaPackaging/BinaryBuilder2.jl`
    Updating git-repo `https://github.com/staticfloat/libstdcxx_jll.jl`
    Updating git-repo `https://github.com/staticfloat/LLVMLibunwind_jll.jl`
    Updating git-repo `https://github.com/staticfloat/LLVMCompilerRT_jll.jl`
    Updating git-repo `https://github.com/staticfloat/libLLVM_jll.jl`
    Updating git-repo `https://github.com/JuliaPackaging/OutputCollectors.jl`
    Updating git-repo `https://github.com/staticfloat/Mingw_jll.jl`
[...]

PR #4355, opened 2025-08-06
Revert "Error if the stdlib version requested mismatches current version"

KristofferC:
Reverts JuliaLang/Pkg.jl#4335

One should be able to add stdlibs at other versions than julia bundles.


IanButterworth:
> One should be able to add stdlibs at other versions than julia bundles.

Only if it's an upgradable stdlib, I believe. 

At least, try it out on previous versions. It silently installs a version other than the one you request.

KristofferC:
> Only if it's an upgradable stdlib, I believe.

Why? We load devved Pkg all the time, it is not upgradable. 

>  It silently installs a version other than the one you request.

Maybe, but that feels like a different bug then? Why shouldn't it be possible? For packages in sysimage I agree but we already have a check and error for that.

IanButterworth:
Dev is a different thing. This is Pkg.add with a name and version spec. 

If all stdlibs were upgradable we wouldn't need upgradable stdlibs?

KristofferC:
> This is Pkg.add with a name and version spec.

Not only (as seen in https://github.com/JuliaLang/Pkg.jl/issues/4352), it is also for `Pkg.add` with a URL (which should in theory work fine)?

Upgradable stdlibs is for when the version is "floating" (it is not devved, nor added by URL).

IanButterworth:
I don't really follow. We don't register stdlibs in General unless they're upgradable. How is adding them with a name and version spec other than that which they ship with supposed to work?

IanButterworth:
I'm happy to revert this and figure out the right thing to do with more time.

IanButterworth:
I mentioned on slack but I think the fix for this new check is to throw a ResolverError so that the tiered resolver can swallow the error and try more permissive update strategies.

KristofferC:
I don't think `up` uses a tiered approach by default and I think this also already throws a ResolverError.

KristofferC:
> I don't really follow. We don't register stdlibs in General unless they're upgradable. How is adding them with a name and version spec

My point in https://github.com/JuliaLang/Pkg.jl/pull/4355#issuecomment-3161754626 was that this code path fires even for stdlibs devved and added by URL. It doesn't only check the VersionSpec case but also the VersionNumber case which applies to dev/url. 

https://github.com/JuliaLang/Pkg.jl/blob/3cfe013fe8392b710edbdcde19802eef2fe31556/src/Operations.jl#L541

KristofferC:
To be explicit:

```diff
diff --git a/Project.toml b/Project.toml
index cdc10d120..729bd92b1 100644
--- a/Project.toml
+++ b/Project.toml
@@ -3,7 +3,7 @@ uuid = "44cfe95a-1eb2-52ea-b672-e2afdf69b78f"
 keywords = ["package management"]
 license = "MIT"
 desc = "The next-generation Julia package manager."
-version = "1.13.0"
+version = "1.14.0"
```

```
(@v1.13) pkg> dev ./Pkg.jl
   Resolving package versions...
ERROR: Cannot add stdlib `Pkg` with version specification `1.14.0`.
The current Julia version v1.13.0-DEV.968 uses `Pkg` v1.13.0.
```

PR #4363, opened 2025-08-13
Mention that standard libraries have a version number since 1.11

felixcremer:
Since julia 1.11 standard libraries have a version number and that should be reflected in the docs or this sentence should be removed.

### DIFF
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -58,7 +58,7 @@ Status `~/.julia/environments/v1.10/Manifest.toml`
   [4ec0a83e] Unicode
 ```
 
-Since standard libraries (e.g. ` Dates`) are shipped with Julia, they do not have a version.
+Since julia 1.11 standard libraries (e.g. ` Dates`) have a version.
 
 To specify that you want a particular version (or set of versions) of a package, use the `compat` command. For example,
 to require any patch release of the v0.21 series of JSON after v0.21.4, call `compat JSON 0.21.4`:


docs/src/managing-packages.md
```diff
@@ -58,7 +58,7 @@ Status `~/.julia/environments/v1.10/Manifest.toml`
   [4ec0a83e] Unicode
 ```
 
-Since standard libraries (e.g. ` Dates`) are shipped with Julia, they do not have a version.
+Since julia 1.11 standard libraries (e.g. ` Dates`) have a version.
```
felixcremer:
```suggestion
Before 1.11 standard libraries (e.g. ` Dates`) did not have dedicated version numbers.
```

felixcremer:
I updated the text with your suggestion.

felixcremer:
The output of all pkg modes in this page should be updated then.

IanButterworth:
If this specific example is updated I think we can merge this. Without that it's a bit odd.

Generally we need to update all Pkg output across the docs. (Or set up something automatic) That's a much bigger job.

felixcremer:
I updated the add and st example to 1.13 and I got confused by the following compat and installation of a certain version examples. If you go through the whole markdown page from top to bottom you get totally different results. this is because we first set the compat version to 0.21.4 and then try to install 0.21.1 and this just doesn't work. I am wondering whether we should be switching these two sections to make it clearer and to be able to go through the whole file and get the same results.

docs/src/managing-packages.md
```diff
@@ -40,25 +41,31 @@ It is possible to add multiple packages in one command as `pkg> add A B C`.
...
+  [a63ad114] Mmap v1.11.0
+  [de0858da] Printf v1.11.0
+  [9a3f8284] Random v1.11.0
+  [ea8e919c] SHA v0.7.0
+  [fa267f1f] TOML v1.0.3
+  [cf7118a7] UUIDs v1.11.0
+  [4ec0a83e] Unicode v1.11.0
+```
+
+Before 1.11 standard libraries (e.g. ` Dates`) did not have dedicated version numbers.
```
IanButterworth:
```suggestion
Note that before 1.11 standard libraries (e.g. ` Dates`) did not have dedicated version numbers.
```

I think it's helpful from a language perspective, but doesn't need to be as eyecatching as a `!!! note`

Issue #4232, opened 2025-05-08
Make PDF doc available


ickc:
The [Julia doc](https://docs.julialang.org/en/v1/) makes the PDF available, but the [pkgdocs](https://pkgdocs.julialang.org/v1/) does not. There's a pdf target already in the project, however. Could you build it and deploy to the gh-pages too?

ickc:
Actually, pdf are already built and uploaded in https://github.com/JuliaLang/Pkg.jl/blob/gh-pages-pdf/v1.11.6/Pkg.jl.pdf
I guess the only remaining issue is that it is not linked from the documentation site?

PR #4375, opened 2025-08-22
Add pdf link to the docs

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4232

Made using Claude


IanButterworth:
I added the built docs to the docs job as an artifact, and this seems to be working.

IanButterworth:
Screenshot 2025-08-22 at 10 26 32 PM

PR #4378, opened 2025-08-27
Add support for reporting an upstream version

IanButterworth:
i.e. for JLLs whose package version is most likely not the same as the upstream version.

Currently we don't record the upstream version reliably from BB, but it's proving painful for advisory tracking (cc. @mbauman)

Here ImageMagick_jll has been modified to have a made up `upstream_version` field in its Project.toml here

Screenshot 2025-08-27 at 1 22 02 PM

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,10 @@
 Pkg v1.13 Release Notes
 =======================
 
+- Added support for an optional `upstream_version` field in Project.toml and Manifest.toml files to track the version
+  of upstream software that a package wraps or is based on. This is particularly useful for wrapper packages like JLLs
+  where the package version might differ from the upstream software version. The upstream version is displayed in
+  `Pkg.status` output alongside the package version when present.
 - Project.toml environments now support a `readonly` field to mark environments as read-only, preventing modifications.
   ([#4284])
 - `Pkg.build` now supports an `allow_reresolve` keyword argument to control whether the build process can re-resolve
--- a/docs/src/toml-files.md
+++ b/docs/src/toml-files.md
@@ -102,6 +102,24 @@ Note that Pkg.jl deviates from the SemVer specification when it comes to version
 the section on [pre-1.0 behavior](@ref compat-pre-1.0) for more details.
 
 
+### The `upstream_version` field
+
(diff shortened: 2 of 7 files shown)


jeremiahpslewis:
Would this be a good place / time to also add a canonical name field for easier matching with upstream?

IanButterworth:
Yeah and that is within the discussion in https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/639

mbauman:
Pretty much everything I thought I knew about JLLs was wrong.  This is wildly challenging.

* There are JLLs that redistribute _many_ upstream components, all bundled together (and not split out into separate dependent JLLs). For example, https://github.com/JuliaBinaryWrappers/oneAPI_Support_jll.jl.
* There are JLLs that combine multiple forks for different platforms at potentially different versions. For example, [OpenSSH_jll@v10.0.1+0](https://github.com/JuliaBinaryWrappers/OpenSSH_jll.jl/tree/OpenSSH-v10.0.1%2B0) ships OpenBSD/ssh version 10.0p1 *and* PowerShell/Win32-OpenSSH version 9.5.0.0p1 (for both 32- and 64-bit, pay no mind to the Win32 there).
* There are JLLs that ship different versions from the same upstream project.  For example, [Git_jll@v2.51.0+0](https://github.com/JuliaBinaryWrappers/Git_jll.jl/tree/Git-v2.51.0%2B0) ships `git-for-windows/git` version 2.51.0 for 64-bit and 2.48.1 for 32-bit (this is in addition to shipping `git/git` for other platforms).
* There are JLLs that were not built by BinaryBuilder, but instead have manually crafted Artifacts that may point _somewhere other than_ a GitHub release asset. For example, Gurobi_jll provides its [Artifact binaries](https://github.com/jump-dev/Gurobi_jll.jl/blob/master/Artifacts.toml) directly from anaconda's distribution.

DilumAluthge:
I'm assuming @giordano has already seen this discussion, but pinging him just in case

DilumAluthge:
Instead of a single version number, could we store a dict with more detailed info?

giordano:
While adding stuff to the registry can be useful, I feel like for what Matt wants it'd be better to dump all the info we want into a file in the JLL repo, so that we don't need to care about cluttering the registry.  Someone needs to champion designing the data we need and the format.  I mentioned in https://github.com/JuliaPackaging/BinaryBuilder.jl/issues/639#issuecomment-974914252 that some information (perhaps all?) has to be platform-specific, as Matt realised the hard way today.

Issue #4376, opened 2025-08-23
Adding multithreading option for Apps


Iddingsite:
Hey!

I have been having fun so far playing with the new App interface. Really nice addition! I encountered an issue though, while wanting to launch my App with multithreading, which I believe is not an uncommon need.

To my understanding, there is currently no easy way to just write:

```shell
MyApp -t auto
```

or other flags to tell Julia how to start.

Currently, my workaround looks like that:

```julia
function (@main)(ARGS)

    if length(ARGS) > 0

        # Check for --threads or -t flag
        i = 1
        while i <= length(ARGS)
            if (ARGS[i] == "--threads" || ARGS[i] == "-t") && i < length(ARGS)
                nstr = ARGS[i+1]
                if nstr == "auto"
                    n = Sys.CPU_THREADS
                else
                    n = parse(Int, nstr)
                end
                if Threads.nthreads() != n
                    println("Restarting with $n threads...")
                    # Remove the thread flag and its value from ARGS for restart
                    new_args = copy(ARGS)
                    splice!(new_args, i:i+1)
                    cmd = `$(Base.julia_cmd()) -t $n -m MAGEMinApp $(new_args...)`
                    run(cmd)
                    return 0
                end
            end
            i += 1
        end

    return 0
end
```

Which works with only the App installed, but has the downside of requiring to launch the App twice.

I think it would be a nice addition to have something cleaner!

KristofferC:
There are two things that could be done:

- Allow apps to set default options which are used when the app is launched (for example `-tauto`).
- Allow setting custom options when lanching the app, PackageCompiler apps do this with the `--julia-args` option (args after that are passed as julia command line arguments).

Iddingsite:
I see. From a user perspective, what I was looking for was option 1. Because that is what I am familiar with, how to start Julia with the flags I want. But that is my opinion!

PR #4381, opened 2025-09-03
allow apps to customize default julia flags and also enable passing in custom julia flags to any app

KristofferC:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4376. cc @Iddingsite.


Iddingsite:
That looks awesome, thx a lot for the fast PR!

PR #4392, opened 2025-09-23
Adapt delayed delete mechanism

Liozou:
This is the companion PR to https://github.com/JuliaLang/julia/pull/59635, and should be merged only if the julia PR is.

During `Pkg.gc()`, instead of deleting the files stored in the `delayed_delete_dir()`, the proposed mechanism consists in deleting the files listed in the `delayed_delete_list()`.


KristofferC:
This is in phase with what ended up being merged, right?

Liozou:
Yes, this looks fine. I don't understand the CI crash though.

KristofferC:
```
FFMPEG.exe("-f", "lavfi", "-i", "testsrc=duration=1:size=128x128:rate=10", "-f", "null", "-"): 0.314764 seconds (4.69 k allocations: 209.887 KiB)
Assertion failed: rcbuf->ref > 0, file nghttp2_rcbuf.c, line 88
```

not sure what is up with that.. I'll try rerun it.

Issue #4372, opened 2025-08-21
REPL pkg mode error on tab completion


felixcremer:
I got the following error on the first  tab completion after opening julia.
This only happened on the first tab completion after updating the julia nightly version.

```julia
(@v1.13) pkg> add JS┌ Error: Error in the keymap
│   exception =
│    AssertionError: !(isdefined(vinfo, :uncompressed_compat))
│    Stacktrace:
│      [1] initialize_uncompressed!(pkg::Pkg.Registry.PkgInfo, versions::Base.KeySet{VersionNumber, Dict{VersionNumber, Pkg.Registry.VersionInfo}})
│        @ Pkg.Registry ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/src/Registry/registry_instance.jl:126
│      [2] initialize_uncompressed!
│        @ ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/src/Registry/registry_instance.jl:108 [inlined]
│      [3] compat_info
│        @ ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/src/Registry/registry_instance.jl:158 [inlined]
│      [4] complete_remote_package!(comps::Vector{String}, partial::String; hint::Bool)
│        @ REPLExt ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/ext/REPLExt/completions.jl:67
│      [5] complete_remote_package!
│        @ ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/ext/REPLExt/completions.jl:57 [inlined]
│      [6] complete_add_dev(options::Dict{Symbol, Any}, partial::String, i1::Int64, i2::Int64; hint::Bool)
│        @ REPLExt ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/ext/REPLExt/completions.jl:157
│      [7] complete_add_dev
│        @ ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/ext/REPLExt/completions.jl:152 [inlined]
│      [8] #42
│        @ ~/.julia/juliaup/julia-nightly/share/julia/stdlib/v1.13/Pkg/ext/REPLExt/completions.jl:232 [inlined]
[...]

PR #4394, opened 2025-09-24
add a lock when mutating `PkgInfo`

KristofferC:
since we cache the registries, multiple threads can mutate the same pkginfo simultaneously

Should fix https://github.com/JuliaLang/Pkg.jl/issues/4372.

### DIFF
--- a/src/Registry/registry_instance.jl
+++ b/src/Registry/registry_instance.jl
@@ -62,6 +62,8 @@ struct PkgInfo
 
     # WeakDeps.toml
     weak_deps::Dict{VersionRange, Dict{String, UUID}}
+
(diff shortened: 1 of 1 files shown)


IanButterworth:
I could've sworn I did this already...

IanButterworth:
Oh, I guess https://github.com/JuliaLang/Pkg.jl/pull/4341 was incomplete

IanButterworth:
LGTM my only thought is that the other PR put the lock inside the init function. We should probably follow the same pattern for both. This way might be cleaner?

KristofferC:
I am not sure what concrete change that would be? Where would the lock be put? Over what region?

PR #4396, opened 2025-09-25
fix so that the `workspace` option to `instantiate` actually works

### LINKED ISSUE #3857: `instantiate` always instantiates whole workspace
https://github.com/JuliaLang/Pkg.jl/blob/162634c5615d12b889e4b64f3cff95d1c377f189/src/API.jl#L1274-L1276

will download all packages + artfacts from packages in the manifest (which will be the whole workspace), not limited to the currently active project.

KristofferC:
fixes https://github.com/JuliaLang/Pkg.jl/issues/3857

test written with Claude


visr:
Does this need a 1.12 and 1.13 backport label?

KristofferC:
Yes, thanks

giordano:
This hasn't been backported to v1.12 yet because it needs manual handling (#4615)?

PR #4399, opened 2025-09-28
Fix loop with large number of CPUs in atomicity test

green-br:
Add cpu number to temporary name inside unique temp dir for atomicity test otherwise it fails on large cpu counts found on Arm.  tempname was likely producing name clashes across CPUs - is it guaranteed to be unique? e.g. thread-safe?  mktempdir made it clear the directory it was counting the files in.


martinholters:
Should `touch(tempname(...))` rather be `mktemp(...)`? That should be safe against name clashes without further shenanigans, no?

PR #4411, opened 2025-10-12
Add `--dry-run` for add/update etc.

IanButterworth:
We previously had `preview` mode, which was [removed](https://github.com/JuliaLang/Pkg.jl/pull/1367) and replaced with a more robust `undo`/`redo`, but I think there's still benefit of having something like `preview` but instead utilize the improved `undo` mechanics automatically at the end of the operation.

i.e. if `undo` is robust, then this should be.

I also thought `--dry-run` was more typical, given that's what cargo has.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Pkg v1.13 Release Notes
 =======================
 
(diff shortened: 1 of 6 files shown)


IanButterworth:
This needs a bit more thought because undo states are only set if something changes, so it could currently undo too far back.

Issue #4390, opened 2025-09-18
`st` should show if package is precompiled yet or not


adienes:
basically title. might be useful to know if `using Foo` will have to launch a precompilation job or not

PR #4415, opened 2025-10-13
Add precompile indicator to status

IanButterworth:
Closes https://github.com/JuliaLang/Pkg.jl/issues/4390

It seems like `Base.isprecompiled` isn't reliable though..

Screenshot 2025-10-12 at 10 10 18 PM


DilumAluthge:
Maybe it's just not reliable for stdlibs in the sysimage?

DilumAluthge:
Anyway I like the overall idea.

Is Preferences underlined intentionally?

KristofferC:
Can this be benchmarked on a somewhat large environment. Especially on Windows? The is also a cache that can be used that the precompile code uses?

KristofferC:
https://github.com/JuliaLang/julia/pull/59670 caused us not to be able to pass a cache to `Base.isprecompiled` anymore.

Issue #4417, opened 2025-10-13
[REPL] Pkg mode autocompletion of action on single package is broken


giordano:
In the REPL, enter Pkg mode with `]`, then

```
activate --temp
add Example
rm <TAB>
```

Normally this would automatically complete the single package present in the environment to

```
rm Example
```

but in julia nightly this autocompletes to

```
Examplerm
```

which is quite broken.

```julia-repl
julia> versioninfo()
Julia Version 1.13.0-DEV.1297
Commit d3f2952a6e* (2025-10-12 15:48 UTC)
Platform Info:
  OS: macOS (arm64-apple-darwin24.5.0)
  CPU: 8 × Apple M1
  WORD_SIZE: 64
  LLVM: libLLVM-20.1.8 (ORCJIT, apple-m1)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 4 virtual cores)
```

KristofferC:
Introduced in 566ee44e609fd09382f5a. I am guessing by https://github.com/JuliaLang/Pkg.jl/commit/844cb9cab4006d6fc67494032ea0d79dca7eff44. cc @IanButterworth

PR #4373, opened 2025-08-22
Fix completion at the end of a complete line

IanButterworth:
Fixes https://github.com/JuliaLang/julia/issues/58690
Fixes https://github.com/JuliaLang/julia/issues/58103

Using Claude.

PR #4418, opened 2025-10-15
Fix completion on empty command

IanButterworth:
Fixes https://github.com/JuliaLang/Pkg.jl/issues/4417

Means https://github.com/JuliaLang/Pkg.jl/pull/4373 should get the `backport 1.12` label back

@giordano @KristofferC


giordano:
Any easy test to add?

IanButterworth:
Needs https://github.com/JuliaLang/Pkg.jl/pull/4436

PR #4425, opened 2025-10-18
Drop the REPL search keymap

tecosaur:
Pairs with https://github.com/JuliaLang/julia/pull/59819, which see.

### DIFF
--- a/ext/REPLExt/REPLExt.jl
+++ b/ext/REPLExt/REPLExt.jl
@@ -152,6 +152,9 @@ function create_mode(repl::REPL.AbstractREPL, main::LineEdit.Prompt)
     hp.mode_mapping[:pkg] = pkg_mode
     pkg_mode.hist = hp
 
+    skeymap = if !isdefined(REPL, :History)
+        last(LineEdit.setup_search_keymap(hp)) # TODO: Remove
+    end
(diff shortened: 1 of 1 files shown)


KristofferC:
Don't there need to be a keybind in the Pkg REPL to initiate the reverse search?

tecosaur:
That's now provided by `LineEdit.history_keymap`

KristofferC:
Ok! Could we put this within e.g. a `isdefined(REPL, :History)` check so that the keybind still works until eventually the new search history PR is merged?

tecosaur:
Sure

ext/REPLExt/REPLExt.jl
```diff
@@ -152,7 +152,9 @@ function create_mode(repl::REPL.AbstractREPL, main::LineEdit.Prompt)
     hp.mode_mapping[:pkg] = pkg_mode
     pkg_mode.hist = hp
 
-    search_prompt, skeymap = LineEdit.setup_search_keymap(hp)
+    skeymap = if isdefined(REPL, :History)
```
KristofferC: Should this be `if !isdefined`?

tecosaur: :facepalm: yes

Issue #3989, opened 2024-08-12
Incorrect resolution of local path dependencies included with `[source]` in a root project setup


JanisErdmanis:
There seems to be an oversight in the implementation of `[sources]` for the local path and its interaction with a workflow that uses a root project. In particular, I have a temporary dependency put within the existing package `ShuffleProofs`, and I have the following `[sources]` section:
```
[sources]
SigmaProofs = {path = "SigmaProofs"}        
```
When activating the project at `ShuffleProofs,` the dependency is correctly located, indicating that the setup is done accurately. 

I use the root project setup extensively and have the following directory structure:
```
├── AppBundler
├── CryptoGroups
├── CryptoPRG
├── CryptoSignatures
├── HistoryTrees
├── Manifest.toml
├── PeaceFounder
├── PeaceFounderAdmin
├── PeaceFounderClient
├── PeaceFounderCompiled
├── PeaceFounderDemo
├── Project.toml
├── ShuffleProofs
├── TallyProofs
└── [peacefounder.org](http://peacefounder.org/)
```
where every dependency is added with `dev` in the project. When I activate this root project and load `ShuffleProofs`, it tries to find `SigmaProofs` in the wrong location:
```
(Github) pkg> test ShuffleProofs
     Testing ShuffleProofs
ERROR: expected package `SigmaProofs [f8559b4c]` to exist at path `/Users/jerdmanis/BtSync/PeaceFounder/Github/SigmaProofs
```
where instead, it should be located at `/Users/jerdmanis/BtSync/PeaceFounder/Github/ShuffleProofs/SigmaProofs`.  For now, I have resolved the situation by making a soft link in the root project.

KristofferC:
Maybe dup of https://github.com/JuliaLang/Pkg.jl/issues/3842

Issue #3842, opened 2024-03-13
Bug in relative paths in `[sources]`


KristofferC:
IIUC, currently, all relative paths for packages (`pkg.path`) are considered relative to the manifest file. But in cases they are provided by a `[sources]` entry they should in fact be treated relative to the project file. This only matters if the project and manifest file are in different directories. That situation can occur when using the `manifest = ` entry in the project file or when using the workspace functionality in #3841.

This is a bit annoying because I think it means we need to remember from where the path entry was retrieved.

PR #4427, opened 2025-10-18
make `.path` field consistently be relative manifest and convert to project relative upon writing to a project file

KristofferC:
fixes https://github.com/JuliaLang/Pkg.jl/issues/3842
fixes https://github.com/JuliaLang/Pkg.jl/issues/3989


KristofferC:
Something is going wrong with packages with a `manifest` entry..

Issue #2194, opened 2020-11-07
Add ability for packages to be "unlisted" in registry


oxinabox:
Idea: 
We should add a field in the registry for to mark packages as "unlisted".
This can be applied to abandoned packages, or to packages that gave been deprecated  or renamed.

The effect of this woud be to make it so that `pkg> add ...` REPL does not tab-complete to that name.
Possibly some other things might like to use it, e.g. JuliaObserver, or JuliaHub might do something with it.

To go beyond this, perhaps we could have a free text field that the unlisting reason could be given.
Which would be displayed when people directly call `pkg> add`.
And that could give a warning message explaing e.g. that it was renamed and that other one should be installed instead,
or that it has been discontinued and a good alternative might be...
Such a message would not be displayed when it is added indirectly.

Naming things is hard, so perhaps this should be `listed = false` or `visiblity = supressed`,
or even `unlisted-message = "Reason this is unlisted"` and if not present then it is listed.
 or something.

StefanKarpinski:
I like `listed = false` but we could also have something like `status = "deprecated"`.

KristofferC:
Maybe `deprecated = true`.

oxinabox:
Another use of this is that registry CI should ignoring unlisted entries for purposes of checking for too similar names.

Maybe even if two packages are registered with the same name but one is unlisted it should not present the menu to chose them

StefanKarpinski:
I think all of that falls under the concept of "deprecated". Although at that point, it feels like we may want to just split deprecated packages into a `Deprecated` registry. I believe that any client using the pkg protocol will install all the registries that the pkg server advertises, so we could introduce more registries pretty smoothly.

oxinabox:
Is it ok that packages in a Deprecated registry would depend on packages in the General registry *and* packages in the General registry would depend on packages in the Deprecated registry?
I think that works fine?

---
One problem i have with the word deprecated is to me it implies there is a replacement.
"X is deprecated in favour of Y".
But for some things they might be unlisted without replacement.
Like if everyone decided it was a terrible idea.
Or if something is super-buggy and the author admits that, but has no time or interest in fixing it.

KristofferC:
I don't think a deprecation necessarily implies an existing replacement (although that is common).

> Although at that point, it feels like we may want to just split deprecated packages into a Deprecated registry. 

Then it would have to be installed by default? Or otherwise you can't instantiate a manfiest with a deprecated package.

[...]

PR #4433, opened 2025-10-19
allow for a `metadata` table in a `Package.toml` in a registry and have a `deprecated` sub-table indicate the package is deprecated

KristofferC:
These packages will not tab complete to and will show up as deprecated in pkg status. The `[deprecation`] table can have arbitrary keys but Pkg itself does not use any of them right now.

Fixes https://github.com/JuliaLang/Pkg.jl/issues/2194

cc @oxinabox

### DIFF
--- a/test/registry.jl
+++ b/test/registry.jl
@@ -5,6 +5,8 @@ using Pkg, UUIDs, LibGit2, Test
 using Pkg: depots1
 using Pkg.REPLMode: pkgstr
 using Pkg.Types: PkgError, manifest_info, PackageSpec, EnvCache
+using Pkg.Operations: get_pkg_deprecation_info
+
 using Dates: Second
 
 using ..Utils
@@ -409,6 +411,18 @@ end
             example1_info = Pkg.Registry.registry_info(example1_entry)
             @test !Pkg.Registry.isdeprecated(example1_info)
             @test example1_info.deprecated === nothing
+
+            # Test get_pkg_deprecation_info function
+            deprecated_pkg_spec = Pkg.Types.PackageSpec(name = "DeprecatedExample", uuid = pkg_uuid)
+            normal_pkg_spec = Pkg.Types.PackageSpec(name = "Example1", uuid = example1_uuid)
+
+            dep_info = get_pkg_deprecation_info(deprecated_pkg_spec, registries)
+            @test dep_info !== nothing
+            @test dep_info["reason"] == "This package is no longer maintained"
+            @test dep_info["alternative"] == "Example"
+
+            normal_info = get_pkg_deprecation_info(normal_pkg_spec, registries)
+            @test normal_info === nothing
         end
     end
 
--- a/docs/src/registries.md
+++ b/docs/src/registries.md
@@ -114,7 +114,11 @@ One use of the `[metadata]` table is to mark packages as deprecated using `[meta
 - Be excluded from tab-completion suggestions
 - Still be installable and usable
 
-The `[metadata.deprecated]` table can contain arbitrary metadata fields. Common fields include:
+The `[metadata.deprecated]` table can contain arbitrary metadata fields. Two special fields are recognized by Pkg and displayed when using `pkg> status --deprecated`:
+- `reason`: A string explaining why the package is deprecated
+- `alternative`: A string suggesting a replacement package
+
+Example:
 
 ```toml
 name = "MyPackage"
@@ -126,7 +130,7 @@ reason = "This package is no longer maintained"
 alternative = "ReplacementPackage"
 ```
 
-The specific fields within `[metadata.deprecated]` are not currently used by Pkg itself, but are stored to allow other tools and registries to utilize this metadata.
+Other fields can be added to `[metadata.deprecated]` for use by registries or other tools.
 
 ### Registry Compat.toml
 
--- a/src/API.jl
+++ b/src/API.jl
@@ -1314,14 +1314,15 @@ end
 
 @deprecate status(mode::PackageMode) status(mode = mode)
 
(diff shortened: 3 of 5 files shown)


tecosaur:
I like this. One thought: is it worth having the optional ability to suggest a different package as a successor? I'm thinking this could be useful in cases like `LightGraphs` → `Graphs` (rename), `JSON2` → `JSON3` (breaking new version as new package), `PushVectors` → `Julia 1.11+` (functionality now available in Julia), etc.

KristofferC:
> The [deprecation] table can have arbitrary keys but Pkg itself does not use any of them right now.

So yes, there is a possibility to add something like a successor or a reason for deprecation or whatever. But for the initial PR I didn't want to spend too much time thinking about exactly what metadata should be included or how it should be presented.

But I guess `"reason"` and `"alternative"` could be "standard" keys to have.

tecosaur:
> But for the initial PR I didn't want to spend too much time thinking about exactly what metadata should be included

Entirely reasonable. Attaching some sort of brief message is the only thing I can think of beyond just having a list of packages. Whether it be now or some point in the future, I think just having one free-form field might suffice (predicated on the main way the message is used being shoving it in package developers' faces, I don't think automated tooling would need to know anything more than "package X is deprecated").

DilumAluthge:
I personally would prefer a more general solution for metadata (or "tags", as Eric has suggested) in the registry (cc: @ericphanson @mbauman), instead of just having syntax specific for package deprecation only.

KristofferC:
How is this not "general"? It is just a key in the TOML file. Other metadata would have other keys. Or am I missing something?

mbauman:
Yeah, but it's a good motivating example, because it's quite clear that the additional metadata beyond the <kbd>deprecated</kbd> "tag" would be quite valuable.  I kinda like aggregating all these sorts of "informational" metadata into single top-level key, but this means that it's something more complicated than the simple `tags = ["deprecated"]`.

ericphanson:
This is also per package, whereas some metadata should probably be per-version

KristofferC:
> Yeah, but it's a good motivating example, because it's quite clear that the additional metadata beyond the deprecated "tag" would be quite valuable. I kinda like aggregating all these sorts of "informational" metadata into single top-level key, but this means that it's something more complicated than the simple tags = ["deprecated"].

I am a bit confused but do I understand it right that if I call this table `metadata.deprecated`, it is good because you want it all to be under one big table?

Metadata for versions would presumably go into the `Versions.toml` file (where there already is "metadata" like `yanked`). Do you also want it to be `metadata.yanked` and have that itself be a table (instead of a key-value entry as it is now)?

mbauman:
I’m not necessarily against it, just trying to think through the relative benefits of this vs something different. It just feels like commingling (potentially) arbitrary tag names alongside semantically meaningful key names will cause us headaches in the future.

KristofferC:
I'm fine putting it within a `metadata` table. One reason for making the manifest v2 format was that we earlier had package names as primary keys, which made it hard to add extra stanzas, so there is arguably some precedence there.

mbauman:
Yeah, I think it'd seem prudent to add an extra level of nesting here.  Then there's a choice — this could either be an array of tables or a table of tables.  That is, either:

```toml
[[metadata]]
name = "deprecated"
reason = "This package is no longer maintained"
alternative = "Example"

[[metadata]]
name = "special"
```

Or:

```toml
[metadata.deprecated]
reason = "This package is no longer maintained"
alternative = "Example"

[metadata.special]
```

I think I slightly favor the table of tables.  And I think I may slightly prefer the name `tags` to `metadata`, but that's _**real**_ bikesheddy.

KristofferC:
> I think I slightly favor the table of tables

Me too, 100%

KristofferC:
Last commit makes that change

KristofferC:
added docs as well

DilumAluthge:
> I think I slightly favor the table of tables. And I think I may slightly prefer the name tags to metadata, but that's real bikesheddy.

I'm a little worried that "tags" might make people think of Git tags, but that might not be enough of a problem to justify using the longer "metadata".

KristofferC:
FWIW I also think metadata is more apt.

ericphanson:
I think metadata seems reasonable, that also leaves room if we want metadata.tags to be something more specific later

KristofferC:
image

DilumAluthge:
I like the way it suggests a different package to use.

Idea: Instead of just suggesting `ModernExample`, could we suggest a specific version range for `ModernExample`?

As a real-world example:

> JSON2 vX.Y.Z [deprecated] (Reason: This package is no longer maintained) (Alternative: JSON >= v1.0.0)

Because in this example, we wouldn't want users to inadvertently install an older version of JSON.jl.

This might be too much added complexity, though.

DilumAluthge:
Another idea: If `isinteractive()` is true, and if the user uses the REPL interface to do `] add DeprecatedPackage`, could we print a warning, or even prompt the user for confirmation?

KristofferC:
>  even prompt the user for confirmation?

A prompt has the risk to become annoying, and we already print the `[deprecated]` in the status output that is shown on add. But a "bigger" warning could be printed on a direct pkg add. But maybe in a follow up PR.

test/registry.jl
```diff
@@ -41,8 +41,6 @@ function setup_test_registries(dir = pwd())
         )
         write(
             joinpath(regpath, "Example", "Deps.toml"), """
-            ["0.5"]
-            julia = "0.6-1.0"
```
mbauman, line 45: I don't understand this change?

KristofferC: It's just a completely bogus entry, the `Deps.toml` file don't have compat info like that. It has name -> uuid entries, e.g.:

https://github.com/JuliaRegistries/General/blob/master/A/AAindex/Deps.toml

I guess nothing read that file before.

Issue #4422, opened 2025-10-16
`Pkg.add` no longer works with SHAs from PRs


MilesCranmer:
I think the change in https://github.com/JuliaLang/Pkg.jl/pull/2330:

```diff
- const refspecs = ["+refs/*:refs/remotes/cache/*"]
+ const refspecs = ["+refs/heads/*:refs/remotes/cache/heads/*"]
```

Is causing the following issue in AirspeedVelocity.jl: https://github.com/MilesCranmer/AirspeedVelocity.jl/issues/113. I am still diagnosing this but I just wanted to post it to see if anybody had suggestions for a quick fix.

~~I guess the solution might be on my side, to change the URL, rather than rely on a cross-repository revision lookup?~~ Edit: Well I guess this is just two remotes, so git should handle it?

MilesCranmer:
Ugh. Actually it seems like requiring a different repository URL to compare PR with master is going to be a pain and a half to implement. Much easier before when it was just comparing two revisions like git. Is there any way I can manually specify the refspecs? Or any way to make Julia aware of both Git remotes? @KristofferC @IanButterworth

MilesCranmer:
In the PR I see that @DilumAluthge made this suggestion at one point -

> I think that in addition to `refs/heads/*`, you'll also need `refs/remotes/*` or something like that?

Do you think this would this fix it?

MilesCranmer:
Copying from #2330:

> @MilesCranmer: Will do that for now. Can the change be reverted? It seems it fundamentally alters the behavior of `Pkg.add`. I understand there's this speedup, but that seems to only be because it removes some functionality, not that it's more efficient.
> 
> I think it would be better to have a special keyword for this rather than changing the standard behavior, perhaps `heads=false` by default, and `heads=true` to enable this reduced refspec

> @KristofferC: Can you elaborate the exact situation that causes this to happen? It doesn't be wide spread so what is the unique thing about this situation?

It is seen when installing a specific revision from a PR, when that PR's commit is based in a fork of the repository. For example: there is a PR to my repository https://github.com/MilesCranmer/SymbolicRegression.jl/pull/515 from a fork. One of the commits in this PR is f154755ba9b06bd1bb6251b4172d015fbade5e22.

On Julia 1.12, it fails to get the commit:

```julia-repl
julia> ]add https://github.com/MilesCranmer/SymbolicRegression.jl#f154755ba9b06bd1bb6251b4172d015fbade5e22
[...]

PR #4441, opened 2025-10-20
fall back to fetching more refspects in case the restricted set doesnt find the rev

KristofferC:
Should fix https://github.com/JuliaLang/Pkg.jl/issues/4422 without being slow in general.


DilumAluthge:
This seems reasonable to me. Most people get the fast behavior, but in the worst case you still get the ref you want, just with potential slowness.

Issue #4361, opened 2025-08-12
pkg> add PackageInManifest should not resolve


fredrikekre:
When adding a package that is already in the manifest it would be preferable if Pkg just inserted the entry in the projects `[deps]` section. The benefits are that i) we don't have to resolve (and "risking" that other changes are made) and ii) we never need to re-precompile anything.

fredrikekre:
This would then behave similarly to `pkg> rm PackageInProject` which just removes the dependency.

Issue #3890, opened 2024-05-03
[workspace] Preserve top level dependencies better


fredrikekre:
When adding package that is already in the manifest because of a top level dependency of a different project (or as a transient dependency, but those almost always track the registry) we should try to preserve the version. For example:
```
$ tree .
.
├── Manifest.toml
├── Project.toml
└── test
    └── Project.toml

$ cat Project.toml
[deps]
Example = "7876af07-990d-54b4-ab0e-23690620f79a"

[workspace]
projects = ["test"]

$ julia -e 'using Pkg; Pkg.status()'
Status `/tmp/tmp.C9bJWtPVPR/Project.toml`
  [7876af07] Example v0.5.4 `https://github.com/JuliaLang/Example.jl.git#master`

$ julia --project=test -e 'using Pkg; Pkg.status()'
Status `/tmp/tmp.C9bJWtPVPR/test/Project.toml` (empty project)

$ julia --project=test -e 'using Pkg; Pkg.add("Example")'
   Resolving package versions...
    Updating `/tmp/tmp.C9bJWtPVPR/test/Project.toml`
  [7876af07] + Example v0.5.3
    Updating `/tmp/tmp.C9bJWtPVPR/Manifest.toml`
  [7876af07] + Example v0.5.3

$ julia -e 'using Pkg; Pkg.status()'
Status `/tmp/tmp.C9bJWtPVPR/Project.toml`
  [7876af07] Example v0.5.3
```

I think this would be fixed by pinning top level dependencies of all projects in the first stage of the tiered resolver.

fredrikekre:
`PRESERVE_ALL` should probably consider the full manifest (not just the subgraph for the project) then? And then perhaps there should be a `PRESERVE_WORKSPACE_DIRECT` before `PRESERVE_DIRECT` that preserve all top level deps in all workspace projects.

fredrikekre:
Yea, I thought that adding a package that was in the manifest already was more of a no-op like `rm` is.

Perhaps `add` of a package that is already in the manifest should just not resolve at all like `rm`? Wouldn't result in precompilation at least which would be pretty nice.

KristofferC:
Currently `add Package` is kind of used as a combined `free + add` in case the package is added by URL or devved. A bit convenient at times but perhaps not worth it.

KristofferC:
Yes, I don't think people would complain about that behavior.

PR #4445, opened 2025-10-21
avoid resolving on add when a compatible package exists in the environment

KristofferC:
fixes https://github.com/JuliaLang/Pkg.jl/issues/4361
fixes https://github.com/JuliaLang/Pkg.jl/issues/3890

There might be some package test that relies on this re-resolving. We'll see. Also, this is a very nice feature.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1903,6 +1903,49 @@ function _resolve(
     end
 end
 
+function can_skip_resolve_for_add(pkg::PackageSpec, entry::Union{PackageEntry, Nothing})
+    # Can't skip if package not in manifest
+    entry === nothing && return false
+
+    # Can't skip if pinned (needs special handling in resolution)
+    entry.pinned && return false
+
+    # Can't skip if tracking path or repo
+    (entry.path !== nothing || entry.repo.source !== nothing || pkg.repo.source !== nothing) && return false
+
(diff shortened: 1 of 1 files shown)


src/Operations.jl
```diff
@@ -1903,17 +1903,62 @@ function _resolve(
...
+        haskey(ctx.env.project.compat, pkg.name) && continue
+        v = ctx.env.manifest[pkg.uuid].version
+        v === nothing && continue
+        pkgversion = string(Base.thispatch(v))
+        set_compat(ctx.env.project, pkg.name, pkgversion)
+        push!(compat_names, pkg.name)
+    end
+    if !isempty(compat_names)
+        printpkgstyle(ctx.io, :Compat, "entries added for $(join(compat_names, ", "))")
+    end
```
IanButterworth, line 1945: I just noticed we used to unconditionally print this.. fastest bugfind and fix ever.

IanButterworth: Backport?

KristofferC: > I just noticed we used to unconditionally print this.

Yeah, I totally meant to fix that... 😇  😊 

> Backport?

Maybe, but I kind of want it to sit on master for a while before putting it into a patch release.

Issue #3845, opened 2024-03-14
Specifying registries needed for a project in the Project file


KristofferC:
From https://github.com/JuliaLang/Pkg.jl/pull/3783#issuecomment-1974794079

> Could it make sense to be able to specify a registry in the sources block? Maybe with a separate registries block to have the url. That way a Project/Manifest does not need to depend on the registries already being installed.

`[sources]` made it easier to resolve a manifest with unregistered dependencies without having to manually add these dependencies by URL. However, you still need to make sure that the registries are available. So the next step would be to store the set of registries needed in the project file as well so that they are automatically installed for you.

This issue is about how to specify that information in the project file.

The first question is if we need to store the UUID of the registry somewhere. It's kind of silly since there are so few registries. If we then only look at the registry name we could have:

```toml
[registries]
CustomRegistry = "https://my.custom.registry/CustomRegistry"
...
```

or maybe this should be a dict and be

```toml
[registries]
CustomRegistry = {url = "https://my.custom.registry/CustomRegistry"}
...
```

------

Should custom registries be auto-added to the project file once a package is added to the project that is not available in general?

---------
cc @ericphanson, @Seelengrab

Seelengrab:
I don't have much of an opinion on how exactly the data is stored, but IMO having the UUID doesn't really cost us all that much but is a gain UX wise. We should already check for uniqueness when doing `]registry add` right?

> Should custom registries be auto-added to the project file once a package is added to the project that is not available in general?

I think so, yeah. Without that information, the project is still not instantiable from just Project.toml alone. Making `General` less special in that regard seems fruitful to me, so that other installations can e.g. completely remove `General` and just have their (perhaps vetted) registry. In that sense, `General` should also be added to `[registries]` as soon as a package is added that's sourced from it.

vchuravy:
Copying what I just posted in Slack here:

> One thought with regards to [sources] I just came across a package that has dependencies that exist in a different registry. It would be nice if we had a section in the Project.toml that would prompt the user upon a Pkg.add(url) to add the corresponding registries.
> Maybe in [sources] we could even state the registry the package is coming from allowing for version resolution over just tracking a branch.

> Should custom registries be auto-added to the project file once a package is added to the project that is not available in general?

Hm, my gut went with asking before adding the registry, but maybe auto-adding is fine.

PR #4460, opened 2025-10-22
Track registries in the manifest for added packages 

KristofferC:
This allows manifests to be instantiated even if non-default manifests are required and are not installed

Fixes https://github.com/JuliaLang/Pkg.jl/issues/378 (but not https://github.com/JuliaLang/Pkg.jl/issues/3845).

Tests written initially by Claude Code and iterated on by me.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1723,27 +1723,6 @@ function handle_package_input!(pkg::PackageSpec)
     return pkg.uuid = pkg.uuid isa String ? UUID(pkg.uuid) : pkg.uuid
 end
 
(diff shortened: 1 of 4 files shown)


vchuravy:
This is great! Does this also handle the case when running instantiate on a fresh Julia installation and General is not yet installed? I have a memory of Julia complaining at me.

KristofferC:
> Does this also handle the case when running instantiate on a fresh Julia installation and General is not yet installed? I have a memory of Julia complaining at me.

Yes, but I think that should already be handled now

PR #4468, opened 2025-10-26
Make ]test use --depwarn=yes #4410 (fixed)

### LINKED ISSUE #4410: Make `]test` use `--depwarn=yes`

vishal2005025:
Title: Ensure `]test` uses `--depwarn=yes` by default (fixes #4410)

Summary
-------
This change makes `Pkg.test` run package tests with deprecation warnings enabled by default by ensuring the test subprocess receives `--depwarn=yes` unless the caller explicitly provides a `--depwarn` option.

Problem
-------
Previously the `--depwarn` flag in the Julia subprocess flags was not guaranteed to be the final occurrence on the command line, and user-supplied `julia_args` could silence or change deprecation-warnings behavior unexpectedly. Some callers (and tests) rely on `--depwarn=yes` for visibility of deprecations, while a few tests intentionally pass `--depwarn=no`. We need a default that enables deprecation warnings but still respects explicit user intent.

What I changed
-------------
- File modified: `src/Operations.jl`
- Function: `gen_subprocess_flags(source_path::String; coverage, julia_args::Cmd)`
- Behavior:
  - Inspect `julia_args` for an explicit `--depwarn` flag.
  - If no `--depwarn` is present, append `--depwarn=yes` to the generated flags so it becomes the final (and therefore effective) setting for the test subprocess.
  - If the caller provided `--depwarn=...`, that value is respected (no override).

Why this approach
-----------------
- Command-line flag precedence depends on the final occurrence. Appending `--depwarn=yes` ensures the default is applied.
- Respecting explicit `--depwarn` preserves existing test cases that intentionally disable deprecation warnings (e.g., some entries in `test/` that pass `julia_args = \`--depwarn=no\``).
- The change is small, local, and low-risk.

How to validate locally
-----------------------
Run these from the repo root (PowerShell):

- Confirm default behavior appends `--depwarn=yes`:
```powershell
julia --project -e 'using Pkg; println(Pkg.Operations.gen_subprocess_flags(pwd(); coverage=false, julia_args=``))'

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -2543,25 +2543,46 @@ function gen_subprocess_flags(source_path::String; coverage, julia_args::Cmd)
         has_depwarn = false
     end
 
-    # Render julia_args to a clean space-joined string to avoid interpolating
-    # the raw Cmd object which can produce extraneous empty tokens (e.g. "''").
-    julia_args_str = try
-        isempty(julia_args.exec) ? "" : join(string.(julia_args.exec), " ")
+    # Split any entries in julia_args.exec that may contain multiple flags
+    # (e.g. "--check-bounds=yes --compiled-modules=yes ...") into individual
+    # tokens so they are passed as separate argv entries to the subprocess.
+    tokens = String[]
+    try
+        for e in julia_args.exec
+            append!(tokens, split(String(e)))
+        end
     catch
-        string(julia_args)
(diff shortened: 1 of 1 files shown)


LilithHafner:
This is AI slop. Do not open another PR to a Julia repo using AI without disclosing that usage.

PR #4472, opened 2025-10-27
ask the package server for zstd compressed data

KristofferC:
Together with https://github.com/JuliaPackaging/PkgServer.jl/pull/220 this allows Pkg to download registries, packages and artifacts as zst archives which based on benchmarks decompress significantly faster and also tend to be smaller.

Having the registry stored as a zst archive is not a backwards compatibility issue because even on older Pkgs it can decompress these with 7z


KristofferC:
I think this is fine, we will see when the new PkgServer is deployed if we automagically start getting zstd registries (and artifacts + packages should also download as that but that is kind of all happening behind the scenes).

PR #4473, opened 2025-10-27
make an app interface to Pkg

KristofferC:
https://github.com/user-attachments/assets/82e899f6-4fb9-43c5-a89d-b0eada500107

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -49,3 +49,6 @@ Tar = "1.10"
 UUIDs = "1.11"
 julia = "1.12"
 p7zip_jll = "17.5"
+
+[apps]
+pkg = {}
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -1012,6 +1012,161 @@ function _auto_precompile(ctx::Types.Context, pkgs::Vector{PackageSpec} = Packag
     end
 end
 
+#############
+# CLI Entry #
+#############
+
+function print_help()
+    println("pkg - Julia package manager command-line interface\n")
+    println("Full documentation available at https://pkgdocs.julialang.org/\n")
+
+    printstyled("OPTIONS:\n", bold=true)
+    println("  --project=PATH    Set the project environment (default: current project)")
+    println("  --offline         Work in offline mode")
+    println("  --help            Show this help message")
+    println("  --version         Show Pkg version\n")
+
+    printstyled("SYNOPSIS:\n", bold=true)
+    println("  pkg [opts] cmd [args]\n")
+    println("Multiple commands can be given on the same line by interleaving a ; between")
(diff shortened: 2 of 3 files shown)


IanButterworth:
Neat. How does this work with using different julia versions?

KristofferC:
> How does this work with using different julia versions?

Well, as will all Pkg apps they are more or less locked to a specific julia version (the one you added it with) unless you set the env variable as described in https://pkgdocs.julialang.org/dev/apps/#Overriding-the-Julia-Executable. Then you could set it to e.g. `julia` and use juliaup to chose the default and all your apps will follow that. But then you have to recompile.

Ideally pkg could just simulate running at the same version as the manifest was resolved with, but for that we need HistoricalStdlibs etc.

stevengj:
An alternative would be to just have `julia -pkg <command...>`, which would make the `julia` version easy and explicit?

KristofferC:
With https://github.com/JuliaLang/Pkg.jl/pull/4495 we can resolve for other julia versions natively.

stevengj:
Still seems more painful than just doing `<myjulia> -pkg <command...>`

Pkg seems special enough that it is worth adding a new command-line flag to `julia`.

stevengj:
Note also that this conflicts with the [FreeBSD `pkg` command](https://docs.freebsd.org/en/books/handbook/ports/), the [Solaris `pkg`](https://docs.oracle.com/cd/E23824_01/html/E21802/gihhp.html) command, and perhaps others.

You could call it `julia-pkg`, of course, but in that case why not `julia -pkg`?

KristofferC:
> which would make the julia version easy and explicit?

You don't want the julia version explicit, you want to use the one that is in the manifest. So if you want `julia -pkg` you kind of need juliaup support then (https://github.com/JuliaLang/juliaup/pull/1315). Julia and Pkg are also different projects, I don't think coupling their CLI options together is a good idea.

stevengj:
> Julia and Pkg are also different projects, I don't think coupling their CLI options together is a good idea.

They are kind of tightly coupled in practice … it's not like we are talking about some random package here.

RomeoV:
Just noting that `pkg` is a very... general name for an executable. I have many package managers installed for various parts of my system. Perhaps `jlpkg` would work better? Or is there an easy way to install an app under another name?

EDIT: Oops seems that has been addressed above already.

visr:
On the topic of naming; how about `jl`?
I took the liberty of copying this PR into https://github.com/visr/Jl.jl, and added `init` (since activate doesn't write TOMLs), and `run` which makes it easy to run scripts by activating and instantiating automatically.

This uses ideas from [uv](https://docs.astral.sh/) and applies them to Julia. All the heavy lifting is already done by Pkg + this PR.

PR #4484, opened 2025-11-01
Fix warning about collect_delay during precompile

Keno:
However, I will note that there is still a reference in the docs that should be updated as well https://github.com/JuliaLang/Pkg.jl/blob/master/docs/src/managing-packages.md?plain=1#L522-L554.

### DIFF
--- a/src/REPLMode/command_declarations.jl
+++ b/src/REPLMode/command_declarations.jl
@@ -512,7 +512,7 @@ compound_declarations = [
             :name => "gc",
             :api => API.gc,
             :option_spec => [
-                PSA[:name => "all", :api => :collect_delay => Hour(0)],
+                PSA[:name => "all", :api => :collect_delay => nothing],
                 PSA[:name => "verbose", :short_name => "v", :api => :verbose => true],
             ],
             :description => "garbage collect packages not used for a significant time",
--- a/test/new.jl
+++ b/test/new.jl
@@ -2765,7 +2765,7 @@ end
         @test isempty(opts)
         api, opts = first(Pkg.pkg"gc --all")
         @test api == Pkg.gc
-        @test opts[:collect_delay] == Hour(0)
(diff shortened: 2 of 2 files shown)


src/REPLMode/command_declarations.jl
```diff
@@ -512,7 +512,7 @@ compound_declarations = [
             :name => "gc",
             :api => API.gc,
             :option_spec => [
-                PSA[:name => "all", :api => :collect_delay => Hour(0)],
+                PSA[:name => "all", :api => :collect_delay => nothing],
```
KristofferC, line 515: can't it be removed completely?

Keno: Yes. That would make it unavailable as an option in the REPL mode but that may be desirable. I just wanted to do the minimal change to get rid of the warning.

KristofferC: Can leave it like this for now (but have to fix the apparent CI failure that tests this specifically for some reason.

IanButterworth:
Still happening on master after https://github.com/JuliaLang/julia/commit/63185d76ad63d4be4cf61a889c901da9da2b4d38

```
    JULIA stdlib/release.image
Precompiling for 2 compilation configurations...
Precompiling packages finished.
  106 dependency configurations successfully precompiled in 85 seconds
  2 dependencies had output during precompilation:
┌ Pkg
│  ┌ Warning: The `collect_delay` parameter is no longer used. Packages are now deleted immediately when they become unreachable.
│  └ @ Pkg.API ~/Documents/GitHub/julia/usr/share/julia/stdlib/v1.14/Pkg/src/API.jl:616
└
┌ Pkg
│  ┌ Warning: The `collect_delay` parameter is no longer used. Packages are now deleted immediately when they become unreachable.
│  └ @ Pkg.API ~/Documents/GitHub/julia/usr/share/julia/stdlib/v1.14/Pkg/src/API.jl:616
└
```

PR #4490, opened 2025-11-04
Improve docs for test-specific deps with workspace approach

### LINKED ISSUE #4483: How to add parent package to test-specific deps with workspace method?
The example in the docs only shows a package with `Test` as a single test dep. What happens when `HelloWorld` is a dependency too?

See https://discourse.julialang.org/t/test-deps-with-workspaces/133591

gdalle:
Fixes #4483 by explaining how to add the parent package to `test/Project.toml` (which is now necessary with the workspace approach)


KristofferC:
Thanks!

PR #4494, opened 2025-11-05
use the same value of `--check-bounds` as the process

KristofferC:
There has been discussions that a full re-precompilation of the environment is not worth it by default.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -2595,7 +2595,6 @@ function gen_subprocess_flags(source_path::String; coverage, julia_args::Cmd)
     return ```
         --code-coverage=$(coverage_arg)
         --color=$(Base.have_color === nothing ? "auto" : Base.have_color ? "yes" : "no")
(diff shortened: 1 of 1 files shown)


src/Operations.jl
```diff
@@ -2595,7 +2595,7 @@ function gen_subprocess_flags(source_path::String; coverage, julia_args::Cmd)
     return ```
         --code-coverage=$(coverage_arg)
         --color=$(Base.have_color === nothing ? "auto" : Base.have_color ? "yes" : "no")
-        --check-bounds=yes
+        --check-bounds=$(("auto", "yes", "no")[Base.JLOptions().check_bounds + 1])
```
vtjnash:
```suggestion
```
The `Base.julia_cmd()` function already generates the correct arguments for the subprocess to fork

KristofferC:
Cannot backport this to 1.12, too big of a change

PR #4496, opened 2025-11-07
keep using gzip on Windows for the registry

KristofferC:


KristofferC:
There might be a better way to do this..

KristofferC:
Maybe not... I mean the obvious thing is of course to just have two parallel registries with two parallel toml files but it's kind of ugly...

PR #4498, opened 2025-11-07
Fix formatting of warning admonition

moble:
Admonition box is empty unless following text is indented with *four* spaces, not two.  This PR just adds those two spaces.


KristofferC:
Thanks, can we change this to info while we are at it? Warning feels too string here in my opinion.

moble:
> Thanks, can we change this to info while we are at it? Warning feels too string here in my opinion.

Done.

PR #4499, opened 2025-11-07
Fix some tests when run in base

Keno:
Part of https://github.com/JuliaLang/julia/issues/60058. When we don't have a log file setup, the default IO is set to devnull. However, this causes `Pkg.status` to early out and skip some validity checks that we are testing for. Fix this by passing an appropriate io directly to the API. For the REPL test, which don't have an IO argument, override the global default IO, but to make that more robust also switch it to a ScopedValue (which Pkg is already using elsewhere).

### DIFF
--- a/ext/REPLExt/precompile.jl
+++ b/ext/REPLExt/precompile.jl
@@ -14,7 +14,7 @@ let
         original_load_path = copy(LOAD_PATH)
         __init__()
         Pkg.UPDATED_REGISTRY_THIS_SESSION[] = true
-        @Base.ScopedValues.with Pkg.DEFAULT_IO => Pkg.unstableio(devnull) begin
+        Base.ScopedValues.@with Pkg.DEFAULT_IO => Pkg.unstableio(devnull) begin
             withenv("JULIA_PKG_SERVER" => nothing, "JULIA_PKG_UNPACK_REGISTRY" => nothing) do
                 tmp = Pkg._run_precompilation_script_setup()
                 cd(tmp) do


ext/REPLExt/precompile.jl
```diff
@@ -14,25 +14,26 @@ let
...
-        withenv("JULIA_PKG_SERVER" => nothing, "JULIA_PKG_UNPACK_REGISTRY" => nothing) do
-            tmp = Pkg._run_precompilation_script_setup()
-            cd(tmp) do
-                try_prompt_pkg_add(Symbol[:notapackage])
-                promptf()
-                term = FakeTerminal()
-                repl = REPL.LineEditREPL(term, true)
-                REPL.run_repl(repl)
-                repl_init(repl)
+        @Base.ScopedValues.with Pkg.DEFAULT_IO => Pkg.unstableio(devnull) begin
```
KristofferC: These even syntax highlights badly :P. I'm sure runic will complain about this at some point so might as well sawp it now heh.

Keno: FWIW, I don't really agree that the other macrocall form is preferable, but I'm ok to change it.

KristofferC:
I'm a bit scared of the `ScopedValues` stuff in general, but I guess we will see if it causes some issues down the line.

src/Pkg.jl
```diff
@@ -61,7 +61,7 @@ const UPDATED_REGISTRY_THIS_SESSION = Ref(false)
 const OFFLINE_MODE = Ref(false)
 const RESPECT_SYSIMAGE_VERSIONS = Ref(true)
 # For globally overriding in e.g. tests
-const DEFAULT_IO = Ref{Union{IO, Nothing}}(nothing)
+const DEFAULT_IO = Base.ScopedValues.ScopedValue{IO}()
```
DilumAluthge, line 64: @Keno I suspect that this line broke PkgEval. See https://github.com/JuliaLang/julia/commit/1067db80b96cafb9da6864afbfdd9d6e27665c33 for further discussion.

Essentially, I think it broke this section in the PkgEval script: https://github.com/JuliaCI/PkgEval.jl/blob/2b001d04dde0c33a22680ce76d22518a2b31de3e/scripts/evaluate.jl#L54-L55

```julia
    io = IOBuffer()
    Pkg.DEFAULT_IO[] = io
```

What's the best way to fix this? Modify PkgEval to check the type of `Pkg.DEFAULT_IO` and branch accordingly?

PR #4504, opened 2025-11-09
test: isolate more tests that call subprocesses

Keno:
The base `make test-Pkg` invocation sets JULIA_DEPOT_PATH to the builtin DEPOT_PATH. However, Pkg tests assume that this DEPOT_PATH do not have any registries. There were various places where we were accidentally using this DEPOT_PATH and installing a registry there. Work around this by unsetting it, isolating some subprocesses and adding a check on CI that fails if any pre-set JULIA_DEPOT_PATH gets modified.

Ref https://github.com/JuliaLang/julia/issues/60058

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -65,11 +65,6 @@ jobs:
           # See https://github.com/actions/virtual-environments/issues/712
           echo "TMP=${USERPROFILE}\AppData\Local\Temp" >> ${GITHUB_ENV}
           echo "TEMP=${USERPROFILE}\AppData\Local\Temp" >> ${GITHUB_ENV}
-          echo "JULIA_DEPOT_PATH=${GITHUB_WORKSPACE}\do_not_modify_depot;" >> ${GITHUB_ENV}
-      - name: Set JULIA_DEPOT_PATH (non-windows)
-        if: runner.os != 'Windows'
-        run: |
-          echo "JULIA_DEPOT_PATH=${GITHUB_WORKSPACE}/do_not_modify_depot:" >> ${GITHUB_ENV}
       - uses: actions/checkout@v5
       - uses: julia-actions/setup-julia@v2
         with:
@@ -82,13 +77,6 @@ jobs:
         env:
           JULIA_PKG_SERVER: ${{ matrix.pkg-server }}
           JULIA_TEST_VERBOSE_LOGS_DIR: ${{ github.workspace }}
-          JULIA_DEBUG: loading
-      - name: Check that we did not modify JULIA_DEPOT_PATH
-        run: |
-          if [ -d "${{ github.workspace }}/do_not_modify_depot/registries" ]; then
-            echo "Error: Tests modified the JULIA_DEPOT_PATH"
-            exit 1
-          fi
       - uses: actions/upload-artifact@v5
         if: ${{ always() }}
         with:
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -19,6 +19,8 @@ module PkgTestsInner
         error("The wrong Pkg is being tested")
     end
 
+    const original_depot_had_registries = isdir(joinpath(Base.DEPOT_PATH[1], "registries"))
+
     ENV["JULIA_PKG_PRECOMPILE_AUTO"] = 0
     ENV["JULIA_PKG_DISALLOW_PKG_PRECOMPILATION"] = 1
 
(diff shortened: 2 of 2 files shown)


DilumAluthge (commented):
CI changes look fine.

.github/workflows/test.yml
```diff
@@ -77,6 +77,15 @@ jobs:
         env:
           JULIA_PKG_SERVER: ${{ matrix.pkg-server }}
           JULIA_TEST_VERBOSE_LOGS_DIR: ${{ github.workspace }}
+          JULIA_DEPOT_PATH: ${{ github.workspace }}/do_not_modify_depot
```
KristofferC: Will this include the stdlib depot? We don't want to have to recompile all the stdlib packages.

DilumAluthge: Ah, good catch. I suppose we want this:

```suggestion
          JULIA_DEPOT_PATH: '${{ github.workspace }}/do_not_modify_depot:'
```

Based on my reading of: https://docs.julialang.org/en/v1/manual/environment-variables/#JULIA_DEPOT_PATH

.github/workflows/test.yml
```diff
@@ -77,6 +77,15 @@ jobs:
...
+      - name: Check that we did not modify JULIA_DEPOT_PATH
+        run: |
+          if [ -d "${{ github.workspace }}/do_not_modify_depot/registries" ]; then
+            echo "Error: Tests modified the JULIA_DEPOT_PATH"
+            exit 1
+          fi
+        env:
+          JULIA_DEPOT_PATH: ${{ github.workspace }}/do_not_modify_depot
```
DilumAluthge: Lines 87 and 88 aren't necessary and can be deleted, right?

.github/workflows/test.yml
Keno: Needs to be `;` on windows. How do I make this OS specific?

.github/workflows/test.yml
Keno: Originally I had written this in Julia, but Claude thought it would be simpler to just check the path directly in bash and it was right, so yes.

.github/workflows/test.yml
DilumAluthge: Oh god. Uh.... I guess define the environment variable programmatically instead of hardcoding it at the top. E.g. two different run steps, one for Windows, one for non-Windows. And in that run step, use GITHUB_ENV to define the environment variable for future steps.

.github/workflows/test.yml
```diff
@@ -57,14 +57,19 @@ jobs:
...
         if: matrix.os == 'windows-latest'
         run: |
           git config --global core.autocrlf false
           git config --global core.eol lf
           # See https://github.com/actions/virtual-environments/issues/712
           echo "TMP=${USERPROFILE}\AppData\Local\Temp" >> ${GITHUB_ENV}
           echo "TEMP=${USERPROFILE}\AppData\Local\Temp" >> ${GITHUB_ENV}
+          echo "JULIA_DEPOT_PATH=${GITHUB_WORKSPACE}\do_not_modify_depot;" >> ${GITHUB_ENV}
+      - name: Set JULIA_DEPOT_PATH (non-windows)
+        if: matrix.os != 'windows-latest'
```
DilumAluthge: Slightly more robust:

```suggestion
        if: runner.os != 'Windows'
```

.github/workflows/test.yml
```diff
@@ -57,14 +57,19 @@ jobs:
             julia-version: 'nightly'
             pkg-server: "pkg.julialang.org"
     steps:
-      - name: Set git to use LF and fix TEMP on windows
+      - name: Set git to use LF, fix TEMP, set JULIA_DEPOT_PATH (windows)
         if: matrix.os == 'windows-latest'
```
DilumAluthge:
```suggestion
        if: runner.os == 'Windows'
```

.github/workflows/test.yml
IanButterworth: Huh.. that never occurred to me and I've fixed so much of this kind of thing.. (or it did and I forgot, which is possible)

Keno:
Sigh, the CI mechanism here is not working, because obviously the Pkg tests use `Pkg.test`, which will already touch the depot, unlike either of the base mechanisms. I guess I will instead add a check to runtests.jl that if it didn't have the `registries` folder at the start, it doesn't have it at the end either. That won't catch it here, but it would catch it in base CI.

Keno:
Good to merge?

KristofferC:
It's probably better than it was before so go ahead

PR #4507, opened 2025-11-11
Prefer adding already loaded dep versions

IanButterworth:
Here JSON v1.3.0 exists. In the 2nd env Pkg.add prefers adding the versions that are already loaded so the env graph can be used in the current session without more precompilation.

Note that it's a preference via a high weighting in the resolver, not a hard rule.. so it shouldn't introduce resolver errors.

Screenshot 2025-11-16 at 11 08 25 AM

Developed with Codex and Claude

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@ Pkg v1.13 Release Notes
 
 - `Pkg.test` now respects the `--check-bounds` setting from the parent Julia session instead of forcing `--check-bounds=yes`.
 
+- `Pkg.add` now prefers versions of packages that are already loaded in the current Julia session when resolving
+  dependencies. This helps maintain compatibility with code already running in your session. The behavior can be
+  disabled using `Pkg.add(pkg; prefer_loaded_versions=false)`. ([#4507])
 - Project.toml environments now support a `readonly` field to mark environments as read-only, preventing modifications.
   ([#4284])
 - `Pkg.build` now supports an `allow_reresolve` keyword argument to control whether the build process can re-resolve
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,9 @@ Pkg v1.13 Release Notes
   silently in the background while returning to the REPL), `c` to cancel, `i` for a profile peek, `v` to toggle
   verbose mode, `?`/`h` for help, and `Ctrl-C` to interrupt. After detaching, use `pkg> precompile --monitor` to
   reattach, `--stop` to stop gracefully, or `--cancel` to cancel immediately. ([#4602])
+- `Pkg.add` now prefers versions of packages that are already loaded in the current Julia session when resolving
+  dependencies. This helps maintain compatibility with code already running in your session. The behavior can be
+  disabled using `Pkg.add(pkg; prefer_loaded_versions=false)`. ([#4507])
 - Project.toml environments now support a `readonly` field to mark environments as read-only, preventing modifications.
   ([#4284])
 - `Pkg.build` now supports an `allow_reresolve` keyword argument to control whether the build process can re-resolve
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,9 +17,10 @@ Pkg v1.13 Release Notes
   silently in the background while returning to the REPL), `c` to cancel, `i` for a profile peek, `v` to toggle
   verbose mode, `?`/`h` for help, and `Ctrl-C` to interrupt. After detaching, use `pkg> precompile --monitor` to
   reattach, `--stop` to stop gracefully, or `--cancel` to cancel immediately. ([#4602])
-- `Pkg.add` now prefers versions of packages that are already loaded in the current Julia session when resolving
-  dependencies. This helps maintain compatibility with code already running in your session. The behavior can be
-  disabled using `Pkg.add(pkg; prefer_loaded_versions=false)`. ([#4507])
+- The Pkg REPL `add` command now prefers versions of packages that are already loaded in the current Julia
+  session when resolving dependencies. This helps maintain compatibility with code already running in your
+  session. The functional `Pkg.add` API keeps the previous (loading-independent) behavior by default for
+  reproducibility; pass `Pkg.add(pkg; prefer_loaded_versions=true)` to opt in. ([#4507])
 - Project.toml environments now support a `readonly` field to mark environments as read-only, preventing modifications.
   ([#4284])
 - `Pkg.build` now supports an `allow_reresolve` keyword argument to control whether the build process can re-resolve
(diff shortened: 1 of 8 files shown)


KristofferC:
I'm not sure something like this should be on by default. I think it can get quite confusing if Pkg behaves differently depending on stateful things like what stuff you happen to have loaded.

IanButterworth:
I agree that it changes behavior in a potentially confusing way but it adds a message when it does, and status will show when packages are outdated. Seems worth it for avoiding all the confusing reprecompilation

IanButterworth:
This has a design issue that I've not yet been able to fix.. it's pinning the direct dep, rather than preferring it, like it does for the indirect deps.

KristofferC:
> Seems worth it for avoiding all the confusing reprecompilation

An alternative could be to have precompilation take into account currently loaded versions, or?

IanButterworth:
I prefer this approach because it does more to respect semver. Loaded package versions could be incompatible (compat, semver etc.) with the versions being added in the env. Hiding that issue further is a worse option, I think.

IanButterworth:
@StefanKarpinski is Resolver.jl likely to land in Pkg soon? I ask because I wonder if this will be easier to do with that? The current resolver does ok with the indirect deps, but I'm having a real headache allowing the resolver the flexibility to prefer the loaded version of the main dep being added, not have it pinned as the only option.

PatrickHaecker:
This is a very interesting feature and surely useful in some situations. However, I have doubts, too, that it should be active by default.

The way I understand it, you `add` a package and you can get an older version than what your packages of your environment are compatible with because you have some old package loaded.

Yes, it's indicated, that a new version is available, but that's easy to miss. And sure, you save same compilation time, but in the worst case you lose much more time by debugging an error which is long fixed in a more recent version of the package.

Won't that lead to the recommendation that you should always restart your Julia process before you add a package when you want to be up-to-date? Is that what we want to advocate?

Krastanov:
The vast majority of current questions I get from students confused about Pkg and "why is julia so slow and bloated" would be solved by this PR.

I think it is fantastic and much more understandable that `add` and `up` will become much more distinct after this RP. If anything, I would argue that this should have been the natural default from the very beginning: `add` should add packages consistent with the current environment, and only `up` should do the much more dangerous (and slow) operation of updating to newer versions.

IanButterworth:
> `add` should add packages consistent with the current environment

Specifically speaking these are deps that aren't in the current environment but are in the current session (where they are unchangeable because they are loaded). `add` already adds packages that are consistent with the current environment (where possible, via the tiered resolver)

PatrickHaecker:
> `add` already adds packages that are consistent with the current environment

> add and up will become much more distinct after this PR

Good points, so we already won't get the most up-to-date version which could be used given the set of installed packages (ignoring their version). So if you want this, it's always `add` followed by `update` anyway already. Nevertheless, with this PR it would be `add`, REPL restart, `update` in general.

I guess we should at least add this changed behavior to the documentation with this PR.

IanButterworth:
@KristofferC would you be open to trying this out for 1.14? I think it's better than the current status.. i.e. try to give the user something that will work in this session, without further precompilation in this session or future ones, respecting semver, with warning about loaded packages being preferred, and the out of date indicators.

KristofferC:
I still don't really like the statefulness on such a temporary state as the packages currently loaded, and from what I understand, this is only partially helping to the point where it is perhaps not super likely to be useful in practice. For example, if you instead of doing `add JSON@1.2; using JSON; activate --temp; add JSON` you did `add JSON@1.2; using JSON; activate env_with_JSON_1.3` you would immediately end up in the bad case again. And I think it is more likely that you swap to an existing environment than to a temp environment.

The core issue is the workflow where you load packages from different environments that have been resolved independently. That should imo be the "point of attack". Identify, why those cases occur and how workflows and features can be added so that that is basically never needed. 

But even so, maybe this PR moves us to a better local minima.. I don't really know.

IanButterworth:
My take on the scenarios.

1. This PR helps
```
add JSON@1.2
using JSON
activate --temp
add JSON
```

2. Yeah doesn't really help. That session is borked.
```
add JSON@1.2
using JSON
activate env_with_JSON_1.3
```

2. startup.jl-related - Super common - This PR helps
```
# startup.jl loads Revise from the default env with Preferences v1.5.1
add Foo # depends on Preferences
# Latest Preferences v1.5.2 installed
# precompilation + loaded/new session warning
using Foo
# precompilation
```


So, IMO, an improvement. 
cc. @topolarity

topolarity:
My $.02 (as a Pkg outsider):
> The core issue is the workflow where you load packages from different environments that have been resolved independently.

Agreed.

> That should imo be the "point of attack". Identify, why those cases occur and how workflows and features can be added so that that is basically never needed.

I think that's fair, but there are a few different sources of environment "tearing" like this. `] activate` is one big source of problems, but I think @IanButterworth is probably right that "Without touching `activate`, modify your environment, things may tear vs. your loaded environment" is another of the fundamental problematic cases (environment tearing "across time")

I don't think we can solve that tearing "across time" in general TBH - dynamically modifying an environment is at odds with running live (at least as long as Julia cannot safely "unload" packages) and you'll inevitably "tear" eventually. I think we have to give up on live env editing / immediately loadable changes when that happens (at least if you want to avoid the "frankenstein" loading environment) - https://github.com/JuliaLang/Pkg.jl/issues/4637 is one stab in that direction.

But when you _don't_ tear, you get to "have your cake and eat it too", i.e. edit your environment live _and_ keep a coherent loading environment to use it immediately. Things like this PR do seem to help with that

KristofferC:
Yes, but as an example, say I have a script whose goal is to produce some environment, it would be surprising to me that the packages I load would influence the environment I end up creating. For example, in the Pkg tests right now, we maybe end up loading some JSON or some package (I am not sure). Now every Pkg operation will be "tainted" by that load which might effect the environments we create in the rest of the test process. So I do believe there is value in that the (default) resolution process is more "pure" in the sense that the inputs are the current env, the registries, and it gives you an output based on that. 

Regarding the startup example, perhaps a possibility is instead to have precompilation take loaded packages into account. So you would then have:

```
# startup.jl loads Revise from the default env with Preferences v1.5.1
add Foo # depends on Preferences
# Latest Preferences v1.5.2 installed
# precompilation
using Foo
# no precompilation because precompilation produced valid package images for the actual resulting state, not the environment
```

But that of course has the drawback that if you load packages, do a `precompile` and then restart julia, you might need to precompile again.

IanButterworth:
The current balance is
Pkg.precompile -> respect the active env but warn that loading may hit more precompilation and restart is recommended
using precompile -> just make it work by paying the additional precompilation price (we could warn about not respecting env versions here.. but that seems OTT)

I don't like the idea of Pkg.precompile silently making it work because it could be at odds with the semver/compat of the current env.

topolarity:
> Yes, but as an example, say I have a script whose goal is to produce some environment

Would it be useful to split the default behavior, so that `] add` would (try to) be loading-consistent by default, while `Pkg.add(...)` is more "pure"?

IanButterworth:
That seems reasonable to me. I'll implement that.

IanButterworth:
@topolarity this is implemented now

> Would it be useful to split the default behavior, so that ] add would (try to) be loading-consistent by default, while Pkg.add(...) is more "pure"?

I personally think it's worth giving this a go alongside https://github.com/JuliaLang/Pkg.jl/pull/4679 as part of https://github.com/JuliaLang/Pkg.jl/issues/4680

KristofferC:
Let's try this then :).

PR #4510, opened 2025-11-12
Revert "test: isolate more tests that call subprocesses"

Keno:
Reverts JuliaLang/Pkg.jl#4504. CI is failing on master and rebased PRs, everything worked locally though, so try to see if it was my PR.


Keno:
This does not appear to have been the cause of the failing CI.

PR #4511, opened 2025-11-12
See if tests would pass with larger timeout

Keno:
Tests are timing out on master, but as far as I can tell, they're not actually stuck - something has just gotten massively slower after a julia upgrade. See if tests would pass with a larger timeout.


Keno:
https://github.com/JuliaLang/julia/pull/60114 I think

PR #4513, opened 2025-11-13
Backport fixes for conflicts between `[workspace]` and `[sources]` relative paths to release-1.12

MasonProtter:
Backport of https://github.com/JuliaLang/Pkg.jl/pull/4427 which fixes conflicts between `[workspaces]` and `[sources]` (e.g. https://github.com/JuliaLang/Pkg.jl/issues/3842). 

In order to to get it, I also needed to backport ~~https://github.com/JuliaLang/Pkg.jl/pull/4225~~ and https://github.com/JuliaLang/Pkg.jl/pull/4229

I also needed to modify the CI to run on `v1.12-nightly` instead of `nightly`, since otherwise the REPL extension failed to precompile.

### DIFF
--- a/test/sources.jl
+++ b/test/sources.jl
@@ -21,7 +21,7 @@ temp_pkg_dir() do project_path
                     cp("Project.toml.bak", "Project.toml"; force=true)
                     cp("BadManifest.toml", "Manifest.toml"; force=true)
                     Pkg.resolve()
-                    @test Pkg.project().sources["Example"] == Dict("url" => "https://github.com/JuliaLang/Example.jl")
+                    @test Pkg.project().sources["Example"] == Dict("rev" => "master", "url" => "https://github.com/JuliaLang/Example.jl")
                     @test Pkg.project().sources["LocalPkg"] == Dict("path" => "LocalPkg")
                 end
             end


KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/4513/commits/fb8b74dffd0a75d1f0a52df47b76c19d5cbded38 should not be in, right? That's a feature.

MasonProtter:
So at least with my local testing, I couldn't get this to work without that commit, presumably because of the way it modified `update_source_if_set(env, pkg)`

But maybe I was just being dumb, or there's a more surgical way we can take only the parts of that commit that are needed to make this work. I'll try a branch without it again and see what happens.

MasonProtter:
Ugh looks like there actually is a problem here too that's not just on Windows, I had accidentally only tested the version that didn't have the commit I actually wanted :facepalm:. Sorry for the premature ping on Slack.

MasonProtter:
Okay, so this is actually looking promising other than 2 things:
1. the `[sources]` test in https://github.com/MasonProtter/Pkg.jl/blob/mp/backport-sources-fix/test/sources.jl#L24 was expecting 
```julia
(Pkg.project()).sources["Example"] == Dict("url" => "https://github.com/JuliaLang/Example.jl")
```
but got
```
Dict(rev = "master", "url" => "https://github.com/JuliaLang/Example.jl")
```
which I think is okay if I'm understanding correctly, because that's what we are already doing on the master branch: https://github.com/JuliaLang/Pkg.jl/blob/master/test/sources.jl#L25, so I will update the test to do that

2. The workspace tests are failing at https://github.com/JuliaLang/Pkg.jl/blob/1a2b5ac0b81605c0a56fcff264520073d2fc4c03/test/workspaces.jl#L142C17-L142C47 during `generate` because we already `generate`'d `PrivatePackage` earlier in the test suite, so I think that can just be removed?

MasonProtter:
Okay, sorry for the noise previously, @KristofferC. This was my first time doing manual backporting and I made some dumb mistakes along the way, but I think it's looking good now.

MasonProtter:
Okay, I think I correctly got rid of that stray change to `pin`. It looks like what happened was that I messed up my local git history, and that change from `pin` got moved onto the wrong commit, so when I cherry picked the commit, it brought that change along with it... New to `git cherry-pick`, sorry!

MasonProtter:
Hmm, that change to `pin` was actually necessary for some of the tests. Should I change the tests, or backport the change to `pin`?

MasonProtter:
Bump @KristofferC

KristofferC:
Is the diff in https://github.com/JuliaLang/Pkg.jl/pull/4513/commits/55455a26656a3aa538c809c580d0497ba0e6fa37 weird? It shows it adds a large number of workspace tests but those are already on 1.12 as far as I can see?

KristofferC:
Actually, I think the whole testing block in the workspace file has gotten duplicated?

We should run runic on the 1.12 branch as well so that backports are conflicting less.

MasonProtter:
Okay, so I started over again with a fresh clone and re-did the `cherry-pick`s. 

1. It turns out that the `pin` change *was* in the commit https://github.com/JuliaLang/Pkg.jl/commit/2097cdb86a5063f065067fc5e0baba17b84d95e8 afterall, I don't know why Github was telling us before that it wasn't, so I've re-associated it with that cherry-picked commit
2. Instead of letting git's diff stuff handle the conflicts with the workspace test (sorry about that, I should have checked it more carefully), I just rejected the automatic diff and manually applied the change in https://github.com/JuliaLang/Pkg.jl/pull/4427/files#diff-f95997eecce5f2754dd5a12ee49dc25041c2b377cb8239112ec267a5abf7f0c0, I *think* that should do the right thing

I think the reason the diff was messed up for the workspace was that it merged funny. i.e., compare the diff in the link above against the diff in the merge commit: https://github.com/JuliaLang/Pkg.jl/commit/55455a26656a3aa538c809c580d0497ba0e6fa37#diff-f95997eecce5f2754dd5a12ee49dc25041c2b377cb8239112ec267a5abf7f0c0

______

Regarding Runic, do you want it applied after these commits, or before them?

KristofferC:
> Regarding Runic, do you want it applied after these commits, or before them?

Neither, it is just something that we should do later, to simplify future backports.

MasonProtter:
Sounds good. Thanks for your patience as I bumbled through this! :P

KristofferC:
https://github.com/JuliaLang/julia/pull/60158

KristofferC:
I think this enabled auto adding of source entries to Project.toml when using e.g. `dev`. I was kind of worried about that to be honest but I repressed it... :sob:

MasonProtter:
Ah crap, yeah it does appear to do that. I think this is the same problem I was having with `pin` and the workspace change, where the merge commit https://github.com/JuliaLang/Pkg.jl/commit/2097cdb86a5063f065067fc5e0baba17b84d95e8 is different from the PR diff in https://github.com/JuliaLang/Pkg.jl/pull/4229 so when I cherry picked that merge commit, I ended up taking chunks of code from that sources PR...

Uh, what's the best way to address this now that it's released?

KristofferC:
Revert this and 1.12.3 I guess...

MasonProtter:
I've opened a PR to try and address the problem without reverting the whole thing: https://github.com/JuliaLang/Pkg.jl/pull/4513

PR #4515, opened 2025-11-15
Use `Sys.STDLIB` instead of hardcoding path to Julia stdlib

nalimilan:
This is cleaner and will allow changing this path at JuliaLang/julia#54353.

### DIFF
--- a/src/utils.jl
+++ b/src/utils.jl
@@ -27,11 +27,11 @@ end
 const URL_regex = r"((file|git|ssh|http(s)?)|([\w\-\.]+@[\w\-\.]+))(:(//)?)([\w\.@\:/\-~]+)(\.git)?(/)?"x
 isurl(r::String) = occursin(URL_regex, r)
 
-stdlib_path(stdlib::String) = joinpath(Sys.STDLIB::String, stdlib)
+stdlib_path(stdlib::String) = joinpath(Sys.STDLIB, stdlib)
 
 function pathrepr(path::String)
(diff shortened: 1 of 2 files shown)


src/utils.jl
```diff
@@ -27,12 +27,11 @@ end
...
 isurl(r::String) = occursin(URL_regex, r)
 
-stdlib_dir() = normpath(joinpath(Sys.BINDIR::String, "..", "share", "julia", "stdlib", "v$(VERSION.major).$(VERSION.minor)"))
-stdlib_path(stdlib::String) = joinpath(stdlib_dir(), stdlib)
+stdlib_path(stdlib::String) = joinpath(Sys.STDLIB::String, stdlib)
 
 function pathrepr(path::String)
     # print stdlib paths as @stdlib/Name
-    if startswith(path, stdlib_dir())
+    if startswith(path, Sys.STDLIB::String)
```
KristofferC: I think `::String` is already annotated on the global.

nalimilan: Yeah. I added the type assertion because it's there in the current version (for `BINDIR` but the situation is the same). It was added in https://github.com/JuliaLang/Pkg.jl/pull/2049. Indeed in Julia 1.6.5 `Sys.BINDIR` wasn't inferred without it. But now it seems to work so I've pushed a commit to drop the assertions.

KristofferC: Yes, type annotations on declaration of non const global didn't exist back then

PR #4518, opened 2025-11-17
release-1.12: init jlls we use if they are not initialized (happens when building a julia itself with Pkg trying to be included in the Base sysimage)

KristofferC:


IanButterworth:
Julia should be calling all package inits in the right order during sysimage load, no?

KristofferC:
Julia calls no inits when the sysimage is *built*

PR #4520, opened 2025-11-18
Add support for `syntax.julia_version` in Project/Manifest

Keno:
Pkg.jl side of https://github.com/JuliaLang/julia/pull/60018. Written by Claude.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -374,6 +374,43 @@ function update_manifest!(env::EnvCache, pkgs::Vector{PackageSpec}, deps_map, ju
     return record_project_hash(env)
 end
 
+"""
+    get_project_syntax_version(p::Project) -> VersionNumber
+
+Extract the syntax version from a Project.
+
+This function determines which version of Julia syntax a package uses, following
+this precedence order:
+
+1. If `syntax.julia_version` is present in the Project.toml, use that value
+2. If `compat.julia` is specified, use the minimum version from the compat range
+3. Otherwise, default to the current Julia VERSION
+
+This information is used to populate the `syntax.julia_version` field in the
+Manifest.toml, allowing Base's loading system to parse each package with the
+correct syntax version.
+"""
+function get_project_syntax_version(p::Project)::VersionNumber
+    # First check syntax.julia_version entry in Project.other
+    if p.julia_syntax_version !== nothing
+        return VersionNumber(syntax_table["julia_version"])
+    end
+
+    # If not found, default to minimum(compat["julia"])
+    if haskey(p.compat, "julia")
+        julia_compat = p.compat["julia"]
+        # Get the minimum version from the first range
+        if !isempty(julia_compat.val.ranges)
+            first_range = first(julia_compat.val.ranges)
+            lower_bound = first_range.lower
+            return VersionNumber(lower_bound.t[1], lower_bound.t[2], lower_bound.t[3])
+        end
+    end
+
+    # Finally, if neither of those are set, default to the current Julia version
+    return VERSION
+end
+
 # This has to be done after the packages have been downloaded
 # since we need access to the Project file to read the information
 # about extensions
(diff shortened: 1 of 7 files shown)


src/Operations.jl
```diff
@@ -374,6 +374,46 @@ function update_manifest!(env::EnvCache, pkgs::Vector{PackageSpec}, deps_map, ju
...
+2. If `compat.julia` is specified, use the minimum version from the compat range
+3. Otherwise, default to the current Julia VERSION
+
+This information is used to populate the `syntax.julia_version` field in the
+Manifest.toml, allowing Base's loading system to parse each package with the
+correct syntax version.
+"""
+function get_project_syntax_version(p::Project)::VersionNumber
+    # First check syntax.julia_version entry in Project.other
+    if haskey(p.other, "syntax")
```
KristofferC: I don't think this should be in `p.other`. That is typically for "non official" project data. Should have a proper field `p.syntax_version` or something.

src/Operations.jl
```diff
@@ -374,6 +374,46 @@ function update_manifest!(env::EnvCache, pkgs::Vector{PackageSpec}, deps_map, ju
...
+            return VersionNumber(syntax_table["julia_version"])
+        end
+    end
+
+    # If not found, default to minimum(compat["julia"])
+    if haskey(p.compat, "julia")
+        julia_compat = p.compat["julia"]
+        # Get the minimum version from the first range
+        if !isempty(julia_compat.val.ranges)
+            first_range = first(julia_compat.val.ranges)
```
KristofferC, line 404: Are these guaranteed to be sorted like this?

Keno: Yes, the `VersionSpec` sorts them

src/Operations.jl
Keno: Done

src/Operations.jl
```diff
@@ -374,6 +374,43 @@ function update_manifest!(env::EnvCache, pkgs::Vector{PackageSpec}, deps_map, ju
...
+3. Otherwise, default to the current Julia VERSION
+
+This information is used to populate the `syntax.julia_version` field in the
+Manifest.toml, allowing Base's loading system to parse each package with the
+correct syntax version.
+"""
+function get_project_syntax_version(p::Project)::VersionNumber
+    # First check syntax.julia_version entry in Project.other
+    if p.julia_syntax_version !== nothing
+        return VersionNumber(syntax_table["julia_version"])
```
KristofferC, line 396: I'm a bit confused, what is `syntax_table`?

Keno: I typo'd but more importantly, I'm missing test coverage, will fix.

Keno:
Good to go?

src/Operations.jl
```diff
@@ -374,6 +374,43 @@ function update_manifest!(env::EnvCache, pkgs::Vector{PackageSpec}, deps_map, ju
...
+        # Get the minimum version from the first range
+        if !isempty(julia_compat.val.ranges)
+            first_range = first(julia_compat.val.ranges)
+            lower_bound = first_range.lower
+            return VersionNumber(lower_bound.t[1], lower_bound.t[2], lower_bound.t[3])
+        end
+    end
+
+    # Finally, if neither of those are set, default to the current Julia version
+    return VERSION
```
KristofferC, line 411: This seems kind of odd to me to tie this to the running julia version, or? The package doesn't know what version it will be run on, so how can we assign properties to it based on that?

Keno: There was a desire that packages without `[compat]` section should get latest syntax (like scripts and the REPL) to avoid confusing people when they move such code into packages. The idea here was that whatever julia version that generates the manifest would fix the syntax version and then it'd continue to work on other julia versions with the same syntax, but I agree with you it's a bit odd. Other consistent options would be:

- Always return 1.13 here and update base to match
- Omit this entry, let each julia version use its syntax version, in base loading distinguish on the toplevel julia_version entry whether it was omitted because it's a pre-1.14 manifest (in which case base should default to 1.13 syntax) or whether it's missing because lack of compat

I think the second option probably be the easiest mental model in practice, but it's more complicated on the implementation side.

KristofferC (approved):
let him cook

src/Operations.jl
KristofferC: We can go with this and see what happens

PR #4523, opened 2025-11-18
run runic on 1.12 and add the pre commit hook + check

KristofferC:
Will simplify 1.12 backports


KristofferC:
Internet is broken so this fails on some random workers but the fact that some workers passed on a formatting change PR makes me confident enough this is ok.

PR #4528, opened 2025-11-22
Isolate more tests

Keno:
As seen in https://github.com/JuliaLang/julia/pull/60197, the test I added to make sure that `make test-Pkg` doesn't corrupt the builtin depot is firing properly - we had several more tests that were improperly isolated. A related issue is that if `Preferences` is not in the depot that starts the test we were trying to Pkg.add it, causing the same problem. Fix that by treating it like the other external packages (Aqua, HistoricalStdlibVersions) and making the corresponding tests optional.

### DIFF
--- a/test/repl.jl
+++ b/test/repl.jl
@@ -845,16 +845,8 @@ end
 
 @testset "JuliaLang/julia #55850" begin
     isolate(loaded_depot = true) do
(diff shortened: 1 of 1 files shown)


DilumAluthge:
Note: easier to review with "whitespace changes" hidden.

PR #4537, opened 2025-11-25
Only check for left over registry installs outside of buildkite

IanButterworth:
It makes sense to run this test locally and on this repos github actions, but not on base tests where parallel tests are using the depot.

@Keno I couldn't identify a better identifier for that situation than `!Base.get_bool_env("BUILDKITE", false)`. I thought `Base.runtests` set some `JULIA_BASE_TESTS` env var, but I can't find such a thing..

### DIFF
--- a/test/runtests.jl
+++ b/test/runtests.jl
@@ -102,7 +102,11 @@ module PkgTestsInner
         end
 
         # Make sure that none of our tests have left temporary registries lying around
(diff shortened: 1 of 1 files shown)


Keno:
No, the test here is good, the base tests are bad - Pkg.jl assumes that this does not happen while it runs and will intermittently fail otherwise. https://github.com/JuliaLang/julia/pull/60232

PR #4545, opened 2025-11-28
Clamp julia syntax version to 1.13 when creating manifests

Keno:
Pkg side of [julia#60275](https://github.com/JuliaLang/julia/pull/60275). Base clamps this when loading explicitly from a Project.toml, so to the extent that a manifest is a compacted summary of Base's loading behavior, it should clamp as well. Of course, this isn't strictly necessary after the Base PR merges, but I think it makes sense to have the manifest reflect what loading would do, rather than what was in the Project.toml.

### DIFF
--- a/test/manifests.jl
+++ b/test/manifests.jl
@@ -648,4 +648,24 @@ end
     end
 end
 
+@testset "Syntax version clamping" begin
+    # Test that old compat/syntax declarations are clamped to v1.13 (NON_VERSIONED_SYNTAX)
+    test_cases = [
+        ("[compat]\njulia = \"1\"", v"1.13"),        # ancient compat clamped
+        ("[compat]\njulia = \"1.6\"", v"1.13"),      # old compat clamped
+        ("[compat]\njulia = \"1.14\"", v"1.14"),     # new compat not clamped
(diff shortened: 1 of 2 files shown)


test/manifests.jl
```diff
@@ -648,4 +648,101 @@ end
...
 end
 
+@testset "Syntax version clamping" begin
+    @testset "get_project_syntax_version clamps to NON_VERSIONED_SYNTAX" begin
+        # Test that old compat declarations are clamped to v1.13
+        # This prevents packages with ancient compat like "julia = 1" from
+        # having their syntax version set too low (which would cause the parser
+        # to reject newer syntax features like `import ... as`)
+
+        mktempdir() do test_dir
```
KristofferC: This feels like kind of "bloaty" tests. Ok, to have one full manifest test for `compat` and one for `syntax` but then the rest can maybe just be unit style tests that can be tested on a single line?

Keno: I do think all four cases are reasonable, but yeah, the implementation is a bit bloaty. Will have claude condense it.

Keno:
@KristofferC good to go?

Issue #4546, opened 2025-11-29
[windows] The system cannot find the batch label specified - __next


ghyatzo:
looking that the `.bat` files generated when installing an app there seems to be no `:__next` label anywhere.

KristofferC:
https://github.com/JuliaLang/Pkg.jl/blob/7a82e43947ac6ed9687069158c762c85eb7ccad8/src/Apps/Apps.jl#L587 is not there? What julia version?

ghyatzo:
eeeer, sometimes I feel like an LLM. it is indeed there. apologies for the noise.
The error is real though.

```
julia> versioninfo()
Julia Version 1.12.1
Commit ba1e628ee4 (2025-10-17 13:02 UTC)
Build Info:
  Official https://julialang.org release
Platform Info:
  OS: Windows (x86_64-w64-mingw32)
  CPU: 12 × 13th Gen Intel(R) Core(TM) i7-1355U
  WORD_SIZE: 64
  LLVM: libLLVM-18.1.7 (ORCJIT, alderlake)
  GC: Built with stock GC
Threads: 1 default, 1 interactive, 1 GC (on 12 virtual cores)
Environment:
  JULIA_PKG_DEVDIR = ~/Dev/+Julia
```

ghyatzo:
popped up trying to use the https://github.com/aviatesk/TestRunner.jl app.
Other apps seems to work fine.

the error coming from the `.bat` file I assumed the issue was with the shims generated by Pkg.

This is the actual shim:
<details><summary>Actual shim</summary>
<p>

```
@echo off
setlocal EnableExtensions DisableDelayedExpansion

    REM  This file is generated by the Julia package manager.
REM  Shim version: 1.1

rem --- Environment (no delayed expansion here to keep '!' literal) ---
set "JULIA_LOAD_PATH=C:\Users\camillo.schenone\.julia\environments\apps\TestRunner"
set "JULIA_DEPOT_PATH=C:\Users\camillo.schenone\.julia;C:\Users\camillo.schenone\.julia\juliaup\julia-1.12.1+0.x64.w64.mingw32\local\share\julia;C:\Users\camillo.schenone\.julia\juliaup\julia-1.12.1+0.x64.w64.mingw32\share\julia"

rem --- Allow overriding Julia executable via environment variable ---
if defined JULIA_APPS_JULIA_CMD (
    set "julia_cmd=%JULIA_APPS_JULIA_CMD%"
) else (
[...]

PR #4547, opened 2025-11-29
normalize line endings for .bat files on Windows to use \r\n conistently

KristofferC:
Hopefully fixes https://github.com/JuliaLang/Pkg.jl/issues/4546


KristofferC:
Apps test succeded on that one windows worker that failed.

PR #4556, opened 2025-12-05
add workspace option to more functions

KristofferC:


IanButterworth:
Maybe this should be backported to 1.12. Seems like a bugfix in a way. 

https://discourse.julialang.org/t/psa-github-dependabot-now-supports-julia/134997/11?u=ianshmean

KristofferC:
Yeah, just gotta add a test and this should be good to go.

giordano:
Friendly bump 🙂

PR #4558, opened 2025-12-09
fix: fixup historical stdlibs in manifest

benlorenz:
`p.deps` in this branch is a `Vector{Base.UUID}` so `name in p.deps` will always be false (but it doesn't error like before https://github.com/JuliaLang/Pkg.jl/pull/4264).
The type is from [`StdlibInfo`](https://github.com/JuliaPackaging/HistoricalStdlibVersions.jl/blob/main/src/StdlibInfo.jl#L15) in HistoricalStdlibVersions.

I don't really know how to test that this works correctly.

cc: @Keno @giordano 

It would also be great if this could make it's way into 1.12 since we are running into the original issue with `MethodError: no method matching haskey(::Vector{Base.UUID}, ::String)`.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -425,8 +425,8 @@ function fixups_from_projectfile!(ctx::Context)
             pkg.weakdeps = Dict{String, Base.UUID}(stdlibs[uuid].name => uuid for uuid in p.weakdeps)
             # pkg.exts = p.exts # TODO: STDLIBS_BY_VERSION doesn't record this
             # pkg.entryfile = p.entryfile # TODO: STDLIBS_BY_VERSION doesn't record this
-            for (name, _) in pkg.weakdeps
-                if !(name in p.deps)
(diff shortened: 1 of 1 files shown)


giordano:
If this works, I'd be ok with hot-patching this in BinaryBuilder with an `@eval` guarded by version checks.

IanButterworth:
A test would be good.

benlorenz:
I don't know how to fully test this, the only test that was added in #4264 still seems to work (in the sense that it doesn't error).

I am still trying to figure out what conditions would be needed for the inner `if` to **not** run.
I think it needs to be resolving for a `stdlib` which has a `weakdep` on a package that is also listed in the `deps` of that stdlib (both lists using the data in `HistoricalStdlibVersions`).
If I am reading the file correctly there is no such case and I'm not sure how that would make sense.

IanButterworth:
Yeah I think you're right

IanButterworth (approved):

PR #4563, opened 2025-12-11
[release-1.12] Backports for 1.12

IanButterworth:
- [x] #4264 
- [x] #4558 
- [x] #4568 

Need manual backport:
- [ ] #4402 
- [x] #4467 
- [x] #4463 
- [x] #4505 

Contains multiple commits, manual intervention needed:
- [ ] #4427 
- [x] #4418 
- [x] #4534 

Non-merged PRs with backport label:
- [x] #4559 
- [x] #4555 
- [x] #4541

PR #4568, opened 2025-12-16
[release-1.12] Fix `AssertionError: normpath(entry.path) == normpath(path)`. Add test.

IanButterworth:
https://github.com/JuliaLang/Pkg.jl/pull/4539 removed the code that syncs sources from manifest entries but kept the assertions checking that sources match manifest. This causes `AssertionError` when instantiating workspace projects with `[sources]` entries (e.g.,` path = ".."`) since the assertions check invariants that are no longer maintained.

Fixes the error:
```
ERROR: AssertionError: normpath(entry.path) == normpath(path)
```

### DIFF
--- a/test/test_packages/WorkspaceSourcesParent/Project.toml
+++ b/test/test_packages/WorkspaceSourcesParent/Project.toml
@@ -1,0 +1,6 @@
+name = "WorkspaceSourcesParent"
+uuid = "11111111-1111-1111-1111-111111111111"
+version = "0.1.0"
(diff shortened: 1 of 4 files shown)


IanButterworth:
Testing the added test without the fix https://github.com/JuliaLang/Pkg.jl/pull/4569 (confirmed)

KristofferC (approved):

IanButterworth:
Folded into #4563

PR #4582, opened 2025-12-29
fix `Core.Box`es in the package

KristofferC:
Found using the script in https://github.com/JuliaLang/julia/pull/60478

PR #4584, opened 2025-12-30
avoid Core.Box in `API.jl`

KristofferC:
I'm confused about the state of CI in https://github.com/JuliaLang/Pkg.jl/pull/4582 so I will have to do the commits file by file or something to figure out what is going on.


KristofferC:
Hm, maybe nightly broke something...

PR #4593, opened 2026-01-15
Fix a JET warning in `tree_hash`

lgoettgens:
This fixes the following JET warning:
```julia
┌ tree_hash(::Type{HashType}, root::AbstractString; debug_out::Union{Nothing, IO}, indent::Int64) where HashType @ Pkg.GitTools /home/lgoe/code/julia/Pkg.jl/src/GitTools.jl:382
│ no matching method found `take!(::Nothing)` (1/2 union split): Pkg.GitTools.take!(child_stream::Union{Nothing, IOBuffer})
```


lgoettgens:
If possible, it would also be great to get this backported to 1.13 and 1.12.

fingolfin (approved):

PR #4601, opened 2026-01-24
Highlight when instantiate is called on an empty env

IanButterworth:
Because the user probably forgot to use `--project` etc. and is unknowingly `instantiating` the empty default env on CI.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1278,9 +1278,7 @@ function instantiate(
     end
 
     # Check for empty environment early
-    project_empty = !isfile(ctx.env.project_file) || isempty(ctx.env.project.deps)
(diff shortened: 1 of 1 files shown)


visr:
Currently if you call instantiate on an empty environment, it writes only a Manifest.toml the first time, and adds a Project.toml the next time. That is probably unintentional, and I guess with this it would never write either.

I wonder how one is supposed to write empty projects though. In the context of https://github.com/JuliaLang/Pkg.jl/pull/4473 it would be nice to be able to `init` an env, since a CLI call to `activate` doesn't have side effects. So I do activate followed by instantiating twice 😅. Not that that should be relied upon or block this PR, just wanted to mention it.

PR #4602, opened 2026-01-25
precompile: Support detaching and monitoring etc.

### LINKED ISSUE #4359: Pkg.precompile() needs to be moved to background task
Doing the "Alan test" today, we really need to move Pkg.precompile() to just launch a precompile task and immediately return fast. Right now, it adds a ton of latency to getting to something useful, at least half of which work is typically useless. I've always had to entirely disable it because of this problem (`JULIA_PKG_PRECOMPILE_AUTO=0`), but if it only ran in the background as an idle task (stopping/restarting whenever the user does any other pkg operation or precompile operation so that it can prioritize what the user actually needs), then user configuration would no longer be necessary (since we already have the pidfile locks now to avoid conflicts and so it no longer must be blocking to avoid errors).

IanButterworth:
See https://github.com/JuliaLang/julia/pull/60943

Closes #4359

### DIFF
--- a/src/Pkg.jl
+++ b/src/Pkg.jl
@@ -266,6 +266,19 @@ end
 !!! compat "Julia 1.13"
     The `Pkg.precompile(f, args...; kwargs...)` do-block syntax requires at least Julia 1.13.
 
+During interactive precompilation the following keyboard controls are available:
+
+  * **`d`/`q`/`]`** — Detach. Returns to the REPL while precompilation continues in the background.
+    Use `pkg> precompile --monitor` to reattach, `--stop` to stop, or `--cancel` to cancel.
+  * **`c`** — Cancel. Kills all subprocesses; prompts for Enter to confirm.
+  * **`i`** — Info. Sends a profiling signal to subprocesses for a profile peek.
+  * **`v`** — Toggle verbose mode. Shows timing, worker PID, CPU%, and memory per compiling package.
+  * **`?`/`h`** — Show keyboard shortcut help.
+  * **Ctrl-C** — Interrupt. Sends SIGINT to subprocesses and displays their output.
+
+!!! compat "Julia 1.14"
+    Keyboard controls during precompilation require at least Julia 1.14.
+
 # Examples
 ```julia
 Pkg.precompile()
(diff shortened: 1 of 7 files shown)


src/Pkg.jl
```diff
@@ -96,6 +96,19 @@ function should_autoprecompile()
     end
 end
 
+# Background precompilation state
+mutable struct BackgroundPrecompileState
+    task::Union{Nothing, Task}
+    stop_requested::Bool
+    cancel_requested::Bool
```
vchuravy: Where are these two read?

src/Pkg.jl
```diff
@@ -1006,10 +1019,196 @@ end
...
+    # Stop any existing background precompilation
+    lock(BACKGROUND_PRECOMPILE.lock) do
+        if BACKGROUND_PRECOMPILE.task !== nothing && !istaskdone(BACKGROUND_PRECOMPILE.task)
+            BACKGROUND_PRECOMPILE.stop_requested = true
+            # Wait a short time for it to stop gracefully
+            for _ in 1:10
+                istaskdone(BACKGROUND_PRECOMPILE.task) && break
+                sleep(0.1)
+            end
+        end
```
vchuravy: This is perhaps flaky because there is no gurantuee that the task will stop gracefully?

src/Pkg.jl
```diff
@@ -1006,10 +1019,196 @@ end
...
+    end
+
+    # Capture necessary context for background task
+    # We need to copy the project file path and package info, not the IO handles
+    project_file = ctx.env.project_file
+    pkg_names = String[pkg.name for pkg in pkgs]
+
+    # Launch new background precompilation
+    lock(BACKGROUND_PRECOMPILE.lock) do
+        BACKGROUND_PRECOMPILE.task = @async begin
```
vchuravy: Should be `Threads.@spawn`

IanButterworth:
@vchuravy this isn't ready for code review. It's just a demo of the interface. Sorry, I should've been clearer up top.

IanButterworth:
Started moving the code over to Base https://github.com/JuliaLang/julia/compare/master...IanButterworth:julia:ib/background_precompile

vtjnash:
Will close https://github.com/JuliaLang/Pkg.jl/issues/4359

PR #4605, opened 2026-01-26
Add version note for [workspace] section

### LINKED ISSUE #4506: State in documentation when certain `Project.toml` features were added
Looking at <https://pkgdocs.julialang.org/v1/toml-files> if there were some points of the form "Added in Pkg.jl version X / Julia version Y"

Like, I am now reading up on [the `[workspace]` section](https://pkgdocs.julialang.org/v1/toml-files/#The-[workspace]-section) and I think this is new in Julia 1.12? But maybe not?

I think this kind of info would be really useful to have.

Similar for e.g. `[sources]` which I think is also fairly new (added in Julia 1.11 perhaps?)

gamila-wisam:
fixes #4506
Adds a compatibility note to the [workspace] section in docs/src/toml-files.md indicating that workspaces require Julia 1.12+.

### DIFF
--- a/docs/src/toml-files.md
+++ b/docs/src/toml-files.md
@@ -112,7 +112,7 @@ the section on [pre-1.0 behavior](@ref compat-pre-1.0) for more details.
 ### The `readonly` field
 
 !!! compat
-    The `readonly` field requires Julia 1.12+.
+    The `readonly` field requires Julia 1.13+.
 
 The `readonly` field is a boolean that, when set to `true`, marks the environment as read-only. This prevents any modifications to the environment, including adding, removing, or updating packages. For example:
 
@@ -290,9 +290,6 @@ Workspace can be nested: a project that itself defines a workspace can also be p
 In this case, the workspaces are "merged" with a single manifest being stored alongside the "root project" (the project that doesn't have another workspace including it).
 
 ### The `[extras]` section (legacy)
-
-!!! compat
-    Extras require Julia < 1.13; for Julia 1.13+ use [workspaces](@ref Workspaces).
 
 !!! warning
     The `[extras]` section is a legacy feature maintained for compatibility. For Julia 1.13+,


gamila-wisam:
This PR adds the same compatibility note for [workspace] as in #4581, but places it right after the section heading so it is immediately visible.
The previous PR seems to have unrelated CI failures (docs-only changes should not affect tests).
This should resolve #4506.

inkydragon (commented):
> Like, I am now reading up on the [workspace] section ...

You may need to check all these sections to see
if compatibility notes need to be added.
I recommend starting with Julia v1.10 (the current LTS version),
or even earlier versions like Julia v1.6 (the previous LTS version).

gamila-wisam:
Hi @inkydragon ,
I’ve added the compatibility notes for the other Project.toml sections.
When you have a chance, could you please review the changes and let me know if everything looks good? Thanks!

(to clarify, the CI has passed.
The codecov/projectFailing failing is appearing because this PR only changes documentation, the [sources] compatibility note was moved to match the style of other sections.
This caused a slight change in coverage report line offsets.
so no broken tests and everything is working as expected.)

docs/src/toml-files.md
```diff
@@ -111,6 +111,9 @@ the section on [pre-1.0 behavior](@ref compat-pre-1.0) for more details.
 
 ### The `readonly` field
 
+!!! compat
+    The `readonly` field requires Julia 1.12+.
```
KristofferC: Is this true?

docs/src/toml-files.md
```diff
@@ -285,6 +291,9 @@ In this case, the workspaces are "merged" with a single manifest being stored al
 
 ### The `[extras]` section (legacy)
 
+!!! compat
+    Extras require Julia < 1.13; for Julia 1.13+ use [workspaces](@ref Workspaces).
```
KristofferC: This seems redundant with the information just below it?

docs/src/toml-files.md
gamila-wisam: Updated!, double-checked and it's julia 1.13+

docs/src/toml-files.md
gamila-wisam: Updated!, originally i added it to make it extra clear, but I see now that was unnecessary and just causing redundancy

gamila-wisam:
Docs CI tests passes , but nightly tests fails due to an upstream issues(including GitHub fetch errors), i already rebased onto current master, but the nightly fails remains.

gamila-wisam:
@KristofferC Hi, can you take a look when you have a moment please?

PR #4620, opened 2026-02-12
`release-1.10`: Update a test manifest to Julia 1.10

DilumAluthge:
This manifest was generated with Julia 1.9, so this PR updates it to 1.10.

Also, currently, this manifest gets flagged by some vulnerability scanners because of old versions of Curl and MbedTLS.[^1] Upgrading the manifest fixes this (because Curl and MbedTLS no longer appear in the updated manifest).

[^1]: Yes, you can add the manifest to an ignorelist, but in this case it's easy enough for us to just fix the manifest.


DilumAluthge:
This might not be destined for success.

PR #4625, opened 2026-02-25
Naming guidelines: Add new guideline that wrapper packages should indicate their wrapper-ness in some way

DilumAluthge:
This is the result of internal discussion among the General registry maintainers.


DilumAluthge:
Needs backport, so it can show up in the `v1` version of the docs sooner.

PR #4632, opened 2026-03-05
Suppress precompilation messages in Pkg subprocesses via `-q`

topolarity:
As pointed out by @vtjnash in https://github.com/JuliaLang/julia/pull/61064#issuecomment-4001629363, pre-compilation during `Pkg.test` (arguably unlike "typical" Julia executions) is not really unexpected or exceptional.

If we want to be able to tell the user that this (bad?) pre-compilation is happening, then we need to make the auto-precompile reliable enough that this is properly "exceptional" and does not clog users' terminals and PkgEval logs.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1814,7 +1814,7 @@ function gen_build_code(build_file::String; inherit_project::Bool = false, add_s
     # *unless* the parent julia session is started with --startup=yes explicitly.
     startup_flag = Base.JLOptions().startupfile == 1 ? "yes" : "no"
     return ```
-    $(Base.julia_cmd()) -O0 --color=no --history-file=no
+    $(Base.julia_cmd()) -q -O0 --color=no --history-file=no
     --startup-file=$startup_flag
     $(inherit_project ? `--project=$(Base.active_project())` : ``)
     --eval $code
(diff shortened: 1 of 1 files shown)


src/Operations.jl
```diff
@@ -2957,7 +2957,16 @@ testfile(source_path::String) = joinpath(testdir(source_path), "runtests.jl")
...
     threads_arg = with_threads ? `--threads=$(get_threads_spec())` : ``
-    cmd = `$(Base.julia_cmd()) $threads_arg $(flags) --eval $code`
+    # Suppress "Precompiling..." messages via `-q` to avoid spamming users with
+    # precompilation work that is routine and not actionable from their perspective.
+    #
+    # FIXME: This is only necessary because auto-precompile before test() is not
+    # reliable and may precompile packages that still need re-precompilation in
+    # the test subprocess. If that design is changed to be properly predictive,
+    # precompilation during test execution becomes rare and indicates a real
+    # problem, so this logging can be enabled again.
```
vtjnash, line 2967: Why do we still have this code here at all anymore? We used to do it because precompile lived in Pkg, but we moved that code to Base since it mades a lot of PkgEval tests a lot slower and broke some others (which didn't expect Pkg to be loaded)

topolarity: All Pkg implements is https://github.com/JuliaLang/Pkg.jl/blob/7c29673db182d4aafd137f4aaeb127d0fe6550a2/src/API.jl#L1235-L1248

The comment is about the precompile logic, regardless of where it lives

PR #4633, opened 2026-03-12
Switch to Julia 1.12 @main function declaration syntax

PatrickHaecker:
Julia LTS (1.10) does not support this syntax.
Julia 1.11 is no longer maintained.
So Julia 1.12+ syntax can be used.


PatrickHaecker:
Build failure seems unrelated.

PR #4635, opened 2026-03-12
show what project is constraining a package version in outdated mode to status when in a workspace

KristofferC:
Example:

Before:
```
(PrecompileExt) pkg> st --workspace --outdated
...
⌅ [7876af07] Example v0.5.3 (<v0.5.5) [compat]
```

After:
```
(PrecompileExt) pkg> st --workspace --outdated
...
⌅ [7876af07] Example v0.5.3 (<v0.5.5) [compat: BarEnv]
```


IanButterworth:
`[compat entry in BarEnv]` might be clearer?

KristofferC:
Clearer yes but it becomes a bit long maybe. I don't think "entry" is needed at least.

IanButterworth:
Yeah maybe `[compat in BarEnv]` is a good balance. Otherwise it could be interpreted in either direction.

PR #4638, opened 2026-03-16
docs: tweak [workspace] section

IanButterworth:
I think a couple of things can be confusing so worth docs:
- The need for the [sources] reference to the main package
- That the test Project isn't just test-specific deps, but all deps used by tests

### DIFF
--- a/docs/src/creating-packages.md
+++ b/docs/src/creating-packages.md
@@ -309,9 +309,21 @@ projects = ["test"]
 ```
 
 When using workspaces, the package manager resolves dependencies for all projects in the workspace together, and creates a single `Manifest.toml` next to the base `Project.toml`. This provides better dependency resolution and makes it easier to manage test-specific dependencies.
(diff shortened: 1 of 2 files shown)


IanButterworth:
@giordano as you've been trying out workspaces, is there anything else you think we should add?

KristofferC:
This is described here I would argue https://pkgdocs.julialang.org/dev/creating-packages/#Test-specific-dependencies

KristofferC:
There could be a small note about not inheriting but I think a full test example is a bit much here, maybe a link towards the end?

IanButterworth:
Ah the dev docs are better than the [1.12 docs](https://pkgdocs.julialang.org/v1.12/creating-packages/#Test-specific-dependencies) on this. I guess there's a backport missing.

Updated based on suggestions.

IanButterworth:
The 1.12 docs page was just out of date because we hadn't tagged since 1.12.3. Updated now.

KristofferC (approved):

PR #4641, opened 2026-03-23
allow file paths for pkg server URL

### LINKED ISSUE #4640: Allow file paths for JULIA_PKG_SERVER variable
Could the regex in [get_server_dir](https://github.com/JuliaLang/Pkg.jl/blob/master/src/PlatformEngines.jl#L80) be modified to not throw warnings on "file:///" urls?

In my initial testing this seems to be the only required change to get an isolated computer working against a "file based" Julia package server when setting the ENV var JULIA_PKG_SERVER=file:///some/local/path

`
-m = match(r"^\w+://([^\\/]+)(?:$|/)", server)

+m = match(r"^\w+:///?([^\\/]+)(?:$|/)", server)
`

KristofferC:
Fixes #4640

### DIFF
--- a/test/platformengines.jl
+++ b/test/platformengines.jl
@@ -238,6 +238,13 @@ end
                 end
             end
         end
+
+        # file:// URLs (issue #4640)
+        for path in ["/some/local/path", "/srv/pkg"]
+            server = "file://$(path)"
+            url = "$(server)/foo"
+            test_server_dir(url, server, basename(path))
+        end
     end
 
     called = 0
--- a/test/platformengines.jl
+++ b/test/platformengines.jl
@@ -240,10 +240,10 @@ end
         end
 
         # file:// URLs (issue #4640)
-        for path in ["/some/local/path", "/srv/pkg"]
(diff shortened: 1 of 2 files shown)


test/platformengines.jl
```diff
@@ -238,6 +238,13 @@ end
                 end
             end
         end
+
+        # file:// URLs (issue #4640)
+        for path in ["/some/local/path", "/srv/pkg"]
```
pfitzseb: Worth putting a test here for windows paths too. Should look like `file:///c%3A/foo/bar` I think.

test/platformengines.jl
```diff
@@ -238,6 +238,13 @@ end
                 end
             end
         end
+
+        # file:// URLs (issue #4640)
+        for (path, expected) in [("/some/local/path", "some"), ("/srv/pkg", "srv"), ("/c%3A/foo/bar", "c%3A")]
```
pfitzseb, line 243: The `expected` dirs are a bit problematic here since `/some/local` and `/some/local2` will overwrite each other. This should use the full path instead of only the first path segment (since the hostname is empty in a `file://$host/$path` url).

PR #4642, opened 2026-03-24
Fix race condition potentially overwriting user configuration

PatrickHaecker:
If the user set, e.g.,
```julia
atreplinit() do repl
    # Robust against older julia versions
    if hasfield(typeof(repl.options), :auto_insert_closing_bracket)
        repl.options.auto_insert_closing_bracket = false
    end
end
```
in their `startup.jl` it has no effect in VS Code's REPL due to VS Code's plugin loading `Pkg.jl` which is calling `REPL.setup_interface(repl)` inside an `atreplinit` hook, which bakes keymaps before user options are applied. The proper approach is to let `run_frontend` handle interface creation and defer mode initialization until the interface exists.

Note, that this fix is necessary, but not sufficient to make it work, because the same race condition needs to be fixed at least in [TerminalPager.jl](https://github.com/ronisbr/TerminalPager.jl/pull/83) and in [julia-vscode](https://github.com/julia-vscode/julia-vscode/pull/4063).

I used Claude Opus 4.6 while developing this fix.

### DIFF
--- a/ext/REPLExt/REPLExt.jl
+++ b/ext/REPLExt/REPLExt.jl
@@ -369,8 +369,15 @@ function __init__()
     else
         atreplinit() do repl
             if isinteractive() && repl isa REPL.LineEditREPL
-                isdefined(repl, :interface) || (repl.interface = REPL.setup_interface(repl))
-                repl_init(repl)
+                # Do not call `REPL.setup_interface` here — it will be called
+                # by `run_frontend`, which respects user options set in other
+                # `atreplinit` hooks (e.g. `auto_insert_closing_bracket`).
+                @async begin
+                    while !isdefined(repl, :interface)
+                        sleep(0.1)
+                    end
+                    repl_init(repl)
+                end
             end
         end
     end
--- a/ext/REPLExt/REPLExt.jl
+++ b/ext/REPLExt/REPLExt.jl
@@ -369,15 +369,15 @@ function __init__()
     else
         atreplinit() do repl
             if isinteractive() && repl isa REPL.LineEditREPL
-                # Do not call `REPL.setup_interface` here — it will be called
-                # by `run_frontend`, which respects user options set in other
-                # `atreplinit` hooks (e.g. `auto_insert_closing_bracket`).
-                @async begin
-                    while !isdefined(repl, :interface)
-                        sleep(0.1)
+                # Let run_frontend build keymaps after all atreplinit hooks run.
+                errormonitor(
+                    @async begin
+                        while !isdefined(repl, :interface)
+                            sleep(0.01)
+                        end
+                        repl_init(repl)
                     end
-                    repl_init(repl)
-                end
+                )
             end
         end
     end


ext/REPLExt/REPLExt.jl
```diff
@@ -359,8 +359,15 @@ function __init__()
     else
         atreplinit() do repl
             if isinteractive() && repl isa REPL.LineEditREPL
-                isdefined(repl, :interface) || (repl.interface = REPL.setup_interface(repl))
-                repl_init(repl)
+                # Do not call `REPL.setup_interface` here — it will be called
+                # by `run_frontend`, which respects user options set in other
+                # `atreplinit` hooks (e.g. `auto_insert_closing_bracket`).
+                @async begin
```
KristofferC: I feel there should be some way to register something to run once the interface is created o something like that, instead of needing a sleep loop like this.

PatrickHaecker: I fully agree. This PR is in the spirit of "at least make it work correctly" not in the spirit of "this is how the optimum solution should look like".

When I implemented this pattern for TerminalPager.jl, I did some analysis on my average notebook and typically it did not sleep at all during a regular REPL start, so it was more of a "if things are unusual still make them work correctly" thing. I haven't done any timing analysis on the `Pkg.jl` race condition, but I was reliably able to reproduce the problem when starting the REPL in VS Code, so I'd expect it to sleep at least once on my computer in this use case.

As much as I want to have the registration interface, I guess this would need to come earliest with Julia 1.14 and would need to tickle down until it hits the long term supported Julia release until we can get rid of this sleep loop.

IanButterworth: I think this is related https://github.com/JuliaLang/julia/pull/54890

PatrickHaecker: I don't understand all relevant code well enough to state whether https://github.com/JuliaLang/julia/pull/54890 will fix this, but given that we have a synchronization problem here, a synchronization mechanism where only one user at a time gets access to a lock with the REPL guaranteed to get access initially seems to solve the problem.

However, I suggest to not let the perfect be the enemy of the good. Starting with Julia 1.13, we [recommend a way](https://docs.julialang.org/en/v1.13-dev/stdlib/REPL/#Automatic-bracket-insertion) which is broken by the current implementation of Pkg.jl, so the REPL in a regular terminal behaves differently (if PKG is not loaded on startup) than the VS Code terminal which is quite annoying.

We can change the sleep time and/or we can progressively increase the sleep time if you want to be sure that the unnecessary relative wait time is limited, but I think we should implement something close to the current PR in Pkg now and then everyone is invited to improve on it in and for 1.14 onward.

PatrickHaecker:
@KristofferC, @IanButterworth:
Is there anything getting worse if this PR were merged? I agree that it is not perfect, but the way I see it, it will not make a performance difference in the good case (no sleep at all) and it will only make things slower (as mentioned: timings can be adjusted) when they were broken or only worked by luck.

Yes, it won't be the final form of the code, but it at least fixes the currently incorrect behavior.

PatrickHaecker:
Note, that Revise has the same problem. Shall we adapt the solution from there (50 ms and making it precompilable) and finally merge this?
The time until the release of Julia 1.13 might be limited and then the race condition will hit everyone who wants their REPL to only type the characters of the keys they actually press.

Issue #4637, opened 2026-03-15
Pkg "strict" mode for well-behaved `[compat]` + precompilation


topolarity:
Julia's stacked environments are very cool because they let you quickly jump between multiple active projects in the same session, but they have the downside that they interact quite poorly with Pkg.precompile and pkgimages in general since Pkg often cannot predict loading behavior with them.

I love the new `[workspaces]` feature, because it solves a ton of problems associated with this workflow:
  - developers can still quickly hop between projects without a restart
    (this is even faster than with stacked environments, since precompilation is shared!)
  - `[compat]` is not violated since the loaded environment actually went through a `] resolve`
  - `] precompile` works equally well for tests and the main project (and overall compiles _much_ less!)
  - `Pkg.test()` failures are easily reproduced and have a clear environment (no need for TestEnv.jl!)

but there are still a few gaps:
 1. Dynamic environment changes are still an issue. The warning improved in https://github.com/JuliaLang/julia/pull/61293 is nice, but the end-to-end Pkg workflow encourages you to do the "easy" thing and attempt a `using` only to immediately hit a ton of unexpected pre-compilation. The problem is that the environment was "torn" since it now has packages of a different version loaded which do not match your environment (or what will load next session). Since your environment is now "torn", this is completely one-off, never-to-be-reused-again work. Furthermore, since you are running in a potentially non-`[compat]` (and non-reproducible environment), so you may hit bugs and choose to back out anyway.
 2. Loading is still affected by `] activate` if you hop between stacked environments instead of between `[workspace]` projects. That leads to the same "torn environment" issues alluded to above when you intersperse `using` with `activate` changes. (i.e. moving between stacked environments is much worse than moving between workspace projects)
 3. Implicit environments such as [package directories](https://docs.julialang.org/en/v1/manual/code-loading/#Environments) still exist.

The problem is that these dynamic / implicit environment changes (1) lead to many "unique" environments that require one-off pre-compilation work and are often broken anyway due to `[compat]` issues, and (2) prevent `Pkg.precompile` from being able to predict which packages your project(s) will actually need / want to load at run time.

As a developer, it'd be useful to me to have a "strict" Pkg workflow which:
  - doesn't allow me to `] activate` anything that would "tear" my environment
  - after any package changes / upgrades that "tear" the environment, "disconnects" the updated Manifest and loads based on only the "old" environment
    (i.e. require a restart if you want to use the new packages, or keep running with the old one)
[...]

PR #4643, opened 2026-03-25
🤖 Prototype: add overlay project support

KristofferC:
Prototyped the overlay project feature I mentioned in https://github.com/JuliaLang/Pkg.jl/issues/4637#issuecomment-4070173063 with Claude.
This still needs code loading support to merge in the overlay project into the workspace.
The goal here is to avoid using the environment stack for dev tools (which has a seprate manifest) and instead always resolve the dev tools together with your active project.
This does not help "torn environments" if you load packages, do package operations and then load more packages. 
Also, it is a bit ugly that if you share a manifest with someone they get all these devtool packages in there, maybe there should be two manifests, one that is everything reachable from the actual active project and one that is the "orphan" dpeendencies assuming the overlay environment would be removed...

TODO:

- Code loading support
- Precompilation support

Usage:

```
julia> ENV["JULIA_PKG_OVERLAY"]
"@devtools"

(@devtools) [overlay] pkg> activate @devtools
  Activating project at `~/.julia/environments/devtools`

(@devtools) [overlay] pkg> st
Status `~/.julia/environments/devtools/Project.toml` (empty project)

(@devtools) [overlay] pkg> add BenchmarkTools, Revise
    Updating `~/.julia/environments/devtools/Project.toml`
  [6e4b80f9] + BenchmarkTools
  [295af30f] + Revise
    Updating `~/.julia/environments/devtools/Manifest.toml`
  [6e4b80f9] + BenchmarkTools
  [295af30f] + Revise

(@devtools) [overlay] pkg> activate --temp
  Activating new project at `/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/jl_sjMOea`

(jl_sjMOea) pkg> st --workspace
Status `/private/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/jl_sjMOea/Project.toml`
Status /Users/kc/.julia/environments/devtools/Project.toml
  [6e4b80f9] BenchmarkTools
  [295af30f] Revise

(jl_sjMOea) pkg> add Example
   Resolving package versions...
   Installed JuliaInterpreter ─ v0.10.11
   Installed Revise ─────────── v3.14.0
    Updating `/private/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/jl_sjMOea/Project.toml`
  [7876af07] + Example v0.5.5
    Updating `/private/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/jl_sjMOea/Manifest.toml`
  [7876af07] + Example v0.5.5

(jl_sjMOea) pkg> st -m
Status `/private/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/jl_sjMOea/Manifest.toml`
  [7876af07] Example v0.5.5

(jl_sjMOea) pkg> st -m --workspace
Status `/private/var/folders/2c/rk2fkhgn6c35lr6qp7cll4t00000gn/T/jl_sjMOea/Manifest.toml`
  [6e4b80f9] BenchmarkTools v1.6.3
  [da1fd8a2] CodeTracking v3.0.0
  [34da2185] Compat v4.18.1
  [807dbc54] Compiler v0.1.1
  [7876af07] Example v0.5.5
  [682c06a0] JSON v1.4.0
```

🤖 description:

-----------
Julia's stacked environments (LOAD_PATH / `@v#.#`) allow loading dev tools alongside project deps, but they use separate manifests. This can lead to incompatible versions being loaded, wasted precompilation, and [compat] violations. Workspaces solve this with a shared manifest, but require explicit [workspace] entries in every project.

An overlay project bridges this gap: a globally-configured Project.toml whose deps get merged into every active environment as if it were a workspace member. Dev tools like Debugger, JET, and Cthulhu can be listed once in the overlay and resolved together with whatever project you're working on — sharing a single manifest and respecting [compat].

Configuration (checked in order, first wins):
  - Pkg.OVERLAY_PROJECT[] = "~/.julia/overlay"
  - ENV["JULIA_PKG_OVERLAY"] = "@devtools"

Paths starting with @ are expanded via Base.load_path_expand, so "@devtools" resolves to ~/.julia/environments/devtools/Project.toml.

The implementation injects the overlay into env.workspace in the EnvCache constructor (~5 lines), which means all downstream code — dep loading, compat intersection, resolve hashing, precompilation — works automatically since it already iterates over workspace members.

When the overlay project itself is active (pkg> activate @devtools), only `add` and `rm` are supported — they write directly to the overlay's Project.toml without resolution or manifest updates. Other operations (develop, update, pin, free) error with a clear message. The REPL prompt shows [overlay] when the overlay project is active.

------------

### DIFF
--- a/docs/src/environments.md
+++ b/docs/src/environments.md
@@ -152,6 +152,27 @@ Shared environments can be activated with the `--shared` flag to `activate`:
 Shared environments have a `@` before their name in the Pkg REPL prompt.
 
 
+## Overlay Projects
+
+As an alternative to shared environments for development tools, you can use an
+[overlay project](@ref overlay-projects). An overlay project's dependencies are resolved
+together with whatever project you activate, sharing a single manifest. This avoids the version
+incompatibilities and redundant precompilation that can occur with stacked environments. See the
+[Overlay Projects](@ref overlay-projects) section in the `Project.toml` documentation for details.
+
+When using an overlay project, you typically want to disable the default stacked environment
+(`@v1.11`) since the overlay already provides your dev tools through the shared manifest.
+To do this, set `JULIA_LOAD_PATH` to only include the active project and stdlib:
+
+```sh
+# In your shell profile (e.g. .bashrc, .zshrc)
+export JULIA_LOAD_PATH="@:@stdlib"
+```
+
+The default value of `JULIA_LOAD_PATH` is `@:@v#.#:@stdlib`, where `@v#.#` is the versioned
+shared environment. Removing it prevents stacked environment packages from being loadable,
+which is what you want when the overlay project is the sole source of your extra dependencies.
+
 ## Environment Precompilation
 
 Before a package can be imported, Julia will "precompile" the source code into an intermediate more efficient cache on disc.
--- a/docs/src/managing-packages.md
+++ b/docs/src/managing-packages.md
@@ -553,6 +553,21 @@ To run a typical garbage collection with default arguments, simply use the `gc` 
 
 Note that only packages in `~/.julia/packages` are deleted.
 
+## Overlay Project
+
+The `JULIA_PKG_OVERLAY` environment variable specifies a path to an [overlay project](@ref overlay-projects) —
+a `Project.toml` whose dependencies are merged into every active environment's resolution. This
+is useful for globally available dev tools that should be resolved together with your project.
+The same can be set programmatically via `Pkg.OVERLAY_PROJECT[]`, which takes precedence over
+the environment variable.
+
+```julia
+ENV["JULIA_PKG_OVERLAY"] = expanduser("~/.julia/overlay")
+
+# Or use @ to reference a named environment in your depots
+ENV["JULIA_PKG_OVERLAY"] = "@devtools"  # expands to ~/.julia/environments/devtools
+```
+
 ## Offline Mode
 
 In offline mode, Pkg tries to do as much as possible without connecting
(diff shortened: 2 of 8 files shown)


vchuravy:
Neat! Would be good to add a test that preference merging works with an overlay project.

KristofferC:
> Neat! Would be good to add a test that preference merging works with an overlay project.

Yes, I forgot to mention that. I am actually not sure it works with workspaces?

KristofferC:
Things that can be discussed:

- How impactful is this in terms of #4637? It avoids having your debug tools load potentially invalid dependencies. It also hopefully avoids the situation where you start julia, load a debug tool, then swap to another environment.
- Is a global variable + env the best way to set an overlay project?
- Should there be support for multiple overlay projects? Implementation-wise, it is easy.

StefanKarpinski:
Very cool!

> It also hopefully avoids the situation where you start julia, load a debug tool, then swap to another environment.

I'm not clear how it avoids this. Can you explain?

> Is a global variable + env the best way to set an overlay project?

Feels like an extra layer of indirection. We could just pick `@overlay` or `@devtools` as the name.

KristofferC:
> Feels like an extra layer of indirection. We could just pick @overlay or @devtools as the name.

It feels to me like you would want this path to be customizable to easily be able to swap between different overlay projects (or to disable it) etc.

PR #4649, opened 2026-04-01
fix `]app up` no-op

KristianHolme:
### Fixes
  - **`Pkg.Apps.update()` with no arguments and `pkg> app up`:** both hit an empty argument list. The vector method now forwards `[]` to
  `update(nothing)` instead of doing zero iterations.
  - **Update all:** with `pkg === nothing`, the loop used `if pkg === nothing || info.name !== pkg.name; continue`, which skipped every package. The
  condition is now `pkg !== nothing && info.name !== pkg.name`.
  - **Source path after updating each app env:** use `source_path(manifest_file, manifest_entry)` where `manifest_file` is the per-app env manifest, so
  relative `path` entries (e.g. `"."`) resolve correctly.
  

 ( 🤖 Done with Cursor)

### DIFF
--- a/src/Apps/Apps.jl
+++ b/src/Apps/Apps.jl
@@ -276,6 +276,9 @@ end
 
 update(pkgs_or_apps::String) = update([pkgs_or_apps])
 function update(pkgs_or_apps::Vector)
+    if isempty(pkgs_or_apps)
+        return update(nothing)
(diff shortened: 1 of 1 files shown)


KristianHolme:
seems like this is fixed now, closing.

PR #4651, opened 2026-04-03
Speed up Pkg.status and Pkg.add

IanButterworth:
- Eliminate redundant EnvCache() construction in show_update by building the struct directly from existing fields instead of re-reading/parsing Project.toml and Manifest.toml from disk
- Add snapshot_originals kwarg to EnvCache to skip deepcopy of project/manifest for read-only sub-environments (collect_developed!)
- Pre-build reverse dependency map in print_status to avoid O(n²) manifest scanning in status_compat_info
- Use Dict-based UUID lookup in diff_array instead of O(n) findfirst

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -588,7 +588,7 @@ isfixed(pkg) = !is_tracking_registry(pkg) || pkg.pinned
 
 function collect_developed!(env::EnvCache, pkg::PackageSpec, developed::Vector{PackageSpec})
     source = source_path(env.manifest_file, pkg)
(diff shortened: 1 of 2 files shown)


KristofferC:
> Eliminate redundant EnvCache() construction in show_update by building the struct directly from existing fields instead of re-reading/parsing Project.toml and Manifest.toml from disk

These should not have to btouch disk because we have a cache for toml files so if the same file is read twice and it is not changed it should give the cached dictionary. Still better not to have to do it though.

Any perf numbers?

IanButterworth:
Opened too eagerly. Basically noise

PR #4658, opened 2026-04-12
Fix workspace test path: use test env Context for instantiate and precompile

IanButterworth:
Previously, `Pkg.precompile()` was called without a Context, so it used the currently active (parent) project instead of the test project. The `instantiate` call also created a throwaway Context each time.

Create a shared `test_ctx` pointing at the test env and pass it to both `instantiate` and `precompile` so they operate on the correct project.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -3072,8 +3072,10 @@ function test(
         if testdir(source_path) in dirname.(keys(ctx.env.workspace))
             proj = Base.locate_project_file(abspath(testdir(source_path)))
             env = EnvCache(proj)
-            # Instantiate test env
-            Pkg.instantiate(Context(env = env); allow_autoprecomp = false)
+            # Use a Context pointing at the test env so that instantiate and
+            # precompile operate on the test project rather than the parent.
+            test_ctx = Context(env = env; io = ctx.io)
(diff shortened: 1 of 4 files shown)


src/Operations.jl
```diff
@@ -3072,16 +3072,18 @@ function test(
...
+            Pkg.instantiate(test_ctx; allow_autoprecomp = false)
             status(env, ctx.registries; mode = PKGMODE_COMBINED, io = ctx.io, ignore_indent = false, show_usagetips = false)
             flags = gen_subprocess_flags(source_path; coverage, julia_args)
 
             if should_autoprecompile()
                 cacheflags = parse(CacheFlags, read(`$(Base.julia_cmd()) $(flags) --eval 'show(Base.CacheFlags())'`, String))
                 # Don't warn about already loaded packages, since we are going to run tests in a new
                 # subprocess anyway.
-                Pkg.precompile(; io = ctx.io, warn_loaded = false, configs = flags => cacheflags)
+                Pkg.precompile(test_ctx; io = ctx.io, warn_loaded = false, configs = flags => cacheflags)
```
KristofferC, line 3086: 👍

KristofferC (dismissed):

PR #4660, opened 2026-04-16
Add `maxfailures` option to `Pkg.test`

VanitasCodes:
This is the Pkg side of the `maxfailures` feature. The Test stdlib side lives in JuliaLang/julia#61560 and needs to be merged first for this to work end to end.

**What it does**

Adds a `maxfailures` keyword argument to `Pkg.test`. When not set the behavior is completely unchanged: all tests run to completion. When set to a non-negative integer `n`, the test run stops after `n` cumulative failures or errors:

- `maxfailures=0` stops on the first failure (zero failures tolerated)
- `maxfailures=2` stops after the second failure

**How it works**

Rather than injecting code into the test subprocess, we set `JULIA_TEST_MAXFAILURES` in the parent environment before launching the subprocess. The subprocess inherits the variable, and the Test stdlib reads it via `OncePerProcess` on first use, matching the pattern used by `JULIA_TEST_FAILFAST`. This keeps all internal functions (`gen_test_code`, `run_test_subprocess`, `run_test_subprocess_in_env`, `run_sandboxed_tests!`) completely unchanged.

Also available from the Pkg REPL:

```
pkg> test --maxfailures=n
```

Validation happens before any subprocess is launched, so a negative value throws a `PkgError` immediately.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -2675,21 +2675,10 @@ function free(ctx::Context, pkgs::Vector{PackageSpec}; err_if_free = true)
     end
 end
 
-function gen_test_code(source_path::String; test_args::Cmd, maxfailures::Int = 0)
+function gen_test_code(source_path::String; test_args::Cmd)
     test_file = testfile(source_path)
-    maxfailures_setup = if maxfailures > 0
-        """
-        let _Test = Base.require(Base.PkgId(Base.UUID("8dfed614-e22c-358e-b65c-d98082ae9e1e"), "Test"))
-            _Test.global_failure_count[] = 0
-            _Test.global_failure_limit = $maxfailures
-        end
-        """
-    else
-        ""
-    end
     return """
     $(Base.load_path_setup_code(false))
-    $maxfailures_setup
     cd($(repr(dirname(test_file))))
     append!(empty!(ARGS), $(repr(test_args.exec)))
     include($(repr(test_file)))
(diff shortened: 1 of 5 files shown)


VanitasCodes:
@oscardssmith Ready for review.

test/pkg.jl
```diff
@@ -252,6 +252,57 @@ temp_pkg_dir() do project_path
...
+            iob = IOBuffer()
+            try; Pkg.test("MaxFailuresPkg"; io = iob); catch; end
+            out = String(take!(iob))
+            @test occursin("D", out) && occursin("E", out)
+
+            # maxfailures=0 means no limit (same as default).
+            iob = IOBuffer()
+            try; Pkg.test("MaxFailuresPkg"; maxfailures = 0, io = iob); catch; end
+            out = String(take!(iob))
+            @test occursin("D", out) && occursin("E", out)
```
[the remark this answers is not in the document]
VanitasCodes: We could fix this by having the testsets print unique markers like `RUNNING_D` and `RUNNING_E` and check for those instead. @oscardssmith does that approach sound good to you or do you have a better suggestion?

test/pkg.jl
```diff
@@ -252,6 +252,57 @@ temp_pkg_dir() do project_path
...
+
+            # maxfailures=2 should stop after the 2nd failure.
+            iob = IOBuffer()
+            try; Pkg.test("MaxFailuresPkg"; maxfailures = 2, io = iob); catch; end
+            out = String(take!(iob))
+            @test occursin("Max failures reached", out)
```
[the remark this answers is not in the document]
VanitasCodes: The coupling here is intentional. "Max failures reached" is exactly what `maxfailures_print` outputs in the Test stdlib so if that message ever changes the test should fail because the behavior changed too.

src/REPLMode/command_declarations.jl
```diff
@@ -10,17 +10,20 @@ compound_declarations = [
...
-                    test [--coverage] [pkg[=uuid]] ...
+                    test [--coverage] [--maxfailures=n] [pkg[=uuid]] ...
 
                 Run the tests for package `pkg`, or for the current project (which thus needs to be
                 a package) if `pkg` is omitted.  This is done by running the file `test/runtests.jl`
                 in the package directory. The option `--coverage` can be used to run the tests with
                 coverage enabled. The `startup.jl` file is disabled during testing unless
-                julia is started with `--startup-file=yes`.
+                julia is started with `--startup-file=yes`. The option `--maxfailures=n` stops the
+                test run after `n` cumulative failures and errors. When `n` is 0 (the default)
```
oscardssmith: when n=0 we should fail after the first failure

src/Operations.jl
```diff
@@ -2976,24 +2987,25 @@ end
 testdir(source_path::String) = joinpath(source_path, "test")
 testfile(source_path::String) = joinpath(testdir(source_path), "runtests.jl")
 
-function run_test_subprocess(io::IO, flags::Cmd, source_path::String, test_args::Cmd; with_threads::Bool)
-    code = gen_test_code(source_path; test_args)
+function run_test_subprocess(io::IO, flags::Cmd, source_path::String, test_args::Cmd; with_threads::Bool, maxfailures::Int = 0)
```
oscardssmith: I don't think anything in this file should take a kwarg.

src/Operations.jl
```diff
@@ -2675,10 +2675,21 @@ function free(ctx::Context, pkgs::Vector{PackageSpec}; err_if_free = true)
...
+    maxfailures_setup = if maxfailures > 0
+        """
+        let _Test = Base.require(Base.PkgId(Base.UUID("8dfed614-e22c-358e-b65c-d98082ae9e1e"), "Test"))
+            _Test.global_failure_count[] = 0
+            _Test.global_failure_limit = $maxfailures
+        end
+        """
+    else
+        ""
+    end
```
oscardssmith: this seems pretty ugly.

VanitasCodes: Would you prefer we use an env var like `JULIA_TEST_MAXFAILURES` instead of code injection, similar to how `JULIA_TEST_FAILFAST` works?

VanitasCodes:
@oscardssmith Ready for review. I believe I have addressed most of the concerns.

VanitasCodes:
The PR description was stale, it still described the original code injection approach. Updated it to reflect what the code actually does now. Copilot was reviewing the description against the code and finding mismatches that don't exist in the code itself.

VanitasCodes:
@oscardssmith I was wondering if you have had time to review this one yet.

PR #4666, opened 2026-04-25
Trying out adding a short summary to resolver errors

IanButterworth:
Just a sketch of trying out whether a helpful summary could be added. (also extending the colors because collisions are common)

Screenshot 2026-04-25 at 12 26 20 AM

-

Screenshot 2026-04-25 at 12 26 54 AM

-

Screenshot 2026-04-25 at 12 27 01 AM

-

Screenshot 2026-04-25 at 12 27 09 AM

### DIFF
--- a/src/Resolve/graphtype.jl
+++ b/src/Resolve/graphtype.jl
@@ -619,9 +619,21 @@ function wipe_snapshots!(graph::Graph)
     return graph
 end
 
-# system colors excluding whites/greys/blacks and error-red
-const CONFLICT_COLORS = [1:6; 10:14];
-pkgID_color(pkgID) = CONFLICT_COLORS[mod1(hash(pkgID), end)]
+# 256-color palette excluding whites/greys/blacks and error-red.
+const CONFLICT_COLORS = [
+    26, 28, 31, 32, 33, 34, 35, 36, 37, 38,
+    39, 40, 41, 42, 43, 44, 45, 49, 50, 51,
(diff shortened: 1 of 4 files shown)


src/Resolve/graphtype.jl
```diff
@@ -619,9 +619,21 @@ function wipe_snapshots!(graph::Graph)
...
-pkgID_color(pkgID) = CONFLICT_COLORS[mod1(hash(pkgID), end)]
+# 256-color palette excluding whites/greys/blacks and error-red.
+const CONFLICT_COLORS = [
+    26, 28, 31, 32, 33, 34, 35, 36, 37, 38,
+    39, 40, 41, 42, 43, 44, 45, 49, 50, 51,
+    62, 63, 69, 70, 75, 76, 81, 82, 83, 87,
+    99, 105, 111, 112, 117, 118, 123, 129, 135,
+    141, 147, 153, 159, 165, 171, 177, 183, 189,
+    195, 201, 207, 213, 219,
+]
```
oxinabox, line 630: The original color list i wanted to use was:
`[1:5; 10:14; 21:51; 55:119; 124:142; 160:184; 196:220])` which I did some experimenting with to end up with (but it was so long ago i don't really remember)

@KristofferC (iirc) argued down to just using the 16 system colors.
Because those are the ones that are theme-able in most terminals. So should be set to be readable colors in any reasonable terminal theme. (and if not its not our fault)
where as the 256-colors are defined absolutely IIRC

PR #4678, opened 2026-05-07
instantiate: add `update_on_mismatch` kwarg to fall back to `update` on a stale manifest

jkrumbiegel:
There's currently no Pkg API for the workflow "use the existing manifest if it works, otherwise regenerate it."
Each of the three closest options covers something else:

- `Pkg.instantiate` errors or warns when the manifest can't be used as-is.
- `Pkg.resolve` doesn't install.
- `Pkg.update` always regenerates, even when the manifest was fine.

For tooling and helper environments (formatters, doc-render scripts, language-server bootstraps) what I actually
want is the first option. I don't care about the exact dependency versions; I want the env to keep working across
teammates editing `Project.toml` and Julia minor bumps.

One simple example where I hit this is Makie's [`tooling/formatter/format.jl`](https://github.com/MakieOrg/Makie.jl/blob/master/tooling/formatter/format.jl)): I want to be able to run
`julia ./tooling/formatter/format.jl` and have the formatter just work, no matter which Julia version I happen to
have active that day, no matter whether the formatter's `Project.toml` has been bumped since I last ran it. The
local `Manifest.toml` typically sticks around across git switches because it's gitignored, so any change someone
made on another branch (a new dep, a tightened compat) leaves my local manifest stale rather than getting reset.
Today that script is a plain `Pkg.instantiate()` and breaks the moment the existing manifest is no longer usable.

Just calling `Pkg.update()` instead isn't the answer either: every run may do a registry update (which by itself
takes a noticeable moment), and any time a dep version actually moves it triggers precompilation. Dep versions
moving isn't constant but it happens often enough that, over the lifetime of running the script, the cost adds
up. The whole point of the script is to be quick to invoke.

In Quarto's Julia engine I ended up [reimplementing this kind of logic by hand](https://github.com/PumasAI/quarto-julia-engine/blob/main/_extensions/julia-engine/ensure_environment.jl)),
inspecting `is_manifest_current` and the manifest's recorded `julia_version` and switching between `instantiate`
and `update` myself.

## What this adds

A new `update_on_mismatch::Bool = false` keyword argument to `Pkg.instantiate`, with a corresponding REPL flag
`-u` / `--update_on_mismatch`. When set, `instantiate` falls back to `Pkg.update()` whenever the manifest cannot
be used as-is, instead of warning or erroring. The default (`false`) preserves the existing behavior.

```julia
Pkg.instantiate(update_on_mismatch = true)
```

```
pkg> instantiate -u
```

The fallback fires when any of the following holds (`Operations.manifest_is_mismatched`):

- the manifest is in v1 format (predates the `julia_version` entry, so Julia compatibility can't be verified);
- the manifest's recorded `julia_version` is missing or differs from the current Julia at minor-version granularity;
- `is_manifest_current(env) === false` (project hash recorded in the manifest doesn't match the current project);
- a direct dep in `Project.toml` has no corresponding entry in the manifest. This is caught explicitly because v2.0
  and v2.1 manifests pre-date the `project_hash` entry (see `test/manifest/formats/v2.0/Manifest.toml` and
  `v2.1/Manifest.toml`, neither of which contains it), so `is_manifest_current` returns `nothing` for those and
  can't catch project drift on its own.

These mirror the conditions that the existing `instantiate` flow currently warns or errors about. With the flag,
each branch point falls back to `up(ctx)` instead.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -1339,7 +1339,7 @@ function instantiate(
             saved_initial_snapshot[] = true
         end
         printpkgstyle(ctx.io, :Update, "manifest does not match project or Julia version, falling back to `Pkg.update()`", color = Base.info_color())
-        up(ctx; update_registry)
+        up(ctx; update_registry, mode = workspace ? PKGMODE_MANIFEST : PKGMODE_PROJECT)
         allow_autoprecomp && Pkg._auto_precompile(ctx, already_instantiated = true)
         return
     end
--- a/src/API.jl
+++ b/src/API.jl
@@ -1293,7 +1293,8 @@ function instantiate(
         ctx::Context; manifest::Union{Bool, Nothing} = nothing,
         update_registry::Bool = true, verbose::Bool = false,
         platform::AbstractPlatform = HostPlatform(), allow_build::Bool = true, allow_autoprecomp::Bool = true,
-        workspace::Bool = false, julia_version_strict::Bool = false, kwargs...
+        workspace::Bool = false, julia_version_strict::Bool = false,
+        update_on_mismatch::Bool = false, kwargs...
     )
     Context!(ctx; kwargs...)
     if Registry.download_default_registries(ctx.io)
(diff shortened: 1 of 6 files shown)


src/API.jl
```diff
@@ -1324,6 +1327,23 @@ function instantiate(
...
+        # for `Pkg.undo` to revert to (the convenience macro at the top of this file
+        # does this for other mutating ops, but `instantiate` is not in its list).
+        # Call with no args so the snapshot reads fresh state from disk; passing
+        # `ctx.env` would share the manifest reference that `up` mutates in place.
+        if !saved_initial_snapshot[]
+            add_snapshot_to_undo()
+            saved_initial_snapshot[] = true
+        end
+        printpkgstyle(ctx.io, :Update, "manifest does not match project or Julia version, falling back to `Pkg.update()`", color = Base.info_color())
+        up(ctx; update_registry)
```
IanButterworth: `workspace` needs to be forwarded here.
```suggestion
        up(ctx; update_registry, mode = workspace ? PKGMODE_MANIFEST : PKGMODE_PROJECT)
```

IanButterworth (dismissed):

IanButterworth (approved):

PR #4679, opened 2026-05-08
activate: warn if different versions are loaded

IanButterworth:
Screenshot 2026-05-08 at 11 58 39 AM


IanButterworth:
This might need some polish, but I think it's worth trying out on master.

PR #4689, opened 2026-06-08
PlatformEngines: protect auth token refresh download

Keno:
Use ArgTools.FileSpec for the package-server auth refresh download so the temporary token response is created with mode 0o600 before parsing it. This avoids exposing the refreshed access token through default temporary-file permissions while preserving ordinary AbstractString download behavior elsewhere.

The refresh response carries a short-lived package-server access token used as the bearer credential for authenticated package-server requests. The refresh token remains in the existing auth.toml and is sent only as the Authorization header to the refresh URL.

This issue was assigned ID JLSEC-2026-610.

This issue was reported by the Anthropic CVD program through the following IDs:
- ANT-2026-7HMYMN3Y
- ANT-2026-EJYX8E8B

Depends on https://github.com/JuliaIO/ArgTools.jl/commit/f77276a25334585e7adce14491f281404cdc23d7.


src/PlatformEngines.jl
```diff
@@ -189,7 +190,7 @@ function get_auth_header(url::AbstractString; verbose::Bool = false)
     refresh_token = auth_info["refresh_token"]::String
     refresh_auth = "Authorization" => "Bearer $refresh_token"
     try
-        download(refresh_url, tmp, auth_header = refresh_auth, verbose = verbose)
+        download(refresh_url, FileSpec(tmp; mode = 0o600), auth_header = refresh_auth, verbose = verbose)
```
StefanKarpinski, line 193: Should this just be the default? I guess we respect the user's umask, which should be secure but may not be...

Keno: I think for general package downloads, it makes sense to respect the umask. That allows downloaded packages to be viewed by others in the group and if that's undesirable, the user should adjust the umask. Just the auth token is more sensitive and we should protect it more, like SSH private keys.

DilumAluthge:
Just a note to myself:
- [x] Wait for https://github.com/JuliaLang/julia/pull/62041 to get merged
- [x] Wait for the nightly to be updated
- [x] Re-run CI here

src/PlatformEngines.jl
Keno, line 206: @StefanKarpinski Should we be deleting `tmp` on the success path here, rather than only on the failure paths? It seems like this leaks auth tokens in tmp files?

DilumAluthge: It sounds like we should put the `rm(tmp; force = true)` inside a `finally` block, so that it gets done no matter what?

DilumAluthge: That being said, it looks like `tmp` comes from `tmp = tempname()`, and the default behavior of `tempname()` is that we "delete the returned path automatically when the [Julia] process exits".

src/PlatformEngines.jl
DilumAluthge: I don't feel too strongly either way - it seems that the important part is for the auth token to be `chmod 600`. Can we stick with the current approach in this PR for now, in the interest getting this PR merged quickly?

DilumAluthge:
@Keno Could you run the Runic code formatter on this PR?

Keno:
CI failure is a bizarre interaction with JuliaSyntaxHighlighting and will be fixed once https://github.com/JuliaLang/julia/commit/8779b2b225dc5c9a11dfb05f8fcc9f8960a06441 gets promoted to nightly.

DilumAluthge:
CI is all green here.

@Keno What's the status here? Are there any remaining blockers?

Keno:
Good to go from my perspective

DilumAluthge:
Let's merge this.

@StefanKarpinski Let us know if there are any additional changes you want, and we can make a follow-up PR.

PR #4696, opened 2026-06-16
respect compat on stdlibs

KristofferC:
this is relevant if you make your own julia distro with custom stdlibs


KristofferC:
https://github.com/JuliaLang/Pkg.jl/blob/ab06f4b740695cbe207e6ff7b8eebed5e9d55995/src/Resolve/graphtype.jl#L308-L316 is a bit of an issue

PR #4698, opened 2026-06-20
Support multiple projects to `instantiate` at once

PatrickHaecker:
Support to `instantiate` several environments in one call by passing (possibly) multiple paths (`Pkg.instantiate("env1", "env2")`, or `pkg> instantiate env1 env2`). The registries are read from disk once and shared across all environments, and the active environment is left unchanged.

This is more intuitive, more convenient and faster.

Claude Opus 4.8 supported in creating the PR.


PatrickHaecker:
Any chance to backport to Julia 1.13? This seems like a very obvious lack of functionality a user would expect to have available when working with monorepos.

KristofferC:
If this is only a wrapper of looping over environments and doing a serial `Pkg.instantiate` it doesn't seem very worthwhile to me. If this is to be actually useful I think it should gather up information from all environments and then doing something like a single "mega" instantiate over all does environments in one pass.

PR #4699, opened 2026-06-21
Fix workspace instantiate without existing manifest

PatrickHaecker:
In an environment which is part of a workspace where the workspace lacks the manifest, all packages in the manifest were instantiated when only the packages of the environment should have been instantiated.

Claude Opus 4.8 supported in creating the PR.


PatrickHaecker:
@IanButterworth: May I kindly request your review here? You seem to have been the most active person in the relevant surrounding code. In my opinion this bug should be fixed at least in Julia 1.13 and 1.14.

PR #4706, opened 2026-06-24
Support running tests in parallel

PatrickHaecker:
Given [Julia has a jobserver](https://github.com/JuliaLang/julia/commit/e319a303a7401f0151246323b53c03f9255b8586) now, we can finally support parallel testing without starting too many processes during precompilation. This does not help avoiding oversubscription during running the tests, but it is at least a start worth making use of, given that there are a lot of single-threaded runtests.jl.

With this change, you can start your tests with
```
Pkg.test(pkgnames; ntasks = Sys.CPU_THREADS)
```
to test multiple packages in parallel.

I used Claude Opus 4.8 while generating this PR.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -568,6 +568,7 @@ function test(
         force_latest_compatible_version::Bool = false,
         allow_earlier_backwards_compatible_versions::Bool = true,
         allow_reresolve::Bool = true,
+        ntasks::Int = 1,
         kwargs...
     )
     julia_args = Cmd(julia_args)
@@ -594,6 +595,7 @@ function test(
         force_latest_compatible_version,
         allow_earlier_backwards_compatible_versions,
         allow_reresolve,
+        ntasks,
     )
     return
 end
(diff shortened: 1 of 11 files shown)


KristofferC:
I am personally not sure if the complexity of this carries its weight here. Does it have a big advantage over just spawning multiple julias? Also, tests themselves might run in parallel and then you kind of have an issue. In fact, the whole multiple test target API for Pkg is not that useful I feel, it probably got added because the previous Pkg supported it.

PatrickHaecker:
Sorry, for not being able to keep up with all PRs, but time is limited and I don't want to multi-task too much.

The big advantage is efficiency:
- When you have a lot of cores, but not a lot of RAM, starting a process per core makes the system swap and often ends up in OOM
- Each test process does the precompilation from scratch, which is really annoying if tests share a lot of `CodeInstance`s. Just adding precompilation is typically not the solution, because it's kind of unrealistic during development and it does not help, because that also happens serialized before tests start

So it does not need to be this implementation, but I think people nowadays expect fast tests and currently Julia with Pkg.jl does really not shine there. There are reasons in Julia why this is difficult, but I think it's even more important that things like multi-threading are used.

Yes, there are packages, but a good solution in Pkg.jl seems to be the cleanest solution.

PR #4707, opened 2026-06-27
GitTools: Correct incorrect hash computation.

Keno:
`GitTools.blob_hash` computed the git blob header length for a symlink with `length(readlink(path))` (the number of characters) rather than the number of bytes. For symlink targets containing non-ASCII (multi-byte UTF-8) characters this produced a git object hash that disagrees with canonical git (and with `Tar.tree_hash`), so the resulting tree hash was wrong.

This is a bit of a mess because existing Artifacts.toml files embed hashes computed with the incorrect algorithm. Thus, for compatibility:

1. Accept both spellings when validating artifacts.
2. Provide a kwarg in create_artifact to select whether to create hashes with the wrong algorithm. This defaults to the correct algorithm, which would cause so-created Artifacts.toml files to fail to load on older julia. My preference would be to keep that default and backport this fix to all supported release branches in short order, with the explicit kwarg there if people need it. It is somewhat rare to have non-ASCII symlinks in artifacts, but e.g. BinaryBuilder does have such artifacts.


Keno:
@KristofferC @StefanKarpinski ping

KristofferC:
The implementation is surely good but I feel there is just a lot of "noise" in docstrings, etc about this little bugfix. It's in like 3-4 docstrings and comments. I realize that there are hashes that are computed with the old one (how many of them are there though?) but the noise to signal still feels high.

Keno:
Sure, we can drop docstring - the question is more is this the strategy that we want or do we want something else.

KristofferC:
Correcting the bug + verify with old + new seems reasonable to me at least. But yeah, I guess we need quick backports to not cause new artifacts with the fix to fail loudly?

PR #4709, opened 2026-06-28
Assert on conflicting sources in workspaces

PatrickHaecker:
This implements the former TODO to assert that there are no conflicting sources within a workspace. Before this PR you could define different kind of conflicts within a workspace, e.g. one project depending on branch `a` of package `P` and another project depending on branch `b` of package `P`. `Pkg` then selected either `a` or `b` kind of randomly violating one of the constraints without noticing the user.

The user gets an error message in case of conflict where the conflicting sources are listed.

Claude Opus 4.8 wrote the initial version of the patch and the test case.

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -164,6 +164,38 @@ end
 merge_pkg_source!(target::PackageSpec, source::PackageSpec) =
     merge_pkg_source!(target, source.path, source.repo)
 
+# In a workspace the same package can be a direct dependency of several projects,
+# each of which may point it at a source via `[sources]`. Merging those entries
+# only makes sense if they agree; otherwise the resulting source would depend on
+# the order the projects happen to be iterated in, so reject the configuration.
+function assert_no_conflicting_sources(pkgs)
+    paths = Set{String}()
+    repo_sources = Set{String}()
+    revs = Set{String}()
+    subdirs = Set{String}()
+    for pkg in pkgs
+        isnothing(pkg.path) || push!(paths, pkg.path)
+        isnothing(pkg.repo.source) || push!(repo_sources, pkg.repo.source)
+        isnothing(pkg.repo.rev) || push!(revs, pkg.repo.rev)
+        isnothing(pkg.repo.subdir) || push!(subdirs, pkg.repo.subdir)
+    end
+    conflicts = String[]
+    if !isempty(paths) && (!isempty(repo_sources) || !isempty(revs))
+        push!(conflicts, "both a path and a repository")
+    end
+    showvals(vals) = join((repr(v) for v in sort!(collect(vals))), ", ")
+    length(paths) > 1 && push!(conflicts, "paths $(showvals(paths))")
+    length(repo_sources) > 1 && push!(conflicts, "repositories $(showvals(repo_sources))")
+    length(revs) > 1 && push!(conflicts, "revisions $(showvals(revs))")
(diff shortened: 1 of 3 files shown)


PatrickHaecker:
@KristofferC : May I kindly request your review here? You added the TODO 2 years ago and therefore can probably judge best whether the fix is what you had envisioned back then.

In my opinion this bug should be fixed at least in Julia 1.13 and 1.14. We lost quite some time debugging it, because when you start the debugging with the mental model that using Julia with Pkg will end up fully reproducible, you exclude the very possibility of this bug.

KristofferC:
I haven't had time to read this properly myself, but here are some notes you might want to think about from the big brother of Opus (Fable). If possible, I suggest you make PRs using Fable because Opus really produces bad code, unfortunately. Also, having gpt review the changes by Fable as a last step usually helps catch some stuff.

----

**1. False positive from stale manifests — blocks the resolve that would fix it (main issue).** The specs passed to `assert_no_conflicting_sources` aren't purely project-declared sources: in `load_project_deps`, a member with no `[sources]` entry inherits `entry.path` / `entry.repo` from the manifest (src/Operations.jl:247-248). So with members A and B both depending on `SharedDep` where only A has a source: after one resolve, editing A's source path makes the next `Pkg.resolve()` fail with "conflicting sources specified by different projects in the workspace" — listing the old manifest path that no project specifies. I verified this against the branch:

```
first resolve: OK
second resolve threw: Package `SharedDep [22222222]` has conflicting sources
specified by different projects in the workspace: paths "variant1/SharedDep",
"variant2/SharedDep".
```

The error is both misleading (only one project declares a source) and self-perpetuating: every subsequent `resolve`/`up`/`add` hits the same check, so the user has to delete the manifest to escape. The check should compare only sources actually declared in project files — e.g. run it on the raw `get_path_repo` results before the manifest fallback is applied, or record on the spec whether its source came from a project.

**2. `merge_pkg_source!` still drops `subdir`.** The check accepts "one member specifies a subdir, the other doesn't" (one-element set), but `merge_pkg_source!` (src/Operations.jl:153-163) merges `path`, `repo.source`, and `repo.rev` — never `repo.subdir`. If the kept spec is the one without the subdir, it's silently lost. Pre-existing, but this PR is explicitly about making the merge sound, so adding `subdir` to the merge would complete it.

**3. "Both a path and a repository" can fire with no repository.** The condition includes `!isempty(revs)` (src/Operations.jl:183), so a single member writing `{path = "..", rev = "x"}` (which `get_path_repo` doesn't reject) reports "both a path and a repository" across the workspace even when no url exists anywhere. Cosmetic, but the message would be confusing.

**4. Absolute paths skip normalization.** `project_path_to_manifest_path` returns absolute paths as-is (src/Types.jl:475), so an absolute source in one member and an equivalent relative one in another compare unequal and get reported as conflicting paths. Minor edge case.

**5. The error now fires from read-only paths too.** `diff_array` calls `load_direct_deps` for `Pkg.status(workspace = true)` (src/Operations.jl:3443), so on a conflicted workspace `status` throws instead of showing state. Arguably fine for an invalid configuration, but worth a conscious decision.

Smaller notes: the test only exercises the path-conflict arm — a repo/rev-conflict case would cover the other branches, and `@test_throws` with a match pattern would be tidier than the manual try/catch. And `assert_no_conflicting_sources` could early-return when `length(pkgs) == 1` to skip four `Set` allocations per UUID in the common no-duplicate case.

PR #4719, opened 2026-07-07
Make `dependencies` and `project` workspace-aware with a flag

PatrickHaecker:
This seems to be a logical extension of available functionality to better support workspaces. This would also allow [reducing the private API surface of `PkgToSoftwareBOM.jl` significantly](https://github.com/SamuraiAku/PkgToSoftwareBOM.jl/pull/48).

Claude Opus 4.8 helped in creating this PR.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,11 @@
 Pkg v1.14 Release Notes
 =======================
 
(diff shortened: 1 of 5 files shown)


KristofferC (dismissed):
Seems reasonable to me

PatrickHaecker:
@KristofferC: Thanks!

I think the code coverage check is a false positive, but I just rebased to get a second run.

So is this ready to be merged?

PR #4720, opened 2026-07-07
Remove accidentally-quadratic scans in Pkg operations

PatrickHaecker:
There seem to be quite some operations with quadratic complexity which should not have quadratic complexity.  In a typical environment this might not be too noticeable. Nevertheless, the needed patterns are already used in some places, so it even increases consistency to cover all.

This made heavy use of Claude Opus 4.8, but I reviewed everything, did some changes and I think this now looks correct.

Claude's summary:

Several core routines used a `Vector{PackageSpec}` (or the manifest) as a lookup
table, so a `findfirst`/`findall`/`any`/`∩` scan ran inside a loop that itself
scales with the number of packages — O(n²) on `add`/`up`/`rm`/`resolve`/`status`
etc. Each is replaced with a one-pass `Set{UUID}`/`Dict{UUID,…}` index, or (for
`rm`/`status`) a reverse-dependency map. Behavior is unchanged (first-occurrence
order and source merging preserved); only the lookup complexity drops from O(n)
to O(1).

## Fixes

- **`load_direct_deps`** — deduped/merged direct deps via `findall` per unique
  UUID → single pass keyed by a `Dict{UUID,Int}`.
- **`load_project_deps` / `load_manifest_deps`** — skipped already-present packages
  with `findfirst` over a growing vector → `Set{UUID}` membership.
- **`resolve_versions!`** — `findfirst` per resolved version and `any` per
  developed/fixed package → `Dict{UUID,Int}` index + maintained `Set{UUID}`.
- **`collect_developed!`** — `any` over the `developed` vector per candidate →
  threaded `Set{UUID}`.
- **`rm`** — reverse-dependency closure rescanned the whole manifest each pass and
  allocated a set via `drop ∩ deps` per entry → reverse-dependency map + worklist
  BFS.
- **`status_compat_info`** — scanned the whole manifest for dependents, once per
  package → reverse-dependency map (`manifest_dependents_map`) built once in
  `print_status` and threaded in.
- **`diff_array`** — `findfirst` per UUID over old/new lists → index each list into
  a `Dict` once.
- **`append_all_pkgs!`** (API) — `findfirst` per workspace dependency → maintained
  `Dict{UUID,Int}`.

## Benchmarks

Each hot spot implemented both ways (old = master, new = this PR) on identical
manifest-shaped data (UUID keys, avg ~5 deps/pkg), reporting the minimum of 50
runs. Speedups (old/new):

| hot spot | N=100 | N=300 | N=900 | N=3000 |
|---|--:|--:|--:|--:|
| `load_manifest_deps` dedup | 1.1× | 3.4× | 9.0× | 22× |
| `load_direct_deps` merge | 3.9× | 10× | 25× | 72× |
| `diff_array` index | 1.6× | 3.5× | 13× | 41× |
| `status` dependents scan | 3.8× | 12× | 25× | 85× |
| `rm` reverse-dep closure | 52× | 124× | 251× | 616× |

Absolute times for the two biggest (old → new):

| hot spot | N=300 | N=900 | N=3000 |
|---|--:|--:|--:|
| `status` dependents scan | 0.41 → 0.03 ms | 3.79 → 0.15 ms | 63 → 0.74 ms |
| `rm` reverse-dep closure | 7.87 → 0.05 ms | 68.2 → 0.26 ms | 780 → 1.6 ms |

`rm` dominates because its old inner loop allocated a fresh set per manifest entry
per fixpoint pass.

## Testing

No behavior change intended. Validated end-to-end in an isolated depot exercising
`add` (resolve), `status`/`status --outdated`, `develop` (→ `collect_developed!`),
`update`, `rm` (reverse-dep closure correctly prunes orphaned transitive deps),
and `resolve`.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -422,15 +422,17 @@ function append_all_pkgs!(pkgs, ctx, mode; workspace::Bool = false)
             push!(pkgs, PackageSpec(name = name, uuid = uuid, path = path, repo = repo))
         end
         if workspace
+            uuid_to_idx = Dict{UUID, Int}(p.uuid => i for (i, p) in pairs(pkgs) if !isnothing(p.uuid))
             for (project_file, project) in ctx.env.workspace
                 for (name::String, uuid::UUID) in project.deps
                     path, repo = get_path_repo(project, project_file, ctx.env.manifest_file, name)
-                    existing = findfirst(p -> p.uuid == uuid, pkgs)
+                    existing = get(uuid_to_idx, uuid, nothing)
                     if existing !== nothing
                         Operations.merge_pkg_source!(pkgs[existing], path, repo)
                         continue
                     end
                     push!(pkgs, PackageSpec(name = name, uuid = uuid, path = path, repo = repo))
+                    uuid_to_idx[uuid] = length(pkgs)
                 end
             end
         end
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -174,21 +174,23 @@ function load_direct_deps(
         append!(pkgs_direct, load_project_deps(project, path, env.manifest, env.manifest_file, pkgs; preserve))
     end
 
-    unique_uuids = Set{UUID}(pkg.uuid for pkg in pkgs_direct)
-    for uuid in unique_uuids
-        idxs = findall(pkg -> pkg.uuid == uuid, pkgs_direct)
-        # TODO: Assert that projects do not have conflicting sources
-        pkg = pkgs_direct[idxs[1]]
-        idx_to_drop = Int[]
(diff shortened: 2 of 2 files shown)


KristofferC:
This hasn't really been a problem in practice (as you can see, for reasonable environments the absolute numbers are kind of low). But fixing is good :)

PatrickHaecker:
Yes, I fully agree. I was kind of relieved to see that it's not really a problem in reality. But I've observed quite some performance problems with larger environments recently (although around the VS Code plugin) and as I was active in Pkg.jl recently, I wanted to check how it copes. Overall I'd say it's a low-hanging, but also (luckily) rather small fruit. :-)

PR #4722, opened 2026-07-07
Add HyperLogLog Over RSA anonymous client-counting header

StefanKarpinski:
## What

Attach a `Julia-Pkg-HLL-RSA` header to package-server requests so a server (ie the Julia community) can estimate the number of **distinct clients** using it, per resource class, **without being able to identify or track any individual client** — even with full access to its own request logs.

This implements the client half of the *HyperLogLog Over RSA* protocol. The design, security argument, and privacy guarantees are written up at https://karpinski.org/HyperLogLogOverRSA.jl/ (see "Deployment in Julia's Pkg client").

## How it works

- The client fetches a per-server *ring certificate* (`$server/hll_rsa.toml`), verifies it is "fingerprint-free" (has the intended two-prime structure, so it can't be used to tag clients) without learning its factorization, then generates and stores a random master key `x₀` at `~/.julia/servers/<host>/hll_rsa.toml`. Re-checked at most once per session and once per day.
- On each request in a counted resource class (`/registries`, `/registry/$uuid`, `/package/$uuid`, `/artifact/$hash`) it attaches `Julia-Pkg-HLL-RSA: <ring-id>,<token>`, where the token is a freshly re-randomized encrypted HyperLogLog value. Two tokens from the same client are unlinkable, yet all decode — only by the ring's offline owner, using the secret factorization — to that client's single stable value for the class.
- Pure-Julia `BigInt` crypto, no new dependencies. Nothing here can break a request or bother the user; the whole stack fails safe.

## Header values

- `<ring-id>,<token>` — the count token for a counted request
- `noclass` — a request in no counted class (so an *absent* header means exactly one thing: the client opted out)
- `error,<reason>` — a failure in the HLL stack (`fetch` / `verify` / `internal`), so a broken certificate or client bug is visible rather than silent

## Opt-out

On by default. Set `JULIA_PKG_SERVER_HLL_RSA` to a false value (`0`, `false`, `no`, …) to send no header at all. Documented in `managing-packages.md`.

## Testing

`test/hll_rsa.jl` (registered in `runtests.jl`) covers resource-class mapping, opt-out, token round-trips decoded by a test-only server-side decoder against reference values, persistence, the error-header vs. opt-out distinction, and a certificate-verification deep dive that perturbs a valid cert in each distinct way. Green (100 tests) against this checkout under Julia 1.14-nightly, and the `get_metadata_headers` hook is smoke-tested end-to-end.

---
🤖 Co-authored with [Claude Code](https://claude.com/claude-code) (see the `Co-Authored-By` trailer).


KristofferC:
> a server can estimate the number of distinct clients using 

Why is this important and for whom? Why would the server care, it's just serving files, or? Do other package managers do that?

StefanKarpinski:
> Why is this important and for whom?

Have you read https://karpinski.org/HyperLogLogOverRSA.jl/dev/? Motivation is specifically addressed at https://karpinski.org/HyperLogLogOverRSA.jl/dev/01-executive-summary/#Why-this-is-worth-doing.

> Do other package managers do that?

No, because prior to the existence of this entirely new protocol, there has been no way of estimating user counts without tracking users.

StefanKarpinski:
Discussion on discourse: https://discourse.julialang.org/t/rfc-proposal-for-estimating-unique-julia-clients-without-tracking/138027

docs/src/managing-packages.md
```diff
@@ -611,3 +611,13 @@ For clarification, some sources are not provided by Pkg server
...
 For the deployment of pkg server, please refer to [PkgServer.jl](https://github.com/JuliaPackaging/PkgServer.jl).
+
+### Anonymous client counting
+
+When talking to a package server, the client attaches a `Julia-Pkg-HLL-RSA` header that lets the server
+estimate how many distinct clients use it, without being able to identify or track any individual client —
+even with full access to its own request logs. This uses the HyperLogLog Over RSA protocol; see the
+[writeup](https://karpinski.org/HyperLogLogOverRSA.jl/) for the design and its privacy guarantees.
+
+It is enabled by default. To opt out, set `JULIA_PKG_SERVER_HLL_RSA` to a false value (e.g. `0`, `false`, or
```
Antipath1, line 622: Shipping this as an opt-out default is a massive violation of user trust. I am completely opposed to this ever being merged, but pushing it through without cryptographic peer review from actual, credentialed cryptographers is completely reckless. (Vetting a novel cryptographic protocol via a Discourse thread, a Slack channel and the LLM Claude is not a security audit).

No matter how clever the client-side cryptography is, this code turns the Julia package manager into a telemetry agent that pings a corporate server (JuliaHub). You are proposing to collect geographic distribution, OS breakdown, and version usage, which is extremely valuable business intelligence for JuliaHub and framing it as a favor to FOSS developers seeking funding for the packages (even though no other package manager has to do it, apparently).

If this is genuinely for the community, make it opt-in. Do not force users to manually edit their environments just to stop their package manager from tracking them. I massively distrust JuliaHub with this data, and this code should not be merged in this state or ever really, but especially not in this state.

CHANGELOG.md
```diff
@@ -11,6 +11,10 @@ Pkg v1.14 Release Notes
   recompilation easier to spot. ([#4679])
 - During package source installation, Pkg now reports when a package has an Artifacts.toml but no artifacts match the
   current platform. ([#4646])
+- Pkg now attaches an anonymous unique-client-counting header (`Julia-Pkg-HLL-RSA`) to package-server requests,
```
Antipath1, line 14: This should not be buried in a changelog. If this massive shift in the ecosystem's trust model is ever forced through (which it absolutely should not ever be), it needs to be at the very top of the list in every single Julia forum.

A fundamental change to telemetry requires dedicated, standalone announcements on Discourse, Slack, and Zulip and not be hidden as a small bullet point in a file full of Pkg.jl updates. At a bare minimum, this should print a big red warning in the REPL so users actually know they are being enlisted as telemetry nodes. At the minimum.

But even with maximum visibility, I fundamentally oppose merging this at all ever.

Antipath1 (dismissed):
> Why is this important and for whom?

Knowing the exact size, geographic distribution, OS breakdown, packages and version usage of the Julia user base is incredibly valuable business intelligence for JuliaHub. Pretending this massive infrastructure shift is about better funding proposals for packages is incredibly transparent, deeply frustrating and insulting to our intelligence.



>  No, because prior to the existence of this entirely new protocol, there has been no way of estimating user counts without tracking users.

Or maybe other FOSS package managers haven't tried to implement this not because they were waiting for a stroke of cryptographic genius, but because they actually want users to trust their package managers not to phone home and respect them enough to not turn their open source package managers into corporate spyware.

This proposal is frankly a massive overreach. As a chemistry student whose research relies heavily on this language, I have invested immense trust into this ecosystem, and I absolutely refuse to blindly trust that JuliaHub won't use my machine as a data-harvesting node. I do not care how mathematically sound the client-side cryptography is, because you cannot fix a massive institutional trust deficit using math. I massively distrust JuliaHub's motivations here, and forcing this into the ecosystem, *especially as an opt-out*, is a terrible idea.

StefanKarpinski:
> @Antipath1: Knowing the exact size, geographic distribution, OS breakdown, packages and version usage of the Julia user base is incredibly valuable business intelligence for JuliaHub. Pretending this massive infrastructure shift is about better funding proposals for packages is incredibly transparent, deeply frustrating and insulting to our intelligence.

Several points here:

- This infrastructure isn't owned or operated by JuliaHub.
- We (the JuliaLang project) already have rough geographic distribution, OS breakdown, packages and version usage data _by request count_, just not by number of unique clients.
- This information has zero business value to JuliaHub aside from the fact that being able to show that Julia is popular overall is useful to reassure people that a Julia-focused business is legitimate.

Frankly, these days since Julia is regularly in the top 20 on various language indices and our rough estimates of the number of Julia users from other sources have climbed into the millions, showing overall Julia usage is just not that much of an issue. To the extent that this helps demonstrate Julia popularity, this provides the same value to JuliaHub as it does to any company or organization betting on Julia. What is much harder is showing the popularity of individual Julia projects, and many of them have really struggled with this, which is what this aims to solve. 

Also: how do we know that there are millions of Julia users? Well, the Julia VS Code extension has over a million users, which we know thanks to user-level telemetry collected by Microsoft, which somehow no one seems particularly upset about. That is the same company that also knows every time you clone a GitHub repo. Will they tell the Julia community that info? Nope.
> Or maybe other FOSS package managers haven't tried to implement this not because they were waiting for a stroke of cryptographic genius, but because they actually want users to trust their package managers not to phone home and respect them enough to not turn their open source package managers into corporate spyware.

Python published raw PyPI request logs with IP addresses unmasked _for decades_ until GDPR made it illegal to do so. Go's package servers are literally hosted and run by Google and they collect the level of telemetry you'd expect based on that (they don't actually tell you what they collect). Our stance before and after this is significantly more privacy-respecting than most package managers. And again, this is the Julia project; JuliaHub has nothing to do with it; this doesn't benefit JuliaHub specifically in any way.

> This should not be buried in a changelog. If this massive shift in the ecosystem's trust model is ever forced through (which it absolutely should not ever be), it needs to be at the very top of the list in every single Julia forum.

> A fundamental change to telemetry requires dedicated, standalone announcements on Discourse, Slack, and Zulip and not be hidden as a small bullet point in a file full of Pkg.jl updates. At a bare minimum, this should print a big red warning in the REPL so users actually know they are being enlisted as telemetry nodes. At the minimum.

It's not like this changelog is the only place it's mentioned — I'm not exactly trying to keep this a secret. I've posted about it on Slack and Discourse. I'm even giving a [JuliaCon talk ](https://pretalx.com/juliacon-2026/talk/JXMW8H/) about it. Is there somewhere else you think I should post it to?

> Shipping this as an opt-out default is a massive violation of user trust. I am completely opposed to this ever being merged, but pushing it through without cryptographic peer review from actual, credentialed cryptographers is completely reckless. (Vetting a novel cryptographic protocol via a Discourse thread, a Slack channel and the LLM Claude is not a security audit).

I'm definitely planning on submitting it for peer review. That takes a long time, unfortunately, and one of the major issues with deploying something in Julia clients is that the lead time on that is long and unlike servers, we cannot simply deploy new code and have it take effect immediately. If it's what the community seems to prefer, we can ship this in clients without turning the server-side on until it's peer reviewed. If it passes peer review, then we can turn on the server side and instantly be getting client count estimates from all Julia versions since 1.14. If we wait until peer review happens and it's published and then implement it in clients, we won't be getting any data for about five years. (Note that people can easily verify that severs are not doing the protocol by checking if the https://pkg.julialang.org/hll_rsa.toml endpoint actually exists or not; or snooping on their own actual Pkg traffic with something like WireShark, to see if the headers are being sent.)

A word about peer review in general. Yes, it's a good idea and it might catch something, but having three typically somewhat disinterested reviewers peruse a paper seems to me significantly less likely to catch issues than having a largish community of interested and highly mathematically minded people scrutinize it. So sure, peer review is a thing we should do, but it's not some magical process that is guaranteed to catch problems. Research papers with massive issues get published all the time even in respectable journals.

Regarding opt-out vs opt-in: if it's opt-in, it has zero value and there's no point in doing it. Why? Because approximately nobody changes defaults for anything. Some tiny percentage of users will opt-in. If we knew what percentage opted in, then yes, we could estimate from there. But we don't know what the percentage is, so if this were opt-in it would be totally useless.

> This proposal is frankly a massive overreach. As a chemistry student whose research relies heavily on this language, I have invested immense trust into this ecosystem, and I absolutely refuse to blindly trust that JuliaHub won't use my machine as a data-harvesting node. I do not care how mathematically sound the client-side cryptography is, because you cannot fix a massive institutional trust deficit using math. I massively distrust JuliaHub's motivations here, and forcing this into the ecosystem, especially as an opt-out, is a terrible idea.

> But even with maximum visibility, I fundamentally oppose merging this at all ever.

I hear you loud and clear and respect that position and appreciate you expressing it here. If this turns out to be the viewpoint of even a significant minority of the community, we will not ship this.

Regarding "I absolutely refuse to blindly trust that JuliaHub won't use my machine as a data-harvesting node": as I've said above, this isn't being done by JuliaHub and doesn't particularly benefit JuliaHub; there's also no blind trust involved — this is open source and you can see exactly what the code on your system is doing. Calling this "data-harvesting node" seems like a pretty significant overstatement, given that this is explicitly and carefully designed to respect and preserve individual privacy as much as possible (more even—literally more than would previously have been possible).

Meanwhile, do you use any closed source software? Google? If not, respect, but I bet you do. You use GitHub at least. You have no idea what that software is doing and it _is_ actually turning all your devices into data harvesting nodes. But the recipients of that data are companies, and they don't tell you what they're collecting or how, let alone show you the code and explain what they're doing, or allow it to benefit open source projects and communities.

CHANGELOG.md
StefanKarpinski: Posted on Discourse and Slack and there will be a [JuliaCon talk](https://pretalx.com/juliacon-2026/talk/JXMW8H/).

docs/src/managing-packages.md
StefanKarpinski: Opt-in makes this feature useless, as discussed in my comment below. This either ships opt-out or it shouldn't ship at all. Either way, nothing to change here; either this PR gets merged or it doesn't.

Antipath1:
If the core concern is that users won't actively seek out hidden settings to change a default, there is a solution that respects both developer funding needs and user consent: a first-run prompt.


Instead of burying a telemetry decision in a changelog and relying on an opt-out default, the first time a user opens the REPL or runs a package command in a new installation, they should be presented with a simple, one-time prompt:



Do you consent to a privacy-preserving, non-identifying way to collect

basic usage for Julia packages? The algorithm is described in [link to

peer-reviewed paper], and this helps package authors secure funding

(maybe cite more reasons if there are any). [y/N]


This is the exact approach taken by many modern CLI tools and open-source frameworks. It solves the issue of users ignoring defaults by requiring an explicit choice, and it addresses the ethical issue by asking for permission before using a user's machine to generate metrics.


If the cryptographic protocol is as sound as claimed, and the benefits to the FOSS ecosystem are clearly stated, the community will gladly press 'y' at I believe a high enough percentage for the data to be useful.

StefanKarpinski:
How would one estimate unique clients given an unknown opt-in rate? Most such telemetry prompts are for actual telemetry, as in collecting data about *how* people use a tool, not just counting how many use it.

Antipath1:
> How would one estimate unique clients given an unknown opt-in rate? Most such telemetry prompts are for actual telemetry, as in collecting data about _how_ people use a tool, not just counting how many use it.

You estimate the total by cross-referencing the opt-in cohort against the baseline metrics you already have: server-side download requests, CI traffic, and the VS Code extension numbers you mentioned earlier. Data scientists constantly model population sizes from partial opt-in data.

But fundamentally, if FOSS funding requires harvesting hardware metrics from users without their explicit permission, then it is the expectations around metric gathering that need to change, not the users' right to consent. FOSS communities have to accept undercounting as the necessary and acceptable cost of respecting user privacy.

This is the cost of respecting the user's autonomy and privacy.

> I hear you loud and clear and respect that position and appreciate you expressing it here. If this turns out to be the viewpoint of even a significant minority of the community, we will not ship this.

Well then, consider my minority position made clear, Stefan.

StefanKarpinski:
> This is the cost of respecting the user's autonomy and privacy.

Assume that we implement all of the protections that are being discussed on the discourse thread (timestamp coarsening, removal of IP hashes, etc.). Suppose you are a user who does not know about this option and does not opt out of it. How would unknowingly leaving this turned on violate your privacy? Concretely. Real question. Please lay out the actual privacy issue that this causes.

StefanKarpinski:
I'm leaving this PR in draft state while I address various issues that have been brought up by both human and AI review. These are all be opened as [issues](https://github.com/StefanKarpinski/HyperLogLogOverRSA.jl/issues) on the HyperLogLogOverRSA.jl repo. Once those are all addressed and I get feedback from a few more people who have graciously volunteered to lend their cryptographic expertise to reviewing this, only then will I consider switching this from draft state.

PR #4725, opened 2026-07-17
Don't create redundant `[sources]` for workspace-internal deps

PatrickHaecker:
A dependency that is itself an environment in the same workspace is located
through workspace membership. This is already working before the PR.

In this PR we stop adding a (completely redundant) `[sources]` entry for such
dependencies. We do respect, however, the existing entries to avoid changing
project files which the user might want to be preserved.

An especially good (and common) example why the previous behavior has not
made too much sense (to me) is the following:
Assume a package which is a workspace which has a test directory with a
separate environment. Per default `Pkg` created a source entry in the test
environment to locate the package in "..".

So we effectively told the user:
> Don't worry, as you are in the test environment we've located the root of the
> workspace automatically, but please tell me explicitly where the package is
> you want to test, because I can't find that automatically although I already
> found it as it is identical to the workspace environment.

I used Claude Opus 4.8 when creating this PR.

This PR is based on #4709, so should be merged after #4709.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,9 @@
 Pkg v1.14 Release Notes
 =======================
 
+- Pkg no longer creates a `[sources]` entry for a dependency that is itself a project in the same workspace. Such a
+  dependency is already located through workspace membership, so the entry was redundant. Existing `[sources]` entries
+  are still honored.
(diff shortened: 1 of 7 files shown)


PatrickHaecker:
The failing windows test is unrelated.

PR #4738, opened 2026-07-27
Use Resolver.jl's SAT-based resolver by default

KristofferC:
This switches Pkg's default resolver to [Resolver.jl](https://github.com/StefanKarpinski/Resolver.jl): exact SAT-based resolution (PicoSAT) with an explicit preference order, and a real diagnosis when a resolve is unsatisfiable. The legacy maxsum resolver stays selectable with `JULIA_PKG_RESOLVER=maxsum`.

Resolver.jl comes in through `[sources]`, pinned at current `main` (StefanKarpinski/Resolver.jl@54856db). Since the first version of this PR, Resolver.jl grew the `Problem`/`Diagnosis` API, interchangeability classes, and a much faster descent, which is what made this worth redoing.

## What

- **`src/SATResolve.jl`**: adapter from `Operations.deps_graph` output to Resolver.jl's `PkgData` universe and `Problem`, replicating the legacy `Resolve.Graph` semantics (weak deps constrain but don't install, first-registry-wins, the "ignore incompatible stdlib compat" rule, julia as a synthetic single-version required package, dev'd/repo-tracked packages as single-version packages, loaded-version preference via `order`).
- **Only the strong-dependency closure is built.** `deps_graph` hands over the whole weak-dependency closure (~1500 packages / 42k versions even for `add JSON`) and Resolver builds a conflict matrix for everything it is given. A package reachable through weak edges alone can never be installed, so it is left out; this is the single largest speedup.
- **User constraints are modelled as constraints, not baked into the universe**: requirement version specs and project `[compat]` entries go in as `compat`, pins as `pin`, and the active/workspace project itself is folded into them instead of being a fake package. That is what lets the diagnosis say "your compat restricts …", "your pin restricts …", "unpin X" rather than "project requires …".
- **UNSAT diagnostics**: `Resolver.resolve` returns a `Diagnosis` with independent conflicts, a minimal proof chain each, verified fixes with what each would allow, a note on the fixes that do not work, and upstream suggestions. Pkg rebuilds it over package names and renders Resolver's own report; fixes that would "drop dependency julia" are filtered out as not actionable. A requirement whose spec matches no version at all is reported directly with the versions that do exist, without a SAT round.
- The diagnosis only runs when the failure is final: intermediate tiers of `tiered_resolve` pass `diagnose_unsat = false`.

## Speed

Resolve step of a fresh `Pkg.add` against the live General registry, julia nightly, warm process, best of 3 (`deps_graph` included in both):

| add | maxsum | SAT |
|---|---|---|
| JSON | 0.37 s | 0.01 s |
| Plots | 0.35 s | 0.06 s |
| DifferentialEquations | 0.42 s | 0.31 s |
| Symbolics | 1.36 s | 0.13 s |
| Makie | 0.34 s | 0.10 s |
| StatsPlots | 0.37 s | 0.07 s |
| DataFrames + CSV | 0.34 s | 0.02 s |
| Plots + DiffEq + Makie + Symbolics + DataFrames + Flux (411 deps) | 2.13 s | 0.33 s |

Manifests are identical to maxsum in every case above. The SAT descent itself is ≤ 30 ms everywhere; what remains of the SAT time is Resolver's T1 matrix build (`pkg_info`) on large closures (DiffEq: 0.2 s of the 0.34 s), which is the remaining upstream lever. First call in a fresh process (compile latency included): 0.26 s SAT vs 0.61 s maxsum for `add Plots`.

Unsatisfiable resolves, including the diagnosis: 0.1–0.5 s where maxsum takes 0.7–0.9 s to produce its log; the worst case tried (`OrdinaryDiffEq@5` + `SciMLBase@2`, a large closure) is 1.2 s vs 0.7 s, of which 0.36 s is the diagnosis.

## Error messages

`add DataFrames@1.7 PrettyTables@1`:

```
Unsatisfiable requirements detected — 1 conflict:

Conflict 1: DataFrames
  • your compat restricts DataFrames to 1.7.0–1.7.1
  • DataFrames 1.7.0–1.7.1 requires PrettyTables 2.1.0–2.4.0
  • your compat restricts PrettyTables to 1.0.0–1.3.1
  Fix it by any one of:
    1. relax your compat on PrettyTables
       → allows: DataFrames 1.7.1, PrettyTables 2.4.0
    2. relax your compat on DataFrames
       → allows: DataFrames 1.3.6, PrettyTables 1.3.1
    3. drop dependency DataFrames
       → allows: PrettyTables 1.3.1
  Note: dropping dependency PrettyTables does not help.
```

A pinned `StaticArrays 0.12.6` and `add Makie@0.24`:

```
Conflict 1: Makie
  • your compat restricts Makie to ≥0.24.0
  • Makie ≥0.24.0 requires StaticArrays ≤0.12.5, ≥1.0.0 (through MathTeXEngine
    and GeometryBasics)
  • your pin restricts StaticArrays to 0.12.6
  Fix it by any one of:
    1. unpin StaticArrays
       → allows: Makie 0.24.14, StaticArrays 1.9.20
    2. relax your compat on Makie
       → allows: Makie 0.21.18, StaticArrays 0.12.6
    3. drop dependency Makie
       → allows: StaticArrays 0.12.6
  Note: dropping dependency StaticArrays does not help.
```

A dev'd package `Foo` with `JSON = "0.20"` in its compat and `add JSON@1`:

```
Conflict 1: Foo
  • Foo 0.1.0 requires JSON 0.20.0–0.20.2
  • your compat restricts JSON to ≥1.0.0
  Fix it by any one of:
    1. relax your compat on JSON
       → allows: Foo 0.1.0, JSON 0.20.2
    2. drop dependency Foo
       → allows: JSON 1.8.0
  Note: dropping dependency JSON does not help.
  Upstream fix: a release of Foo supporting JSON 1.8.0 would fix this; 0.1.0,
    its latest, supports only 0.20.0–0.20.2.
    → would allow: JSON 1.8.0
```

`julia_version = 1.6.7` and `add Makie@0.24` (maxsum blames `ComputePipeline` here):

```
Conflict 1: julia and Makie
  • julia 1.6.7 constrains Makie to ≤0.23.0
  • your compat restricts Makie to ≥0.24.0
  Fix it by any one of:
    1. relax your compat on Makie
       → allows: Makie 0.23.0
    2. drop dependency Makie
```

Two independent conflicts in one `add` are reported as two numbered conflicts ("pick a fix for each"); a conflict three packages deep is stated with the packages it goes through ("SciMLBase 2.0.0–2.155.2 constrains OrdinaryDiffEq to ≤4.21.1, ≥6.0.0 (through ArrayInterface)") where maxsum prints a 30-line restriction tree.

## Tests

`api`, `add`, `update`, `pkgtest`, `workspaces`, `sources`, `develop`, `force_latest_compatible_version`, `stdlib_compat`, `historical_stdlib_version` and `resolve` pass locally. Test changes: the `resolver backends` testset (maxsum toggle, env-var validation, direct "no such version" report, diagnosis rendering), backend-agnostic wording in two existing assertions, and the host depot appended for the three subprocess tests that load the dev Pkg in an isolated depot (`Utils.host_deps_depot`).

## Caveats / follow-ups

- Pkg as a stdlib can't ship non-stdlib deps: landing this for real needs Resolver.jl and `libpicosat_jll` (MIT, one C file) vendored or made stdlibs, plus a registered Resolver.jl release instead of `[sources]`.
- `Pkg.test`'s `allow_reresolve` path still runs the diagnosis before retrying; threading `diagnose_unsat` through `Pkg.resolve` would avoid that.
- Remaining SAT cost is Resolver's T1 build on large closures; the descent is no longer a factor.

### DIFF
--- a/Project.toml
+++ b/Project.toml
@@ -20,12 +20,16 @@ Logging = "56ddb016-857b-54e1-b83d-db4d58db5568"
 Markdown = "d6f4376e-aef5-505a-96c1-9c027394607a"
 Printf = "de0858da-6303-5e67-8744-51eddeeeb8d7"
 Random = "9a3f8284-a2c9-5f02-9a11-845980a1fd5c"
+Resolver = "ecd60e94-5748-11e9-04dc-1fe0f95e4121"
 SHA = "ea8e919c-243c-51af-8825-aaa63cd721ce"
 TOML = "fa267f1f-6049-4f14-aa54-33bafae1ed76"
 Tar = "a4e569a6-e804-4fa4-b0f3-eef7a1d5b13e"
 UUIDs = "cf7118a7-6976-5b1a-9a39-7adc72f591a4"
 Zstd_jll = "3161d3a3-bdf6-5164-811a-617609db77b4"
 p7zip_jll = "3f19e933-33d8-53b3-aaab-bd5110c3b7a0"
+
+[sources]
+Resolver = {url = "https://github.com/StefanKarpinski/Resolver.jl", rev = "kc/pkg-integration"}
 
 [weakdeps]
 REPL = "3fa0cd96-eef1-5676-8a61-b3b8758bbffb"
@@ -46,6 +50,7 @@ Markdown = "1.11"
 Printf = "1.11"
 REPL = "1.11"
 Random = "1.11"
+Resolver = "0.2"
 SHA = "0.7, 1"
 TOML = "1"
 Tar = "1.10"
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -9,6 +9,7 @@ using Random: randstring
 import LibGit2, Dates, TOML
 
 using ..Types, ..Resolve, ..PlatformEngines, ..GitTools, ..MiniProgressBars
+import ..SATResolve
 import ..depots, ..depots1, ..devdir, ..set_readonly, ..Types.PackageEntry
 import ..Artifacts: ensure_artifact_installed, artifact_names, extract_all_hashes,
     artifact_exists, select_downloadable_artifacts, mv_temp_dir_retries
@@ -748,6 +749,15 @@ function get_compat_workspace(env, name)
     return compat
 end
 
+# Which resolver backend to use: the SAT-based resolver from Resolver.jl
+# (default) or the legacy maxsum resolver in `Resolve`.
+function resolver_backend()
+    backend = get(ENV, "JULIA_PKG_RESOLVER", "sat")
+    backend == "sat" && return :sat
+    backend == "maxsum" && return :maxsum
+    return pkgerror("unknown resolver backend `$(backend)` in JULIA_PKG_RESOLVER; expected `sat` or `maxsum`")
+end
+
 # Resolve a set of versions given package version specs
 # looks at uuid, version, repo/path,
 # sets version to a VersionNumber
@@ -755,7 +765,8 @@ end
 # all versioned packages should have a `tree_hash`
 function resolve_versions!(
         env::EnvCache, registries::Vector{Registry.RegistryInstance}, pkgs::Vector{PackageSpec}, julia_version,
-        installed_only::Bool, preferred_versions::Dict{UUID, VersionNumber} = Dict{UUID, VersionNumber}()
+        installed_only::Bool, preferred_versions::Dict{UUID, VersionNumber} = Dict{UUID, VersionNumber}();
+        diagnose_unsat::Bool = true
     )
     installed_only = installed_only || OFFLINE_MODE[]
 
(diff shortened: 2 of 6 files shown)


StefanKarpinski:
Rendering nits should be improved as well as @JeffBezanson's request to change "your compat leaves" to "your compat allows only".

KristofferC:
"restricts Package to... " is another possibility

KristofferC:
> drop requirement Makie

This is also a little bit obscure for your average user (what is "requirement")

StefanKarpinski:
I like "restricts P to" and I can replace "requirement" with "dependency". Any other wording suggestions? I can do them as a batch.

KristofferC:
"Blocked fixes:" is a little bit obscure as well imo. Blocked on what?

Other than that, I think it looks pretty good.

StefanKarpinski:
So the concept of "blocked fixes" is that they're fixes that people might *think* would work but which don't actually work for some (non-obvious) reason. What would be a better way to explain that? Or is that not something we should be presenting at all? Otherwise people might think "maybe I can just drop PrettyTables as a dependency".

KristofferC:
I am not against the idea, just that the word "blocking" feels wrong. Also, saying why the change does not help would also be useful, for example formulating it as `dropping requirement PrettyTables (conflict remains via DataFrames)`.

Some possibilities of how this could be changed.:

- Change "blocked" to something like "Won't help", "Insufficient fixes".
- Make it inline, so something like: 

```
  Fix it by any one of:
    1. relax your compat on PrettyTables → ...
    2. relax your compat on DataFrames → ...
    3. drop requirement DataFrames → ...
  Note: dropping PrettyTables alone won't help — DataFrames still requires PrettyTables 2.x.
```

StefanKarpinski:
More wording tweaks merged in https://github.com/StefanKarpinski/Resolver.jl/pull/121.

KristofferC:
Post updated with latest wordings

PR #4740, opened 2026-07-28
release-1.11: Backport #4716

Keno:
The Pkg server may serve the General registry zstd-compressed rather than gzip-compressed; the registry test hard-coded `General.tar.gz`.


KristofferC:
I am not sure we care about 1.11 at this point.

PR #3492, opened 2023-05-23
handle if a usage file is missing a "time" entry

IanButterworth:
Fixes #3491

A more general fix that could also be done is https://github.com/JuliaLang/Pkg.jl/pull/3454

PR #4743, opened 2026-07-30
Atomically update usage TOMLs and handle malformed files

### LINKED ISSUE #4742: Error key "time" not found
I can't reprodue it but it happens second time, and removing ~/.julia helps.

```
(@v1.12) pkg> gc
RROR: KeyError: key "time" not found                                                                                                                                                                    [472/1841]
Stacktrace:                                                                                              
  [1] getindex(h::Dict{String, Any}, key::String)  
    @ Base ./dict.jl:477                                                                                                                                                                                           
  [2] (::Pkg.API.var"#210#211"{Dict{String, Set{String}}, Dict{String, Dates.DateTime}})(filename::String, info::Dict{String, Any})
    @ Pkg.API ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:660
  [3] _broadcast_getindex_evalf                  
    @ ./broadcast.jl:699 [inlined]                  
  [4] _broadcast_getindex                                                                                
    @ ./broadcast.jl:672 [inlined]                                                                       
  [5] _getindex                                                                                          
    @ ./broadcast.jl:620 [inlined]                                                                       
  [6] getindex                                                                                           
    @ ./broadcast.jl:616 [inlined]                                                                       
  [7] macro expansion                              
    @ ./broadcast.jl:995 [inlined]                  
  [8] macro expansion                            
    @ ./simdloop.jl:77 [inlined]                    
  [9] copyto!                                                                                            
    @ ./broadcast.jl:994 [inlined]                                                                       
 [10] copyto!                                                                                            
    @ ./broadcast.jl:947 [inlined]                  
 [11] copy                                                                                               
    @ ./broadcast.jl:919 [inlined]               
 [12] materialize(bc::Base.Broadcast.Broadcasted{Base.Broadcast.DefaultArrayStyle{1}, Nothing, Pkg.API.var"#210#211"{Dict{String, Set{String}}, Dict{String, Dates.DateTime}}, Tuple{Base.RefValue{String}, Vector{
Any}}})                                                                                                  
    @ Base.Broadcast ./broadcast.jl:894
 [13] (::Pkg.API.var"#reduce_usage!#205")(f::Function, usage_filepath::String)
    @ Pkg.API ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:627
 [14] gc(ctx::Pkg.Types.Context; collect_delay::Dates.Day, verbose::Bool, force::Bool, kwargs::@Kwargs{})
    @ Pkg.API ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:658
 [15] gc
    @ ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:595 [inlined]
 [16] gc()
    @ Pkg.API ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/API.jl:595
 [17] do_cmd(command::Pkg.REPLMode.Command, io::Base.TTY)
    @ Pkg.REPLMode ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/REPLMode/REPLMode.jl:414
 [18] do_cmds(commands::Vector{Pkg.REPLMode.Command}, io::Base.TTY)
    @ Pkg.REPLMode ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/src/REPLMode/REPLMode.jl:403
 [19] do_cmds(repl::REPL.LineEditREPL, commands::String)
    @ REPLExt ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/ext/REPLExt/REPLExt.jl:116
 [20] on_done(s::REPL.LineEdit.MIState, buf::IOBuffer, ok::Bool, repl::REPL.LineEditREPL)
    @ REPLExt ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/ext/REPLExt/REPLExt.jl:130
 [21] #invokelatest_gr#232
    @ ./reflection.jl:1295 [inlined]
 [22] invokelatest_gr
    @ ./reflection.jl:1289 [inlined]
 [23] (::REPLExt.var"#create_mode##0#create_mode##1"{REPL.LineEditREPL})(s::REPL.LineEdit.MIState, buf::IOBuffer, ok::Bool)
    @ REPLExt ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/Pkg/ext/REPLExt/REPLExt.jl:155
 [24] run_interface(terminal::Base.Terminals.TextTerminal, m::REPL.LineEdit.ModalInterface, s::REPL.LineEdit.MIState)
    @ REPL.LineEdit ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/LineEdit.jl:2854
 [25] run_frontend(repl::REPL.LineEditREPL, backend::REPL.REPLBackendRef)
    @ REPL ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:1663
 [26] (::REPL.var"#61#62"{REPL.LineEditREPL, REPL.REPLBackendRef})()
    @ REPL ~/.julia/juliaup/julia-1.12.6+0.x64.linux.gnu/share/julia/stdlib/v1.12/REPL/src/REPL.jl:650
```

Maybe it's some internal bug? Couldn't find it reported in issues either, so in case someone can confirm I am opening this one.

Regards.

jakobnissen:
The files {artifact,scratch,manifest}_usage.toml were written by incremental appending instead of atomically. This means a user Ctrl-C-ing Pkg could leave these files in a malformed state. Malformed files would throw an internal KeyError as expected fields were not written before interruption.

Fix this both by using the existing `atomic_toml_write` to atomically update these files to prevent corruption, and also by handling malformedness more consistently. Malformed TOML files were already handled, but not in the `.gc()` function.

Closes #4742 

This PR was made using an AI agent and manually reviewed.

### DIFF
--- a/src/API.jl
+++ b/src/API.jl
@@ -750,9 +750,14 @@ function gc(ctx::Context = Context(); collect_delay::Union{Period, Nothing} = no
     # Combine the per-depot live paths for the mark-and-sweep phase below.
     all_manifest_tomls = Set(f for (_, files) in manifest_usage_by_depot for f in keys(files))
     all_artifact_tomls = Set(f for (_, files) in artifact_usage_by_depot for f in keys(files))
-    all_scratch_dirs = Set(f for (_, dirs) in scratch_usage_by_depot for f in keys(dirs))
     all_unknown_scratch_parents = Set(
         f for (_, paths) in unknown_scratch_parents_by_depot for f in paths
+    )
+    # Unknown-parent records are deliberately excluded from the condensed usage map,
+    # but must still enter marking so `process_scratchspace` can preserve them.
+    all_scratch_dirs = union(
+        Set(f for (_, dirs) in scratch_usage_by_depot for f in keys(dirs)),
+        all_unknown_scratch_parents,
     )
 
     function process_manifest_pkgs(path)
--- a/src/API.jl
+++ b/src/API.jl
@@ -13,7 +13,7 @@ import Base: StaleCacheKey
 
 import ..depots, ..depots1, ..logdir, ..devdir, ..printpkgstyle, .._autoprecompilation_enabled_scoped, ..manifest_rel_path
 import ..Operations, ..GitTools, ..Pkg, ..Registry
-import ..can_fancyprint, ..pathrepr, ..isurl, ..PREV_ENV_PATH, ..atomic_toml_write, ..safe_realpath
+import ..can_fancyprint, ..pathrepr, ..isurl, ..PREV_ENV_PATH, ..safe_realpath
 using ..Types, ..TOML
 using ..Types: VersionTypes
 using Base.BinaryPlatforms
@@ -648,16 +648,21 @@ function gc(ctx::Context = Context(); collect_delay::Union{Period, Nothing} = no
     # Collect both last known usage dates, as well as parent projects for each scratch space
     scratch_usage_by_depot = UsageByDepotDict()
     scratch_parents_by_depot = Dict{String, Dict{String, Set{String}}}()
+    unknown_scratch_parents_by_depot = Dict{String, Set{String}}()
 
-    # Fold every recorded event in one usage log into caller-owned condensed state.
-    # A parse failure returns false so GC can stop rather than treat the log as empty.
-    function reduce_usage!(f::Function, usage_filepath)
-        parsed_usage = Types.read_usage_file(usage_filepath)
-        parsed_usage === nothing && return false
-        for (filename, infos) in parsed_usage
-            f.(Ref(filename), infos)
+    function condense_index_usage(usage_file::String)
+        return Types.with_usage_file(usage_file; create = false, malformed = :preserve) do entries
+            usage = UsageDict()
+            valid_usage = Types.foreach_usage_entry(entries, usage_file) do filename, info
+                time = Types.usage_time(info, usage_file, filename)
+                usage[filename] = max(get(usage, filename, DateTime(0)), time)
+            end
+            valid_usage || return nothing
+            filter!(((path, _),) -> Pkg.isfile_nothrow(path), usage)
+            empty!(entries)
+            merge!(entries, Dict(path => [Dict("time" => time)] for (path, time) in usage))
+            return usage
         end
-        return true
     end
 
     # Read and rewrite each existing usage log as one locked transaction. This prevents
@@ -668,22 +673,7 @@ function gc(ctx::Context = Context(); collect_delay::Union{Period, Nothing} = no
         # the manifest log with that condensed representation.
         manifest_usage_file = joinpath(logdir(depot), "manifest_usage.toml")
         manifest_usage_result = if isfile(manifest_usage_file)
-            Types.with_usage_file_lock(manifest_usage_file) do
-                usage = UsageDict()
-                reduce_usage!(manifest_usage_file) do filename, info
-                    time = Types.usage_time(info, manifest_usage_file, filename)
-                    usage[filename] = max(get(usage, filename, DateTime(0)), time)
-                end || return nothing
-                filter!(((path, time),) -> Pkg.isfile_nothrow(path), usage)
-                condensed_usage =
-                    Dict(path => [Dict("time" => time)] for (path, time) in usage)
-                atomic_toml_write(
-                    manifest_usage_file,
-                    condensed_usage,
-                    sorted = true
-                )
-                return usage
-            end
+            condense_index_usage(manifest_usage_file)
         else
             UsageDict()
         end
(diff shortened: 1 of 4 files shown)


IanButterworth:
I haven't reviewed the code but I find it very surprising that this wasn't fixed by #2661 #2793 #3492 #3928

jakobnissen:
Those fixes work, but they don't fix the code path hit in #4742. Previous PRs fixed up `manifest_usage.toml` and `artifact_usage.toml` whereas this one fixes up `scratch_usage.toml`.
Maybe that suggests all three should use the same mechanism. I'll give that a go.

IanButterworth:
Ah yeah. That'd be good.

jakobnissen:
Things that might be interesting to review:
* Here, GC aborts on malformed log file, whereas the other operations instead overwrite the malformed file, skipping malformed entries. The reason is that it's much worse to accidentally GC the wrong thing. Is that right behaviour?
* The internal interfaces in `src/Types.jl` - `with_usage_file`, `write_env_usage` and `foreach_usage_entry`. Is their behaviour understandable and sensible?

jakobnissen:
I don't like how these internal TOML files are parsed into untyped `Dict`s, but parsing them into typed structs is out of scope for this PR.

jakobnissen:
This is ready for review now, @IanButterworth

IanButterworth:
Sorry for the delay on this. Thanks for doing it. There are some findings I think worth addressing:

Claude:

---

The description overstates the problem: `manifest_usage.toml` and `artifact_usage.toml` were already written atomically on master (pidlock + `atomic_toml_write` in `write_env_usage`, since #2793 and #4293). The only appending writer in Pkg was the `build_versions` scratch append this PR removes, and Scratch.jl also appends to `scratch_usage.toml`, which this PR can't reach, so corrupted files will keep appearing and the reader hardening is the load-bearing part. It also undersells a real fix: `gc` previously rewrote the condensed logs without holding the pidlock across its read and write, so it could clobber a concurrent writer's update with a stale snapshot. Suggested replacement body:

```markdown
`Pkg.gc()` threw an internal `KeyError` when a usage log contained a partial entry.
`scratch_usage.toml` is appended to non-atomically, both by `build_versions` in Pkg and by
Scratch.jl, so an interrupted write can leave an entry without its `time` key, or invalid
TOML. `manifest_usage.toml` and `artifact_usage.toml` were already written atomically, but
`gc` rewrote them without holding the pidfile lock across its read and write, so it could
overwrite a concurrent writer's update with a stale snapshot.

This PR replaces the appending scratch writer with a locked read-modify-write via the new
`Types.with_usage_file`, now shared by all usage-log access including `gc`, and makes the
readers tolerate missing or invalid `time` and `parent_projects` fields, repairing what
they can and conservatively preserving what they cannot decode. Scratch.jl still appends,
so malformed files can still occur; the reader hardening handles them.

Closes #4742
```

Two issues from review:

1. `write_env_usage` ignores `foreach_usage_entry`'s return value (Types.jl:824). On the first structurally invalid record, iteration stops but the file is still rewritten with only the entries visited so far plus the new update. Since dict iteration order is arbitrary, one junk record can silently delete valid usage timestamps for other manifests/artifacts, letting a later `gc` treat them as unreachable sooner. Skip the bad record and continue, or abort the write.

2. On your open question: aborting `gc` on a malformed log is the right call (collecting on bad data is worse), but it needs to be loud, and it can currently be permanent. `gc` returns with only a `@warn` from deep inside; no collection happens and nothing says it was skipped. Manifest/artifact logs self-heal on the next resolve, but an unparseable `scratch_usage.toml` on a machine that never builds packages is never rewritten, so `gc` stays disabled forever, roughly the #4742 reporter's situation. Suggest erroring with the file path and remedy, or quarantining an unparseable file (rename to `.broken`), since a file that can't be parsed protects nothing anyway.

Minor: `unknown_scratch_parents` holds scratch paths whose parents are undecodable, not parent paths, so the name misleads.

PR #4747, opened 2026-08-17
Run artifact selectors in resolved environments

### LINKED ISSUE #3225: Feature request: loading dependencies and/or artifacts from package hooks
## Context

Package hooks (e.g. `.pkg/select_artifacts.jl`) are currently designed to execute in a fixed environment where it's only possible to load stdlibs. This has been done to simplify installation, https://github.com/JuliaLang/Pkg.jl/pull/2024#issuecomment-760534026, because (quoting @staticfloat) it makes it possible to install packages and their dependencies without requiring an iterative approach.

I do however think that I'm in a situation where this capability, or at least the capability to access some artifact during execution of the artifact selection hook, is required.

## Issue

The problematic package is, as always, `CUDA.jl`. The situation is as follows:

- `CUDA.jl` requires binaries for the runtime, and the driver. Both are (to be) provided by a JLL;
- The driver JLL packs a single .so, while the runtime JLL has many artifacts (one for each version of CUDA);
- After loading the driver, we can query the supported CUDA version and use that to select an artifact.

This means that the `CUDA_Runtime_jll` artifact selection hook needs to be able to download the driver artifact. However, that's not possible, because we can't load `CUDA_Driver_jll` from the hook as the environment is not yet set-up. The situation could be simplified by removing `CUDA_Driver_jll` and putting its artifact into `CUDA_Runtime_jll` and loading it directly, but that currently does not work either. @staticfloat suggested making it possible to load lazy artifacts, which would mostly resolve this issue.

(I'd still be in favor of the ability to load actual dependencies though, because in reality we don't just want to load `CUDA_Driver_jll`, but a full-blown package, because the driver selection logic is [more complicated](https://github.com/JuliaGPU/CUDA_Driver.jl/blob/main/src/CUDA_Driver.jl) than just loading an .so from an Artifact, and we'd like this functionality to be reusable across several packages. But that's not crucial, and I guess we could inline it all into the `CUDA_Runtime_jll`.)

## Workaround

I only ran into this when registering `CUDA_Runtime_jll`, because once the environment has been instantiated the dependency _is_ available and can be loaded from the package hook (which seems to use the global environment). That means that the unavailability of `CUDA_Driver_jll` only happens when initially installing `CUDA.jl`.

Given that the above issue is so rare, can't I just for now (as a temporary workaround) do:

```
$ head .pkg/platform_augmentation.jl
```
```julia
try
    using CUDA_Driver
catch err
    using Pkg
    redirect_stdout(stderr) do
        Pkg.activate(; temp=true)
        Pkg.add("CUDA_Driver")
    end

    using CUDA_Driver
end

# the rest of the hook
```

Sure, this is pretty cursed, but it works (and lets me continue working on the JLLification of our GPU stack). Or is there something fundamentally broken about this?

cc @vchuravy

maleadt:
Artifact hooks run before Pkg writes the newly resolved manifest. Dependency imports can therefore load stale versions or fail on a fresh installation, making installation-time artifact selection disagree with runtime initialization.

Run hooks against a temporary copy of the resolved environment. Hooks can use plain `using` for their package's dependencies; those dependency names take precedence over the active project's, while source locations come from the resolved manifest. Preserve active-project and workspace preferences, and exclude checkout manifests, checkout preferences, and unrelated stacked environments.

Install static artifacts first, then selector artifacts in dependency order, with parallel downloads within each level. Cache successful selections for the session and reuse them during status checks. Hooks retain `--compile=min` and inherit the parent's optimization level so dependency precompilation is reusable. A separate prerequisite commit fixes installation checks to inspect every selected artifact.

Validation: the final artifact suite passes 204 assertions on nightly, covering dependency loading and identity, bootstrap ordering, manifest relocation, workspace preferences, checkout isolation, and cache invalidation. Closure-box detection and formatting checks pass. The preceding implementation also passed all 30 test files across a full run and targeted reruns, and installed and imported `CCCL_C_jll` 3.5.0+0 in a fresh depot. Those broader checks were not repeated after the dependency-load-path update; cross-platform CI is still needed.

Hooks execute before build scripts, must reserve stdout for TOML, and must tolerate dependencies that cannot initialize on the host during cross-installation. Cache inputs exclude arbitrary system state and included augmentation files; changes outside the key require a new session.

Fixes #3225.

### DIFF
--- a/docs/src/artifacts.md
+++ b/docs/src/artifacts.md
@@ -299,4 +299,13 @@ end
 
 This ensures that the same artifact is used by your code as Pkg attempted to install.
 
-Artifact selection hooks are only allowed to use `Base`, `Artifacts`, `Libdl`, and `TOML`. They are not allowed to use any other standard libraries, and they are not allowed to use any packages (including the package to which they belong).
+Hooks may use standard libraries and load their package's `[deps]` with `using MyDependency`.
+Pkg supplies the resolved dependency versions and installs their artifacts before running the hook.
+Preferences come from the active project and its workspace parents; unrelated stacked environments and the package checkout's own manifest and preferences are excluded.
+
+Hooks run before package build scripts and should not load their own package, whose artifacts have not yet been selected.
+Dependencies must be loadable without a build step, and hooks should handle initialization failures when installing for another platform.
+Keep diagnostics, including those from dependencies, on `stderr` so they do not corrupt the TOML output.
+
+Successful selections are cached for the Julia session and invalidated by changes to the environment, preferences, selector script, or artifact TOML file.
+Changes to other inputs, such as system state or included augmentation files, require a new session.
--- a/docs/src/artifacts.md
+++ b/docs/src/artifacts.md
@@ -309,4 +309,4 @@ Keep diagnostics, including those from dependencies, on `stderr` so they do not 
 
 Successful selections are cached for the Julia session and invalidated by changes to the environment, preferences, selector script, or artifact TOML file.
 Changes to other inputs, such as system state or included augmentation files, require a new session.
-`Pkg.status` never runs hooks: it reports a package with a hook as downloaded once its source is present, and leaves checking the selected artifacts to `Pkg.instantiate`.
+`Pkg.status` and offline resolution never run hooks: they treat a package with a hook as downloaded once its source is present, and leave checking the selected artifacts to `Pkg.instantiate` and the installing operations.
--- a/docs/src/artifacts.md
+++ b/docs/src/artifacts.md
@@ -303,7 +303,7 @@ Hooks may use standard libraries and load their package's `[deps]` with `using M
 Pkg supplies the resolved dependency versions and installs their artifacts before running the hook.
 Preferences come from the active project and its workspace parents; unrelated stacked environments and the package checkout's own manifest and preferences are excluded.
 
-Hooks run before package build scripts and should not load their own package, whose artifacts have not yet been selected.
+Hooks run before package build scripts and cannot load their own package, whose artifacts have not yet been selected; Pkg reports an error when a hook does.
 Dependencies must be loadable without a build step, and hooks should handle initialization failures when installing for another platform.
 Keep diagnostics, including those from dependencies, on `stderr` so they do not corrupt the TOML output.
 
--- a/docs/src/artifacts.md
+++ b/docs/src/artifacts.md
@@ -309,3 +309,4 @@ Keep diagnostics, including those from dependencies, on `stderr` so they do not 
 
 Successful selections are cached for the Julia session and invalidated by changes to the environment, preferences, selector script, or artifact TOML file.
 Changes to other inputs, such as system state or included augmentation files, require a new session.
+`Pkg.status` never runs hooks: it reports a package with a hook as downloaded once its source is present, and leaves checking the selected artifacts to `Pkg.instantiate`.
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1597,7 +1597,6 @@ function check_artifacts_downloaded(pkg_root::String; platform::AbstractPlatform
             if !artifact_exists(Base.SHA1(artifacts[name]["git-tree-sha1"]))
                 return false
             end
-            break
         end
     end
     return true
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -1840,8 +1840,12 @@ end
 function check_artifacts_downloaded(
         pkg_root::String;
         platform::AbstractPlatform = HostPlatform(), selector_project::Union{Nothing, String} = nothing,
-        env_key::Union{Nothing, UInt} = nothing
+        env_key::Union{Nothing, UInt} = nothing, run_selectors::Bool = true
     )
+    # Without running the selector there is no way to know which artifacts it would pick,
+    # so callers that opt out treat such packages as downloaded. Installation still runs
+    # the selector, so a missing selection is corrected there.
+    run_selectors || !has_artifact_selector(pkg_root) || return true
     collected_artifacts = try
         collect_artifacts(pkg_root; platform, selector_project, env_key)
     catch err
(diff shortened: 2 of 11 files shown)


maleadt:
@KristofferC Could you take a look? Would be great to have this in 1.13, in case this becomes LTS.

KristofferC:
> Would be great to have this in 1.13, in case this becomes LTS.

I really doubt 1.13 will be LTS but will still look

KristofferC:
Tangentially, one thing I am a bit worried about with these hooks is that we run them in e.g. `Pkg.status` to tell if the environment is instatiated. If they start loading a bunch of deps that will be much slower. Maybe it could be cached or something but still has to be paid once on every session...

src/Operations.jl
```diff
@@ -1350,11 +1361,12 @@ function collect_artifacts(pkg_root::String; platform::AbstractPlatform = HostPl
...
-                # has not been updated yet, if we try to load any dependencies, it may fail.
-                # Therefore, this project inheritance is really only for Preferences, not dependencies.
-                # We only guarantee access to the `stdlib`, which is why we set `add_stdlib` here.
-                select_cmd = Cmd(`$(gen_build_code(selector_path; inherit_project=true, add_stdlib=true)) --compile=min -t1 --startup-file=no $(triplet(platform))`)
+                select_cmd = if selector_project === nothing
+                    gen_build_code(selector_path; inherit_project = true, add_stdlib = true)
+                else
+                    gen_build_code(selector_path; project = selector_project, load_path = ["@", "@stdlib"])
+                end
+                select_cmd = Cmd(`$select_cmd --compile=min -t1 --startup-file=no $(triplet(platform))`)
```
KristofferC: I don't know if the `--compile=min` is problematic when actually loading proper deps.

maleadt:
> I really doubt 1.13 will be LTS but will still look

In that case, this can brew some more.

src/Operations.jl
maleadt: Compile mode is not part of `Base.CacheFlags` AFAIU. There was a problem because of an inherited `-O0` though: Julia’s cache matching accepts a higher optimization level than requested, but not a lower one, so an `-O0` cache was rejected by the usual `-O2` process and the dependency was compiled twice. Selectors now keep `--compile=min` but run at the parent process’s optimization level. In a local benchmark, the resulting cache was reusable on first normal load, while keeping `--compile=min` made the selector itself slightly faster.

maleadt:
> Tangentially, one thing I am a bit worried about with these hooks is that we run them in e.g. `Pkg.status` to tell if the environment is instatiated. If they start loading a bunch of deps that will be much slower. Maybe it could be cached or something but still has to be paid once on every session...

Successful selections are now cached for the current Julia session, keyed by the artifact files, target platform, resolved environment content, preferences, and hook contents. This avoids rerunning the hook during the status display and subsequent operations.

I also changed `is_instantiated` to check all package sources and statically selected artifacts before starting any hook. Hooks are then checked in dependency order, so an incomplete environment is detected without spawning selectors that cannot initialize their dependencies.

vchuravy:
Summoning @staticfloat for a brief look.

docs/src/artifacts.md
```diff
@@ -299,4 +299,13 @@ end
 
 This ensures that the same artifact is used by your code as Pkg attempted to install.
 
-Artifact selection hooks are only allowed to use `Base`, `Artifacts`, `Libdl`, and `TOML`. They are not allowed to use any other standard libraries, and they are not allowed to use any packages (including the package to which they belong).
+Hooks may use standard libraries and load their package's `[deps]` with `using MyDependency`.
+Pkg supplies the resolved dependency versions and installs their artifacts before running the hook.
+Preferences come from the active project and its workspace parents; unrelated stacked environments and the package checkout's own manifest and preferences are excluded.
+
+Hooks run before package build scripts and should not load their own package, whose artifacts have not yet been selected.
```
vchuravy: > should not load their own package

Can we make this an error?

KristofferC:
> Successful selections are now cached for the current Julia session

I am still thinking of the CLI use where each invocation is a new process. I don't know if it makes sense to cache the result even between processes or if "quick" operation like status should not run the custom artifact selectors just to show if the env is instantiated or not.

maleadt:
> don't know if it makes sense to cache the result even between processes or if "quick" operation like status should not run the custom artifact selectors just to show if the env is instantiated or not.

I originally had a session-persistent cache, but removed it to avoid complexity. I can explore not running these on cheap operations though.

staticfloat:
This seems good to me; I did a cursory inspection of the code and don't see anything obviously wrong.  I had to solve a similar problem in BB2 where I didn't want to keep on re-resolving the same JLLs over and over again.  I ended up hashing all the inputs to the resolution algorithm (including current registry commit) and caching that, but that gets busted every time General updates, so I ended up using Claude to write some code to analyze the registry and hash [only the slice of the registry that is involved in the resolution of that environment](https://github.com/JuliaPackaging/BinaryBuilder2.jl/commit/51a594be14277f78628936f6cde6c5afbfb559aa).  This is extremely new code (only on a branch) but I think it's what I want to go with, as in principle I think it's the right thing to do.

I'd be pretty happy if you stole this code from me and embedded it directly into `Pkg` for use in caching the artifact selection logic, and then I can just make use of it from BB2.  I'm thinking something like "given these root-level package names, hash this `RegistryInstance`'s content for those packages and all recursive dependents".  Given that hash (and combining with whatever other inputs you need, such as julia version, target triplet, etc...) we should be able to give a pretty stable resolution result.

maleadt:
@KristofferC Agreed; `Pkg.status` and offline resolution no longer run artifact selectors. They treat a package with a selector as downloaded once its source is present, and leave checking the selected artifacts to `instantiate` and the installing operations.

@staticfloat Thanks for the pointer. I decided against on-disk caching, since with CUDA.jl it depends on host state Pkg can't see (e.g. the NVIDIA driver). On the package side we handle this by calling `Base.include_dependency` on the driver library so precompilation is invalidated, but wiring an equivalent into Pkg would need hooks to declare what they depend on, which seems clunky. With status no longer running hooks, I don't think a cross-process cache is needed anymore.

docs/src/artifacts.md
maleadt: I think so, but it's annoying since we do want preferences from the package to stick here. I have a hacky version locally, but `__init__`s still run, so I'm not sure it's worth the mess.

PR #4754, opened 2026-08-27
Fix #4750: don't override registry deps with parent's [sources] entries

### LINKED ISSUE #4750: [1.13+] Failure to instantiate when environment contains subpackage from registry and sourced main package with sourced subpackage in its environment
Probably due to https://github.com/JuliaLang/Pkg.jl/pull/4151 @IanButterworth 

The situation is the following. KernelInterface is a released standalone package in a subdir of the unreleased KernelAbstractions 0.10 dev branch. Attempting to add the main branch of KernelAbstractions with the released version of KernelInterface errors as shown below. KernelAbstractions has a local source entry for KernelInterface and removing that from the KernelAbstractions Project.toml seems to make the error go away.

MWE:
```julia-repl
(jl_8Zbu9a) pkg> activate --temp; add KernelInterface KernelAbstractions#main
  Activating new project at `/var/folders/4g/lnkpkf3s4rxd_wbl8vwnqs4r0000gn/T/jl_0mfuB0`
    Updating git-repo `https://github.com/JuliaGPU/KernelAbstractions.jl.git`
   Resolving package versions...
ERROR: could not find source path for package KernelInterface based on manifest /private/var/folders/4g/lnkpkf3s4rxd_wbl8vwnqs4r0000gn/T/jl_0mfuB0/Manifest.toml
```
The Manifest mentioned in the error is either deleted automatically or never created.

See also https://buildkite.com/julialang/metal-dot-jl/builds/3339/table?jid=01a029cb-7173-496b-8137-f5f5c8f89b23&tab=output

rajat12826:
## Problem

When a repo-tracked package (e.g., `KernelAbstractions#main`) has a dependency in its `[sources]` section with a local path (e.g., `KernelInterface = {path = "lib/KernelInterface"}`), and the user's environment also has the same package added from the **registry**, the `add` operation fails with:ERROR: could not find source path for package KernelInterface based on manifest ...

This was introduced/surfaced by #4151 which added an explicit `nothing` check in `fixups_from_projectfile!`.

## Root Cause

In `collect_fixed!`, when processing a repo-tracked package's dependencies via `collect_project`, deps with paths from the parent's `[sources]` section were unconditionally added to the `fixed` dict — even when the same UUID was already present in the top-level `pkgs` list as a registry-tracking package.

This caused the resolver to skip assigning the package a version/tree_hash (since fixed packages are excluded from `Resolve.resolve()` output). The registry-tracking `PackageSpec` in `pkgs` was left with `tree_hash=nothing` and `path=nothing`, causing `source_path()` to return `nothing` in `fixups_from_projectfile!`.

## Fix

Added a guard clause in `collect_fixed!` that checks if a dep discovered via a parent package's `[sources]` already exists as a **registry-tracking** package in the full `pkgs` list. If so, the registry version takes precedence and the path-based dep is skipped from the fixed set.

The `collect_fixed!` function now accepts an optional `all_pkgs` parameter (the full package list including registry-tracking packages), which `resolve_versions!` passes at the call site.

## Test

Added a regression test in `test/sources.jl` with a new test fixture `RepoWithSourcedRegistryDep` that reproduces the exact scenario:
- A repo package with `Example` in `[deps]` and `[sources] Example = {path = "lib/Example"}`
- Adding both `Example` (from registry) and the repo package in the same `Pkg.add` call
- Verifies `Example` remains registry-tracked and the operation succeeds

### DIFF
--- a/src/Operations.jl
+++ b/src/Operations.jl
@@ -629,7 +629,7 @@ function collect_developed(env::EnvCache, pkgs::Vector{PackageSpec})
     return developed
 end
 
-function collect_fixed!(env::EnvCache, pkgs::Vector{PackageSpec}, names::Dict{UUID, String}, julia_version)
+function collect_fixed!(env::EnvCache, pkgs::Vector{PackageSpec}, names::Dict{UUID, String}, julia_version, all_pkgs::Vector{PackageSpec} = pkgs)
     deps_map = Dict{UUID, Vector{PackageSpec}}()
     weak_map = Dict{UUID, Set{UUID}}()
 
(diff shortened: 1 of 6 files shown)


KristofferC:
This can be closed based on merging https://github.com/JuliaLang/Pkg.jl/pull/4756 AFAIU

PR #4760, opened 2026-09-02
Run tests in parallel with ParallelTestRunner and other test time optimizations

KristofferC:
Each test file runs in its own worker process scheduled by
ParallelTestRunner, with a serial fallback for Base.runtests where test
deps aren't available. Network traffic stays limited and independent of
the number of workers:

- a local caching proxy in front of the pkg server (shared by all test
  processes) downloads each registry/package/artifact resource at most
  once per cache lifetime; locally the cache persists across runs
- the registry and the loaded depot are downloaded/populated once up
  front and shared by all workers via PKG_TESTS_CACHE_DIR

new.jl's "Depot setup" now runs against a private depot; the shared
loaded depot is populated by Utils.populate_loaded_depot! instead.

Co-Authored-By: Claude Fable 5 <noreply@anthropic.com>


KristofferC:
I want to run this with:

- a first time
- a second time (where the time durations are recorded and should be run in better order)
- tweak the number of cpus (e.g. for mac which now runs with 1 cpu)

christiangnrd:
I would suggest adding the `--verbose` flag since I've found the "Init time" column to be the best way of knowing if a mac mini is overwhelmed from high memory demand

KristofferC:
Noticed we are not running with the cache workflow so added that.

christiangnrd:
PTR 2.8.2 is [released](https://github.com/JuliaTesting/ParallelTestRunner.jl/releases/tag/v2.8.2)...

KristofferC:
Oh, ok,

giordano:
`new` should probably be split up into multiple files

image

KristofferC:
Yep, still working on this.

christiangnrd:
Mentioning it here since you're overhauling CI, but macos-14 is [deprecated](https://github.com/actions/runner-images/issues/13518) and being fully removed November 2nd so you might want to use macos-26 or just macos-latest

KristofferC:
Timings now:

```
┌───────────────────────────┬────────┬──────────┬─────────┐
│            Job            │ master │ PR #4760 │ Speedup │
├───────────────────────────┼────────┼──────────┼─────────┤
│ ubuntu x64, pkg server    │ 29:15  │ 6:40     │ 4.4x    │
├───────────────────────────┼────────┼──────────┼─────────┤
│ ubuntu x86, pkg server    │ 36:30  │ 7:21     │ 5.0x    │
├───────────────────────────┼────────┼──────────┼─────────┤
│ ubuntu x64, no pkg server │ 32:00  │ 8:20     │ 3.8x    │
├───────────────────────────┼────────┼──────────┼─────────┤
│ macOS-14 aarch64          │ 24:35  │ 4:36     │ 5.3x    │
├───────────────────────────┼────────┼──────────┼─────────┤
│ windows x64               │ 44:41  │ 8:56     │ 5.0x    │
├───────────────────────────┼────────┼──────────┼─────────┤
│ windows x86               │ 54:14  │ 9:40     │ 5.6x    │
└───────────────────────────┴────────┴──────────┴─────────┘
```

code coverage is murderous for us so I have disabled it for now (can maybe be picked up after https://github.com/JuliaLang/julia/pull/62724)

KristofferC:
FWIW, this is another case where 3 macOS runners are hugely beneficial.

christiangnrd:
> FWIW, this is another case where 3 macOS runners are hugely beneficial.

Yeah the 2GiB per worker is good for the kind of tests PTR was originally built for, but I agree that we need a better way to help PTR decide how many workers would be appropriate for a testsuite.

I'm even forcing 3 jobs on macos in the [PR](https://github.com/JuliaTesting/ParallelTestRunner.jl/pull/178) to use PTR in PTR :)

PR #4762, opened 2026-09-03
Make `atomic_toml_write()` preserve the process umask

JamesWrigley:
Previously this would use the 0o600 permissions of `mktemp()`, which for files like the registry `General.toml` would make them only readable by the creator, which causes errors when depots are on shared filesystems meant to be used by multiple users. Auth files are kept with 0o600.

Behaviour changed in 1.13 with #4293. Written with help from Claude :robot:


JamesWrigley:
(rebased to fix merge conflicts)

PR #4764, opened 2026-09-04
Allow a `[sources]` entry to give both a `path` and an `url`/`rev` repo (fix #4089)

### LINKED ISSUE #4089: Allow simultaneous `path` and `url` under [sources] with `path`-priority
I have a situation where I'd like to use the [sources] entry of Project.toml, to indicate that a package, X, should first be looked for at some `path` - and, if not present there, be looked for at some `url`. I had hoped this might be possible to do by something like
```toml
[sources]
X = {path="..", url="https://github.com/user/X.jl"}
```
but that's not possible currently, by design, cf. https://github.com/JuliaLang/Pkg.jl/blob/69c6de01909d53b0a4514b11715f829fe0569d98/src/project.jl#L104-L106

I think it would be nice to be able to do this though: it would allow one to develop multiple interdependent private packages locally, preferring local versions specified in `path` if the exist, and then distribute them to collaborators, who might not have a local versions (then pulling from `url`).

Basically, I think it would be natural to let `path` take precedence if both `path` and `url` are present and a directory with a project exists at `path`.

thchr:
This is a stab at implementing the feature I hoped for in #4089: if a `[sources]` entry contains *both* a `path` and an `url`, then Pkg will first attempt to use `path` - and, if that folder doesn't exist, fall back to `url`.

Generated with Claude Opus, and checked & revised subsequently. Robot summary (~ commit message) below.

I'm not very familiar with the code here, so there could well be issues, despite my checking. E.g., I highlight uncertainties on whether:
- the implementation handles present `rev` but absent `url` correctly, 
- the new tests are idiomatic and robust,
- the handling of the error message if neither `path` nor `url` could be found.

---

🤖 An entry listing both `path` and `url` was rejected when the project file was read. Allow it, with the meaning asked for in #4089: the `path` is used when it exists locally, the repository when it does not, so a local checkout can shadow the repository without the entry being edited.

Downstream uses all assume only one of the two is set, so the choice is made in the one place they funnel through, `get_path_repo`, which still returns either a path or a `GitRepo` but never both. It also raises the error for when neither is there, as far as that can be told locally. `write_env` and `update_source_if_set` both reconstruct an entry from whichever source is in use, and now leave an entry with a fallback alone so the unused key survives the next write.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -11,6 +11,9 @@ Pkg v1.14 Release Notes
   recompilation easier to spot. ([#4679])
 - During package source installation, Pkg now reports when a package has an Artifacts.toml but no artifacts match the
   current platform. ([#4646])
+- A `[sources]` entry may now give both a local `path` *and* a repo `url`/`rev` to fall back on. The path is used if it
+  exists and the repository otherwise, so a local checkout can shadow the repository without the entry being edited.
+  ([#4764])
 
 Pkg v1.13 Release Notes
 =======================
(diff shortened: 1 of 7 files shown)


thchr:
The `windows-latest` test error is in the artifacts code; so unrelated, I assume.

PR #4774, opened 2026-09-11
precompile: add --noskip and --force

IanButterworth:
Claude:

---

Exposes the new `skip_dependents` and `force` keywords of `Base.Precompilation.precompilepkgs` from JuliaLang/julia#63122. By default Base now skips packages whose dependencies failed to precompile; `--noskip` attempts them anyway, for packages that only load a dependency on some platforms. `--force` recompiles packages whose cache files are already fresh. Standard libraries are not forced: their cache files would land in the user depot and shadow the ones julia ships from then on, so the earlier `--forceall` is dropped.

Keywords are passed straight through, so this must wait for the julia PR to merge.


KristofferC:
Forceall including stdlibs seems a bit obscure. Couldn't it be a separate parameter (include_stdlibs) or something like that?

IanButterworth:
Actually I think allowing forcing the stdlibs is a bad idea because they'll end up in the user depot and then be loaded rather than the cached ones from there out.

PR #4775, opened 2026-09-12
artifacts: keep `^C` working under Julia 1.14's scope cancellation

IanButterworth:
Claude:

---

Julia 1.14 delivers `^C` by cancelling the scope the evaluation runs in rather than throwing an `InterruptException` into it, and that broke interrupting an artifact installation in two ways.

A `pkg>` command used to run on the REPL's frontend task, outside any evaluation. With no episode open the press was ignored, so `pkg> add FFMPEG, Gtk` followed by `^C` simply ran to completion. The commands of an input line now run on the REPL backend through `REPL.call_on_backend`, as `julia>` input does, so the backend's per-evaluation episode covers them and Pkg does not touch the episode machinery. Whatever escapes the command's own error reporting, such as the cancellation (which makes that reporting throw too), is printed with `REPL.print_response` the way `julia>` mode prints it.

From `julia>` mode the cancellation did arrive, but the progress bar task only recognised `InterruptException`, so it died with an `Unhandled Task ERROR: CancellationRequest`, and the write that restores the cursor was itself cancelled, leaving the cursor hidden. That cleanup now runs with cancellation masked.

The interrupt test from #4757 gains a variant that cancels the scope the way `^C` does on 1.14; its fake terminal now refuses writes in a cancelled scope like a real one, which is what makes the cursor check fail without the fix. A new test runs a `pkg>` command through a REPL backend and cancels its episode. Both only run where Base has the cancellation machinery. Checked on a 1.14.0-DEV build in a real REPL: `^C` during `pkg> add FFMPEG, Gtk` now stops the downloads, restores the cursor and returns to a usable prompt.

**Needs JuliaLang/Downloads.jl#299 in julia first.** Once the cancellation reaches the in-flight transfers, the current Downloads frees a cancelled request's easy handle while libcurl's multi still uses it: `^C` during `pkg> add FFMPEG, Gtk` on a 1.14 nightly with this PR alone printed `curl_multi_socket_action: 8` and then segfaulted in `Curl_checkheaders`. With that Downloads change, the same `^C` cancels cleanly, with no task errors. So that this can be tried together with it, the last commit points `[sources]` at that branch; it is to be dropped before merging, once the Downloads change is in julia's nightly. Until julia's Pkg is bumped past #4776, the stock Pkg running `Pkg.test` on CI cannot instantiate a git-sourced stdlib from scratch, so CI is expected to be red here; the run on the fix commit alone, with the stock Downloads, was green.


ext/REPLExt/REPLExt.jl
```diff
@@ -121,7 +121,11 @@ function do_cmds(repl::REPL.AbstractREPL, commands::Union{String, Vector{Command
         if commands isa String
             commands = prepare_cmd(commands)
         end
-        return REPLMode.do_cmds(commands, repl.t.out_stream)
+        # the error display below has to run outside the episode's scope: once that is
+        # cancelled, writing to the terminal would throw again
+        return with_interrupt_episode() do
```
Keno: This smells a little. A SIGINT episode should be around one semantic unit of front end work.

PR #4780, opened 2026-09-14
Fix `Pkg.add` with a tag `rev` under `JULIA_PKG_USE_CLI_GIT=true`

### LINKED ISSUE #4779: JULIA_PKG_USE_CLI_GIT=true breaks tag revs on 1.13: CLI git fetch throws before the #4441 refspec fallback can run
Adding a package at a tag `rev` regressed in Julia 1.13 when `JULIA_PKG_USE_CLI_GIT=true`. The restricted-refspec fallback added in #4441 is present in 1.13 but is unreachable on the CLI-git path, because that path turns a failed `git fetch` into a thrown `pkgerror` instead of returning.

### Repro

Public, no credentials needed. `v0.5.3` in Example.jl is a tag that does not point at the default branch tip:

```julia
# JULIA_PKG_USE_CLI_GIT=true
import Pkg
Pkg.add(url="https://github.com/JuliaLang/Example.jl", rev="v0.5.3")
```

```
fatal: couldn't find remote ref refs/heads/v0.5.3
ERROR: The command `git -C ~/.julia/clones/<hash> fetch -q --depth=1 https://github.com/JuliaLang/Example.jl +refs/heads/v0.5.3:refs/cache/heads/v0.5.3` failed,
error: ProcessFailedException(...ProcessExited(128))
```

Matrix, each cell run against a fresh `JULIA_DEPOT_PATH` so no clone cache is reused:

| Julia | `JULIA_PKG_USE_CLI_GIT` | result |
|---|---|---|
| 1.12.7 | true | OK (installed) |
| 1.12.7 | false | OK (installed) |
| 1.13.0 | false | OK (installed) |
| **1.13.0** | **true** | **FAIL — couldn't find remote ref `refs/heads/v0.5.3`** |

Only the CLI-git cell on 1.13 fails.

The tag must not point at the default branch tip. If it does, the initial `git clone --depth=1` already carries the tag, `get_object_or_branch` succeeds, and no fetch is attempted — which is why this looks intermittent across repos.

### Cause

`Types.jl` (1.13.0) narrows the refspec for anything that isn't a commit hash:

```julia
elseif !looks_like_commit_hash(rev_or_hash_str)
    specific_refspec = ["+refs/heads/$(rev_or_hash):refs/cache/heads/$(rev_or_hash)"]
    GitTools.fetch(ctx.io, repo, repo_source_typed; refspecs = specific_refspec, depth = 1)
```

and then applies the #4441 fallback:

```julia
obj_branch = get_object_or_branch(repo, rev_or_hash)
# If still not found, try with broader refspec as fallback
if obj_branch === nothing
    GitTools.fetch(ctx.io, repo, repo_source_typed; refspecs = refspecs_fallback)   # Types.jl:1051
```

That fallback only runs if the first fetch **returns**. Under LibGit2, fetching a refspec that matches nothing is a no-op that returns normally, so `obj_branch === nothing` → fallback → `+refs/*:refs/cache/*` finds the tag. Under CLI git, `GitTools.fetch` raises (`GitTools.jl:242`):

```julia
cmd = `git $args`
try
    run(pipeline(cmd; stdout = devnull))
catch err
    Pkg.Types.pkgerror("The command $(cmd) failed, error: $err")
end
```

`git` exits 128 for an unknown ref, `pkgerror` throws, and the fallback five lines down in the caller never executes.

Same root as #4422 (from the #2330 refspec narrowing); #4441 fixed the LibGit2 path only.

### Impact

CI setups that set `JULIA_PKG_USE_CLI_GIT=true` (common for SSH auth to private repos) cannot resolve any `[sources]`/`Pkg.add` entry whose `rev` is a tag once they move to 1.13. Workaround is replacing tag revs with the commit SHA, which resolves through the `looks_like_commit_hash` branch instead.

### Versions

Julia 1.13.0 / Pkg 1.13.0 (failing), Julia 1.12.7 (working), Linux x86_64.

krynju:
Fixes #4779.

## Problem

Since Pkg 1.13, adding a package at a tag fails when `JULIA_PKG_USE_CLI_GIT=true`:

```julia
Pkg.add(url="https://github.com/JuliaLang/Example.jl", rev="v0.5.3")
```
```
fatal: couldn't find remote ref refs/heads/v0.5.3
ERROR: The command `git -C ~/.julia/clones/<hash> fetch -q --depth=1 https://github.com/JuliaLang/Example.jl +refs/heads/v0.5.3:refs/cache/heads/v0.5.3` failed, error: ProcessFailedException(...)
```

The same command works with LibGit2, and worked with CLI git on 1.12.

## Cause

`handle_repo_add!` first fetches a non-hash `rev` with a narrow `+refs/heads/<rev>` refspec (#2330). For a tag, that ref does not exist on the remote. The two git backends disagree on what happens next:

- **LibGit2**: fetching a refspec that matches nothing is a silent no-op. The rev is still missing, so the broader `+refs/*:refs/cache/*` fallback fetch added in #4441 runs and finds the tag.
- **CLI git**: `git fetch` exits 128 with `couldn't find remote ref`. `GitTools.fetch` turns that into a `PkgError`, so the fallback a few lines later is never reached.

#4441 only ever fixed the LibGit2 path. This is the same root cause as #4422.

## Fix

- `src/Types.jl`: the narrow fetch (PR ref / branch / commit hash) is wrapped in a `try`. A `PkgError` from it is treated as "not found" and control continues to the existing lookup and fallback fetch. If the rev really does not exist, the fallback path still raises `Did not find rev ... in repository` as before; any other failure (network, auth) is raised by the fallback fetch.
- `src/GitTools.jl`: the CLI `git fetch` now captures stderr and appends it to the `PkgError` message. Previously the error carried only a bare `ProcessFailedException`, and the now-recoverable first fetch would otherwise print a misleading `fatal: couldn't find remote ref` line to the terminal on a successful add.

## Test

New testset `tag rev missing from clone cache` in `test/repo.jl`, inside the existing loop that runs the download tests with `JULIA_PKG_USE_CLI_GIT` unset and `"true"`. It needs no network:

1. add a package from a local git repo, which populates the clone cache;
2. create a new commit upstream and tag it `v1.0.0`;
3. `Pkg.add(url=..., rev="v1.0.0")` and check the resolved tree hash is the tagged commit's;
4. check that `rev="does-not-exist"` still throws `PkgError`.

The issue's shallow-clone scenario cannot be reproduced against a local repo (local and `file://` sources disable `--depth`), but this scenario hits exactly the same code path. Without the fix the CLI-git iteration fails with `couldn't find remote ref refs/heads/v1.0.0`; with it both iterations pass.

## Backport

The `src/` changes apply cleanly to `release-1.13`. The test does not, because `test/repo.jl` only exists on master (#4760); on 1.13 the same loop lives in `test/new.jl` and the test block ports there verbatim.


KristofferC:
LGTM (I removed a, what I felt was, unnecessary comment)

krynju:
> LGTM (I removed a, what I felt was, unnecessary comment)

Thanks, I'll let you know when this is ready. There was some CI failure on the original commit

krynju:
Should be good, CI passes

PR #4793, opened 2026-09-16
CI: re-enable code coverage

KristofferC:
Coverage was disabled in #4760 because \`--code-coverage\` prevents julia from using the native code of package images, so Pkg gets compiled from scratch in every test worker and spawned julia process. This turns it back on (including the \`julia-processcoverage\` and codecov upload steps that were removed) so the actual cost with the parallel test runner can be measured against master.


DilumAluthge (approved):

KristofferC:
Things are still kind of slowed down by this. Not sure I care enough about coverage on every PR to care about that.

PR #4797, opened 2026-09-16
Improve time to first Pkg operation

KristofferC:
Reading a compressed registry meant decompressing the tarball and walking
~75k tar entries into a Dict on every session that touched the registry
(~90 ms plus GC). Instead, build a per-registry cache file once
(registries/.cache/<Name>-<treehash>.v1.bin) with a per-package index and
the uncompressed TOML files, and mmap it so a session only pays for the
packages it uses. Loading the registry drops from ~90 ms to ~2 ms; first
Pkg.status()/Pkg.add go from ~125 ms to ~19 ms. JULIA_PKG_REGISTRY_CACHE=false
restores the old path.

Also avoid runtime compilation on the first operation: Resolve.logstr no
longer specializes on the type of stderr, the TOML printer used for usage
logs is precompiled explicitly, and gc's usage reduction uses a loop instead
of a broadcast.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>


IanButterworth (approved):

KristofferC:
I want to think about this a little more. But the approach seem reasonable.

IanButterworth:
My claude (opus 5.5) went off trying to do the same thing as this before I pointed out this PR was open. It found these things with this PR:

---

Two bugs, reproduced locally:

1. Cache contents aren't validated past the header. A truncated or zero-filled file (e.g. power loss; the write has no fsync) throws `BoundsError` from `_cache_string` during registry load, outside any `try`. Since the file is keyed by tree hash and never rewritten, every session then fails until `.cache` is deleted by hand. Suggest validating every (offset, len) against the file size at load (~1 ms), deleting and rebuilding on failure, and fsyncing before the rename.

2. Cleanup matches caches by name prefix (`first(split(basename(tar), '.')) * "-"`). Registries `My.A` and `My.B` delete each other's cache and rebuild every session; removing `Foo` deletes `Foo-Bar`'s; and a future `CACHE_VERSION` bump will delete the other version's files, so release and nightly sharing a depot thrash. Suggest stripping the `.tar.gz`/`.tar.zst` extension for the name and deleting only files matching `^$name-[0-9a-f]{40}\.v$(CACHE_VERSION)\.bin$`.

Smaller: temp files from killed writes (~28 MB for General) and caches orphaned by `registry rm` are never collected; `mv(...; force=true)` unlinks before renaming where a plain rename is atomic; the build's catch-all swallows `InterruptException`.

The nightly failures are from this PR: `test/registry.jl:694` expects a pkg-server request, but `pkg_server_registry_urls()` is now lazy, so a git-only registry update no longer makes one. Looks intended, so the test needs updating.

Issue #4783, opened 2026-09-15
Pkg.update() unnecessarily downgrades HDF5_jll from 2.1.2 to 1.14.3 while updating CSV


indexthm:
## Summary

On Julia 1.13.0, an untargeted `Pkg.update()` can select a strictly older `HDF5_jll` even though no compatibility constraint requires the downgrade.

In the minimal reproducer below, updating `CSV` from 0.10.17 to 1.0.0 also downgrades:

- `HDF5_jll` from 2.1.2+0 to 1.14.3+3
- `OpenMPI_jll` from 5.0.11+0 to 4.1.9+0

`CSV` and `SQLite` have no dependency path to `HDF5_jll`. A subsequent targeted update of `HDF5_jll`, without changing the direct dependencies, immediately restores 2.1.2+0. This demonstrates that the downgrade is not needed to satisfy compatibility constraints.

## Reproducer

This reproduces with the General registry containing CSV 1.0.0.

```julia
using Pkg

ENV["JULIA_PKG_PRECOMPILE_AUTO"] = "0"

Pkg.activate(; temp=true)
Pkg.add([
    PackageSpec(name="HDF5", version=v"0.17.3"),
    PackageSpec(name="CSV", version=v"0.10.17"),
    PackageSpec(name="SQLite", version=v"1.8.2"),
])

function installed_version(name)
    only(info.version for (_, info) in Pkg.dependencies() if info.name == name)
end

println("Before update:")
println("  CSV         = ", installed_version("CSV"))
println("  SQLite      = ", installed_version("SQLite"))
println("  HDF5_jll    = ", installed_version("HDF5_jll"))
println("  OpenMPI_jll = ", installed_version("OpenMPI_jll"))

Pkg.update(; update_registry=false)

println("After update:")
println("  CSV         = ", installed_version("CSV"))
println("  SQLite      = ", installed_version("SQLite"))
println("  HDF5_jll    = ", installed_version("HDF5_jll"))
println("  OpenMPI_jll = ", installed_version("OpenMPI_jll"))

Pkg.update(
    ["HDF5_jll"];
    mode=Pkg.PKGMODE_MANIFEST,
    update_registry=false,
)

println("After targeted HDF5_jll update:")
println("  CSV         = ", installed_version("CSV"))
println("  SQLite      = ", installed_version("SQLite"))
println("  HDF5_jll    = ", installed_version("HDF5_jll"))
[...]

PR #4813, opened 2026-09-17
Resolve: let the local optimality pass install a chain of new packages

### LINKED ISSUE #4810: Pkg.resolve chooses `HDF5_jll` v1 instead of v2
I find that `Pkg.resolve` in Julia 1.13 falls back to `HDF5_jll` v1 without any visible reason, instead of using v2.

I am using `HDF5`, which depends on `HDF5_jll`. Its newest version is 2.2.2.

When I create a new Julia setup (`activate --temp`) and `add HDF5` there, then Julia chooses `HDF5_jll @2.2.2`, as expected.

However, when I create a new Julia setup (`activate --temp`) and `add HDF5 Metal CairoMakie` there, then Julia chooses `HDF5_jll @1.14.3` there. This is weird because neither Metal nor CairoMakie depend on `HDF5_jll`, not even indirectly. It seems that `Pkg.resolve` just gets confused and chooses an outdated version of `HDF5_jll`. I cannot see a reason for this because just updating from v1.14.3 to v2.2.2 works fine without changing any other resolved package versions.

This is the pattern I find:

```
[HDF5]                      -> HDF5_jll 2.2.2+0
[HDF5, Metal]               -> HDF5_jll 2.2.2+0
[HDF5, CairoMakie]          -> HDF5_jll 2.2.2+0
[HDF5, Metal, CairoMakie]   -> HDF5_jll 1.14.3+3
```

There is a work-around: I can `add HDF5_jll` as well, and the resolver then prioritizes (?) that package and chooses its latest version.

eschnett:
Fixes #4810.

### Symptom

Adding `HDF5` to an empty environment chooses `HDF5_jll@2.2.2+0` as dependency (the current version). Adding it alongside two packages that have nothing to do with HDF5 gives v1 (an outdated version):

    [HDF5]                     -> HDF5_jll 2.2.2+0
    [HDF5, Metal]              -> HDF5_jll 2.2.2+0
    [HDF5, CairoMakie]         -> HDF5_jll 2.2.2+0
    [HDF5, Metal, CairoMakie]  -> HDF5_jll 1.14.3+3

A solution with `HDF5_jll@2.2.2+0` exists in all four cases. The resolver just fails to find it in the last one.

### Reason

After a solution is found, `enforce_optimality!` makes repeated passes trying to lift packages to newer versions. Within a pass it walks the graph breadth-first and fixes each package's version on first contact.

Packages that were already installed when the pass started are allowed to move up. A package that gets installed as part of the bump is not allowed to move. It is pinned wherever the walk first happens to reach it, and can only be searched downward afterwards. If something discovered later in the same walk needs a higher version of it, the bump is declared impossible and rolled back, even though a consistent assignment exists.

Bumping `HDF5_jll` to 2.x pulls in a chain of new `aws_c_*` packages, which is exactly this situation: an early member of the chain pins a shared dependency at a version that is too low, a later member needs a higher version, and the whole bump is abandoned. Whether this happens depends on the order the walk reaches things, which is probably why adding `Metal` and `CairoMakie` changes the outcome for `HDF5_jll`.

### Correction

When a bump fails because a freshly installed package was clamped below a version it needs, retry once with constraint propagation instead of the order-dependent walk. Pin the attempted version, forbid every already-installed package from moving down, and re-solve. Propagation doesn't care about visiting order, so it can bring in the whole chain at once. The result is only accepted if nothing gets downgraded. The retry is bounded (at most once per (package, version), plus an overall cap).

### Tests

Tested against the live General registry, `HDF5 Metal CairoMakie` now resolves to `HDF5_jll@2.2.2+0`, and `OpenMPI_jll` moves from 4.1.9+0 to 5.0.11+0 (reported in #4783). `HDF5`, `HDF5 Metal`, `HDF5 CairoMakie` and several unrelated environments resolve unchanged. The PR includes a regression test on a hand-built graph with the same shape.

Implemented via AI.


KristofferC:
https://github.com/JuliaLang/Pkg.jl/pull/4814 seems like a better version of this

IanButterworth:
Closing given https://github.com/JuliaLang/Pkg.jl/pull/4814 merged

PR #4816, opened 2026-09-19
Strengthen the admonition against editing `Manifest.toml` by hand

DilumAluthge:
The main goal is to prevent AI agents from editing manifests by hand, but this should also help dissuade humans.

When I was working on https://github.com/JuliaLang/julia/pull/63195, Claude went and manually edited some of the manifest files in the Julia repo to reflect the new LibGit2 version number.

---

Cross-ref:
- https://github.com/JuliaLang/julia/pull/63205


KristofferC:
I don't really like this:

- There may be other tools that edit manifests other than Pkg
- Sometimes, e.g.when debugging, hand editing a manifest might be a good idea. This wording feels like it bans such hand modifications and may confuse an agent more than it helps. 
- Working a lot with agents on Pkg etc I haven't seen this being much of an issue in practice
- I don't really want to change things based on some specific Ai model doing something stupid at some point in time. 

If for some reason this happens a lot for you I think a local AGENTS entry makes more sense

